PR #50, opened 2022-11-04
Issue 39 | Enhancement: rename all the entries from middleware to manager

### LINKED ISSUE #39: Migrate from middleware to manager

In the infrastructure project we have several references where we are using `middleware` where we should use `manager`.

Tansito:
### Summary

Fix #39 

Rename all the entries from the old `middleware` name to the new `manager`.


infrastructure/helm/quantumserverless/charts/manager/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v1
```
IceKhan13, line 1: I think we do not even need it. 
It's better to do expose ping endpoint in manager and add healthcheck in helm chart. We can do it later

Issue #55, opened 2022-11-14
Migrate to kuberay api service


IceKhan13:
### What is the expected enhancement?

Kuberay helm chart provides ApiService to manage ray clusters (CRUD).

We need to migrate to this helm chart and update client to work with those resources.

Issues in project:
- https://github.com/Qiskit-Extensions/quantum-serverless/issues/130
- https://github.com/Qiskit-Extensions/quantum-serverless/issues/131
- https://github.com/Qiskit-Extensions/quantum-serverless/issues/132
- https://github.com/Qiskit-Extensions/quantum-serverless/issues/133

psschwei:
Is this one still needed? If so, I could pick it up.

IceKhan13:
@psschwei awesome! Thank you!
We will post list of issues related to that small epic tomorrow (basically we will split this issue into list of smaller ones) :)

psschwei:
All linked issues have been closed

PR #60, opened 2022-11-14
Bug: client connection

IceKhan13:
### Summary

Hotfix for connection error which were blocking `QuantumServerless` object from creation


IceKhan13:
This is hotfix and will be removed within #55

PR #77, opened 2022-11-22
Issue 68 | Port configuration

### LINKED ISSUE #68: Client: ports configuration for QuantumServerless

There are at least 2 types of ports we need to support:
1. interactive (client) mode for 10001
2. dashboard/jobs for 8265

We now only supporting one port for client mode

In compute resource setting we need to support both ports, but make them with default values.

IceKhan13:
### Summary

Extend port configuration to interactive and job server.

### Details and comments

Closes #68


Tansito (approved):
LGTM!

PR #80, opened 2022-11-29
Issue 70 | Redis middleware

### LINKED ISSUE #70: Create a State middleware to reuse between programs

Allow to `quantum_serverless` library implements a memory state class to reuse between different programs that can run in different pods.

We can start with something simple (key, value), like:

```
quantum_serverless.state.set(key, value) # Atomic operation
@quantum_serverless.state.get(key, value) # Internally k will be the combination of env+k, think about it
```

### LINKED ISSUE #71: Create events interface to notify actions

Allow to quantum_serverless library implements a events-telemetry class to reuse between different programs that can run in different pods.

We can start with something simple (key, value), like:

```
quantum_serverless.events.notify(key, value) # Will be broadcast to all listeners listen by key
@quantum_serverless.events.listener(key) # Internally key will be the combination of env+key
```

IceKhan13:
### Summary

Redis middleware to handle message exchange in the system

-----

Adding `state` to `run_qiskit_remote` will give you extra parameter `state` in the body of the remote function, so you can get and set parameters during execution. 

```python

state_handler = RedisStateHandler(...)

@run_qiskit_remote(state=state_handler)
def func_with_state(state: RedisStateHandler):
    state.set("some_key", {"payload": ...})
    data = state.get("other_key")
    return ...

with serverless:
    res = get(func_with_state()) # no need to pass state as it is handled in decorator

# you can get state after execution
state_handler.get("some_key")
# {"payload": ...}
```

----

Events emitter for telemetry. 

You can manually create `RedisEventHandler` class and emit any messages to any topic. 

```python
events_handler = RedisEventHandler(redis_host, redis_port)

events_handler.publish("some_topic", {"some_key": "some_value"})
```

For all remote functions this class will automatically be created and emit metadata about executed function (resources, etc.)

### Details and comments

Rendered example of usage of state is [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/issue-71/events-bus/docs/guides/06_shared_state.ipynb)

Closes #70 
Closes #71


client/quantum_serverless/core/decorators.py
```diff
@@ -64,7 +72,11 @@ def from_dict(cls, dictionary: dict):
         return Target(**dictionary)
 
 
-def run_qiskit_remote(target: Optional[Union[Dict[str, Any], Target]] = None):
+def run_qiskit_remote(
+    target: Optional[Union[Dict[str, Any], Target]] = None,
+    state: Optional[StateHandler] = None,
+    events_handler: Optional[EventHandler] = None,
```
pacomf, line 78: Can we rename the events_handler by just events? to be consistent with state interface as well

client/quantum_serverless/core/events.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def to_dict(self):
+        """Converts to dictionary."""
+        return asdict(self)
+
+    def __repr__(self):
+        return f"<ExecutionMessage | {self.workload_id}|{self.layer}>"
+
+
+class EventHandler(ABC):  # pylint: disable=too-few-public-methods
```
pacomf, line 66: Can we implement some subscribe interface in case that someone wants to listen some specific events? (or some events by some regexp)

pacomf (changes requested):

pacomf (approved):
LGTM! 🚀

PR #85, opened 2022-12-15
Issue 69 | Nested programs interface

### LINKED ISSUE #69: Create new interface: Nested Programs

Create a new interface called `NestedPrograms` to allow encapsulate the `run_qiskit_remote` methods in a class to bring more powerful logic regarding to reuse/import/export the code in the near future.

The proposal can be something like:

```python
# It will have all/by type default libraries imported

class NestedProgram (...): # To extend in your class
	… # local parameters like visibility : ENUM …

	def __init__(…, env, options): # Options allowed to define the resources to use
		…

	def _import_program(…): # Code imported to run. From local/remote. Overwritten
		…

	def program(in: quantum_serverless.INPUT): # Code to run. Overwritten
		…
		return (quantum_serverless.OUTPUT) out

	@quantum_serverless.run_qiskit_remote(…)
	def run(…, options)		…
		# Run the nested program in the secure and controlled environment
		output = self.program(…)
		…
```

Then, we can create workflows with the new interface, allowing something like

```python
# Invoking Nested Programs

env = { ‘name’:’my_nested_environment_customized’, ’options’: { … } } # It can be a VAR ENV?

myProgram = new MyNestedProgram(env, …)
result_step_1 = myProgram.run(…)


myProgram2A = new NestedProgram(env, {‘from’: ‘cloud’, ‘origin’: ‘ENDPOINT’, ‘options’: { … } }))
future_result_2A = myProgram2A.runAsync(result_step_1)

myProgram2B = new NestedProgram(env, {‘from’: ‘cloud’, ‘origin’: ‘ENDPOINT’, ‘options’: { … } }))
future_result_2B = myProgram2B.runAsync(result_step_1)

[result_2A, result_2B] = quantum_serverless.get([future_result_2A, future_result_2B])

results_step_2  = combine_results_step2(result_2A, result_2B)


myProgram3 = new NestedProgram(env, {‘from’: ‘file’, ‘origin’ ‘PATH|URL’, ‘options’: {…}))
final_result = myProgram3.run(results_step_2)
```

IceKhan13:
### Summary

Nested programs interface

Now we provide a way to execute async nested programs in a following manner

```python
from quantum_serverless import QuantumServerless, Program

serverless = QuantumServerless(...)

my_program = Program(
    name="my_program",
    entrypoint=script.py,
    dependencies=["requests", "pyscf==2.0.3"],
    version="0.0.1"
)

serverless.set_provider("provider")
job = serverless.run_program(my_program) # this will run program on provider
job.status() # to get status
job.logs() # to get logs

serverless.set_provider("another_provider")
another_job = serverless.run_program(my_program) # this will run program on another_provider
```
where `script.py` has our program code of interest for example
```python
# script.py
from quantum_serverless import QuantumServerless, run_qiskit_remote, get

@run_qiskit_remote()
def ultimate():
    return 42


with QuantumServerless():
    result = get([ultimate() for _ in range(10)])

print(result)

```

Moreover we provide interface for program storage backend that can be implemented for example for local storage
```python
from quantum_serverless.core.program import ProgramStorageBackend


class LocalStorageBackend(ProgramStorageBackend):
     ...

program_storage = LocalStorageBackend()

# to save program
program_storage.save_program(my_program)

# to load program
recovered_program = program_storage.get_program("my_program")
```

### Details and comments

Co-authored with @Tansito 

Closes #69

### DIFF
--- a/client/quantum_serverless/__init__.py
+++ b/client/quantum_serverless/__init__.py
@@ -16,6 +16,7 @@ from importlib_metadata import version as metadata_version, PackageNotFoundError
 
 from .core import Provider, run_qiskit_remote, get, put, get_refs_by_status
 from .quantum_serverless import QuantumServerless
+from .core.program import Program
 
 try:
     __version__ = metadata_version("quantum_serverless")
--- a/client/quantum_serverless/core/program.py
+++ b/client/quantum_serverless/core/program.py
@@ -26,15 +26,17 @@ Quantum serverless nested program
 
     Program
 """
+from abc import ABC
 from typing import Optional, Dict, List
 from dataclasses import dataclass
 
 
 @dataclass
-class Program:
+class Program:  # pylint: disable=too-many-instance-attributes
     """Serverless programs.
 
     Args:
+        name: program name
         entrypoint: is a script that will be executed as a job
             ex: job.py
         arguments: arguments for entrypoint script
@@ -45,6 +47,7 @@ class Program:
         version: version of a program
     """
 
+    name: str
     entrypoint: str
     arguments: Optional[Dict[str, str]] = None
     env_vars: Optional[Dict[str, str]] = None
@@ -52,3 +55,38 @@ class Program:
     working_dir: Optional[str] = None
     description: Optional[str] = None
     version: Optional[str] = None
+
+
+class ProgramStorageBackend(ABC):
+    """Base program backend to save and load programs from."""
+
+    def save_program(self, program: Program) -> bool:
+        """Save program in specified backend.
+
+        Args:
+            program: program
+
+        Returns:
+            success state
+        """
+        raise NotImplementedError
+
+    def get_programs(self) -> List[str]:
+        """Returns list of available programs to get.
+
+        Returns:
+            List of names of programs
+        """
+        raise NotImplementedError
+
+    def get_program(self, name: str, **kwargs) -> Program:
+        """Returns program by name of other query criterieas.
+
+        Args:
+            name: name of program
+            **kwargs: other args
+
+        Returns:
+            Program
+        """
+        raise NotImplementedError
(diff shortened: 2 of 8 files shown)


pacomf (approved):
LGTM like first approach!

Some questions:

1) How to concatenate different programs in a workflow? Any use example?
2) How to invoke programs from cloud? I mean, the private/hide version of a program to allow users run a program without get the source code of it. Any thoughts or idea to implement it?
3) In the example, Job can be named like program, right? Because inside of a program we can run several runtime jobs, for example
4) In the section not implemented yet ProgramStorageBackend, what is the meaning of Backend? can we use another naming to avoid confusion?

IceKhan13:
> How to concatenate different programs in a workflow? Any use example?

I think workflows will be inside as tasks and program is just an entrypoint to execution
```python
# program.py

@run_qiskit_remote()
def step_1():
    ...

@run_qiskit_remote()
def step_2():
    ...

@run_qiskit_remote()
def combine_results_step2(step_1_res, step_2_res):
    ...

with serverless:
    result = get(combine_results_step2(step_1(...), step_2(...)))
```


> How to invoke programs from cloud? I mean, the private/hide version of a program to allow users run a program without get the source code of it. Any thoughts or idea to implement it?

late on `Program` class will `local/remote` flag. So, if program is remote then `run_program` will serialize payload and send it over to cloud. 

You will still be able to call `get_program` on private programs to get `Program` instance, but this instance will only have name of the program to run, no actual data

> In the example, Job can be named like program, right? Because inside of a program we can run several runtime jobs, for example

Yes, I'm planning to update docs a little bit in a following PR. Good suggestion on naming and examples!

> In the section not implemented yet ProgramStorageBackend, what is the meaning of Backend? can we use another naming to avoid confusion?

Good suggestion, I will remote Backend from name

PR #93, opened 2023-01-03
Issue 92 | Benchmark program for resources stress tests

### LINKED ISSUE #92: Programs: add benchmark program to test scalability on given compute resources

Add stress test program to test scalability on given compute resources.

Create a program that combines lot of small and big tasks into a big computational graph to check compute resource scaling. Sort of a benchmark program.

IceKhan13:
### Summary

In order to test compute resources we need to have benchmark program.

Below you can see executon graph of a program. Each of boxes is a task in a program. This program is designed to tests vertical and horizontal scalability of compute resources.

benchmark_program

Program accepts list of arguments to control stress levels:
- `n_entries` - number of circuits and observables will be used in each iteration on generation task.
- `depth_of_recursion` - defines how many recursive iterations generation task will have. (Number of generated circuits will be n_entries * depth_of_recursion).
- `n_backends` - number of backends, which defines how many parallel trainspilation and estimations will happen.
- `n_qubits` - number of qubits will be used in a program. Affects backends, circuits and observables.
- `circuit_depth` - depth of generated circuits.
- `size_of_observable` - size of generated observables.
- `n_graphs` - number of times graph above will be executed within one program.

### Details and comments

Closes #92


Tansito (approved):
Looks really nice @IceKhan13 🎉 . Just to confirm but this is just to a have a way to measure the stress of the system. We are not preparing it to integrate it with our tests or something like that (at least by now).

pacomf (approved):
LGTM! For future load, performance and stress tests, we should use some framework to compare results, get limits for memory and cpu and define thresholds to integrate it in our battery of tests at least to run previous to release a new version. Something similar to https://k6.io/ (I think that for our case k6 is not worthy, because it is ready to test backends, apis, ws, etc... but maybe I am wrong https://github.com/grafana/xk6-kubernetes hehe) but that's the idea.

IceKhan13:
@Tansito @pacomf   yes, you are both correct. it's just a simple program that we can tweak to see how resources are scaling with more programs and tasks running. We can use this program inside something like k6 or similar to track results.

PR #95, opened 2023-01-04
Include Redis in our helm configuration

### LINKED ISSUE #84: Helm: Add redis chart

Add optional redis chart to our helm setup

Tansito:
### Summary

Fix #84 

This PR includes **Redis** in our `helm` configuration for future uses.

### Details and comments

We add the [Bitnami's Redis helm configuration](https://artifacthub.io/packages/helm/bitnami/redis) to our dependencies.

Now the project can be built (we had problems with the versions but is fixed in this PR) and installed. The user can use all the configurations provided by Bitnami for `redis` but to simplify the initial setup I created a similar configuration that we have in the `docker-compose`.

Some changes that we are applying here:
- The port for Redis will be the `7000` to avoid collisions with the port used by `ray-dashboard`.
- Redis configuration is in `standalone` to simplify but the user can configure a replica if it wants.
- The `*.tgz` from Redis is ignored by `.gitignore` (I was not totally sure about this decision or it we need to upload this file).


infrastructure/helm/quantumserverless/values.yaml
```diff
@@ -76,7 +95,7 @@ ray-cluster:
   head:
     initArgs:
       dashboard-host: '0.0.0.0'
-      port: '6379'
```
IceKhan13, line 79: I suggest do that in opposite way. As ray is our 'more important' component we should leave it as is, otherwise we have to configure it everywhere in client. While redis is supplementary dependency, which can be turned off. 

What do you think?

Tansito: My way to think here was: "How the dashboard even being from Ray is maybe something secondary  I can change it". But I don't have a strong opinion honestly. I'm fine with your approach too!

IceKhan13: oh, I know the confusion :) Dashboard is a strange name, but actually it is entrypoint to connect to ray head node (GCS server). So if ray need to do some distributed computation, you usually specify 'auto' connection in context initialization and it will connect to that port. I think, the reason for this port is exactly 6379 because they used redis for that purpose

Tansito: Ooooh, I see. Thanks for the clarification, I will change it then!

infrastructure/helm/quantumserverless/Chart.yaml
```diff
@@ -17,3 +17,5 @@ dependencies:
     condition: rayClusterEnable
   - name: manager
     condition: managerEnable
+  - name: redis
+    condition: redisEnable
```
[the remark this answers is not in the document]
Tansito: Not something to fix in this PR but something that I just come to my mind reading again your messsage, @IceKhan13 . Does it make sense that the user  can disable `ray`? Being this the core of the deployment.

I suppose that the user can only deploy the jupyter to have the examples and the notebook running. It can connect it later on with its external instance but I don't know. Just curious about your opinion here.

IceKhan13 (approved):
Looks great!

PR #168, opened 2023-02-06
Use overrides for kuberay-apiserver

psschwei:

Without this, there was an issue with the clusterrole not being named properly (it was `quantum-serverless-kuberay-apiserver` when kuberay was expecting `kuberay-apiserver`).


IceKhan13 (approved):
Thank you!

PR #169, opened 2023-02-07
Repository: custom validators

### LINKED ISSUE #156: Repository: custom validations

Create custom validation fields to verify for example that the structure followed in the JSON fields are the correct one:

- `dependencies` & `tags`: must be a list not a dict.
- `env_vars` & `arguments`: must be a dict not a list.

Epic: #112

Tansito:
### Summary

Added custom validations to support the limitations of the data types in the models. In this case we are adding two custom validators:
- One to check that a list is a list
- Another one to check that a dict is a dict

Both types are valid JSON data types but we need to restrict what kind of JSON we are introducing in those fields.

### Details and comments

Fix #156 
- Implemented a custom validator for the list fields: `dependencies` and `tags`
- Implemented a custom validator for the dict fields: `env_vars` and `arguments`
- Prepared tests to check that the validations work

### DIFF
--- a/repository/api/validators/dict_validator.py
+++ b/repository/api/validators/dict_validator.py
@@ -1,0 +1,34 @@
+from rest_framework import serializers
+from typing import Any, Dict, List, OrderedDict, Union
+
+
+class DictValidator:
+    """
+    TODO: description
+    """
+
+    def __init__(self, fields: List[str], nullable=True):
+        self.fields = fields
+        self.nullable = nullable
+
+    def __call__(self, attrs: OrderedDict[str, Any]):
+        error_messages = {}
+
+        for field in self.fields:
+            error_message = self.validate(field, attrs[field])
+            if error_message is not None:
+                error_messages.update(error_message)
+
+        if error_messages:
+            raise serializers.ValidationError(error_messages)
+
+    def validate(self, field: str, value: Any) -> Union[Dict[str, str], None]:
+        if value is None:
+            if not self.nullable:
+                return {f"{field}": "This field may not be null."}
+        else:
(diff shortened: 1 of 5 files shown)


repository/api/validators/dict_validator.py
```diff
@@ -0,0 +1,34 @@
...
+
+        if error_messages:
+            raise serializers.ValidationError(error_messages)
+
+    def validate(self, field: str, value: Any) -> Union[Dict[str, str], None]:
+        if value is None:
+            if not self.nullable:
+                return {f"{field}": "This field may not be null."}
+        else:
+            # Using `type` instead of `isinstance` to validate that it is a dict and no a subtype
```
Tansito, line 30: Here @IceKhan13 I used `type` instead of `isinstance` to check that we are receiving a `dict` and not an `OrderedDict`, for example. I think is correct but let me know what you think, I had some doubts.

repository/api/validators/list_validator.py
```diff
@@ -0,0 +1,34 @@
...
+
+        if error_messages:
+            raise serializers.ValidationError(error_messages)
+
+    def validate(self, field: str, value: Any) -> Union[Dict[str, str], None]:
+        if value is None:
+            if not self.nullable:
+                return {f"{field}": "This field may not be null."}
+        else:
+            # Using `type` instead of `isinstance` to validate that it is a list and no a subtype
```
Tansito, line 30: Similar to my previous comment about the dict.

IceKhan13 (approved):
Looks good to me! Regarding typing, we can work it out later in the main PR if needed

PR #209, opened 2023-02-15
Add zlib compression to circuit serializer methods

### LINKED ISSUE #206: Add zlib compression by default to circuit serializer methods

We are using qpy to serialize circuits in serializers class.

qpy is a binary format but for bigger circuits the size is a problem. For example, a characterization example with 80 circuits using a 127 qubit backend with 30 layers (depth) per circuit, it involves a ~45mb qpy object. To mitigate that, we can introduce zlib compression, that will decrease the size of the qpy object to ~2mb.

Introduce zlib for serialize and deserialize methods to allow compress and uncompress qpy object serialized

pacomf:
### Summary

Close #206 adding zlib compression by default into circuit serializer/deserializer methods

### Details and comments

QPY compression data is worthy in comparision with the original size of qpy format without compression

### DIFF
--- a/client/quantum_serverless/serializers/serializers.py
+++ b/client/quantum_serverless/serializers/serializers.py
@@ -31,6 +31,7 @@ Quantum serverless serializers
 """
 import base64
 import io
+import zlib
 
 import ray
 from qiskit import QuantumCircuit, qpy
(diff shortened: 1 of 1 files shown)


Tansito (approved):
🚀

PR #212, opened 2023-02-15
Issue 210 | Repository: file upload

### LINKED ISSUE #210: Repository: add artifact field

Repository: add artifact field

Epic https://github.com/Qiskit-Extensions/quantum-serverless/issues/112

#### References

https://stackoverflow.com/questions/30176570/using-django-rest-framework-how-can-i-upload-a-file-and-send-a-json-payload

IceKhan13:
### Summary

Artifact upload

### Details and comments

Closes #210


repository/api/models.py
```diff
@@ -42,3 +42,4 @@ class NestedProgram(models.Model):
     arguments = models.JSONField(null=True, default=empty_dict)
     tags = models.JSONField(null=True, default=empty_list)
     public = models.BooleanField(default=True)
+    artifact = models.FileField(upload_to="artifacts_%Y_%m_%d", null=False, blank=False)
```
Tansito, line 45: More than enough by now. In the future we could create a path per user like:

```
def user_directory_path(instance, filename):
    # file will be uploaded to MEDIA_ROOT/user_<id>/<filename>
    return 'user_{0}/{1}'.format(instance.user.id, filename)

class MyModel(models.Model):
    upload = models.FileField(upload_to=user_directory_path)
```

Reference [here](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.FileField.upload_to). What do you think?

PR #225, opened 2023-02-20
Issue 201 | Repository: swagger

### LINKED ISSUE #201: Program repository: swagger

Add swagger to program repository API

Epic https://github.com/Qiskit-Extensions/quantum-serverless/issues/112

IceKhan13:
### Summary

Swagger/redoc and openapi for repository

### Details and comments

Screenshots:

Screen Shot 2023-02-20 at 2 48 38 PM

Fixes #201

### DIFF
--- a/repository/main/urls.py
+++ b/repository/main/urls.py
@@ -17,12 +17,36 @@ from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.urls import include, path, re_path
+from rest_framework import permissions
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
 
+schema = get_schema_view(  # pylint: disable=invalid-name
+    openapi.Info(
+        title="Program repository API",
+        default_version="v1",
+        description="List of available API endpoint for Program repository.",
+    ),
+    public=True,
+    permission_classes=[permissions.AllowAny],
+)
 
 urlpatterns = [
     path("admin/", admin.site.urls),
     re_path(r"^v1/api/", include(("api.v1.urls", "api"), namespace="v1")),
     path("api-auth/", include("rest_framework.urls", namespace="rest_framework")),
+    # docs
+    re_path(
+        r"^swagger(?P<format>\.json|\.yaml)$",
+        schema.without_ui(cache_timeout=0),
+        name="schema-json",
+    ),
+    re_path(
+        r"^swagger/$",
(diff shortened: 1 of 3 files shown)


repository/main/urls.py
```diff
@@ -17,12 +17,36 @@
...
+from drf_yasg.views import get_schema_view
+from drf_yasg import openapi
 
+schema = get_schema_view(  # pylint: disable=invalid-name
+    openapi.Info(
+        title="Program repository API",
+        default_version="v1",
+        description="List of available API endpoint for Program repository.",
+    ),
+    public=True,
```
Tansito, line 30: I would set this value to `False` to show only those end-points accesible by the user.

I could make log-in works adding in `settings.py` the next lines of code:
```
SWAGGER_SETTINGS = {
    "LOGIN_URL": "/api-auth/login/",
    "LOGOUT_URL": "/api-auth/logout/"
}
```

Tansito (commented):
To specify [security definitions](https://drf-yasg.readthedocs.io/en/stable/security.html#security-definitions) for swagger I was taking a look about [Django's authentication](https://www.django-rest-framework.org/api-guide/authentication/#api-reference) and it seems Django is using `BasicAuthentication` and `SessionAuthentication` by default. The first one is not quite recommended and ideally we should support `TokenAuthentication` but it seems [Django's implementation](https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication) is not the best solution (they propose other libraries).

How all this logic can be affected by keycloak integration I didn't go more in depth but I think it's a good point out to look out.

repository/main/urls.py
IceKhan13: added in https://github.com/Qiskit-Extensions/quantum-serverless/pull/225/commits/d69349292ff90d6bd79c8ccd98a83954c009c818

IceKhan13:
@Tansito i suggest we migrate to token auth after keycloak as it will change internal implementation of authentication mechanist drastically.

repository/main/urls.py
Tansito: I miss the change about `public = False` to hide end-points not accesible if you are not logged, what do you think, @IceKhan13 ?

IceKhan13: Good suggestion! Fixed!

Tansito (approved):
Nice! 🚀

PR #230, opened 2023-02-22
Update renovate configuration

Tansito:
### Summary

The idea of this PR is group opentelemetry packages when renovate tries to update the dependency. We follow [packageRules](https://docs.renovatebot.com/configuration-options/#packagerules) to group them.


IceKhan13 (approved):
Let's try it 🚀  We can always revert

PR #241, opened 2023-02-28
Implement method to create kuberay cluster via apiserver

### LINKED ISSUE #132: Client: Kuberay provider

Implement Kuberay provider that supports working with kuberay API server

```python
class KuberayProvider(Provider):
    # implemet CRUD for kuberay API server
    ...
```

Depends on:
- https://github.com/Qiskit-Extensions/quantum-serverless/issues/131

Related PRs:
- https://github.com/Qiskit-Extensions/quantum-serverless/pull/178#pullrequestreview-1291505555

Epic: https://github.com/Qiskit-Extensions/quantum-serverless/issues/55

psschwei:
Fixes #132 

Implements the `create_compute_resource()` function for `KuberayProvider`.

Copying the docstring here as it outlines the general way the function is structured:

```
        First, create a compute_template based on the defined ComputeResource.
        If the compute_template already exists (which shouldn't be the case,
        exit silently. Otherwise, create the template. This template is
        ephemeral and will be deleted when no longer needed (either the cluster
        was created or we failed to create it and thus need to start over).
        
        Then use that template to create a KubeRay Cluster. If the cluster
        already exists, we exit silently. Users are only able to configure the
        amount of cpu/memory and the number of worker replicas. The full spec
        can be found in the KubeRay API spec under the "protoCluster"
        definition:
        https://github.com/ray-project/kuberay/blob/master/proto/swagger/cluster.swagger.json
```

I made a number of assumptions in how to approach this, happy to revise any that don't seem to be the right ones :smile:

### DIFF
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -276,6 +276,7 @@ class KuberayProvider(Provider):
         name: str,
         host: Optional[str] = None,
         namespace: Optional[str] = "default",
+        img: Optional[str] = "qiskit/quantum-serverless-ray-node:latest-py39",
         token: Optional[str] = None,
         compute_resource: Optional[ComputeResource] = None,
         available_compute_resources: Optional[List[ComputeResource]] = None,
@@ -298,6 +299,7 @@ class KuberayProvider(Provider):
             name: name of provider
             host: host of provider a.k.a managers host
             namespace: namespace to deploy provider in
+            image: container image to use for ray cluster
             token: authentication token for manager
             compute_resource: selected compute_resource from provider
             available_compute_resources: available clusters in provider
@@ -307,6 +309,7 @@ class KuberayProvider(Provider):
         self.host = host
         self.token = token
         self.namespace = namespace
+        self.image = img
         self.compute_resource = compute_resource
         if available_compute_resources is None:
             if compute_resource is not None:
@@ -317,10 +320,8 @@ class KuberayProvider(Provider):
 
     def get_compute_resources(self) -> List[ComputeResource]:
         """Return compute resources for provider."""
-        req = requests.get(
-            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}/clusters",
-            timeout=TIMEOUT,
-        )
+        api_root = f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
+        req = requests.get(api_root + "/clusters", timeout=TIMEOUT)
         if req.status_code != 200 or not req.json():
             return []
 
@@ -351,43 +352,37 @@ class KuberayProvider(Provider):
         definition:
         https://github.com/ray-project/kuberay/blob/master/proto/swagger/cluster.swagger.json
         """
-        compute_resource_name = self.name + "-template"
+        template_name = self.name + "-template"
+        api_root = f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
 
         # Create template from resource
         req = requests.get(
-            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
-            f"/compute_templates/{compute_resource_name}",
+            api_root + f"/compute_templates/{template_name}",
             timeout=TIMEOUT,
         )
-        if req.status_code == 200:
-            print(f"template name {compute_resource_name} already exists")
+        if req.ok:
+            print(f"template name {template_name} already exists")
             return 1
 
         data = {
-            "name": compute_resource_name,
+            "name": template_name,
             "namespace": self.namespace,
             "cpu": resource.resources["cpu"],
             "memory": resource.resources["memory"],
         }
-        req = requests.post(
-            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
-            f"/compute_templates",
-            json=data,
-            timeout=TIMEOUT,
-        )
-        if req.status_code != 200:
+        req = requests.post(api_root + "/compute_templates", json=data, timeout=TIMEOUT)
+        if not req.ok:
             req.raise_for_status()
             return 1
 
         # Create cluster from template
         req = requests.get(
-            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
-            f"/clusters/{self.name}",
+            api_root + f"/clusters/{self.name}",
             timeout=TIMEOUT,
         )
-        if req.status_code == 200:
+        if req.ok:
             print(f"cluster {self.name} already exists")
-            delete_kuberay_template(self.host, self.namespace, compute_resource_name)
+            delete_kuberay_template(self.host, self.namespace, template_name)
             return 1
 
         data = {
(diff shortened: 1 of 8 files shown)


client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +335,127 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+            return 1
+
+        data = {
+            "name": self.name,
+            "namespace": self.namespace,
+            "user": "default",
+            "clusterSpec": {
+                "headGroupSpec": {
+                    "computeTemplate": compute_resource_name,
+                    "image": "rayproject/ray:2.2.0",
```
IceKhan13, line 400: I think this should be our image.

We can have default nightly one 3.9. Also it would be good to be able to pass it through `resource` argument, or even have `**kwags` for `create_compute_resource` method. Or being able to set it up through env variables.

Maybe it will be something like that:
- if passed as parameter -> use this values
- if not passed as parameter, but env variable exist -> use this env var
- if no env var and passed parameter -> use default

client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +335,127 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+        """
+        compute_resource_name = self.name + "-template"
+
+        # Create template from resource
+        req = requests.get(
+            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
+            f"/compute_templates/{compute_resource_name}",
+            timeout=TIMEOUT,
+        )
+        if req.status_code == 200:
```
IceKhan13, line 362:
```suggestion
        if req.ok:
```

I think this will cover correct redirects too

client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +335,127 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+            "cpu": resource.resources["cpu"],
+            "memory": resource.resources["memory"],
+        }
+        req = requests.post(
+            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
+            f"/compute_templates",
+            json=data,
+            timeout=TIMEOUT,
+        )
+        if req.status_code != 200:
```
IceKhan13, line 378:
```suggestion
        if not req.ok:
```

client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +335,127 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+            req.raise_for_status()
+            return 1
+
+        # Create cluster from template
+        req = requests.get(
+            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
+            f"/clusters/{self.name}",
+            timeout=TIMEOUT,
+        )
+        if req.status_code == 200:
```
IceKhan13, line 388:
```suggestion
        if req.ok:
```

client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +335,127 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+        amount of cpu/memory and the number of worker replicas. The full spec
+        can be found in the KubeRay API spec under the "protoCluster"
+        definition:
+        https://github.com/ray-project/kuberay/blob/master/proto/swagger/cluster.swagger.json
+        """
+        compute_resource_name = self.name + "-template"
+
+        # Create template from resource
+        req = requests.get(
+            f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
```
IceKhan13, line 358: maybe create 

```python
self.api_root = f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
``` 
or something which will shorten this line :)

IceKhan13 (changes requested):
I think only image need to be changed, other than that it has everything we need :)

client/quantum_serverless/core/provider.py
psschwei: Did not know about the qiskit image...

IceKhan13: https://hub.docker.com/r/qiskit/quantum-serverless-ray-node/tags

pacomf (approved):

client/quantum_serverless/core/provider.py
```diff
@@ -273,6 +276,7 @@ def __init__(
         name: str,
         host: Optional[str] = None,
         namespace: Optional[str] = "default",
+        img: Optional[str] = "qiskit/quantum-serverless-ray-node:latest-py39",
```
Tansito, line 279: My only worry about this is that it's a place that we will need to remember to change when it involves a change in the image. How do you see move it to the [constants file](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/quantum_serverless/core/constrants.py) by now, @psschwei , @IceKhan13 ?

Probably not the best solution either but I couldn't think in a better approach by now (cost-effective I mean).

client/quantum_serverless/core/provider.py
```diff
@@ -332,16 +336,122 @@ def get_compute_resources(self) -> List[ComputeResource]:
...
+
+        Then use that template to create a KubeRay Cluster. If the cluster
+        already exists, we exit silently. Users are only able to configure the
+        amount of cpu/memory and the number of worker replicas. The full spec
+        can be found in the KubeRay API spec under the "protoCluster"
+        definition:
+        https://github.com/ray-project/kuberay/blob/master/proto/swagger/cluster.swagger.json
+        """
+        template_name = self.name + "-template"
+        api_root = f"{self.host}/apis/v1alpha2/namespaces/{self.namespace}"
```
Tansito, line 356: Maybe I'm wrong but how do you see to move `api_root` to the constructor to just not repeat it in each method? Maybe it has no sense because it can change depending on what the method does but just to mention.

Tansito (approved):
Awesome job @psschwei ! I just had two comments but they are not blocking 😄

client/quantum_serverless/core/provider.py
psschwei: I think I'm fine with that... any opinions on whether the constant should be the tag only or the whole thing? (I lean slightly towards the former but don't have a strong preference)

client/quantum_serverless/core/provider.py
psschwei: I initially was thinking the same thing, but iirc it didn't play nicely with something... let me try again

psschwei: it was pylint that was complaining: `quantum_serverless/core/provider.py:271:0: R0902: Too many instance attributes (8/7) (too-many-instance-attributes)`... think it should be fine to bump that value up by one

client/quantum_serverless/core/provider.py
Tansito: Yeah, I don't have a strong opinion either. For me it sounds better the first option that you are mentioning too 👍

client/quantum_serverless/core/provider.py
Tansito: If we need it and it has sense I'm fine too 😄

client/quantum_serverless/core/provider.py
Tansito: @psschwei I don't know if you want to address this to use the tag instead of the whole string but from my part is approved. Feel free to merge it 🙏

psschwei: of course, I did the opposite in my patch... do what I do, not what I say huh :laughing: 

> Feel free to merge it

I think we need @IceKhan13 to approve (since he originally requested changes)

Tansito: oh, yes, we can wait then!

IceKhan13 (approved):

Issue #246, opened 2023-02-28
Keycloak does not play nicely with Kind


psschwei:
This came up in #189 ... simplest solution is to use something other than Kind locally (e.g. minikube / docker desktop) if you can, but ideally setup should work for kind too, so...

### Steps to reproduce the problem

Spin up a kind cluster, i.e. `kind create cluster`

Do the usual `helm install`, but the local-ip for connecting to keycloak is not easily found

### What is the current behavior?

Depending on your system, there could be a bunch of network interfaces listed, and the user has to basically try each one until one (hopefully) works

### What is the expected behavior?

The IP would be easily discoverable, either a single IP ala minikube (`minikube ip`) or an easily identifiable service

akihikokuroda:
@psschwei How do you access to the service (like Ray dashboard) in the kind cluster from the host system?

psschwei:
either by creating a node port or using kubectl port-forward

akihikokuroda:
After `helm install`, some pods don't come up.
```
aaaaa-kuberay-head-hpj87                                          0/2     Pending   0             74m
``` 
The status is 
```
    message: '0/1 nodes are available: 1 Insufficient cpu. preemption: 0/1 nodes are
      available: 1 No preemption victims found for incoming pod.'
```
Do they come out  for you @psschwei 
Thanks!

psschwei:
They all come up for me... are you running in a VM or somewhere where you have to provide a set amount of resources? (I've seen similar things happen on action runners and minikube where the environment is constrained)

akihikokuroda:
I'm running it in an IBM cloud instance.  Are you using your local system?

akihikokuroda:
It's a small one **2 vCPU | 4 GB**. Kind works fine for the operator development but it may not be enough for this.  I'll try making it bigger.

psschwei:
yeah I think you'll need to bump it up (if I remember right, I needed at least 4 cpus to run QS on minikube)

akihikokuroda:
I increased number of cpu to 4.  Now it's complaining about memory :-). I'll increase it.

BTW, would you try these?
```
docker exec -it kind-control-plane bash
apt-get update
apt-get install net-tools
ifconfig
```
You may see something like this:
```
eth0: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 172.18.0.2  netmask 255.255.0.0  broadcast 172.18.255.255
[...]

PR #255, opened 2023-03-06
🏗️  IBM Cloud deployment improvements

Tansito:
### Summary

**THIS PR IS UNDER WORK**

Testing our current configuration in IBM Cloud I could make it work but with some slightly changes.

### Details and comments

Due to the differences between the deployment in local vs IBM Cloud I made more dynamic our Ingress configuration with the purpose to be able to support AWS too without the need to change the template if not only the values (as a normal user would do).

- I removed `nginxconfig` in favor of adding those annotations in `ray-ingress`.
- `ray-ingress` now is more dynamic so you can add the annotations and host configuration through the values file.
- In `RayCluster` removed unused ports.
- Fix in terraform values that now are required by the project: `machine_type`, `workers_per_zone`, `disable_pse`...
- Removed `helm` integration in `terraform` due to now helm release requires values from the cluster after its creation (investigating if we can obtain them).

### Resources

- [IBM Cloud](https://cloud.ibm.com/docs/containers?topic=containers-managed-ingress-setup#managed-ingress-steps-resource) 
- [NGINX Controller annotations](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#rewrite)

**By now I only tested this configuration in IBM Cloud and without KeyCloak. In the next days I will continue testing keycloak and make it work with local (AWS next iteration).**


Tansito:
From #246 I would like to know your opinions about the changes that I'm doing @psschwei @akihikokuroda 🙏 

Once time I finish to make the project fully work in IBM Cloud and localhost I will open the PR.

akihikokuroda:
I looked through the changes.  It looks good to me.

psschwei:
I got an error when I tried installing on an IKS cluster from this PR:

```console
$ helm -n ray install quantum-serverless --create-namespace .
Error: INSTALLATION FAILED: Ingress.extensions "ray-ingress" is invalid: spec: Invalid value: 
[]networking.IngressRule(nil): either `defaultBackend` or `rules` must be specified
```

Tansito:
How this PR got a bit stale I close it in favour of #447 where I continue this work.

PR #268, opened 2023-03-10
Bump actions python to 3.9

psschwei:
### Summary

We're using Python 3.9 as the default elsewhere, so bumping this in the various actions


Tansito (approved):
Awesome, thank you Paul!

Issue #307, opened 2023-03-22
Sync repository and gateway applications to use same language


Tansito:
### What is the expected enhancement?

`repository` and `gateway` are two `django` applications that currently have some differences in the configurations and approaches used when they were developed. The idea is try to sync them to make them similar in those points where they need to be similar:

- [ ] Folder structure
- [ ] Project configuration
- [ ] Update `repository` admin file with a similar approach used in `gateway`
- [ ] Use same validation in `repository` for `artifact` field

PR #320, opened 2023-03-22
Improved repository code quality

Tansito:
### Summary

Ref #307. This PR tries to use the same approach registering modules in the panel administrar as we are doing in `gateway` and the same validator in the `artifact` field.

### Details and comments
- Updated `admin.py` to use annotations instead of the method
- Updated _NestedProgram_ `artifact` field with `tar` extension validator


pacomf:
@Tansito can you check the tests?

pacomf (approved):

PR #354, opened 2023-03-24
API gateway versioning

### LINKED ISSUE #308: Implement versioning for the end-points

As we are doing in `repository` we want to add versioning to manage changes in the end-points for the `gateway` project.

Tansito:
### Summary

This PR includes support in `gateway` api for different versions and introduce the change in the `client`.

### Details and comments

- Applied versioning structure to support different versions in the API
- Modified the client to start using the versions
- `ViewSets` to support different versions now has dynamic serializers with `get_serializer_class`
- Refactorization of `ray_job_status_to_model_job_status`
- Refactorization of `User` model inside job to remove the pylint error, now we access the user through `get_user_model`
- Updated the tests accordingly 
- Refactorization of `run_program` end-point to be just `run`
- `requests` package is added to requirements

Fix #308


client/quantum_serverless/core/job.py
```diff
@@ -223,6 +226,12 @@ def __repr__(self):
 
 def save_result(result: Dict[str, Any]):
     """Saves job results."""
+
+    version = os.environ.get(GATEWAY_PROVIDER_VERSION)
+    if version is None:
+        logging.warning("Please provide `version` of gateway.")
+        return False
```
psschwei, line 233: thinking out loud, but could we default to the latest version if no version is specified? (so if they want to use the old version they have to explicitly ask for it) 

of course, plenty of reasons why that might be a bad idea...

client/quantum_serverless/core/provider.py
```diff
@@ -494,6 +497,10 @@ def __init__(
         if host is None:
             raise QuantumServerlessException("Please provide `host` of gateway.")
 
+        version = version or os.environ.get(GATEWAY_PROVIDER_VERSION)
```
akihikokuroda, line 500: This may cause using different versions `save_result` and the `provider`

gateway/api/v1/views.py
```diff
@@ -0,0 +1,43 @@
...
+    queryset = NestedProgram.objects.all()
+    serializer_class = v1_serializers.NestedProgramSerializer
+    permission_classes = [permissions.IsAuthenticated]
+
+    @staticmethod
+    def get_serializer_job_class():
+        return v1_serializers.JobSerializer
+
+    def get_serializer_class(self):
+        return v1_serializers.NestedProgramSerializer
```
akihikokuroda, line 30: Should this be `return self.serializer_class`?

client/quantum_serverless/core/job.py
Tansito: I thought the same too. I couldn't decide which would be the best option. I agree with your approach too so let's do that.

client/quantum_serverless/core/provider.py
Tansito: Sorry @akihikokuroda I'm not following you. Both resources use `GATEWAY_PROVIDER_VERSION` to obtain the value. Do you mean that how I'm letting the user provide the version through `args` the user could use different versions?

gateway/api/v1/views.py
Tansito: For this particular case not. Because what I'm doing here is override the serializer. For a new version modify the `View` would be something as easy as:

```py
    # in gateway/api/v2/views.py
        return v2_serializers.NestedProgramSerializer
```

And both versions could live at the same time.

gateway/api/v1/views.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Job view set first version. Use JobSerializer V1.
+    """
+
+    queryset = Job.objects.all()
+    serializer_class = v1_serializers.JobSerializer
+    permission_classes = [permissions.IsAuthenticated, IsOwner]
+
+    def get_serializer_class(self):
+        return v1_serializers.JobSerializer
```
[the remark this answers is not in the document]
Tansito: Similar to my previous comment.

client/quantum_serverless/core/job.py
Tansito: Done here @psschwei [f0e7b0e](https://github.com/Qiskit-Extensions/quantum-serverless/pull/354/commits/f0e7b0e419078e14885446bc11651f08a90d136a), let me know what you think.

gateway/api/v1/views.py
Tansito: Changed here @akihikokuroda [81e70aa](https://github.com/Qiskit-Extensions/quantum-serverless/pull/354/commits/81e70aa64355d7e89620609a8882278f750272e9)

client/quantum_serverless/core/job.py
Tansito: Maybe I could create a new variable `GATEWAY_PROVIDER_VERSION_DEFAULT` as we have with `OT_PROGRAM_NAME` and `OT_PROGRAM_NAME_DEFAULT`. What do you think?

client/quantum_serverless/core/provider.py
akihikokuroda: Yes, if the user gives the arg with the different value from the value set in the env variable, does the `save_result` work correctly?

Tansito: Oh yes, now I see your point and you are totally right. My problem when I was writing this code was that (from my perspective) `save_result` should be refactored and live inside some of the upper classes that contains that information. I didn't do it in this PR because I thought there were too many changes and I would need some time to analyze if that method is as it is for some reason.

So I would like to open an issue instead to solve it in this PR maybe (I'm open to discuss it).

akihikokuroda (approved):
LGTM!!

pacomf (approved):

client/quantum_serverless/core/job.py
Tansito: @psschwei I would like to ping you here to know your opinion about the change before continue with the work 🙏

Tansito: Yeah, with the introduction of the `DEFAULT` value that has more sense. I will write that to see how it looks 👍

client/quantum_serverless/core/constants.py
```diff
@@ -41,6 +41,7 @@
 # gateway
 GATEWAY_PROVIDER_HOST = "GATEWAY_PROVIDER_HOST"
 GATEWAY_PROVIDER_VERSION = "v1"
```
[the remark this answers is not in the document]
psschwei: My main concern was with this line: `version = os.environ.get(GATEWAY_PROVIDER_VERSION)`, which with substitution would evaluate to `version = os.environ.get("v1")`  (or am I confusing myself here on what's actually happening)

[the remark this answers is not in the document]
Tansito: Yes, you are absolutely right. I will change it, good catch 👍

psschwei (commented):
minor nit: in one place we use `if version is None` whereas in the other we use `version = version or os.environ...` do we want to standardize?

Tansito:
Now it should work as expected @psschwei . If a user specifies a version overrides everything, if not it goes the environment and at the end the default value.

I can't add the `version` from the user in the second one because `save_result` is outside `Job` but the logic is similar. My idea is to refactorize that method later in another PR including it in the `Job` class. I think it has sense but I need to review it well in case there was a reason becasue we had it like that.

psschwei (approved):

Tansito:
From what I tested there is no breaking changes in the documentation plus all the changes were internal so merging it right now!

PR #358, opened 2023-03-24
Output black diff to stdout for CI

### LINKED ISSUE #357: Make lint check failures more verbose

Currently, if the linter fails, it will do so with a message like this:

```
lint: commands[0]> black --check .
would reformat quantum_serverless/quantum_serverless.py

Oh no! 💥 💔 💥
1 file would be reformatted, 32 files would be left unchanged.
```

It would be nice to include WHAT changes are needed in the CI by including the `--diff` flag in the command as well.

psschwei:
Fixes #357 

### Summary

Updates the `black` check in the CI to use the `--diff` flag (so as to output the suggested changes to stdout). Should help newer users who run into issues with the linters when submitting PRs.

### DIFF
--- a/client/tox.ini
+++ b/client/tox.ini
@@ -27,7 +27,7 @@ commands =
 envdir = .tox/lint
 skip_install = true
 commands =
-  black --check .
+  black --diff --check .
   pylint -rn quantum_serverless tests
   mypy --install-types --non-interactive .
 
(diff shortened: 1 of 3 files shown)


Tansito (approved):
Really cool @psschwei thank you! 💯

Issue #368, opened 2023-03-27
Add dashboard for repository metrics


psschwei:
Create a dashboard for the repository in prometheus (once #367 merges and the new container image is built with those updates)

Epic: #139

psschwei:
note to self: need to figure out how to add prometheus to the `ALLOWED_HOSTS` for django...

Issue #374, opened 2023-03-29
Add prometheus metrics in gateway


Tansito:
### What is the expected behavior?

Expose prometheus metrics from gateway. 

Epic #139 

Reference: #279 & #367, #381

Issue #315, opened 2023-03-22
Configure API projects to use Postgres


Tansito:
### What is the expected enhancement?

To simplify the process for `repository` and for `gateway` we started the MVP's using SQLite. Preparing the projects for production we would like to configure Postgres as the database to use.

akihikokuroda:
I'll try this one.

IceKhan13:
Thank you @akihikokuroda !

Tansito:
Thank you @akihikokuroda ! In case you want some inspiration I was reading some entries to prepare for production the applications:

- https://dragonprogrammer.com/django-drf-api-production-docker/

akihikokuroda:
@Tansito I'll take a look.  Thanks!

Issue #306, opened 2023-03-22
Migrate off Docker


pacomf:
### What is the expected behavior?

We need to have an alternative to run in local without use docker images and the docker desktop app. Propose some alternatives and evaluate the best one to follow the best standard possible. (@psschwei has some ideas).

Some of the alternatives can be:

- Podman (https://podman.io/) -> Supported by RedHat
- LXD (https://ubuntu.com/lxd) ->  Maintained by Canonical
- Containerd (https://containerd.io/) ->  CNCF-graduated container

psschwei:
giant PLUS ONE to this

akihikokuroda:
On my Mac, my setup is:

- Containerd (replacement for docker engine/daemon)
- nerdctl (replacement for docker CLI)
- Rancher Desktop (replacement for DockerDesktop)

Tansito:
I'm a bit newbie on these environments so I will read your comments and learn 😂

psschwei:
Building images:

Most tools out there require docker / podman, so not sure how much sense it makes to drop Dockerfiles at this point in time. If we really, really want to eliminate all docker tooling, podman probably makes the most sense, BUT podman uses cri-o rather than containerd as the runtime. Not a blocker, but just something to keep in mind.

Running images:

@akihikokuroda has a good solution for running images locally, using rancher.

akihikokuroda:
I'm using Rancher Desktop : It supports Linux, Windows, Mac (both x86 and Apple chips). It runs on the containerd with nerdctl.  nerdctl is almost compatible with docker CLI (it supports Docker compose, too)

FYI: https://www.youtube.com/watch?v=evWPib0iNgY

pacomf:
I agree with use docker engine (with rancher, podman or whatever... just we need to consider the more generic and faster way for run in local), but we need to have some script or similar to run everything from scratch by you... so in 1 command (running the script) anyone can have running the stack in local in any architecture (Apple chip, Intel, etc)... makes sense for you @psschwei ? can we do it?

psschwei:
in theory, yes .... at a minimum, we could detect the user's arch and then find/replace image tags based on that

Tansito:
From what I know if we just declare the image as (just to put an example):

```
FROM nginx:stable-alpine3.17-slim
```

It should build the correct architecture (if the image supports it, of course). We are not adding the `SHA` because that way we would be specifying the arch but, I'm finding it a bit unnecessary lately.

[...]

PR #381, opened 2023-03-30
Configuration to scrape repository metrics

psschwei:
### Summary

Configuration to allow prometheus to scrape metrics from the repository

prereq for #368 

### Details and comments

Added a dependency on `django-allow-cidr` so that we could configure `ALLOWED_HOSTS` to be internal services (i.e. from a `10.` address). Also added a `repository: "true"` label for the repository pod(s) in order for the serviceMonitor to be able to scrape them.


repository/main/settings.py
```diff
@@ -27,6 +27,8 @@
 
 ALLOWED_HOSTS = []
 
+# allow connections from any internal host
+ALLOWED_CIDR_NETS = ["10.0.0.0/8"]
```
Tansito, line 31: One doubt here @psschwei , must be this `net` configurable? Are we expecting that can be changed maybe?

Tansito (commented):
Thanks for this @psschwei I just left one doubt. Btw and just to confirm but this configuration that you are doing will need to be replicated in the gateway I assume. I created an issue about that: #374

repository/main/settings.py
psschwei: Good question. I'm not sure if it needs to be configurable... we'd want to allow access from within the cluster (`10.` addresses) and from users loading things into it (I'm assuming we'll have an API that makes an HTTP request). 

By configurable, I'm assuming you're meaning from an end user (and not someone building their own image from source).

(Lots of assumptions here... )

Tansito:
> One thing I saw is that we're running the Django apps in debug mode

Yes for `repository` and `gateway`. I didn't spend too much time yet preparing both applications for productions because including `DEBUG` we have the issue #315 too so I was thinking that they are stuff that it will happen at the same time (I can be wrong).

repository/main/settings.py
Tansito: > got it ... yeah, we'll probably need to tweak that for different environments going forward... for now, the 10. will cover the private IPs of a k8s cluster.

Understood!

> I'd recommend phasing out docker compose for local dev in favor of kind / rancher

I was mostly using `k8s` until now too but for convenience we were maintaining `docker-compose`. Probably is a discussion that we can have now that #306 is getting more attention, I totally agree 👍

Tansito (approved):
Thank you @psschwei 🙏

repository/main/settings.py
psschwei: Seems like the preferred Django solution for production is to use a domain name to the allowed hosts, though not sure exactly how that would work on a HPC...

PR #389, opened 2023-03-31
Add metrics collection for gateway

### LINKED ISSUE #374: Add prometheus metrics in gateway

psschwei:
### Summary

Fixes #374 

Adds metrics for gateway service

### DIFF
--- a/gateway/main/settings.py
+++ b/gateway/main/settings.py
@@ -28,6 +28,10 @@ DEBUG = int(os.environ.get("DEBUG", 1))
 
 ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "*").split(",")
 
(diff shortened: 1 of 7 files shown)


gateway/main/settings.py
```diff
@@ -28,6 +28,10 @@
 
 ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "*").split(",")
```
psschwei, line 29: @pacomf @Tansito related to our earlier discussion, I think we can use something like this in the repository to allow host configuration. will work on that tomorrow

pacomf (approved):

PR #417, opened 2023-04-10
Renumber getting started notebooks

psschwei:
### Summary

All three of the getting started notebooks began with `01_`, which is not ideal for the file viewer sidebar 

image

### Details and comments

Renumbered the notebooks, so there's a `01`, `02`, and `03`.


Tansito (approved):
Simpler two times better 😄

PR #421, opened 2023-04-10
Dockerfile for ray-node arm64

akihikokuroda:
### Summary
Add a new Docker file that builds ray-node container for linux/arm64 platform

### Details and comments

The base image of the container should be for linux/arm64 platform.  Ray project started building one in March.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@
 
 version=0.0.5
 repository=qiskit
+arch=$(shell uname -p)
 
 notebookImageName=$(repository)/quantum-serverless-notebook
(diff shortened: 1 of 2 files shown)


akihikokuroda:
I'll update the Make file to select the correct Dockerfile for the local system.

akihikokuroda:
@Tansito Thanks for review.  Yes, I saw the issue in Ray project.  So I put 2.3.0 in the Dockerfile for the m1 image.  I'll try if I can upgrade for the amd64, too.

IceKhan13 (approved):
lgtm

Maybe later on we can add separate docker-compose-dev file or condition it somehow to build from apropriate file.

PR #390, opened 2023-03-31
Program rename

Tansito:
### Summary

Fix #386 , #342

### Details and comments

- Merged #347 
- Merged #401 
- Updated references from `NestedProgram` to `Program`

PR #442, opened 2023-04-14
Issue 411 | Client: rename `run_qiskit_remote` to `distribute_task`

### LINKED ISSUE #411: Refactorize `run_qiskit_remote` with `distribute_task`

We are using right now an annotation called `run_qiskit_remote` to run the method annotated in a classical resource orchestrated by the stack. We need to rename it to use `distribute_task` after to discuss it with @jenglick

IceKhan13:
### Summary

Client: rename `run_qiskit_remote` to `distribute_task`

`run_qiskit_remote` is deprecated until 0.2 release

### Details and comments

Closes #411


Tansito (approved):
Thank you @IceKhan13 ! 💯 

Btw I didn't do any `deprecation` with #390 . Should have I done it? Which is our process with these cases?

psschwei:
> Btw I didn't do any `deprecation` with #390 . Should have I done it? Which is our process with these cases?

FWIW, I like the alpha / beta / GA lifcycle we have in Knative (which is itself adapted from the k8s lifecycle): https://knative.dev/docs/serving/configuration/feature-flags/#lifecycle

(if we follow that lifecycle, I'd figure we're still in alpha at this point, which would mean no need to deprecate and could just include that it's a breaking change in the release notes)

PR #466, opened 2023-04-27
Issue 456 | Program: move arguments to run method

### LINKED ISSUE #456: Client: move arguments from `Program` to `run_program` method

Move out arguments from program to run_program method in order to make programs sharable without arguemnts

```python
program = Program(
    entry=''
)

serverles.run_program(program, arguments={})
```

IceKhan13:
### Summary

Program: move arguments to run method

### Details and comments

Closes #456


pacomf (approved):
LGTM!

client/quantum_serverless/core/program.py
```diff
@@ -81,11 +79,13 @@ def from_json(cls, data: Dict[str, Any]):
...
 
     def __post_init__(self):
-        if self.arguments is not None and not is_jsonable(
-            self.arguments, cls=QiskitObjectsEncoder
-        ):
-            raise QuantumServerlessException(
-                "Arguments provided are not json serializable."
+        if self.arguments is not None:
+            warnings.warn(
+                "Passing `arguments` as constructor argument to `Program` is deprecated. "
```
psschwei, line 84: nit: do we have an idea of when we're going to remove this? if we do, it might make sense to include that here (e.g. "it will be removed in v0.0.9" or something like that). but not a big deal if not either

IceKhan13: Agree!

psschwei (approved):

PR #469, opened 2023-04-28
Update kuberay apiserver to v0.5.1 in values.yaml

psschwei:
### Summary

xref #468 

Bumps the kuberay/apiserver to the new helm chart version (v0.5.1)

### DIFF
--- a/infrastructure/helm/quantumserverless/values.yaml
+++ b/infrastructure/helm/quantumserverless/values.yaml
@@ -191,7 +191,7 @@ kuberay-apiserver:
  
   image:
     repository: kuberay/apiserver
-    tag: v0.5.0
+    tag: v0.5.1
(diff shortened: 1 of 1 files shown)


pacomf (approved):

Tansito (commented):
@psschwei I'm having problems with the PR. If I try to execute the update I'm not being able to use the new image due to this error:

```
Error response from daemon: manifest for kuberay/apiserver:v0.5.1 not found: manifest unknown: manifest unknown
```

And looking at the repository and the docker-hub I couldn't find the version of the image (the last one is `0.5.0`), do you know if I should verify something more?

psschwei:
No, you're right... looks like they had an issue with the chart that needed a patch release but no changes to the actual image, so we can close this out.

PR #476, opened 2023-04-28
Issue 474 | Gateway: stop jobs

### LINKED ISSUE #474: GatewayProvider: stop job

Apparently stop job is not implemented for GatewayProvider. We need to implement it.

2 things need to be realized here: client gatewayprovider stop job and new detailed view on job for gateway to stop jobs

IceKhan13:
### Summary

Gateway stop jobs

### Details and comments

Closes #474

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -30,6 +30,7 @@ Quantum serverless job
 import json
 import logging
 import os
+import tarfile
 from typing import Dict, Any, Optional
 from uuid import uuid4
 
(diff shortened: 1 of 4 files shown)


pacomf:
Thinking in maybe a feature issue, not for this PR... can we stop any running quantum job that it is running inside of the ray job as well?

IceKhan13:
Then we need to keep track of all of all of the quantum jobs, which is kinda clunky. Basically we will need to monitor each `<Primitive>.run` call and keep all job ids.

Definitely not this PR, but it's worth thinking about this, you right.

pacomf (approved):

PR #490, opened 2023-05-01
Add circuit-knitting-toolbox to jupyter notebook image

psschwei:
### Summary

Add CKT to jupyter notebook image

### Details and comments

It might be nice to be able to use the circuit knitting toolbox in our jupyter notebook, so this PR adds it. As there isn't a released version yet, it clones the repo and builds from source (which may or may not be the right way to go about it, open to alternatives here).

### DIFF
--- a/infrastructure/docker/Dockerfile-notebook
+++ b/infrastructure/docker/Dockerfile-notebook
@@ -12,6 +12,8 @@ RUN pip install ipywidgets
 RUN cd ../
 RUN rm -r ./qs
 
+RUN git clone https://github.com/Qiskit-Extensions/circuit-knitting-toolbox.git ckt && cd ckt && pip install . && cd ../ && rm -r ./ckt
+
 COPY --chown=$NB_UID:$NB_UID ./docs/tutorials/ ./serverless/tutorials/
 COPY --chown=$NB_UID:$NB_UID ./docs/guides/ ./serverless/guides/
 COPY --chown=$NB_UID:$NB_UID ./docs/getting_started/ ./serverless/getting_started/


psschwei:
The only stable release was the [initial one](https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/releases) from back when the repo was first released, so for now I think the only option is to build off of the main branch if we want to include the CKT now. Once a new release is cut though I agree we should switch over to that. We could also wait to add the CKT until there's a stable release... I don't have a strong opinion either way, so either option works for me.

pacomf:
what do you think @IceKhan13 @caleb-johnson @garrison @jenglick ?

caleb-johnson:
> Perfect scenario would be to have CKT as pypi package so we can install in as dependency passing to program object
> 
> ```python
> Program(
>    ...
>    dependencies=["ckt==0.1.0"]
> )
> ```
> 
> But I do not know exact timeline of release, so maybe it make sense to have it build from source now.
> 
> @caleb-johnson should weight in here

There haven't been any major updates since the initial 0.0.0 release, as we've largely been developing internally. I would try to work off 0.0.0 unless you are having problems.

If you can't get it working with 0.0.0 and don't want to install from the main branch, we plan to release sometime in the next month or so, and we will be on timed releases after that. We treat main as our stable branch, so it wouldn't be the worst thing to install from that branch.

garrison:
> It might be nice to be able to use the circuit knitting toolbox in our jupyter notebook

One potential issue is that ckt depends on cplex, which is not available on all platforms (e.g., ARM).  There is an open PR to fix this at https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/109.  If it would be helpful to you, I would be in favor of cutting a new ckt release (0.1.0) after we merge https://github.com/Qiskit-Extensions/circuit-knitting-toolbox/pull/109, but before we add significant new functionality.

psschwei:
> cutting a new ckt release (0.1.0) 

+1 for me

Tansito:
For me this is a good point from @garrison :

> One potential issue is that ckt depends on cplex, which is not available on all platforms (e.g., ARM).

Right now we are making some progress with M1 chips in `quantum-serverless` and I would wait until that is fixed in `ckt` cause we have people in the team with those chips: cc @pacomf . So from my part I would wait a little bit before introduce it as it is in this PR.

psschwei:
Seems general consensus was to wait for a v0.1 release to add CKT, so going to close this.

Tansito:
Thank you @psschwei to open the discussion anyway. Really appreciate it 🙏

caleb-johnson:
@psschwei the release is coming imminently (should have a release today that resolves this problem and another next week for gate cutting). You *could* reopen and wait

psschwei:
> the release is coming imminently (should have a release today that resolves this problem and another next week for gate cutting)

awesome! will reopen and update the PR when the new version is out

garrison:
circuit-knitting-toolbox 0.1.0 has been released.

psschwei:
Seems like I can't reopen this one for some reason (:shrug:), so will open a new PR instead

PR #499, opened 2023-05-03
Use versioned releases in helm chart instead of nightlies

### LINKED ISSUE #458: Helm charts should use latest stable release instead of nightly / latest

The default image in the helm charts should use the current release (e.g. `0.0.7`) instead of `nightly` and/or `latest` tags for images. 

We should also probably version our helm charts, so that users can choose a version to install (e.g. `helm install --version 0.0.7 quantum-serverless ...`), though that may be a separate issue...

psschwei:
### Summary

Fixes #458 
Updates helm chart to use version `0.0.7` for quantum-serverless components


akihikokuroda (approved):
LGTM

PR #504, opened 2023-05-04
docker-compose file for Mac

### LINKED ISSUE #365: Docker option in local is not working in Mac M1 chip
I try to run the mantra:

```
docker-compose pull
docker-compose up
```

in local with a Mac M1 Max chip and it is not working properly, because some images are not ready to run on this chip (gateway, ray-head,  qs-jupyter).

The main problem is about gateway, all requests are generating 500 errors.:S, maybe it is produced by ray dependency:

https://docs.ray.io/en/master/ray-overview/installation.html#m1-mac-apple-silicon-support

akihikokuroda:
### Summary
Fix #365 

### Details and comments
A new docker-compose-dev-arm64.yml file is added for Apply silicon system.


pacomf (approved):

psschwei (commented):
I wonder if instead of adding another file we could use an argument, e.g. `ARCH`, kind of like we do here for version: https://github.com/Qiskit-Extensions/quantum-serverless/blob/441fa167b8bcd53e9bb76bbc823e2fcb1d8fd63d/docker-compose.yml#L5

pacomf:
It can be great!

Also we can consider to provide an script to run (docker-compose pull and up) and detect it in automatic way in the script, pass to an argument (ARCH) and do it in transparent way.

akihikokuroda:
I can not find a way to have only one docker-compose file so I'll create a small script file.

Makefile
```diff
@@ -47,3 +47,13 @@ push-gateway:
 
 push-repository-server:
 	docker push $(repositoryServerImageName):$(version)
+
+compose-up:
+ifeq ($(arch),arm)
+	docker-compose -f docker-compose-dev-arm64.yml up
+else
+	docker-compose -f docker-compose-dev.yml up
```
Tansito, line 55: Should we use the `profile: full` here?

akihikokuroda: I'll add it. Thanks!  I'm actually using `nerdctl` that doesn't support the `profile` so I didn't need it.

PR #508, opened 2023-05-06
use postgresql as database for gateway and repository-server

### LINKED ISSUE #315: Configure API projects to use Postgres

akihikokuroda:
### Summary
Fix #315

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -84,10 +84,16 @@ services:
       - SITE_HOST=http://gateway:8000
 #      - SETTINGS_AUTH_MECHANISM=default #custom_token
 #      - SETTINGS_TOKEN_AUTH_URL=<URL_FOR_TOKEN_VERIFICATION>
+      - DATABASE_HOST=postgres
+      - DATABASE_PORT=5432
+      - DATABASE_NAME=testkeycloakdb
+      - DATABASE_USER=testkeycloakuser
+      - DATABASE_PASSWORD=testkeycloakpassword
     networks:
       - safe-tier
     depends_on:
       - keycloak
+      - postgres
   repository-server:
     container_name: repository-server
     build:
@@ -101,6 +107,13 @@ services:
       - DJANGO_SUPERUSER_USERNAME=admin
       - DJANGO_SUPERUSER_PASSWORD=123
       - DJANGO_SUPERUSER_EMAIL=admin@noemail.com
+      - DATABASE_HOST=postgres
+      - DATABASE_PORT=5432
+      - DATABASE_NAME=testkeycloakdb
+      - DATABASE_USER=testkeycloakuser
+      - DATABASE_PASSWORD=testkeycloakpassword
+    depends_on:
+      - postgres
     networks:
       - safe-tier
   prometheus:
--- a/gateway/main/settings.py
+++ b/gateway/main/settings.py
@@ -9,7 +9,9 @@ https://docs.djangoproject.com/en/4.1/topics/settings/
 For the full list of settings and their values, see
 https://docs.djangoproject.com/en/4.1/ref/settings/
 """
+import os
 import os.path
+import sys
 from datetime import timedelta
 from pathlib import Path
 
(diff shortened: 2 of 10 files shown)


IceKhan13 (approved):
LGTM

@Tansito be aware of this in your PR https://github.com/Qiskit-Extensions/quantum-serverless/pull/447

pacomf (approved):

gateway/main/settings.py
```diff
@@ -89,17 +92,40 @@
...
+    "version": 1,
+    "disable_existing_loggers": False,
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+        },
+    },
+    "root": {
+        "handlers": ["console"],
+        "level": "DEBUG",
```
Tansito, line 105: We can open an issue and do it later but we should be able to modify the level of the LOGs through an environment variable, for example.

gateway/main/settings.py
```diff
@@ -89,17 +92,40 @@
...
+        "PORT": os.environ.get("DATABASE_PORT", "5432"),
+    },
+    "test": {
         "ENGINE": "django_prometheus.db.backends.sqlite3",
         "NAME": BASE_DIR / "db.sqlite3",
-    }
+    },
 }
 
+if "test" in sys.argv:
```
Tansito, line 127: Just as a doubt @akihikokuroda , now if we want to use `sqlite` what command do we need to use: `python manage.py runserver test`?

infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
```diff
@@ -27,6 +27,10 @@ spec:
       serviceAccountName: {{ include "gateway.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      initContainers:
+        - name: waitpostresql
+          image: actions/pg_isready
+          command: ['sh', '-c', 'until pg_isready -U {{ .Values.database.user | quote }} -d "dbname={{ .Values.database.name }}" -h {{ .Release.Name }}-postgresql -p {{ .Values.database.port | quote }}; do echo waiting for myservice; sleep 2; done']
```
Tansito, line 33: Correct me if I'm wrong @akihikokuroda , @psschwei but now from #505 we are not using or requiring the use of {{ .Release.Name }}, right?

Tansito: And thinking in loud. I would suggest to pass the host entirely as we are doing with the user and so on, I don't know what do you think (thinking that we are going to have the DBs outside the cluster, initially).

Tansito: by the way, won't it be required in this case the password?

gateway/main/settings.py
akihikokuroda: It doesn't give the choice to use `sqlite` for runtime.  It is for using the sqlite for the unit tests.

docker-compose-dev.yml
```diff
@@ -84,10 +84,16 @@ services:
       - SITE_HOST=http://gateway:8000
 #      - SETTINGS_AUTH_MECHANISM=default #custom_token
 #      - SETTINGS_TOKEN_AUTH_URL=<URL_FOR_TOKEN_VERIFICATION>
```
Tansito, line 86: Just as a last comment @akihikokuroda I don't know if we can remove these lines commented if we are not going to need them.

Tansito (approved):

docker-compose-dev.yml
akihikokuroda: These lines were added in https://github.com/Qiskit-Extensions/quantum-serverless/pull/524.  I'll leave them there.

Issue #506, opened 2023-05-05
Result method should be protected


Tansito:
### Steps to reproduce the problem

`result` method in `gateway` has public access and should be protected:

https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/gateway/api/views.py#L162

PR #509, opened 2023-05-08
Add circuit-knitting-toolbox to jupyter notebook image

### LINKED ISSUE #510: Nightly builds for notebook image are failing
https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/4898954389
https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/4904624347
https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/4910438932

psschwei:
Fixes #510

### Summary

Add CKT to jupyter notebook image

### Details and comments

It might be nice to be able to use the circuit knitting toolbox in our jupyter notebook, so this PR adds it. As there isn't a released version yet, it clones the repo and builds from source (which may or may not be the right way to go about it, open to alternatives here).

This should also fix the [failing nightly builds](https://github.com/Qiskit-Extensions/quantum-serverless/actions/workflows/image-build-and-push-nightly.yml) for the notebook image by updating the Makefile to include the refactored notebook locations.

(seems that I can't reopen #490 but this is what I would've done there if I could reopen it)

### DIFF
--- a/infrastructure/docker/Dockerfile-notebook
+++ b/infrastructure/docker/Dockerfile-notebook
@@ -2,23 +2,15 @@ ARG IMAGE_PY_VERSION=3.9
 FROM jupyter/base-notebook:python-$IMAGE_PY_VERSION
 
 USER 0
-RUN apt-get -y update && apt-get -y install gcc build-essential libopenblas-dev git
+RUN apt-get -y update && apt-get -y install gcc build-essential libopenblas-dev
 
 USER $NB_UID
 
 COPY --chown=$NB_UID:$NB_UID ./client ./qs
 RUN cd ./qs && pip install .
-RUN pip install ipywidgets
+RUN pip install ipywidgets circuit-knitting-toolbox
 RUN cd ../
 RUN rm -r ./qs
-
-RUN git clone https://github.com/Qiskit-Extensions/circuit-knitting-toolbox.git ckt && \
-    cd ckt && \
-    git fetch --all --tags && \
-    git checkout tags/0.1.0 -b ckt && \
-    pip install . && \
-    cd ../ && \
-    rm -r ./ckt
 
 COPY --chown=$NB_UID:$NB_UID ./docs/running/notebooks/ ./serverless/running/
 COPY --chown=$NB_UID:$NB_UID ./docs/development/examples/ ./serverless/examples/
--- a/infrastructure/docker/Dockerfile-notebook
+++ b/infrastructure/docker/Dockerfile-notebook
@@ -2,7 +2,7 @@ ARG IMAGE_PY_VERSION=3.9
 FROM jupyter/base-notebook:python-$IMAGE_PY_VERSION
 
 USER 0
-RUN apt-get -y update && apt-get -y install gcc build-essential libopenblas-dev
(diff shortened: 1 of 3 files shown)


psschwei:
weird, changing a dockerfile shouldn't break the tests ... :thinking:

psschwei:
hmm... tests pass locally:

```console
$ tox -elint
lint: commands[0]> black --diff --check .
All done! ✨ 🍰 ✨35 files would be left unchanged.
lint: commands[1]> pylint -rn quantum_serverless tests
PYLINTHOME is now '/home/paulschw/.cache/pylint' but obsolescent '/home/paulschw/.pylint.d' is found; you can safely remove the latter

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

lint: commands[2]> mypy --install-types --non-interactive .
tests/core/test_decorator.py:34: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
Success: no issues found in 35 source files
  lint: OK (12.70=setup[0.13]+cmd[0.29,8.76,3.52] seconds)
  congratulations :) (12.85 seconds)
```

psschwei:
think I found it

psschwei:
looks like the auth token is stored in the `access` field instead of `access_token` now

IceKhan13:
Yes, I just facing this issue in #506

psschwei:
Is that something we changed or was it a dep autoupdate?

IceKhan13:
I'm had the same thought. Maybe we should and let "renovate" handle update 🤔

psschwei:
I was actually just thinking something like `library>=2.0,<3.0`, but I like your idea much better :smile:

infrastructure/docker/Dockerfile-notebook
```diff
@@ -12,8 +12,16 @@ RUN pip install ipywidgets
 RUN cd ../
 RUN rm -r ./qs
 
-COPY --chown=$NB_UID:$NB_UID ./docs/tutorials/ ./serverless/tutorials/
-COPY --chown=$NB_UID:$NB_UID ./docs/guides/ ./serverless/guides/
-COPY --chown=$NB_UID:$NB_UID ./docs/getting_started/ ./serverless/getting_started/
+RUN git clone https://github.com/Qiskit-Extensions/circuit-knitting-toolbox.git ckt && \
```
IceKhan13, line 15: I think ckt team just release to Pypi. We might want to include this requirement in `/client/requirements.txt` 

https://pypi.org/project/circuit-knitting-toolbox/

IceKhan13: or we might not even needed anymore since users can do this now:
```python
Program(
  title="DO ckt",
  entrypoint="...",
  dependencies=["circuit-knitting-toolbox==0.1.0"]
)
```

psschwei: ah, did not know that (I just saw the tags, not the releases). that'll simplify things a good bit.

I'd still lean towards including it in the notebook dockerfile (and not to `requirements.txt`) as that would let non-Programs also have access to it

IceKhan13 (approved):

PR #535, opened 2023-05-11
Issue 525 | Better error handling

### LINKED ISSUE #525: Client + gateway: fix user error messages

**User story**: user should not see large error messages. All errors should be handled and clear concise message should appear for user.

Currently client and gateway returns errors with bodies, which is overwhelming for users.
We need to make error handling less verbose and more specific (like: `check auth credentials`, `authentication failed` or something along those lines).

IceKhan13:
### Summary

Better error handling

### Details and comments

Instead of long-long lists of nested exception client handles it with high level exception with clearer message.

Screenshot 2023-05-11 at 4 47 38 PM

Closes #525

### DIFF
--- a/client/quantum_serverless/utils/errors.py
+++ b/client/quantum_serverless/utils/errors.py
@@ -1,23 +1,53 @@
 """Errors utilities."""
-from typing import Optional
+from typing import Optional, Dict, Union
+
+ErrorCodeType = Union[int, str]
 
 
-def format_err_msg(
-    message: str, code: Optional[int] = None, extra: Optional[str] = None
-):
+def is_http_standard_error(error_code: ErrorCodeType):
+    """Returns true if code is in standard http error codes range."""
+    result = False
+    if isinstance(error_code, int) and (600 > error_code >= 100):
+        result = True
+    return result
+
+
+class ErrorCodes:  # pylint: disable=too-few-public-methods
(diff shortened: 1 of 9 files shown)


client/quantum_serverless/utils/errors.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def format_err_msg(
+    message: str, code: Optional[int] = None, extra: Optional[str] = None
+):
+    """Formats error message.
+
+    Args:
+        message: main body of message
+        code: error code
```
pacomf, line 12: Can we have an error code (an string) to identify the errors (maybe we can have a json or a file with all errors categorized)?

Something like:

error_code = "AUTH1001"
message = "..."

So the error_code is a way to identify the message in some way, because the current code we are using it to pass the http status code, for example, so they are not uniques per error

pacomf: Maybe for a next iteration with the error handling... creating a new issue is enough for now

pacomf (approved):

client/quantum_serverless/utils/errors.py
IceKhan13: I like the idea! Will do it here, no worries :)

pacomf (approved):

PR #539, opened 2023-05-12
Make Job.result a blocking command

### LINKED ISSUE #454: job.result should be blocking
Currently, job.result() is non-blocking, which could cause ungraceful crashes if the result is accessed in subsequent code before the result object is available.

job.result should be a blocking command

caleb-johnson:
Resolves #454 

### Summary
Job.result should be a blocking command by default. When users say `result = myjob.result(job_id)`, they will likely expect the result variable to be populated when it is used in the ensuing code. 

### Details and comments
We will add a flag for users to optionally turn off the blocking component of `result()`, and we will add a flag for enabling logging of a heartbeat while the method waits for the result object to populate.

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -286,13 +286,11 @@ class Job:
                 while waiting for job result to populate
         """
         counter = 0
-        print(self._job_client.status(job_id))
         if wait and self._job_client.result(self.job_id) == "{}":
             logging.info("Waiting for job result.")
             counter = 0
             # Check every 3s, log a heartbeat every 30s
             while self._job_client.result(self.job_id) == "{}":
-                print(self._job_client.status(job_id))
                 time.sleep(3)
                 counter += 1
                 if counter % 10 == 0:
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -276,8 +276,27 @@ class Job:
         """Returns logs of the job."""
         return self._job_client.logs(self.job_id)
 
-    def result(self):
-        """Return results of the job."""
+    def result(self, wait: bool = True, verbose: bool = False):
+        """Return results of the job.
+
+        Args:
+            wait: flag denoting whether to wait for the
+                job result to be populated before returning
+            verbose: flag denoting whether to log a heartbeat
+                while waiting for job result to populate
+        """
+        counter = 0
+        print(self._job_client.status(job_id))
+        if wait and self._job_client.result(self.job_id) == "{}":
+            logging.info("Waiting for job result.")
+            counter = 0
+            # Check every 3s, log a heartbeat every 30s
+            while self._job_client.result(self.job_id) == "{}":
+                print(self._job_client.status(job_id))
+                time.sleep(3)
+                counter += 1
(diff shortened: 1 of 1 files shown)


client/quantum_serverless/core/job.py
```diff
@@ -276,8 +277,25 @@ def logs(self) -> str:
...
+                job result to be populated before returning
+            verbose: flag denoting whether to log a heartbeat
+                while waiting for job result to populate
+        """
+        counter = 0
+        if wait and self._job_client.result(self.job_id) == "{}":
+            logging.info("Waiting for job result.")
+            counter = 0
+            # Check every 3s, log a heartbeat every 30s
+            while self._job_client.result(self.job_id) == "{}":
```
caleb-johnson, line 294: I'd like to also check that the status is `Running` here, but I'm not sure how to find what all status strings I can expect back from the `BaseJobClient`

caleb-johnson:
I didn't see a test module dedicated to `job.py`, so there aren't any unit tests for this yet. If there is some specific way I should test this little delay/heartbeat log, I'd be happy to do it -- I'm just not sure what the test would be or where it would go.

client/quantum_serverless/core/job.py
caleb-johnson: I want to avoid an infinite loop here, so I think it'd be best to key off the `status` as well here.

client/quantum_serverless/core/job.py
```diff
@@ -276,8 +277,25 @@ def logs(self) -> str:
...
+                time.sleep(3)
```
IceKhan13, line 295: maybe also move this sleep thing in `cadence` (or something like this) parameter and make default 3

IceKhan13 (commented):
Tests for a program/job are happening [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/tests/core/test_program.py) 

But it is using Provider which is directly connecting to ray cluster defined in [compose file](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/tests/resources/test-compose.yml). Statuses in ray are a little bit different that in gateway (ray has  str version of `JobStatus.PENDING` and gateway has just `PENDING`)

All states described [here in the model ](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/gateway/api/models.py#L55)
And mapping between Ray and Gateway states are [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/gateway/api/utils.py#L23) 


So I would suggest to do something along those lines:

```python
class Job:
   ...

    def in_terminal_state(self) -> bool:
        """Checks if job is in terminal state"""
        terminal_states = ["STOPPED", "SUCCEEDED", "FAILED"]
        job_status = self.status()
        return  any(terminal_state in job_status for terminal_state in terminal_states)

    def result():
        ...
        while not self.in_terminal_state():
```

psschwei:
Out of curiosity, is there a reason not to use `ray.get()` or `ray.wait()` to block? (I'm not an expert on this part of the code, so feel free to disregard if I'm wandering off-topic...)

caleb-johnson:
Closing in favor of #557 . I wasn't working off a fork for this PR :)

PR #543, opened 2023-05-15
Remove unnecessary custom.css

Eric-Arellano:
This was a bad copy and paste from other Qiskit projects. It only has rules for `.toggle`, which no longer exists in the qiskit_sphinx_theme (which this project isn't using either way.)


Eric-Arellano:
By the way, I'm Eric! I joined IBM Quantum 3 months ago to improve Qiskit documentation, including for the Ecosystem. Let me know if I can help you out with documentation or if you have suggestions to improve things :) 

In June/July, we'll be adding a new "ecosystem theme variant" to qiskit_sphinx_theme so that it better fits projects than the current styling we use on e.g. https://qiskit.org/ecosystem/nature/ (like no top nav bar). There's no expectation that Ecosystem projects must use the theme, but I'll coordinate once this is released to see if you all would be interested and to make the migration for you if you'd like :)

IceKhan13 (approved):
Thank you @Eric-Arellano ! 

Please, let us know when ecosystem theme variant will be available :)

Eric-Arellano:
Will do! Also, you're welcome to start using qiskit_sphinx_theme now if you'd like. See https://qiskit.org/ecosystem/nature/ for an example.

We'll soon be improving the visual design by switching the base theme from a Pytorch fork to Furo, which is a nicer and more modern Sphinx theme. It's used by pip, Black, and attrs.

So, the design will soon improve & we'll add the ecosystem variant theme. But, it's valid to start using the current theme now.

Let me know. I'd be happy to put up a PR :)

PR #545, opened 2023-05-15
Gateway: custom auth, token verification

IceKhan13:
### Summary

Gateway: custom auth, token veification

### Details and comments

Adds token verification for custom token authentication backend


pacomf (commented):
LGTM!

PR #544, opened 2023-05-15
Build and push images to ibm container registry

psschwei:
### Summary

xref: #298 

Add actions to build and push images to IBM Container Registry

### Details and comments

Adds new workflow / actions for IBM Container Registry (as we'll be retiring DockerHub in the near future, I just duplicated the files rather than adding a registry parameter to the existing one as we can remove the Docker ones once we're happy with the ICR images).

For the image tag, we're going from `qiskit` to `icr.io/quantum-public`

For the authentication to IBM Cloud, I used https://github.com/IBM/actions-ibmcloud-cli action, which adds the IBM Cloud CLI and logs in to IBM Cloud / Container Registry.

PR #551, opened 2023-05-16
Using docker login for pushing to ICR

psschwei:
### Summary

xref #544 
Log into ICR via docker

### Details and comments

Using the ibmcloud cli login wasn't passing the credentials down to docker for some reason, so let's cut out the middle man and log directly into ICR from docker.

Tested this locally using `act`, so hopefully this should actually "fix" the issue...


pacomf (approved):
LGTM

psschwei:
Ran this on my fork and it's pushing images, so we should be good to go...

PR #557, opened 2023-05-16
Make core Job.result a blocking command

### LINKED ISSUE #454: job.result should be blocking

caleb-johnson:
Resolves #454 

### Summary
Job.result should be a blocking command by default. When users say `result = myjob.result(job_id)`, they will likely expect the result variable to be populated when it is used in the ensuing code. 

### Details and comments
We will add a flag for users to optionally turn off the blocking component of `result()`, and we will add a flag for enabling logging of a heartbeat while the method waits for the result object to populate. Users may also control how often to check for the result through the `cadence` argument

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -303,9 +303,27 @@ class Job:
         """Returns logs of the job."""
         return self._job_client.logs(self.job_id)
 
-    def result(self):
-        """Return results of the job."""
+    def result(self, wait = True, cadence = 5, verbose = False):
+        """Return results of the job.
+        Args:
+            wait: flag denoting whether to wait for the
+                job result to be populated before returning
+            verbose: flag denoting whether to log a heartbeat
+                while waiting for job result to populate
+            cadence: time to wait between checking if job has
+                been terminated
+        """
+        if wait:
+            while self._in_terminal_state():
+                time.sleep(cadence)
+                if verbose:
+                    logging.info(".")
         return self._job_client.result(self.job_id)
+
+    def _in_terminal_state(self) -> bool:
+        """Checks if job is in terminal state"""
+        terminal_states = ["STOPPED", "SUCCEEDED", "FAILED"]
+        return self.status() in terminal_states
 
     def __repr__(self):
         return f"<Job | {self.job_id}>"


client/quantum_serverless/core/job.py
```diff
@@ -303,10 +303,28 @@ def logs(self) -> str:
...
+            wait: flag denoting whether to wait for the
+                job result to be populated before returning
+            verbose: flag denoting whether to log a heartbeat
+                while waiting for job result to populate
+            cadence: time to wait between checking if job has
+                been terminated
+        """
+        if wait:
+            while self._in_terminal_state():
+                time.sleep(cadence)
```
caleb-johnson, line 318: @psschwei @IceKhan13 

Paul recommended using ray for the timeout here. I always default to using pure Python any time I can. I'm really indifferent. Any thoughts @IceKhan13 ? I guess I'd just want to be consistent with what is being done in the rest of the repo in these cases.

`time.sleep`, `ray.wait`, and `ray.get` brings up exactly one instance apiece in the Github search

IceKhan13: `wait`, `get` works on task references inside of the program (ray job). We are working from outside, so your approach is correct :) 
Just fyi, here is what we have from outside https://docs.ray.io/en/latest/cluster/running-applications/job-submission/jobs-package-ref.html#jobsubmissionclient

client/quantum_serverless/core/job.py
```diff
@@ -303,10 +304,30 @@ def logs(self) -> str:
...
+                job result to be populated before returning
+            cadence: time to wait between checking if job has
+                been terminated
+            verbose: flag denoting whether to log a heartbeat
+                while waiting for job result to populate
+        """
+        if wait:
+            if verbose:
+                logging.info("Waiting for job result.")
+            while self._in_terminal_state():
```
IceKhan13, line 320:
```suggestion
            while not self._in_terminal_state():
```

Other than that looks good to go :)

IceKhan13 (changes requested):

IceKhan13 (approved):

PR #553, opened 2023-05-16
Use ICR for deployments

psschwei:
### Summary

Update docker-compose / helm to use ICR as image repository

### Details and comments

Part of the larger migration to ICR from DockerHub
xref #544

PR #558, opened 2023-05-17
Add ARM builds to github actions

### LINKED ISSUE #523: Enable multi-architecture image build and push in github action

The build and push action builds and push container images for amd64 architecture only.  It's better to build multi-architecture images and push them in the registry.

psschwei:
### Summary

Add ARM builds to github actions
Fixes #523 

### Details and comments

Adds a set of steps to build the ARM ray images to our github actions. 
Pushing to the ICR registry as that should be done for all images in #553
To distinguish the ARM images, an `-arm64` is added to the end of the tag.

### DIFF
--- a/.github/actions/build-and-push-images/action.yml
+++ b/.github/actions/build-and-push-images/action.yml
@@ -34,6 +34,33 @@ runs:
         tags: qiskit/quantum-serverless-ray-node:${{inputs.tag}}-py310
         build-args:
           IMAGE_PY_VERSION=py310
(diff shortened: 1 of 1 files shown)


akihikokuroda:
@psschwei Thanks for doing this.  Does ICR support [multi-archtecture](https://blog.thesparktree.com/docker-multi-arch-github-actions) if so we can pull images with the same tag for all architecture.

I think that the images other than ray node also need to be built.  

Thanks!

PR #562, opened 2023-05-17
Issue 555 | Docker-compose: separate dbs for repository and gateway

### LINKED ISSUE #555: Repository: named tables for models

Gateway and repository has the same application name and same table programs. Therefore migrations in both applications were creating the same table in the same database. Migration passed as it was thinking that those table existed already. 

In order to fix this situation we need to have different names for tables for repository and gateway. This can be achieved by providing `Meta` information for models in `repository`

Example:
```python
# repository/api/models.py

class Program(models.Model):

    class Meta:
        db_table = "repository_program"
```

And then migrations must be regenerated

IceKhan13:
### Summary

Docker-compose: separate dbs for repository and gateway.

Simply placing metadata to models will not work due to gateway and repository are using similar auth and other django related migrations. Therefore having them in single database might potentially be harmful. Therefore adding extra database instance for repository.

We will not need this anywhere else other that docker compose as we will have multiple databases within same postgres instance on deployment. And it is a good practice not to mix applications in one db.

### Details and comments

Closes #555


akihikokuroda (approved):
LGTM

PR #563, opened 2023-05-17
Version bump 0.1.0

IceKhan13:
### Summary

Update to version 0.1.0 for release


psschwei (approved):
:partying_face:

pacomf (approved):

akihikokuroda (approved):

psschwei:
Shoot, just noticed something -- looks like a bunch of the notebooks still have the provider host as localhost instead of gateway...

psschwei:
I'm going to submit a PR to update them (maybe we should patch that in?)

akihikokuroda:
I wonder in what environment the `localhost` works.  I always have to change to `gateway`.

IceKhan13:
I think we should have it as env variable or something like that. And provide this env variable in notebook as `"http://gateway:8000/"`.

```python
GatewayProvider(
    host=os.environ.get("GATEWAY_HOST", "http://localhost:8000/")
)
```

psschwei:
You're going to test my grepping / sed skills with that one :smiley:

akihikokuroda:
I want to make the password for the docker compose and the helm same, too.   But it can be next time.

psschwei:
@akihikokuroda which password(s)?

akihikokuroda:
The password of the gateway.

psschwei:
For helm, would we change that in the keycloak password: https://github.com/Qiskit-Extensions/quantum-serverless/blob/0129ca95703adef53210256ea4ca45b08dca4fb0/infrastructure/helm/quantumserverless/values.yaml#L238 ?

I may be missing something obvious (been a long day :tired_face:), but that was the only user/passw0rd combo I saw in the values ...

akihikokuroda:
@psschwei Yes, it seems correct.

psschwei:
ok, cool, thanks -- I'll update that in my PR too

PR #573, opened 2023-05-19
Enable multi-archetecture CI buid

### LINKED ISSUE #523: Enable multi-architecture image build and push in github action

akihikokuroda:
### Summary
Fix #523 

### Details and comments
Enabled minimum images for multi-arch because it increases the build time.


infrastructure/docker/Dockerfile-ray-qiskit
```diff
@@ -1,9 +1,19 @@
...
 
+FROM rayproject/ray:2.4.0-$IMAGE_PY_VERSION AS ray-node-amd64
 USER $RAY_UID
-
 COPY --chown=$RAY_UID:$RAY_UID ./client ./qs
 RUN cd ./qs && pip install .
 RUN cd ../ 
 RUN rm -r ./qs
+
+FROM rayproject/ray:2.4.0-$IMAGE_PY_VERSION-aarch64 AS ray-node-arm64
```
psschwei, line 10: :thinking:  it might be worth seeing if the ray project would take a PR to make their builds multiarch too... :thinking:

akihikokuroda: I'll check the rayproject CI.

akihikokuroda: It is using `buidkite` for its CI.  It may not be simple as github action.

psschwei (approved):

psschwei:
I'm going to go ahead and merge this (in the off chance the builds fail, there will still be time before the weekend to revert... rolling out changes late on Friday is always a little risky :smile: )

psschwei:
Looks like this more than doubled the time of the build and push job (don't think that's an issue, but just making a note of it)

PR #575, opened 2023-05-20
clean up after dockerfile merge

akihikokuroda:
### Summary
This PR remove 2 unnecessary files for amd64 architecture related files and adjust some other files.


docker-compose-dev.yml
```diff
@@ -18,6 +18,8 @@ services:
     build:
       context: ./
       dockerfile: infrastructure/docker/Dockerfile-ray-qiskit
+      args:
+        TARGETARCH: <Please set amd64 or arm64 here in the docker-compose-dev.yml>
```
psschwei, line 22: Can we use an envvar here, similar to how we set the version here: https://github.com/Qiskit-Extensions/quantum-serverless/blob/c761272e7c8cb3fb97d8fd356e95e0e9ece5e3b5/docker-compose.yml#L5

Also, and just thinking out loud here, are there other places where we set the arch that could use the same envvar? (ie if users could `export DOCKERARCH=arm64` in their `.bashrc` and then not have to worry about it anymore...)

akihikokuroda: @psschwei Thanks for review.  I couldn't make the environment variable work so I put the the message there to show the message in the error when the user uses it first time.

We can use the `uname -m` for must of the cases to get the platform architecture.  I just couldn't make the docker compose work with any variables. :-(

psschwei: This works for x86: `TARGETARCH: "${TARGETARCH:-amd64}"`. Could you try that and run with `TARGETARCH=arm64 docker compose...`?

akihikokuroda: Thanks!  The reason that it doesn't work for me may be I'm using `nerdctl` on mac. I put the env variable here for the most of users.

psschwei: I wonder if you need to declare it as an ARG in the dockerfile and/or explicitly pass it as an [environmental variable](https://docs.docker.com/compose/environment-variables/set-environment-variables/) to compose ...

akihikokuroda: No, it actually causes a problem in the github action: https://github.com/docker/buildx/issues/574

psschwei (approved):

PR #576, opened 2023-05-22
Upgrade container packages to latest versions

### LINKED ISSUE #574: Ensure base images are using latest versions

Some of our base images are a bit out of date, so we should update them to the latest versions.

psschwei:

Fixes #574 

We want to avoid running into security vulnerabilties in our images, so we're upgrading system packages to the latest versions via apt update/upgrade.


akihikokuroda (approved):
LGTM

pacomf (approved):

PR #580, opened 2023-05-23
Rename `run_program` to just `run`

### LINKED ISSUE #435: Rename `run_program` to just `run`

To run `Programs` in `QuantumServerless` we need to execute `run_program` method. To make it more generic we would like to:
- Deprecate `run_program`
- Implement the new `run` method

psschwei:
### Summary

Fixes #435


pacomf:
@psschwei we should rename all appearances in documentation files and in jupyter notebooks samples as well, right?

pacomf (approved):
LGTM!

PR #604, opened 2023-05-26
Add repository docker compose profile

### LINKED ISSUE #599: Simplify quickstart profile

Our current quickstart looks like this (from the main readme):

```
docker-compose pull
VERSION=nightly docker-compose --profile full up
```

We can simplify this to make it easier for users to get started:
* make jupyter part of the default install. This will let us drop the profile flag
* Drop the version envvar, as that should default to the latest version

Making both of those changes will make the quickstart a simple `docker compose up` (we should also drop the hyphen since that's going away soon). 

The following locations should be updated:
- docs/quickstart/index.rst
- docs/deployment/local.rst
- README.md
(I didn't do a thorough search so there might be others)

There might be other locations that used a `--profile jupyter` flag, those would need to be cleaned up too.

psschwei:
### Summary

Fixes #599 
Add a profile to docker compose for repository

### Details and comments

Add a separate `"repo"` profile for installing the repository components
Also update to docker compose v2 (dropping the hyphen in `docker-compose`).

### DIFF
--- a/docs/deployment/local.rst
+++ b/docs/deployment/local.rst
@@ -12,9 +12,10 @@ Once you have Docker and docker-compose installed, you can run the following com
 
         $ VERSION=<VERSION> docker compose [--profile <PROFILE>] up
 
-The available profiles are `full` and `repo`.
+The available profiles are `full`, `jupyter`, and `repo`.
 The repo profile installs core services and the program repository,
-while the full profile installs all core services,
+the jupyter profile installs core services and Jupyter Notebook,
+and the full profile installs all core services,
 Jupyter Notebook, and logging and monitoring systems.
 
 Available version can be found in `GitHub releases`_ page.
--- a/docs/deployment/local.rst
+++ b/docs/deployment/local.rst
@@ -2,7 +2,7 @@
 Local infrastructure setup
 ==========================
 
-To set up Quantum Serverless on your local machine, you will need to use `docker-compose`_.
+To set up Quantum Serverless on your local machine, you will need to use `docker compose`_.
 
 .. _docker-compose: https://docs.docker.com/compose/
 
--- a/docs/deployment/local.rst
+++ b/docs/deployment/local.rst
@@ -10,7 +10,7 @@ Once you have Docker and docker-compose installed, you can run the following com
 
 .. code-block::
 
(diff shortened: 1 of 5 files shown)


pacomf (approved):

docs/deployment/local.rst
```diff
@@ -10,10 +10,10 @@ Once you have Docker and docker-compose installed, you can run the following com
 
 .. code-block::
 
-        $ VERSION=<VERSION> docker-compose [--profile <PROFILE>] up
+        $ VERSION=<VERSION> docker compose [--profile <PROFILE>] up
 
-The available profiles are `full` and `jupyter`.
-The jupyter profile installs core services and Jupyter Notebook,
+The available profiles are `full` and `repo`.
```
caleb-johnson, line 15: OK I'm a little slow on the docker stuff, but are `full` and `repo` available here? You removed `jupyter` and `full` as profiles from `docker-compose`, but didn't replace them.

`full` and `repo` are available in `docker-compose-dev`. Is that the environment the user will be in here?

psschwei: > You removed `jupyter` and `full` as profiles from `docker-compose`, but didn't replace them.

By removing the profile from the jupyter service, it will always be installed, so there's no need to pass a flag to get jupyter. 
 
> `full` and `repo` are available in `docker-compose-dev`. Is that the environment the user will be in here?

Whoops, I forgot to add the repo profile to the main `docker-compose.yml` file ... good catch.

psschwei:
@caleb-johnson just checking in: are you good with merging this now or are there other changes needed?

docs/quickstart/index.rst
```diff
@@ -22,7 +22,7 @@ Step 2: Run infrastructure.
 .. code-block::
    :caption: Run docker compose from the root of the quantum serverless project.
       cd quantum-serverless/
-      docker compose --profile full up
+      docker compose up
```
caleb-johnson, line 25: Should we unpin the version above in quickstart if we do this? (see step 1)

psschwei: good call

psschwei:
Assuming no objections, planning to merge this Wednesday afternoon...

PR #611, opened 2023-05-28
add helm chart package to release artifacts

### LINKED ISSUE #52: Upload helm chart to chart repository

Use chart repository to store helm charts

akihikokuroda:
### Summary
Fix #52 

Add github workflow that builds and adds the helm chart package to the release artifacts when the release is published.

### DIFF
--- a/.github/workflows/publish-helm.yaml
+++ b/.github/workflows/publish-helm.yaml
@@ -1,0 +1,43 @@
+name: Helm Package
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tags: 
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Helm
+        uses: azure/setup-helm@v3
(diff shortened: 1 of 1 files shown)


akihikokuroda:
@psschwei We are already using https://qiskit-extensions.github.io/quantum-serverless for our documentations so I  put the package into the release artifacts.

Tansito:
The idea that @IceKhan13 and me analyzed in the past was to try to use https://artifacthub.io/. To use it we saw that the best option was to use `github` in combination with `github-pages` (most of the applications that we saw were using this approach) but for that we required a dedicated repository to manage it and we were waiting for that. That's the reason because it was `on-hold` that issue.

akihikokuroda:
I believe that it's useful to build and publish the helm package for each release so that the user can install the helm chart without cloning repo, setup helm repos, and etc..

psschwei (commented):
LGTM, will let @Tansito give final approval

Issue #463, opened 2023-04-27
Working with persistent storage


IceKhan13:
### What is the expected behavior?

During program execution as a user I want to be able to save files in some kind of persistent storage allocated for my user specifically. 

As a user I might need it for:
- saving large job results
- saving checkpoints for long running jobs / variational workloads
- accessing previously saved files

Notes:
- think of s3

pacomf:
Remember enable the redis possibility, to work with cache storage as well (the state service but through gateway to protect data between users, something like each key sent by users, add the userID like prefix to ensure that other users cant access to these keys...

We should have 2 kind of storage:
- Cache storage, to share data between tasks or store small content
- Persistent storage, to store checkpoint states or more large content

psschwei:
I can tackle this one.

I'd lean towards configuring storage via the infrastructure rather than the SDK (i.e. via PVCs in Kubernetes over a `write_to_s3()` function call from the client), so it might make sense to wait for this until #423 is in...

pacomf:
We need to take care to protect the access to the information between users, without expose any possible interface to get data from other users, it doesnt matter than users are from the same installation

psschwei:
On the cache vs. persistent storage, how would we differentiate which one is used? Or would that be something for the users to decide? If the latter, perhaps we should start simple with just s3 and then introduce redis or something similar down the line if/when there's a big demand for it...

psschwei:
xref IBM COS PVC instructions: https://cloud.ibm.com/docs/containers?topic=containers-storage_cos_apps

pacomf:
Yeah! Exactly, let’s start with S3

psschwei:
Well... looks like PVCs are implemented in Kuberay yet: https://github.com/ray-project/kuberay/blob/f6a172f3d7e44d4ee22bd058f2c33a69601c39c2/apiserver/pkg/util/cluster.go#L387

There's an [open issue](https://github.com/ray-project/kuberay/issues/1087) , but no clue on when it will land... that's ... _not ideal_ :frowning:

psschwei:
@pacomf by when do we need a persistent storage solution to land?

pacomf:
Probably before July, so maybe we need to start with our own solution, at least for the moment, @IceKhan13 what do you think?

psschwei:
(another choice is we submit a patch to add PVC support... long-term, that may end up being less overall work)

psschwei:
[...]

PR #624, opened 2023-05-31
Adds utilities for working with persistent storage

psschwei:
### Summary

Part of #463 
Add `PersistentStorage` class and functions for reading / writing files

### Details and comments

The `PersistentStorage` class uses the `s3fs` library to create an object linked to a specific COS bucket. Users need to set the region endpoint and bucket name in the constructor and set the access/secret key values via environmental variables.

Once that object is created, users can call `PersistentStorage.persist_data()` and `PersistentStorage.retrieve_data()` to write and read data from the s3 bucket. Note that the `filename` parameter for both write and read methods can be several "subdirectories" down, i.e. `this/is/some/fake/s3/directory/tree/file.txt`.

The example below can be run in a notebook to show how the process works, assuming you have a COS bucket and HMAC (AWS style) credentials 

```python
from quantum_serverless.utils.storage import PersistentStorage
import s3fs

get_ipython().run_line_magic('env', 'ACCESSKEY=<somevalue>')
get_ipython().run_line_magic('env', 'SECRETKEY=<somevalue>')
endpoint="https://s3.<region>.cloud-object-storage.appdomain.cloud"
bucket="<yourbucket>"

s3 = PersistentStorage(endpoint=endpoint, bucket=bucket)

data="""It was the best of times, it was the worst of times,
it was the age of wisdom, it was the age of foolishness, 
it was the epoch of belief, it was the epoch of incredulity, 
it was the season of Light, it was the season of Darkness, 
it was the spring of hope, it was the winter of despair, 
we had everything before us, we had nothing before us, 
we were all going direct to Heaven, we were all going direct the other way
—in short, the period was so far like the present period, that some of its 
noisiest authorities insisted on its being received, 
for good or for evil, in the superlative degree of comparison only. """

s3.persist_data("tale-two-cities.txt", data)

s3.retrieve_data("tale-two-cities.txt")

for b in s3.storage.ls(bucket):
    print(b)

s3.storage.rm(f"{bucket}/tale-two-cities.txt")
```

### DIFF
--- a/client/quantum_serverless/utils/storage.py
+++ b/client/quantum_serverless/utils/storage.py
@@ -28,17 +28,22 @@ Quantum serverless storage utilities
 from io import BytesIO
 import os
 from typing import Optional
-import s3fs
+
 from abc import abstractmethod
+import s3fs
 
 
 class BaseStorage:
+    """Base class for persistent storage."""
+
     @abstractmethod
     def save(self, path: str, data: BytesIO):
+        """Save data."""
         raise NotImplementedError
 
     @abstractmethod
     def load(self, path: str):
+        """Load data."""
         raise NotImplementedError
 
 
@@ -61,12 +66,12 @@ class S3Storage(BaseStorage):
             endpoint_url=self.endpoint, key=self.key, secret=self.secret
         )
 
-    def save(self, filename, data):
+    def save(self, path, data):
         """Store binary data in persistent storage."""
-        with self.storage.open(f"{self.bucket}/{filename}", "wb") as f:
+        with self.storage.open(f"{self.bucket}/{path}", "wb") as f:
             f.write(data)
 
-    def load(self, filename):
+    def load(self, path):
         """Get binary data from persistent storage."""
-        with self.storage.open(f"{self.bucket}/{filename}", "rb") as f:
+        with self.storage.open(f"{self.bucket}/{path}", "rb") as f:
             print(f.read())


client/requirements.txt
```diff
@@ -11,3 +11,4 @@ redis>=4.5.5
 opentelemetry-api>=1.18.0
 opentelemetry-sdk>=1.18.0
 opentelemetry-exporter-otlp-proto-grpc>=1.18.0
+s3fs==2023.5.0
```
IceKhan13, line 14:
```suggestion
s3fs>=2023.5.0
```

client/quantum_serverless/utils/storage.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class PersistentStorage:
+    """Class for storing objects in a non-temporary manner."""
+
+    def __init__(self, endpoint: str, bucket: str):
+        """Long-term storage for serverless computation."""
+        self.endpoint = endpoint
+        self.bucket = bucket
+        self.key = os.getenv("ACCESSKEY")
```
IceKhan13, line 39: maybe also add `key: Optional[str] = None` and `secret: Optional[str] = None` to constructor and do something like this here

```python
self.key = key or os.getenv("ACCESSKEY")
```

IceKhan13 (approved):
I was thinking maybe add `BaseStorage` class and S3 will inherit from it. 
Once we will add PVCs we can create another one, something like `FileStorage`.

Something like this:
```python
class BaseStorage:
    @abstractmethod
    def save(self, path: str, data: BytesIO):
        raise NotImplementedError

    @abstractmethod
    def load(self, path: str):
        raise NotImplementedError

class S3Storage(BaseStorage):
    ...

# once PVC will be in place
class FileStorage(BaseStorage):
    ...
```

But it is up to you :) This PR is already good for now

client/quantum_serverless/utils/storage.py
```diff
@@ -0,0 +1,53 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+=====================================================
```
IceKhan13, line 14: oh, I think this might break sphinx docs :) 

Lines should be the same length, I think

psschwei: weird, the docs test didn't run...

psschwei:
@IceKhan13 went ahead and made the class changes (which reminded me how much I haven't missed OO the past few years :wink: )... if you can, take one more look before I merge.

client/quantum_serverless/utils/storage.py
```diff
@@ -0,0 +1,77 @@
...
+        endpoint: str,
+        bucket: str,
+        key: Optional[str] = None,
+        secret: Optional[str] = None,
+    ):
+        """Long-term storage for serverless computation."""
+        self.endpoint = endpoint
+        self.bucket = bucket
+        self.key = key or os.getenv("AWS_ACCESS_KEY")
+        self.secret = secret or os.getenv("AWS_SECRET_ACCESS_KEY")
```
Tansito, line 64: This is an absolute suggestion/question so, thinking that S3 almost works as an standard in most of the clouds for object storage does it have sense to rename this `env_vars` to:
- `S3_ACCESS_KEY`
- `S3_SECRET_ACCESS_KEY`

With the purpose to use this class with any object storage from any cloud provider without the reference to `AWS`, what do you think @psschwei , @IceKhan13 ?

client/requirements.txt
```diff
@@ -11,3 +11,4 @@ redis>=4.5.5
 opentelemetry-api>=1.18.0
 opentelemetry-sdk>=1.18.0
 opentelemetry-exporter-otlp-proto-grpc>=1.18.0
+s3fs>=2023.5.0
```
Tansito, line 14: I was taking a look and it seems `s3fs` doesn't use `sem-ver`. Should we pin the version? It seems each version doesn't add too much things but just thinking in loud.

Tansito (approved):
Just a couple of questions but awesome job @psschwei 🎉

client/quantum_serverless/utils/storage.py
psschwei: The AWS CLI (which a lot of folks use to interact with s3 stores, including IBM COS) requires these exact envvars, so that's why I didn't pick a more generic value here

client/requirements.txt
psschwei: I'm fine either way (originally had it pinned but switched based on @IceKhan13 's [review](https://github.com/Qiskit-Extensions/quantum-serverless/pull/624#discussion_r1214346135))

client/quantum_serverless/utils/storage.py
Tansito: I see, thank you for the explanation, Paul. I agree then 😄

client/requirements.txt
Tansito: Oh, I didn't see it, sorry. I don't have a strong opinion so @IceKhan13 if you think it's fine let us know and we can resolve this conversation.

pacomf (approved):

PR #629, opened 2023-06-01
Helm version update

IceKhan13:
### Summary

Helm version update


psschwei (commented):
Should also bump gateway, etc.

psschwei (commented):
in Chart.yaml, repository (L43) also needs an update

akihikokuroda (approved):

psschwei (approved):

PR #630, opened 2023-06-02
Drop nightly builds

### LINKED ISSUE #621: Drop nightly support in favor of latest

From @psschwei comment : https://github.com/Qiskit-Extensions/quantum-serverless/pull/513#issuecomment-1548818823

Now that we are creating `latest` images we can discuss if we can drop support for `nightly`.

From my point of view I found the comment interesting and I would drop support for `nightly` just to not support more tags. What do you think @IceKhan13 , @akihikokuroda , @psschwei ?

psschwei:
Fixes #621 
Drops nightly builds

### DIFF
--- a/.github/workflows/icr-image-build-and-push.yml
+++ b/.github/workflows/icr-image-build-and-push.yml
@@ -9,8 +9,6 @@ on:
   push:
     branches:
       - main
-  schedule:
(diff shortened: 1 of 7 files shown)


akihikokuroda (approved):
LGTM

PR #631, opened 2023-06-02
fix broken keycloak in docker compose

psschwei:

Reverts changes made in #622
gateway provider was throwing 500 errors when trying to connect using the newer version, so reverting for the time being


caleb-johnson (approved):
I verified that reverting this change makes the demo tutorials work as expected

PR #632, opened 2023-06-02
Use DynamicClient to manage raycluster

akihikokuroda:
### Summary
Use DynamicClient to manage the raycluster instead of the ray apiserver.

This gives more flexibility creating the raycluster instence.

### DIFF
--- a/gateway/api/ray.py
+++ b/gateway/api/ray.py
@@ -8,6 +8,10 @@ import time
 import uuid
 from typing import Any, Optional
 
+import yaml
+from kubernetes import client, config
+from openshift.dynamic import DynamicClient
+
 import requests
 from ray.dashboard.modules.job.sdk import JobSubmissionClient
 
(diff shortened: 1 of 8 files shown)


gateway/api/ray.py
```diff
@@ -99,54 +69,118 @@ def create_ray_cluster(
...
-        timeout=30,
+    config.load_incluster_config()
+    k8s_client = client.api_client.ApiClient()
+    dyn_client = DynamicClient(k8s_client)
+    raycluster_client = dyn_client.resources.get(
+        api_version="v1alpha1", kind="RayCluster"
     )
-    if not response.ok:
+
+    cluster = """
```
psschwei, line 84: will do a full review on Monday, but had a question: could we change multiple resources together? i.e. something like:

```yaml
apiVersion: someresource
kind: something
metadata:
  name: one
spec:
  image: imageone:latest
---
apiVersion: someotherresource
kind: something_else
metadata:
  name: two
spec:
  image: imagetwo:latest
```

akihikokuroda: I don't think so.  We have to create the dynamic client for each type of the resource.  We can use the structured client for non-CRD resources.

PR #641, opened 2023-06-05
Use ICR for client tests

psschwei:
### Summary

Fix failing client tests, ex: https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5178072985


psschwei:
hang on, that picked up changes I didn't want...

PR #642, opened 2023-06-05
Update client tests to use ICR images

psschwei:
### Summary

Fix failing client tests, ex: https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5178072985


IceKhan13 (approved):
ooooh! that what it is :)

PR #648, opened 2023-06-07
Create CI workflow for running a notebook test in notebook docker environment

caleb-johnson:
### Summary
This PR adds a check to the CI which loads up the full docker profile and runs a notebook test ( `pytest --nbmake ...` ) on all of our notebooks in the `docs` dir against the running `qr-jupyter` container.

### Details and comments

 - This is a lengthy test, so it will only run on pushes to main and/or changes to docker-compose.yml/.dockerignore
 - I have these tests enabled on this PR, but I will enable them only for main branch before merging
 
- [ ] Need to debug why we're getting a  gateway error in notebook tests.


.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,16 @@
...
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - name: Run the full docker profile
+        run: docker compose --profile full up -d
+      - name: Test notebooks in the docker environment
+        shell: bash
+        run: docker exec qs-jupyter "bash" "-c" "pip install nbmake pytest && pytest --nbmake /home/jovyan/serverless/"
```
caleb-johnson, line 16: @psschwei any idea why the gateway doesnt seem to be working in the notebooks when I run these tests?

The gateway container should be up and available from line 13, I'm able to pip install the packages and run the nbmake command fine, the notebooks just fail bc they cant access gateway host

psschwei: I'm hitting a similar issue in reviewing #640 , still investigating

caleb-johnson:
This PR is in a weird state from debugging. going to submit a fresh PR

PR #665, opened 2023-06-08
Rename gateway_scheduler component to scheduler

psschwei:
### Summary

This one's more of a nit, but having components named `gateway` and `gateway-scheduler` make using autocomplete for names a little slower: `G <tab> S` or `G <tab> <what does the hash string start with for the gateway pod>` instead of just `G <tab>` or `S <tab>`. 

So, in this PR we rename the scheduler component to just scheduler, in both docker compose and helm.

### DIFF
--- a/docker-compose-dev.yml
+++ b/docker-compose-dev.yml
@@ -97,8 +97,8 @@ services:
     depends_on:
       - keycloak
       - postgres
-  gateway_scheduler:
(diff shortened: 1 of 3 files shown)


Tansito (approved):
For me it has a lot of sense, thanks @psschwei 👍

pacomf (approved):

Issue #664, opened 2023-06-08
gateway-volume is not writable in Kind


akihikokuroda:
### Steps to reproduce the problem
Run 01_running_program in the Kind cluster

### What is the current behavior?
500 error is reported in the Gateway pod

### What is the expected behavior?
It should run without error.

akihikokuroda:
The hostPath
```
path: "/mnt/data/tmp"
```
is not writeable by the user 1001, group 0

So far I could make this work in Kind by changing the template/pvcs.yaml

1. removing persistentVolume gateway-volume 
2. Change the `storageClassName` to `standard` in PersistentVolumeClaim gateway-claim
3. Change the `accessMode` to `ReadWriteOnce`  in PersistentVolumeClaim gateway-claim

I have to work on to make changes work on other platforms.

psschwei:
Confirming this works for me too on Kind :partying_face:

PR #667, opened 2023-06-08
use default storageclass for kind

akihikokuroda:
### Summary
fix: #664 

Use default storageclass when `platform` is set to `kind`

### Details and comments
The default storageclass in the kind cluster is `standard`.  It works without issue.  When the chart is install in a kind cluster, an additional value `platform` must be set.  So the helm command for kind is like this:
```
helm install quantumserverless --set platform=kind https://github.com/Qiskit-Extensions/quantum-serverless/releases/download/v0.1.2/quantum-serverless-0.1.2.tgz
```


psschwei (approved):

psschwei:
Note: this also works for minikube, so in theory we could use platform=local (probably should've checked minikube before approving, but c'est la vie... :laughing: )

PR #668, opened 2023-06-09
IBM deployment fixes

Tansito:
### Summary

We are preparing a deployment in IBM for this project. In our tests with discovered a couple of issues related with our current configuration that this PR tries to mitigate. List of changes:

- New `pvc` to be used with a `cos` instance
- Fixed `deployment` configuration for `gateway` to work with `IQP`
- New entity in values to start working with `cos`

### DIFF
--- a/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
@@ -69,22 +69,22 @@ spec:
               value: {{ .Values.application.siteHost | quote }}
             - name: RAY_HOST
               value: {{ .Values.application.rayHost | quote }}
-            - name: CLIENT_ID
-              value: {{ .Values.application.keycloak.clientId | quote }}
             - name: SETTINGS_AUTH_MECHANISM
               value: {{ .Values.application.auth.mechanism | quote }}
             - name: SETTINGS_TOKEN_AUTH_URL
-              value: {{ .Values.application.token.url | quote }}
+              value: {{ .Values.application.auth.token.url | quote }}
             - name: SETTINGS_TOKEN_AUTH_VERIFICATION_URL
-              value: {{ .Values.application.token.verificationUrl | quote }}
+              value: {{ .Values.application.auth.token.verificationUrl | quote }}
             - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
-              value: {{ .Values.application.token.verificationField | quote }}
+              value: {{ .Values.application.auth.token.verificationField | quote }}
+            - name: CLIENT_ID
+              value: { { .Values.application.auth.keycloak.clientId | quote } }
             - name: SETTING_KEYCLOAK_URL
-              value: {{ .Values.application.keycloak.url | quote }}
+              value: {{ .Values.application.auth.keycloak.url | quote }}
             - name: SETTING_KEYCLOAK_REALM
-              value: {{ .Values.application.keycloak.realm | quote }}
+              value: {{ .Values.application.auth.keycloak.realm | quote }}
             - name: SETTINGS_KEYCLOAK_CLIENT_SECRET
-              value: {{ .Values.application.keycloak.clientSecret | quote }}
+              value: {{ .Values.application.auth.keycloak.clientSecret | quote }}
             - name: DJANGO_SUPERUSER_USERNAME
               value: {{ .Values.application.superuser.username | quote }}
             - name: DJANGO_SUPERUSER_PASSWORD
--- a/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
@@ -78,7 +78,7 @@ spec:
             - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
               value: {{ .Values.application.auth.token.verificationField | quote }}
             - name: CLIENT_ID
-              value: { { .Values.application.auth.keycloak.clientId | quote } }
+              value: {{ .Values.application.auth.keycloak.clientId | quote }}
             - name: SETTING_KEYCLOAK_URL
               value: {{ .Values.application.auth.keycloak.url | quote }}
             - name: SETTING_KEYCLOAK_REALM
--- a/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
@@ -71,6 +71,14 @@ spec:
               value: {{ .Values.application.rayHost | quote }}
             - name: CLIENT_ID
               value: {{ .Values.application.keycloak.clientId | quote }}
+            - name: SETTINGS_AUTH_MECHANISM
+              value: {{ .Values.application.auth.mechanism | quote }}
+            - name: SETTINGS_TOKEN_AUTH_URL
+              value: {{ .Values.application.token.url | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_URL
+              value: {{ .Values.application.token.verificationUrl | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
+              value: {{ .Values.application.token.verificationField | quote }}
             - name: SETTING_KEYCLOAK_URL
               value: {{ .Values.application.keycloak.url | quote }}
             - name: SETTING_KEYCLOAK_REALM
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -52,10 +52,10 @@ serviceAccount:
 
 podAnnotations: {}
 
-podSecurityContext: {}
-#  fsGroup: 1001
-#  runAsUser: 1001
-#  runAsGroup: 1001
+podSecurityContext:
+  fsGroup: 1001
+  runAsUser: 1001
+  runAsGroup: 1001
 #  fsGroupChangePolicy: "OnRootMismatch"
 
 securityContext: {}
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -9,11 +9,9 @@ application:
   debug: 0
   siteHost: "http://127.0.0.1:8000"
   rayHost: "http://ray:8265/"
-  keycloak:
-    clientId: "quantumserverless"
-    url: "http://keycloak:8080/"
-    realm: "quantumserverless"
-    clientSecret: "secret"
+  auth:
+    token: {}
+    keycloak: {}
   superuser:
     username: "admin"
     password: "passw0rd"
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -53,10 +53,10 @@ serviceAccount:
 
 podAnnotations: {}
 
-podSecurityContext:
-  fsGroup: 1001
-  runAsUser: 1001
-  runAsGroup: 1001
+podSecurityContext: {}
+#  fsGroup: 1001
+#  runAsUser: 1001
+#  runAsGroup: 1001
 #  fsGroupChangePolicy: "OnRootMismatch"
 
 securityContext: {}
--- a/infrastructure/helm/quantumserverless/templates/pvcs.yaml
+++ b/infrastructure/helm/quantumserverless/templates/pvcs.yaml
@@ -1,4 +1,6 @@
-{{ if .Values.cosEnable }}
+{{ if eq .Values.platform "kind" }}
+{{end}}
+{{ if eq .Values.platform "ibm" }}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
--- a/infrastructure/helm/quantumserverless/templates/pvcs.yaml
+++ b/infrastructure/helm/quantumserverless/templates/pvcs.yaml
@@ -1,28 +1,21 @@
-apiVersion: v1
-kind: PersistentVolume
-metadata:
-  name: gateway-volume
-  labels:
-    type: local
-  annotations:
-    pv.beta.kubernetes.io/gid: "1001"
-spec:
-  storageClassName: manual
-  capacity:
-    storage: 2Gi
-  accessModes:
-    - ReadWriteMany
-  hostPath:
-    path: "/mnt/data/tmp"
----
-apiVersion: v1
+{{ if .Values.cosEnable }}
 kind: PersistentVolumeClaim
+apiVersion: v1
 metadata:
   name: gateway-claim
+  annotations:
+    ibm.io/auto-create-bucket: "false"
+    ibm.io/auto-delete-bucket: "false"
+    ibm.io/bucket: {{ .Values.cos.bucket | quote }}
+    ibm.io/quota-limit: "false"
+    ibm.io/endpoint: {{ .Values.cos.endpoint | quote }}
+    ibm.io/tls-cipher-suite: "default"
+    ibm.io/secret-name: {{ .Values.cos.secretName | quote }}
 spec:
-  storageClassName: manual
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: 1Gi
+      storage: 10Gi
+  storageClassName: {{ .Values.cos.storageClassName | quote }}
+{{ end }}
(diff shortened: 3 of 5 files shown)


infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
```diff
@@ -69,14 +69,22 @@ spec:
...
+              value: {{ .Values.application.auth.mechanism | quote }}
+            - name: SETTINGS_TOKEN_AUTH_URL
+              value: {{ .Values.application.auth.token.url | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_URL
+              value: {{ .Values.application.auth.token.verificationUrl | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
+              value: {{ .Values.application.auth.token.verificationField | quote }}
             - name: CLIENT_ID
-              value: {{ .Values.application.keycloak.clientId | quote }}
+              value: { { .Values.application.auth.keycloak.clientId | quote } }
```
akihikokuroda, line 81: Is `{ {` instead of `{{` OK?

infrastructure/helm/quantumserverless/templates/pvcs.yaml
akihikokuroda, line 1: Would you make this work for local setup (kind, Minikube, and etc)?

infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
Tansito: Ups, no, a typo haha. Thanks Aki!

infrastructure/helm/quantumserverless/templates/pvcs.yaml
Tansito: Sadly not yet. But the current configuration doesn't work either in the cloud.

Tansito: In fact this PVC configuration is very specific for IBM Cloud.

psschwei: I think if we merge #667 first and then rebase here we'll be mostly covered. (If we've got kind working, I think that's probably good enough for local dev...)

infrastructure/helm/quantumserverless/charts/gateway/values.yaml
```diff
@@ -52,10 +52,10 @@ serviceAccount:
 
 podAnnotations: {}
 
-podSecurityContext: {}
-#  fsGroup: 1001
-#  runAsUser: 1001
-#  runAsGroup: 1001
+podSecurityContext:
```
psschwei, line 55: Did we need this change for COS?

infrastructure/helm/quantumserverless/templates/pvcs.yaml
```diff
@@ -1,28 +1,21 @@
-apiVersion: v1
```
psschwei, line 1: I wonder if we can do something like this here:

```yaml
{{ if .Values.platform "cos" }}
kind: PersistentVolumeClaim
...
{{ end }}
{{ if eq .Values.platform "kind" }}
kind: PersistentVolumeClaim)     
...
{{ end }}
```

infrastructure/helm/quantumserverless/templates/pvcs.yaml
```diff
@@ -1,28 +1,21 @@
...
+    ibm.io/secret-name: {{ .Values.cos.secretName | quote }}
 spec:
-  storageClassName: manual
   accessModes:
-    - ReadWriteMany
+    - ReadWriteOnce
   resources:
     requests:
-      storage: 1Gi
+      storage: 10Gi
```
IceKhan13, line 19: Let's move this as variable too, for future deployment we will need more probably :)

IceKhan13:
hmmmm, what happened with tests? 🤔

IceKhan13:
let's also fix image pull policy here (I think it is set to Always right now)

psschwei:
> hmmmm, what happened with tests?

tests pass on main, though the unauthorized message pops up there as well...

```
~/quantum/quantum-serverless/gateway on  main via 🐍 v3.9.16 (venv) 
$ tox -epy39
py39: install_deps> pip install -U -r requirements.txt -r requirements-dev.txt
py39: commands[0]> pip check
No broken requirements found.
py39: commands[1]> python manage.py test
Using selector: EpollSelector
Found 12 test(s).
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
..Unauthorized: /api/v1/jobs/
..GET http://test_user-head-svc:8265/ 200
.....Unauthorized: /api/v1/programs/
...
----------------------------------------------------------------------
Ran 12 tests in 3.601s

OK
Destroying test database for alias 'default'...
  py39: OK (33.63=setup[27.66]+cmd[0.76,5.21] seconds)
  congratulations :) (33.75 seconds)
```

infrastructure/helm/quantumserverless/charts/gateway/values.yaml
Tansito: Agree, for COS probably not.

infrastructure/helm/quantumserverless/templates/pvcs.yaml
Tansito: Yeah, to try to introduce @akihikokuroda changes. What do you think about a logic like: `eq "kind",  not eq "kind"` (thinking that `platform: cos` sounds a bit weird?, maybe my impression). I'm assuming that in the cloud you are going to use `cos` and in local not.

psschwei: only issue there would be if you wanted to add a third/fourth/etc option (e.g. AWS)... maybe `platform: ibm`?

IceKhan13:
Tests should be working fine now :) 
Fixed in https://github.com/Qiskit-Extensions/quantum-serverless/pull/669/commits/3d7b0c7cd04ac2ae1c997b24eaf3797ecce64916

pacomf:
@Tansito can you resolve the conflicts?

infrastructure/helm/quantumserverless/templates/pvcs.yaml
Tansito: You're right and I think that would work because probably configurations can be different between platforms too. Thanks @psschwei !

akihikokuroda: @Tansito I think that the default case (not, kind or ibm) is missing.  Putting this just above the last {{ end }} seems good.
```
{{ else }}                                                                                                                                                                                                                                    
apiVersion: v1                                                                                                                                                                                                                                
kind: PersistentVolume                                                                                                                                                                                                                        
metadata:                                                                                                                                                                                                                                     
  name: gateway-volume                                                                                                                                                                                                                        
  labels:                                                                                                                                                                                                                                     
    type: local                                                                                                                                                                                                                               
  annotations:                                                                                                                                                                                                                                
    pv.beta.kubernetes.io/gid: "1001"                                                                                                                                                                                                         
spec:                                                                                                                                                                                                                                         
  storageClassName: manual                                                                                                                                                                                                                    
  capacity:                                                                                                                                                                                                                                   
    storage: 2Gi                                                                                                                                                                                                                              
  accessModes:                                                                                                                                                                                                                                
    - ReadWriteMany                                                                                                                                                                                                                           
  hostPath:                                                                                                                                                                                                                                   
    path: "/mnt/data/tmp"                                                                                                                                                                                                                     
---                                                                                                                                                                                                                                           
apiVersion: v1                                                                                                                                                                                                                                
kind: PersistentVolumeClaim                                                                                                                                                                                                                   
metadata:                                                                                                                                                                                                                                     
  name: gateway-claim                                                                                                                                                                                                                         
spec:                                                                                                                                                                                                                                         
  storageClassName: manual                                                                                                                                                                                                                    
  accessModes:                                                                                                                                                                                                                                
    - ReadWriteMany                                                                                                                                                                                                                           
  resources:                                                                                                                                                                                                                                  
    requests:                                                                                                                                                                                                                               
      storage: 1Gi  
```

pacomf (approved):

infrastructure/helm/quantumserverless/templates/pvcs.yaml
Tansito: Thanks Aki, Iskandar. Solved in: [3e272e5](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/3e272e52cc0c352ef6bf3bac0ced525f24ab84b0)

infrastructure/helm/quantumserverless/templates/pvcs.yaml
Tansito: Solved in: [90e5295](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/90e52951b4cc79aa90d40a72f365f28771e28dff), [92f3d6e](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/92f3d6efc583c4018136a9d949929d605435a4df)

infrastructure/helm/quantumserverless/charts/gateway/values.yaml
Tansito: Solved in [e2029a9](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/e2029a9d291cd2e871a2356cb4ca838252e177b4)

infrastructure/helm/quantumserverless/values-ibm.yaml
```diff
@@ -73,7 +67,7 @@ gateway:
 # Redis configs
 # ===================
 
-redisEnable: true
+redisEnable: false
```
psschwei, line 70: nit: in theory, we could delete all the other config options for components we disable (not sure if that's something you'd want to do though)

psschwei (changes requested):
Something's broken in the non IBM deploy ...

```
Error: INSTALLATION FAILED: template: quantum-serverless/charts/gateway/templates/deployment.yaml:75:31: executing "quantum-serverless/charts/gateway/templates/deployment.yaml" at <.Values.application.auth.token.url>: nil pointer evaluating interface {}.url
```

infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
```diff
@@ -67,14 +67,24 @@ spec:
...
+            - name: RAY_HOST
+              value: {{ .Values.application.rayHost | quote }}
+            - name: SETTINGS_AUTH_MECHANISM
+              value: {{ .Values.application.auth.mechanism | quote }}
+            - name: SETTINGS_TOKEN_AUTH_URL
+              value: {{ .Values.application.auth.token.url | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_URL
+              value: {{ .Values.application.auth.token.verificationUrl | quote }}
+            - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
+              value: {{ .Values.application.auth.token.verificationField | quote }}
```
psschwei, line 79: may need to wrap these in an `if platform == ibm` block

Tansito: mmm... Initially I would say that maybe you're right but it's a bit tricky because why should not we apply same logic for `keycloack`? Or even you want to use both methods? Thinking out loud, maybe we need to start reviewing our configuration.

infrastructure/helm/quantumserverless/values-ibm.yaml
Tansito: yep, for me it sounds good 👍

infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
Tansito: But as first approach to solve the problem maybe we can enable one method or another if we use or not `keycloack`, for example.

psschwei: keycloak uses different fields. for example, there's no `Values.application.auth.token` ... for keycloak it's `Values.application.auth.keycloak`. that's causing the non-ibm deploy to fail because none of the token values are set (so they're `nil` which throws an error). so we probably have to use the platform to select which value to set for these envvars, depending on the install platform (local / ibm / etc.)

Tansito: But I mean, it's the same for `token`, isn't it? There is no `Values.application.auth.keycloack` for `token`.

psschwei: I'd assume so... I only tested with a local install, but I'd assume the same thing would happen for missing keycloak values when deploying on IBM Cloud

Tansito: Weird, for me is working and locally there were never `token` values 🤔. Does this happen for you too, @akihikokuroda ?

psschwei: I got it to deploy successfully by adding `application.auth.token.url: "nil"` to `values.yaml`, so it seems like it's just that one field that's misbehaving...

psschwei: actually, the issue is the missing `token` ... that's what's getting rendered as `nil` and causing the failure

something like this would be totally fine:

```yaml
auth:
  keycloak:
    stuff...
  token: {}
```

(and probably vice-versa for the IBM config)

Tansito: Sounds good to me, I don't understand yet why in one way was happening but in another way not 🤷 . Thanks for solving it anyway, @psschwei !

psschwei: ok, I'm running around in circles here.... 

the default values for the gateway chart set the keycloak stuff, that's why those aren't failing on IBM Cloud 

if we set the token values in that chart (`helm/quantumserverless/charts/gateway/values.yaml`), then this all goes away and no need to add a blank `token: {}` to the main `values-ibm.yaml`

actually, maybe we should just change the gateway defaults to:

```yaml
application:
  keycloak: {}
  token: {}
```

and then set the real values when deploying to a specific environment...

Tansito: Solved with [2e320c0](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/2e320c01c338b77f81e2cc4ffd9c57d14005ae62)!

infrastructure/helm/quantumserverless/values-ibm.yaml
Tansito: Something like this @psschwei ?: [b461c11](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/b461c1118c5258b0b77debb7f748808e2460e0cd)

psschwei: yes!

infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml
Tansito: And solved last comment with [63eef02](https://github.com/Qiskit-Extensions/quantum-serverless/pull/668/commits/63eef020a174a570251118c40cc1c52d2046d73a), thank you for the good catch up @psschwei !

infrastructure/helm/quantumserverless/values.yaml
```diff
@@ -35,18 +35,20 @@ gatewayEnable: true
...
   application:
     siteHost: "http://gateway:8000"
     rayHost: "http://kuberay-head-svc:8265"
-    keycloak:
-      clientId: "gateway-client"
-      url: "http://keycloak:31059"
-      realm: "quantumserverless"
-      clientSecret: GATEWAYSECRET-CHANGEME
+    auth:
+      token: {}
```
psschwei, line 46:
```suggestion
```

handled in default gateway values

infrastructure/helm/quantumserverless/values-ibm.yaml
```diff
@@ -48,11 +42,13 @@ gateway:
...
-      url: "http://keycloak:31059"
-      realm: "quantumserverless"
-      clientSecret: GATEWAYSECRET-CHANGEME
+    auth:
+      mechanism: "custom_token"
+      token:
+        url: "https://auth-dev.quantum-computing.ibm.com/api/users/loginWithToken"
+        verificationUrl: "https://auth-dev.quantum-computing.ibm.com/api/users/me"
+        verificationField: "ibmQNetwork"
+      keycloak: {}
```
psschwei, line 51:
```suggestion
```

we should be able to drop this now

psschwei (approved):
two minor nits, but otherwise LGTM

PR #669, opened 2023-06-09
Gateway: compute resources error fallback

IceKhan13:
### Summary

Gateway: kill compute resource if something went wrong during creation

### DIFF
--- a/gateway/api/ray.py
+++ b/gateway/api/ray.py
@@ -58,7 +58,7 @@ def submit_ray_job(job: Job) -> Job:
 
 def create_ray_cluster(
     user: Any, cluster_name: Optional[str] = None
-) -> ComputeResource:
+) -> Optional[ComputeResource]:
     """Creates ray cluster.
 
     Args:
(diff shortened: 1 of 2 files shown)


gateway/api/ray.py
```diff
@@ -203,20 +206,21 @@ def wait_for_cluster_ready(cluster_name: str):
     url = f"http://{cluster_name}-head-svc:8265/"
     success = False
     attempts = 0
+    max_attempts = 120
```
psschwei, line 209: should we make this configurable?

IceKhan13:
I'm also wondering can we get something like a readiness probe from cluster/pod instead of just calling head node 🤔 
Not for this at this moment, but just a general curiosity

IceKhan13:
@psschwei I found this flick in tests fail for gateway https://github.com/Qiskit-Extensions/quantum-serverless/pull/669/commits/3d7b0c7cd04ac2ae1c997b24eaf3797ecce64916

`random.choice` vs `random.sample` sneaky bastard! :)

psschwei (approved):
Logic makes sense.

Long-term, we probably need to figure out how to test this.

PR #671, opened 2023-06-09
Remove deprecated electronic structure problem example

### LINKED ISSUE #661: ElectronicStructureProblem tutorial returns empty result
- build and run the `jupyter` docker profile
- open jupyter lab on localhost
- run serverless/examples/06_electronic_structure_problem.ipynb

The result is an empty dictionary and the following code block fails because it expects a populated result

The result should be populated, and the following code block should not error out

caleb-johnson:
Resolves #661 

This PR removes the electronic structure problem notebook from `docs/development/examples`. In addition to it being problematic in the CI, the code in the source file for this tutorial is badly deprecated and needs a significant refactor.

Let's remove this tutorial in order to:
 - remove a heavy dependency Nature
 - clean up our CI a bit by not having to deal with this deprecated notebook

Note: I also snuck in a change which updates the quckstart to include cloning of the repo as an optional step. I also switched installation of Docker to a required step, since it is required to complete the quickstart. cc: @akihikokuroda


psschwei (commented):
LGTM but will let @IceKhan13 or @pacomf give final sign off

pacomf (approved):

PR #593, opened 2023-05-24
Update docs

caleb-johnson:
This PR addresses the specific feedback currently in #511 

I can revert any change or add anything we think, I just wanted to open this PR to address those specific items

PR #673, opened 2023-06-11
Create new tutorial showing how to access old jobs. Show logs in two …

caleb-johnson:
…tutorials.

### Summary
We trimmed up tutorial 1 in #593 and lost any mention of printing logs or accessing past jobs through the client.

This PR introduces two changes:
 - Introduces a new tutorial 5 showing how to retrieve jobs through the client using job ID
 - Adds a couple of small print statements in tutorial one to demonstrate job logs
 - Adds a short cell in tutorials 1 and 4 showing how to retrieve logs.
     - The logs will be very short (two lines from program 1), so this won't clutter either notebook.

### DIFF
--- a/docs/running/notebooks/source_files/program_1.py
+++ b/docs/running/notebooks/source_files/program_1.py
@@ -3,6 +3,7 @@ from qiskit.primitives import Sampler
 
 from quantum_serverless import save_result
 
+print("Running program 1...")
 circuit = QuantumCircuit(2)
 circuit.h(0)
 circuit.cx(0, 1)
@@ -14,3 +15,4 @@ sampler = Sampler()
 quasi_dists = sampler.run(circuit).result().quasi_dists
 
 save_result(quasi_dists)
+print("Completed running program 1.")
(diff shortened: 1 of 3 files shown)


docs/running/notebooks/source_files/program_1.py
```diff
@@ -3,6 +3,7 @@
 
 from quantum_serverless import save_result
 
+print("Running program 1...")
```
pacomf, line 6: Should you remove the prints here? :)

caleb-johnson: I actually added these so we would have something short to show in the logs when we show the logs in the tutorials.

I could use the logging package, or I could leave these print statements. What do you think?

pacomf: @akihikokuroda @IceKhan13 do we have some logging package used in other samples? Should we maintain the simple print?

IceKhan13: as long as it spits to stdio :) 
print is fine, anyway most of the users will do that

Issue #596, opened 2023-05-25
Enable TLS among Ray components


akihikokuroda:
### What is the expected enhancement?
Enable TLS for communication among Ray components.  It is recommended to enable TLS for communications among the components in cluster.  
Ref: https://github.com/ray-project/kuberay/blob/master/docs/guidance/tls.md

akihikokuroda:
Ref: https://github.com/ray-project/ray/pull/33737

PR #678, opened 2023-06-12
Enable gRPCTLS w/o cert manager

akihikokuroda:
### Summary
This is part of #596 
This implementation follows [Ray guide](https://github.com/ray-project/kuberay/blob/master/docs/guidance/tls.md).  

### Details and comments
I'll work for the better solution with the [cert manager](https://cert-manager.io/) in the follow up PR.


pacomf (approved):

psschwei (approved):
FWIW, on my machine it took about 25s for the ray head node to become active

Also, in the scheduler logs it's still showing the `http` address... is that something else we'll need to update?

akihikokuroda:
@psschwei Does the head node start take 25s every time?  
This change seems the communication among the ray nodes.  The request from the ray node is not included.  We have to investigate some more about it.

psschwei:
I didn't check multiple runs... I think it may have been cert creation that ate up some time too

PR #681, opened 2023-06-14
Issue 680 | Client: programs widget

### LINKED ISSUE #680: Client: add list widget

Add jupyter widget to list executed programs.

IceKhan13:
### Summary

Adds programs widget

### Details and comments

Closes #680


IceKhan13:
@pacomf here is a minimal implementation. No extra dependencies, pure ipywidgets :) 

Screenshot 2023-06-15 at 12 52 25 PM

pacomf:
Good! Can we add tabs to show the status and other interesting thing without click in the item?

IceKhan13:
you mean add status to each element of the list when collapsed? 
What other interesting things you want? 😄

pacomf:
Yes, in columns:

Name | Program Id | Status | Creation Date

Something like that (with an extra row in the header to name each column)

IceKhan13:
Added status and date, but I'm not sure we need header as it is eating up space.
If we want header, than it's probably better to move to table view, instead of accordion list, which will require to write an ugly html things, which we do not need at this moment IMO. We can always prettify later if needed. 

Screenshot 2023-06-15 at 1 47 29 PM

But if we really need it we can go that route. It will require more work, though

pacomf:
For the first PoC, i think that it can work...

image

IceKhan13:
oh, it's actually very minimal too. Let's do exactly the same thing, to make it consistent, no problems :)

IceKhan13:
Screenshot 2023-06-15 at 4 14 38 PM

Screenshot 2023-06-15 at 3 57 22 PM

pacomf:
here we go! :)

pacomf (approved):
LGTM!

PR #686, opened 2023-06-16
Add retry logic in Gateway for Ray client set up

### LINKED ISSUE #682: Add retry logic in Gateway for Ray client set up
From discussion in #679 :

We saw a number of failures running programs on Kind when we added PVCs to the Ray clusters. The failures are happening in the [`submit_ray_job()`](https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98fad67bebc5f98f47ba988eb19abb3ac10a0e7/gateway/api/ray.py#L25) function.

There's actually 2 places where things can fail:

1. [When we set up the `JobSubmissionClient`](https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98fad67bebc5f98f47ba988eb19abb3ac10a0e7/gateway/api/ray.py#L34)

Seems like the JobSubmissionClient tries to make a connection to the Ray cluster, but if that fails it throws a `NewConnectionError`.

2. [When using the ray client to submit the job](https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98fad67bebc5f98f47ba988eb19abb3ac10a0e7/gateway/api/ray.py#L43)

I didn't copy an example of the failure here, but it does fail.

In both cases, the issue is that a single failure appears to kill the `submit_ray_job()` function, which throws the SIGTERM that kills the ray cluster, which then spins up a new ray cluster. Rinse / lather / repeat.

If instead of failing we retried both of those calls a few times, then the job will succeed. Based on my limited testing, usually less than 5 retries is enough.

That's a massive speedup (I've seen upwards of 30 ray clusters fail on kind, which if you figure each one takes 20s means six minutes of thrashing).

Here's my hacked function with a fix for retrying (and some print statements to tell where I am when things fail):

```python
def submit_ray_job(job: Job) -> Job:
    """Submits job to ray cluster.

    Args:
        job: gateway job to run as ray job

    Returns:
        submitted job
    """
    print("++++++++++ let's go!")
    success = False
    err = "failed"
    runs = 0
    while not success:
        print(f"trying to set up job client x{runs}")
        runs += 1
        try:
            ray_client = JobSubmissionClient(job.compute_resource.host)
            success = True
        except:
            print("client setup failed")
            time.sleep(1)
        if runs > 30:
            print("unable to set up job submission client")
            success = True
            raise Exception("we failed")

    print("job submission client setup")
    program = job.program

    _, dependencies = try_json_loads(program.dependencies)
    print("trying to load json stuff")
    with tarfile.open(program.artifact.path) as file:
        extract_folder = os.path.join(settings.MEDIA_ROOT, "tmp", str(uuid.uuid4()))
        file.extractall(extract_folder)
    print("++++++++++ artifact got")

    entrypoint = f"python {program.entrypoint}"
    print("++++++++++ submitting job")
    success = False
    err = "failed"
    runs = 0
    while not success:
        runs += 1
        if runs > 30:
            success = True
            break
        print("trying job submission")
        try:
            ray_job_id = ray_client.submit_job(
                entrypoint=entrypoint,
                runtime_env={
                    "working_dir": extract_folder,
                    "env_vars": json.loads(job.env_vars),
                    "pip": dependencies or [],
                },
            )
            print("job succeeded!")
            err = None
        except:
            print("job failed")
            err = "failed"
        if err is None:
            print("mark success true")
            success = True

    print("++++++++++ job submitted ++++++++++")
    job.ray_job_id = ray_job_id
    job.status = Job.PENDING
    job.save()

    if os.path.exists(extract_folder):
        shutil.rmtree(extract_folder)

    return job
```

Given that a lot of things happen async on Kubernetes, it's probably worth getting some kind of retry logic in and/or revisiting our readiness / liveness probes for ray clusters...

psschwei:
### Summary

Add retry logic in Gateway for Ray client set up
Fixes #682 

### Details and comments

Also adds a fix to the ray cluster template and an optimization for improving local incremental build speeds for gateway.


IceKhan13 (approved):
Let's fix linter and we are good! We are close to 0.2.0! 

Looks like it's logging formatting
```python
logging.info("bla-lba %s", var)
```
instead of 
```python
logging.info(f"bla-lba {var}")
```

PR #687, opened 2023-06-16
use cert-manager for certification management

akihikokuroda:
### Summary
fix: #596 
 
This change integrate [cert-manager](https://cert-manager.io/) for certificate management of Ray gRPC communication.

### Details and comments

The use of cert-manager is disabled by default.   It can be enable by setting  `gateway.useCertManager` to `true` in the values.yaml file.

The cert-manage can be installed by 
```
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.12.0/cert-manager.yaml
```


psschwei:
For the `gateway/api/ray.py:200` warning, you can add this comment to the line to get the linter to ignore: `# pylint: disable=broad-exception-caught`

For the `gateway/api/ray.py:223:4` warning, I think you can either rename to something using `snake_case` OR tell pylint to ignore it via a variable in the `.pylintrc` (I did it for a single letter variable in a different PR, but not having luck finding it right now)

Will do a proper review on Monday

psschwei:
found it... the `.pylintrc` key is `good-names`

psschwei (approved):
LGTM

Could you also add a bit on enabling cert-manager to the readme?

pacomf (approved):

PR #702, opened 2023-06-21
Change the labels of scheduler deployment and pod

### LINKED ISSUE #701: Can not set port forward for gateway service
1.Install quantum middleware via helm
2. set port forward for gateway service
```
kubectl port-forward service/gateway 3333:8000   
```
The creation of the port forward fails with the following error.
> error: Pod 'scheduler-8465cc9d66-7lf6n' does not have a named port 'http'

Port forward created correctly

akihikokuroda:
### Summary
fix #701 

Create an unique labels for scheduler deployment and pod

### Details and comments
The scheduler deployment and pod are using the same labels as the gateway.  It causes the gateway service attached to both the gateway and scheduler pods.  The gateway service should only be attached to the gateway pods.

### DIFF
--- a/infrastructure/helm/quantumserverless/charts/gateway/templates/_helpers.tpl
+++ b/infrastructure/helm/quantumserverless/charts/gateway/templates/_helpers.tpl
@@ -41,12 +41,25 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


psschwei:
@akihikokuroda you should be able to close / reopen this PR to pick up the fixes in #703 (without needing to rebase etc.)

akihikokuroda:
@psschwei Oh, I didn't know the trick.  I rebased this.

psschwei:
no worries, it does basically the same thing (closing/reopening basically just resets what your diffs are diffing against)

PR #697, opened 2023-06-21
Client: remove deprecated functionality

IceKhan13:
### Summary

Remove deprecated functionality in client

### Details and comments

All deprecation was set to be removed by 0.2.

So good to remove code, not to add :)

PR #703, opened 2023-06-21
fix lint error in client verify process

akihikokuroda:
### Summary
Fix lint error in client verify process

### DIFF
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -321,14 +321,6 @@ class KuberayProvider(Provider):
             available_compute_resources: available clusters in provider
         """
         super().__init__(name)
-        warnings.warn(
-            "`KuberayProvider` is deprecated "
-            "and will be removed in v0.3. "
-            "Please, consider using `GatewayProvider`.",
-            DeprecationWarning,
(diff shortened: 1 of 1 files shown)


client/quantum_serverless/core/provider.py
```diff
@@ -321,14 +321,6 @@ def __init__(
             available_compute_resources: available clusters in provider
         """
         super().__init__(name)
-        warnings.warn(
-            "`KuberayProvider` is deprecated "
-            "and will be removed in v0.3. "
```
psschwei, line 326: I don't think we should drop this before v0.3, probably better to have pylint ignore it via something like `# pylint: disable=undefined-variable`

akihikokuroda: @psschwei thanks for review.  It looks like this should have forgotten to  take out as a part of #697.

psschwei: good catch -- yes, we should add that import back

akihikokuroda: OK.  I'll do that.  Thanks.

psschwei (approved):
LGTM

PR #704, opened 2023-06-21
helm install doc change

akihikokuroda:
### Summary
Change the helm install instructions to install from the github release artifacts instead of the local clone of the repository.   It allows the user to install quantum middleware without cloning the repository and install any released version easily.

### Details and comments

This also change the example of the port forward from the pod IP to the service.  The service names are static so the user can always set port forward in the same way.


Tansito (approved):
Nice change @akihikokuroda ! <3

psschwei (approved):
@Tansito @akihikokuroda in the future, should we consider about adding a default ingress for the gateway so we could drop the port-forward step?

akihikokuroda:
@psschwei Yes, I agree.  Ingress is environment dependent so it may be a little tricky.

PR #717, opened 2023-06-23
Configure security context for gateway

Tansito:
### Summary

From our tests deploying the project with the new `pvc` structure we discovered that if we want use a`cos` with non-root access we need to configure the security context for the `pod` correctly.

This PR configures `gateway` chart with the user and group `1000:100` following the same configuration that we have in our Dockerfile. `fsGroup` and `runAsUser` with the same values.

### Details and comments

I decided to change these values internally and not through values because this configuration comes from the Dockerfile. Something that people will not be able to configure (easily at least).

- Info about PVC configuration for non-root users: https://cloud.ibm.com/docs/containers?topic=containers-storage-cos-understand
- Info about security context configuration: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
- Gateway's Dockerfile: https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/infrastructure/docker/Dockerfile-gateway#L20

### DIFF
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -58,9 +58,9 @@ serviceAccount:
 podAnnotations: {}
 
 podSecurityContext: 
+  fsGroup: 1000
   runAsUser: 1000
   runAsGroup: 100
-  fsGroup: 1000
 #  fsGroupChangePolicy: "OnRootMismatch"
 
 securityContext: {}
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -57,10 +57,10 @@ serviceAccount:
 
 podAnnotations: {}
 
-podSecurityContext: {}
-#  fsGroup: 1001
-#  runAsUser: 1001
-#  runAsGroup: 1001
+podSecurityContext: 
+  runAsUser: 1000
+  runAsGroup: 100
+  fsGroup: 100
 #  fsGroupChangePolicy: "OnRootMismatch"
 
 securityContext: {}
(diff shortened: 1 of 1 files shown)


infrastructure/helm/quantumserverless/charts/gateway/values.yaml
```diff
@@ -57,10 +57,10 @@ serviceAccount:
...
 podAnnotations: {}
 
-podSecurityContext: {}
-#  fsGroup: 1001
-#  runAsUser: 1001
-#  runAsGroup: 1001
+podSecurityContext: 
+  runAsUser: 1000
+  runAsGroup: 100
+  fsGroup: 100
```
IceKhan13, line 63:
```suggestion
  fsGroup: 1000
```

To access a file in IBM Cloud Object Storage with a non-root user, you must set the runAsUser and fsGroup values in your deployment to the **same** value.

Tansito: Thanks Iskandar!

psschwei:
With `fsgroup: 1000` jobs are failing for me locally.

Tansito:
Mmm... Are we going to need to configure the rest of `pv`? 🤔

akihikokuroda:
This work for me with RancherDesktop.

IceKhan13:
@psschwei probably it was failing due to lack of pod security context in scheduler https://github.com/Qiskit-Extensions/quantum-serverless/commit/eab9bcdc7afde0f2f2e4b6dba9b68abc0439fbae#diff-37f992b66d97d74cfef51a8baf2688d8e8e9bcb5ca0133cd6c0bfd06ecbf68cbR150

psschwei:
yeah, checking now

psschwei:
that seems to have been it

psschwei (approved):

PR #720, opened 2023-06-23
user smaller images for initcontainers

### LINKED ISSUE #716: Use smaller images for gateway init containers

from https://github.com/Qiskit-Extensions/quantum-serverless/pull/707#discussion_r1238534767

currently using the ray image, which is 2+ GB... something smaller, like busybox or distroless assuming they have the needed tools, might be more efficient

akihikokuroda:
### Summary
fix #716 

### Details and comments
I picked up images that support both amd64 and arm64 and have necessary packages.

alpine/k8s:1.27.3 - 230 mb
finalgene/openssh:9.1 - 12.5 mb

### DIFF
--- a/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/values.yaml
@@ -24,7 +24,7 @@ application:
     replicas: 1
     minReplicas: 1
     maxReplicas: 4
-    opensslImage: finalgene/openssh:9.1
+    opensslImage: registry.access.redhat.com/ubi8/openssl:8.8-3
     kubectlImage: alpine/k8s:1.27.3
   limits:
     maxJobsPerUser: 2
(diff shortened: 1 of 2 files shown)


psschwei:
I'm a little leery of `finalgene`  as that seems to be just some random person... not blocking, but just figured I'd mention that

akihikokuroda:
I agree.  I couldn't find a good one so far.   We can add the openssl to alpine but I don't want to maintain the image.

akihikokuroda:
I try `registry.access.redhat.com/ubi8/openssl:8.8-3`.

psschwei:
> I don't want to maintain the image

agree, though in a worst case we could add it to the build GHAs to trigger a rebuild on every release (or nightly if that would be better).

But the UBI image looks promising... I'll trade 4MB to trust the maintainer :smile:

akihikokuroda:
The ubi image works.

psschwei (approved):

PR #726, opened 2023-06-26
Issue 700 | Client: UX improvements

### LINKED ISSUE #700: Client: fail fast and UX enhancements

Improve UX of client:
- check artifact file size (under 2-5 Mb)
- when Gateway provider is created with token -> verify token
- check if entrypoint file exists

IceKhan13:
### Summary

Client: UX improvements
- [x] verify token
- [x] check artifact size
- [x] check entrypoint

### Details and comments

Closes #700


psschwei (commented):
blanket question: should we also log the errors (in addition to raising the exception)?

LGTM otherwise

IceKhan13:
@psschwei since it is client lib we do not need to log as we will not be able to access them. If this client will be used somewhere inside of other service, then this service will catch and log those errors

akihikokuroda (approved):
LGTM

Issue #733, opened 2023-06-30
ICR | build and push failing 


akihikokuroda:
### Steps to reproduce the problem
 execute: ICR | build and push action

### What is the current behavior?
ICR | build and push failing 
https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5421890580

### What is the expected behavior?
The action completes successfully

akihikokuroda:
ray-2.5.1-cp39-cp39-manylinux2014_aarch64.whl is missing in [here](https://pypi.org/project/ray/#files) 

2.5.0 has [ray-2.5.0-cp39-cp39-manylinux2014_aarch64.whl ](https://files.pythonhosted.org/packages/03/56/ede2adc089223c659055c84c16c2bfcab064d6113c59d4f8720c913f30a1/ray-2.5.0-cp39-cp39-manylinux2014_aarch64.whl)(54.7 MB [view hashes](https://pypi.org/project/ray/2.5.0/#copy-hash-modal-b0637aa1-97b7-417c-99a3-a9a50dd03ee9)) in [here](https://pypi.org/project/ray/2.5.0/#files)

akihikokuroda:
I opened an issue in Ray (https://github.com/ray-project/ray/issues/36991). Should we move back to 2.5.0?

Tansito:
I think we can close this and add the fix when they deploy a new version.

PR #734, opened 2023-07-03
Downgrade ray version to 2.5.0

Tansito:
### Summary

From #733 , we need to downgrade `ray` version to `2.5.0` to continue with other developments. Let's see how this problem advances.


Tansito:
The error comes from `ray`. They are using `pydantic` and in a recent version they introduced breaking changes. Ref: https://github.com/ray-project/ray/pull/37000

akihikokuroda:
@Tansito All images are built fine.  Thanks!

akihikokuroda (approved):
@Tansito LGTM.  Thanks!

PR #739, opened 2023-07-05
Prepull ray images to all cluster nodes

### LINKED ISSUE #728: Infrastructure: ray cluster image caching

Currently when we create ray cluster for the first time (and when cache is outdates) kuberentes is starting to fetch >2Gb image, which is taking roughly 90 seconds, which is not ideal. We need to think of a way how to cache docker images to speed up process.

psschwei:
### Summary

Fixes #728 
Add a daemonset to prepull the ray images to all nodes in the cluster.

### Details and comments

A [daemonset](https://kubernetes.io/docs/concepts/workloads/controllers/daemonset/) ensures that all nodes run a copy of a pod. Our daemonset is set up using the ray node (using the value from the gateway ray image key) as an init container and the google "pause" container as the main one (the "pause" container more or less just sleeps and has minimal footprint). By using a daemonset, we ensure that anytime a new node is added to the cluster, the ray image will be pulled to it.

### DIFF
--- a/infrastructure/helm/quantumserverless/templates/image-prepuller-daemonset.yaml
+++ b/infrastructure/helm/quantumserverless/templates/image-prepuller-daemonset.yaml
@@ -1,0 +1,20 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: prepuller
+spec:
+  selector:
+    matchLabels:
+      name: prepuller
+  template:
+    metadata:
+      labels:
+        name: prepuller
+    spec:
+      initContainers:
+         - name: prepuller-1
+           image: {{ .Values.gateway.application.ray.nodeImage }}
+           command: ["sh", "-c", "'true'"]
+      containers:
+        - name: pause
+          image: gcr.io/google_containers/pause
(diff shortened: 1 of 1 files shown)


infrastructure/helm/quantumserverless/templates/image-prepuller-daemonset.yaml
```diff
@@ -0,0 +1,20 @@
...
+      labels:
+        name: prepuller
+    spec:
+      initContainers:
+         - name: prepuller-1
+           image: {{ .Values.gateway.application.ray.nodeImage }}
+           command: ["sh", "-c", "'true'"]
+      containers:
+        - name: pause
+          image: gcr.io/google_containers/pause
```
akihikokuroda, line 20: The `latest` image doesn't have the arm64 arch format.  Would you add tag `3.2` that has multiple arch formats.

akihikokuroda:
It may be too much but I wonder if we want to enable **node selector** or **node affinity** in the values.yaml file in some environment.

psschwei:
I'd think I'd lean towards too much for now, unless there's a specific use case that needs it.

akihikokuroda (approved):
LGTM. Thanks!

PR #745, opened 2023-07-07
Execute CI with a new release

Tansito:
### Summary

Now that we are starting to release more often I wanted to improve our DX. With this PR the idea is that once time we publish a release all the required workloads run automatically, without an action from our part.

### DIFF
--- a/.github/workflows/icr-image-build-and-push.yml
+++ b/.github/workflows/icr-image-build-and-push.yml
@@ -6,6 +6,8 @@ on:
       tag:
         description: 'Tag for images'
         required: true
+  release:
+    types: [published]
   push:
     branches:
       - main
@@ -38,3 +40,8 @@ jobs:
         uses: ./.github/actions/icr-build-and-push-images
         with:
           tag: ${{ github.event.inputs.tag }}
+      - name: Build and push on release
+        if : github.event_name == 'release' && github.event.action == 'published'
+        uses: ./.github/actions/icr-build-and-push-images
+        with:
+          tag: ${{ github.event.release.tag_name }}
--- a/.github/workflows/icr-image-build-and-push.yml
+++ b/.github/workflows/icr-image-build-and-push.yml
@@ -40,8 +40,12 @@ jobs:
         uses: ./.github/actions/icr-build-and-push-images
         with:
           tag: ${{ github.event.inputs.tag }}
+      - name: Set version from tag
+        if : github.event_name == 'release' && github.event.action == 'published'
+        id: set-version-from-tag
+        run: echo "VERSION=${GITHUB_REF_NAME/v}" >> $GITHUB_OUTPUT
       - name: Build and push on release
         if : github.event_name == 'release' && github.event.action == 'published'
         uses: ./.github/actions/icr-build-and-push-images
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/icr-image-build-and-push.yml
```diff
@@ -38,3 +40,8 @@ jobs:
         uses: ./.github/actions/icr-build-and-push-images
         with:
           tag: ${{ github.event.inputs.tag }}
+      - name: Build and push on release
+        if : github.event_name == 'release' && github.event.action == 'published'
+        uses: ./.github/actions/icr-build-and-push-images
+        with:
+          tag: ${{ github.event.release.tag_name }}
```
Tansito, line 47: The main problem with this approach is that now our docker images instead to be 0.2.1 will be v0.2.1. Not a problem at all but it's a change. The only option that I found to fix it it was this one:

https://stackoverflow.com/questions/71630227/github-action-how-to-remove-starting-v-from-variable

But I didn't like at all. What do you think?

akihikokuroda: It is used in the helm packaging action here: https://github.com/Qiskit-Extensions/quantum-serverless/blob/9d72667d9e853c561c647907838a7169abd79cc6/.github/workflows/publish-helm.yaml#L23C51-L23C64

psschwei: I don't have a strong preference either way... it's probably better to leave the tags as is rather than introduce a change (though if we _were_ going to make a change, now is probably the time to do it), but it also could be nice if the image tags matched exactly to the git tags... 

If it's just the syntax that's bugging you, there's always  `sed`  :smile:

akihikokuroda: The helm version usually doesn’t have the “v” prefix.  The GitHub tag usually has the “v” prefix. The docker tag is free format :-).

Tansito: Yeah, I think we can find different formats in different platforms. I will change the code to use it, thanks team 😄

.github/workflows/icr-image-build-and-push.yml
```diff
@@ -38,3 +40,12 @@ jobs:
         uses: ./.github/actions/icr-build-and-push-images
         with:
           tag: ${{ github.event.inputs.tag }}
+      - name: Set version from tag
+        if : github.event_name == 'release' && github.event.action == 'published'
+        id: set-version-from-tag
+        run: echo "VERSION=${GITHUB_REF_NAME/v}" >> $GITHUB_OUTPUT
```
psschwei, line 46: nit: should we set this in the build and push images action? (just a thought, might make it easier to follow the flow, but I could go either way...)

Tansito: I see you point. Mmm... Right now in the action we have and input called `tag` that can have different values. We are using `latest` and the specific release version as values on it, depending of the action. I would say in this case at least from my point of view I would prefer the reusability that we have that introduce that value in the `action` (totally my opinion).

IceKhan13 (approved):
Looks good :)

PR #754, opened 2023-07-10
Split helm install into core and observability

psschwei:
### Summary

Create a separate helm chart for observability 

### Details and comments

Will add CI for building the observability helm chart after this merges


psschwei:
Good catch. I've updated the readmes for now, will do the deployment doc with the CI update (after this merges)

akihikokuroda (approved):
LGTM.  Thanks!

PR #777, opened 2023-07-13
mount cert volumes to init container

### LINKED ISSUE #776: Ray cluster: TLS is blocking workers to join cluster

Deploy ray cluster with tls enabled

Workers are handing in non-ready status on `wait for GCS to be ready;`

In init container `wait-gcs-ready`

This check is always failing
>       until ray health-check --address cluster-user-994c3d3c-head-svc.quantum-serverless.svc.cluster.local:6379 > /dev/null 2>&1; do echo wait for GCS to be ready; sleep 5; done

Therefore workers cannot connect to head node a.k.a join cluster.

Workers should be able to join cluster with TLS enabled

akihikokuroda:
### Summary
fix #776

### Details and comments
The init container needs the certificates to communicate to the head node.  The volume mounts were missing.

### DIFF
--- a/infrastructure/helm/quantumserverless/charts/gateway/templates/rayclustertemplate.yaml
+++ b/infrastructure/helm/quantumserverless/charts/gateway/templates/rayclustertemplate.yaml
@@ -264,6 +264,11 @@ data:
               - mountPath: /data
                 name: user-storage
                 subPath: {{`{{ user_id }}`}}
+              - mountPath: /etc/ca/tls
+                name: ca-tls
+                readOnly: true
+              - mountPath: /etc/ray/tls
+                name: ray-tls
             imagePullSecrets: []
 {{- if .Values.useCertManager }}
             serviceAccountName: ray-cluster-sa


akihikokuroda:
@IceKhan13 I think that this fix the issue.  You can try by adding the lines in the configmap template and reinstall the helm chart.

psschwei (approved):
(not for this PR) Is it possible to make the logs for that init container a bit more verbose? Right now if there's an error they just repeat "wait for GCS to be ready" over and over. I didn't see the `ray health-check` command in our repo though, so maybe not...

IceKhan13:
checking :)

IceKhan13 (approved):
Awesome! it's working. Workers are connected to head and workload is distributed to them.

IceKhan13:
Thank you @akihikokuroda!

akihikokuroda:
> Is it possible to make the logs for that init container a bit more verbose?

@psschwei The initContainer is probably added by the Ray operator so the changes may need in the kuberay.  I'll take a look.

Issue #780, opened 2023-07-14
Rename quantumserverless to quantum-serverless


Tansito:
### What is the expected behavior?

Due to not standardize our naming conventions we are generating errors like: #779 . I want to start introducing some conventions into this and use it along the code. My proposal is:
- quantum-serverless
- QuantumServerless

For those places where we require to reference the project. And this development should pontentially fix #779 .

Tansito:
Opinions @IceKhan13 , @psschwei , @akihikokuroda ?

Tansito:
I don't have an opinion yet for use cases like: `qs-observability`.

psschwei:
+1 for using the hyphen whenever possible.

psschwei:
@Tansito would`QuantumServerless` mainly be for use in the python code (which if I remember prefers camelCase over hyphens)?

re: `qs-observability` -- I tried to keep it short, but also add some type of distinguisher (as I can imagine there's a bunch of helm charts called `observability` out there)

Tansito:
It's the main use case in what I was thinking yes, but I don't discard that we could find other ones. But those two possibilities are the most typicall ones I would think.

About the `qs-observability` yeah. When I saw the PR I couldn't think in a better naming either. I'm fine if we consider `qs` as the "official" shorter name, for example 🤷‍♂️

PR #781, opened 2023-07-14
Fix helm publish action

### LINKED ISSUE #779: publish-helm action error

Error in the action `publish-helm`: https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5553652887

Due to errors in our naming convention we are generating errors like what we can see in the [0.3.1 release](https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5553652887).

This action should be executed correctly.

akihikokuroda:
### Summary
fix #779 

### Details and comments
This PR fix the helm publish error w/o changing the package names.

### DIFF
--- a/.github/workflows/publish-helm.yaml
+++ b/.github/workflows/publish-helm.yaml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         release:
-          - 'quantum-serverless'
+          - 'quantumserverless'
           - 'qs-observability'
     steps:
       - name: Set up Helm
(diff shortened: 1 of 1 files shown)


psschwei:
There's a discussion started in #780 to possibly rename the helm directory to `quantum-serverless`... should we merge this now (so we can release the helm chart) and then update again if/when we rename? I'm ok either way.

Tansito:
I don't think we have any kind of urgency so I would prefer to work in a better / definitive solution, I agree 👍

PR #775, opened 2023-07-13
Add default annotation to scrape gateway

psschwei:
### Summary

Adds an entry to `podAnnotations` for the gateway. 

### Details and comments

Annotation is `prometheus.io/scrape: "true"`. Needed to enable default scraping of the gateway pods by Prometheus, Sysdig, etc.

PR #784, opened 2023-07-14
Configure ray pods for prometheus scraping

psschwei:
Adds annotation so that ray pods / services will be scraped by Prometheus by default. Can be disabled in the gateway helm chart if so desired.

xref #775


Tansito (approved):
LGTM 💯

PR #792, opened 2023-07-18
Added encoder/decoder to save_result() and result()

scottmason88:
### Summary
save_result() and result() now support all types that 'Program.arguments' supports.

### Details and comments
I added QiskitObjectsDecoder to result(). I also changed save_result() to use data rather than json in the response, and then used json.dumps to still get the intended json. I did this so I could also use QiskitObjectsEncoder here.


IceKhan13:
@scottmason88 Thank you for PR! I think linter is not happy a little bit, but you can fix this by running

```
pip install black
black client/
```

and we also ask contributors to sign qiskit CLA

scottmason88:
Not sure how I've added a second user to my pull request. Closing and reopening. Sorry all!

PR #793, opened 2023-07-18
Added encoder/decoder to save_result() and result()

### LINKED ISSUE #492: Client: `save_result` should support all types that `Program.arguments` support

Client: `save_result` should support all types that `Program.arguments` support.

Currently `save_result` only support dictionaries. 

This can be achieved just by using encoder and decoder from program arguments.

scottmason88:
### Summary
Fixes #492 
save_result() now supports all types that Program.arguments supports

### Details and comments
I added the encoder to save_result(), changing the json in the response to data and using json.dumps to do the same thing and be able to use the encoder. I added the decoder to result().


IceKhan13 (approved):
LGTM!

Thank you @scottmason88

PR #794, opened 2023-07-18
Add shutdown delay for Ray clusters to gateway

### LINKED ISSUE #783: Add shut-down-delay parameter for Ray clusters

After a job is finished running, we shut down the associated Ray clusters via https://github.com/Qiskit-Extensions/quantum-serverless/blob/6cad5b936a2e004eee82d8db5e402226dd239222/gateway/api/management/commands/free_resources.py#L19-L39

However, there are cases where we might not want to immediately shut down the cluster (debugging, rapid series of jobs, etc.). 

So it might make sense to introduce a `shutdown-delay` option for the `handle()` function. This would take a value in seconds of the time to delay until the pod is deleted. A zero or undefined value would keep the existing functionality of immediate deletion. 

Open question on whether we want to also have something like `shutdown-delay = -1` to mean don't delete the cluster at all... 

This should be a parameter only exposed to operators / admins (probably via a configmap or envvar or label) and not something that users can set.

psschwei:
### Summary

Fixes #783 
Add shutdown delay for Ray clusters to gateway

### Details and comments
Creates a new environmental variable, `RAY_CLUSTER_SHUTDOWN_DELAY`, for the gateway which controls the number of seconds to delay deleting ray clusters once the shutdown job is called. 

Default value of `0` means no delay (so no breaking change in behavior). A value of `-1` means do not delete the cluster at all (useful as a debugging option).


gateway/api/management/commands/free_resources.py
```diff
@@ -27,6 +29,14 @@ def handle(self, *args, **options):
...
             # only kill cluster if not in local mode and no jobs are running there
             if len(alive_jobs) == 0 and not settings.RAY_CLUSTER_MODE.get("local"):
+                # Add a delay in killing the ray cluster
+                # default value is 0, meaning no delay
+                if settings.RAY_CLUSTER_SHUTDOWN_DELAY == -1:
+                    # -1 means don't delete the ray cluster at all
+                    return
+                if settings.RAY_CLUSTER_SHUTDOWN_DELAY > 0:
+                    # a value greater than zero means to delay shutting down
+                    time.sleep(settings.RAY_CLUSTER_SHUTDOWN_DELAY)
```
akihikokuroda, line 39: This call blocks the processing of the scheduler.  The other jobs are not processed during this sleep.

psschwei: good catch. if this was go, I'd call a goroutine here. guess we need to find the python equivalent... :thinking:

psschwei: or, maybe for now to keep things simple, we make it just a don't delete cluster option and punt the variable delay down the road (most of what we need is probably covered by not deleting...)

akihikokuroda: That's fine with me.  With the not delete, the raycluster CRD instances stay in the cluster.  We may want something like [this](https://github.com/Qiskit-Extensions/quantum-serverless/pull/698).

psschwei: yeah, I think that would be a good thing to include -- a way to clear out all the old clusters...
(thinking out loud) maybe we should include a label when the no-delete option is on, so that we can easily find clusters that won't be cleaned up automatically... will need to poke around on that a bit more tomorrow...

psschwei: at first glance, looks like it should be possible to run something like `kubectl delete raycluster -l nodelete=true`, but want to test it out before making the code change

psschwei: Yep, works great with a label, so cleanup is as simple as the command in my previous comment!

akihikokuroda: Looks good.  How about putting a debug log when cluster delete is skipped?

akihikokuroda (approved):
LGTM Thanks!

PR #817, opened 2023-07-26
Issue 566 | Gateway: runtime env variables

### LINKED ISSUE #566: Inject apiToken in the pod environment

If the environment is configured to login with apiToken, we need to have the option to inject this apiToken like VAR ENV where the tasks for a program is executed (compute resources for the user).

We need to avoid store it, just pass to the compute resources, loaded like VAR ENV during the program tasks execution and not remember it when the resource is deallocated after execution.

For example, in case that apiToken is related with an channel account from Qiskit, we can inject it in the VAR ENVs of the compute resources environment to avoid extra authentication to users, and do that authentication to them in a transparent way.

Imagine that authentication process is using the qiskit runtime cloud services... so we can inject the apiToken in the environment to be connection with qiskit runtime cloud services transparent for users, it means, something like:

```
QISKIT_IBM_TOKEN = ${apiToken}
QISKIT_IBM_CHANNEL = ibm_quantum
```

In this way, users in the environment, to import the Runtime primitives service to invoking, they just need to initialize like:

```
from qiskit_ibm_runtime import QiskitRuntimeService
service = QiskitRuntimeService()
```

Without pass extra credentials in the **QiskitRuntimeService**, so it is an easy way to call that remote service.

IceKhan13:
### Summary

Pass qiskit runtime env variables to jobs.

### Details and comments

Closes #566

### DIFF
--- a/gateway/README.md
+++ b/gateway/README.md
@@ -41,3 +41,5 @@ docker build -t qiskit/quantum-serverless-gateway:<VERSION> .
 | RAY_CLUSTER_WORKER_MIN_REPLICAS        | min worker replicas per cluster                                                                                                                                       |
 | RAY_CLUSTER_WORKER_MAX_REPLICAS        | max replicas per cluster                                                                                                                                              |
 | RAY_CLUSTER_MAX_READINESS_TIME         | max time in seconds to wait for cluster readiness. Will fail job if cluster is not ready in time.                                                                     |
+| QISKIT_IBM_CHANNEL                     | Channel that will be set in env variables in jobs for QiskitRuntimeService client                                                                                     |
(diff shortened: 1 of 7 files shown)


IceKhan13:
`QISKIT_IBM_URL` still not supported by `qiskit_ibm_runtime` package. So, it will only work for prod deployment of runtime. For any other url we will need to specify `url` in constructor for `QiskitRuntimeService` object.
```python
service = QiskitRuntimeService() # will work for prod
service = QiskitRuntimeService(url="other_url") # url is not supported from env vars for now
```
But I've created an issue, once it will be implemented we will not need to do any changes.

pacomf (approved):
LGTM, can you add some tests?

IceKhan13:
Added tests! Thanks Paco!

PR #821, opened 2023-07-28
CI improvements

Tansito:
### Summary

Related with the discussion #764 this pull request is intended to apply the next changes:
- Associate `latest` to the last published version
- Publish images only after publishing the release, removing `main`
- Created a `release` pull request to run special actions like the `vscan` before bumping a version

My idea with this last change is to have a `release` tag for version bumps when we want to publish a new version to run actions that mast be run before publishing anything and can be relatively cost to run them in every pull request.


Tansito:
This pull request pretends to improve a little bit our CI experience. I would like to know your opinions 🙏

.github/workflows/icr-image-build-and-push.yml
```diff
@@ -49,3 +41,7 @@ jobs:
         uses: ./.github/actions/icr-build-and-push-images
         with:
           tag: ${{ steps.set-version-from-tag.outputs.VERSION }}
+      - name: Build and push latest
+        uses: ./.github/actions/icr-build-and-push-images
+        with:
+          tag: latest
```
psschwei, line 47: Just a thought: we've already built the image in the previous step, so would it be better to just re-tag that image, rather than building & pushing again?

Tansito: I was thinking the same and absolutely it would be my preferred approach, but I didn't find a way to do it with the action. Do you know solutions for that @psschwei ?

psschwei: probably would need to script the whole build/push process... something like

```
docker build -t <release> ...
docker tag <release> latest ...
docker push <release>
docker push latest
```

Tansito: oh, good catch! ❤️

akihikokuroda (approved):

.github/workflows/icr-image-build-and-push.yml
Tansito: Solved in [b4afcef](https://github.com/Qiskit-Extensions/quantum-serverless/pull/821/commits/b4afcefd82617007fdffdb2bceb8b89d18c05382)

PR #826, opened 2023-07-31
Increase NGINX body-size

### LINKED ISSUE #818: Increase request payload size

Currently maximum size of uploaded file is set to ~2 mb (which is looks like default for nginx) which is limitation for users as artifact from uploaded programs might have large artifacts.

We need to allow users to upload larger artifact files (around 50 mb).

Probably it is need to be configured in deployment nginx

Tansito:
### Summary

Fix #818 
Co-authored-by: @psschwei 

Confirmed with the infrastructure team and added the required label specified in the issue: `proxy-body-size`.

### DIFF
--- a/infrastructure/helm/quantum-serverless/values-ibm.yaml
+++ b/infrastructure/helm/quantum-serverless/values-ibm.yaml
@@ -15,6 +15,7 @@ ingress:
     kubernetes.io/ingress.class: "public-iks-k8s-nginx"
     nginx.ingress.kubernetes.io/proxy-buffers-number: "4"
     nginx.ingress.kubernetes.io/proxy-buffer-size: "512k"
+    nginx.ingress.kubernetes.io/proxy-body-size: 50m
   tls:
     - hosts:
         - "gateway.YOUR_INGRESS_PUBLIC_END_POINT"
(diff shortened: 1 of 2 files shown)


psschwei (approved):

IceKhan13 (approved):
🚀

akihikokuroda:
I got a little confused with the changes in the value.yaml file change for the local install.  There are 2 ingress instances configurations in the value file.  One is installed with Ray operator and the other is installed as a dependency of the quantum-serferless chart.  The changes make the change in the ingress installed with Ray operator that is not enable for the local install.  The other ingress is enabled for the local install.

Tansito:
Mmm... Which instance for Ray operator do you mean, @akihikokuroda ?

Right now I see two different instances:

- Local NGINX for local development, managed by `nginxIngressControllerEnable`
- NGINX configuration for 3rd party cloud platforms

I admit that there we can have a bit of confuse because depending on the provider we are going to need to change that configuration. I'm totally open to improve that (I think @psschwei mentioned before some ways to fix it).
Sorry, I don't know if I'm followinf here, @akihikokuroda 🙏

akihikokuroda:
@psschwei Yes, it is.

PR #840, opened 2023-08-01
Drop aws terraform files

### LINKED ISSUE #761: Update AWS terraform configuration

From #760 we discovered that we have an outdated configuration in AWS terraform files.

I see two approaches that we can take with:
- Delete the entire configuration. We are not using it and if it's needed we can recover it later on.
- Update the configuration removing the package from it.

psschwei:
### Summary

Fixes #761


akihikokuroda (approved):
LGTM

PR #844, opened 2023-08-02
Issue 842 | Gateway: do not delete compute resource

### LINKED ISSUE #842: Gateway: do not delete `ComputeResource` on job update, deactivate instead

Gateway: do not delete `ComputeResource` on job update, deactivate instead.

This will allow us to keep information about program association to specific compute resource.

IceKhan13:
### Summary

Gateway: do not delete compute resource model, just deactivate it

### Details and comments

Closes #842


psschwei (commented):
We probably can drop the `RAY_CLUSTER_NO_DELETE_ON_COMPLETE` envvar from #794 with this change, right ?

akihikokuroda (approved):
LGTM

IceKhan13:
@psschwei this change only saves (prevent from removal) information about compute resource in db. Ray cluster itself is deleted anyway. Therefore `RAY_CLUSTER_NO_DELETE_ON_COMPLETE` is still needed

psschwei (approved):

Tansito (approved):
LGTM! 😄 

Btw @IceKhan13 just a question that came to my mind. Has it sense to add an `updated` field for a similar reason that we added it to the `Job`? I don't think we will need to monitor `ComputeResources` but just asking.

PR #855, opened 2023-08-03
Update `working_dir` too big exception message and docs

### LINKED ISSUE #849: Client: fix error message on artifact

Fix error messages about artifact. Make them more clear and actionable.

#### Details

This will be a common problem: We need to document clearly somewhere that everything in the working directory gets shipped to the remote cluster. Good practice to only put your program files in the working directory.

Update error message to be more specific about the problem. E.g., instead of using “artifact” say something about the files in the working directory are too large. 

The constraints we set on the Programs include a size limit on input files / arguments. In practice, this limit refers to the size of everything in the working directory. We need to make this clear whenever / wherever we talk about Program constraints

psschwei:
### Summary

Fixes #849 

### Details and comments

Note: the api docs don't seem to be picking up the updated Program docstring , but it also looks like those pull from the most recent version, so not sure if that's as expected ...


IceKhan13 (approved):
We actually had pretty good message already 🤔

PR #856, opened 2023-08-03
Add tox as development requirement

### LINKED ISSUE #854: `tox` is not listed as a requirement anywhere
Was setting up a new venv and wasn't able to run the tests because `tox` hadn't been installed... probably should add it to all the requirements-dev.txt files...

psschwei:
Fixes #854

### DIFF
--- a/client/requirements-dev.txt
+++ b/client/requirements-dev.txt
@@ -15,4 +15,4 @@ reno>=3.5.0
 black[jupyter]~=22.12
 requests-mock>=1.9.3
 testcontainers>=3.7.1
-tox>=4.6.0
+tox>=4.0.0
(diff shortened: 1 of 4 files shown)


IceKhan13 (approved):

Tansito (approved):
Thanks @psschwei ! One question here, should we start pinning versions even for `dev` requirements, @IceKhan13 , @psschwei ?

psschwei:
> should we start pinning versions even for `dev` requirements

:thinking: pinning makes things more reproducible / stable, but these are for building, so it might also be nice to be bleeding edge and/or pick up patches/fixes a little quicker ... I could go either way

Tansito:
Well, in general we are relatively fast with updates with `renovate`. From time to time we got stuck in a version but I suppose it's something common. I don't have a strong opinion, either (feel free to merge it).

psschwei:
Let's merge this and then pin them
(I did notice some weird stuff with using the most recent version of tox, which is why I had to do https://github.com/Qiskit-Extensions/quantum-serverless/pull/856/commits/f66909bace3b03b88f3672538ff81aef5bc512ce ... but I don't seem to be able to find the failed tests right now for some reason)

Issue #841, opened 2023-08-01
Infra: OT tracing throughout entire stack


IceKhan13:
### What is the expected behavior?

Currently we have setup to OT within `Program`, which trace through tasks or a job.

What we need is ability to trace request from client -> gateway -> ray cluster -> job and tasks (we have that now).

akihikokuroda:
I'll take a look.

psschwei:
I think you should be able to view OT metrics via prometheus (there's some discussion of it in #280 )

akihikokuroda:
@psschwei That's right but the distributed tracing doesn't go there.  We probably need something for it.

akihikokuroda:
I finally can see the tracing entries of the `program` locally now.  I'll start working for the additional components.

akihikokuroda:
![Screenshot 2023-08-17 at 11 33 36 AM](https://github.com/Qiskit-Extensions/quantum-serverless/assets/16141898/9cd61048-1f6a-40e2-af61-a7d557da3d60)

PR #865, opened 2023-08-07
Enable OT tracing in Ray

akihikokuroda:
### Summary
This is the first part of the #841. 

### Details and comments
This just enables OT tracing in the Ray components.

The propagation from Ray span to Program span and the OT tracing in the Gateway (django) are still under investigation.


client/quantum_serverless/core/tracing.py
```diff
@@ -114,3 +114,17 @@ def _trace_env_vars(env_vars: dict, location: Optional[str] = None):
...
+
+    Passed as an argument at Ray start
+    """
+    resource = Resource(attributes={SERVICE_NAME: "Quantum-Serverless: Ray"})
+    provider = TracerProvider(resource=resource)
+    otel_exporter = BatchSpanProcessor(
+        OTLPSpanExporter(endpoint="http://otel-collector:4317/", insecure=True)
+    )
+    provider.add_span_processor(otel_exporter)
+    trace._set_tracer_provider(provider, log=False)
```
psschwei: think the lint warning can be disabled via something like  `# pylint: disable=W0212` , but should we? not sure what the python best practices are in this space...

psschwei: looks like you fixed it while I was typing that :laughing:

client/quantum_serverless/core/tracing.py
```diff
@@ -114,3 +114,17 @@ def _trace_env_vars(env_vars: dict, location: Optional[str] = None):
...
+
+def setup_tracing() -> None:
+    """Setup Tracing for Ray cluster
+
+    Passed as an argument at Ray start
+    """
+    resource = Resource(attributes={SERVICE_NAME: "Quantum-Serverless: Ray"})
+    provider = TracerProvider(resource=resource)
+    otel_exporter = BatchSpanProcessor(
+        OTLPSpanExporter(endpoint="http://otel-collector:4317/", insecure=True)
```
psschwei, line 127: out of curiosity, would it cause any issues here if an alternative agent host/port is passed in L77 ? I assume those are separate so there's no reason to worry, but figure it doesn't hurt to ask

akihikokuroda: oops,  it's for my debugging.  I'll change it to
```
f"{agent_host}:{agent_port}"
```
Thanks!

psschwei (approved):

Tansito:
Oh, true, I don't know why. Maybe a connectivity problem (sometimes happens).

client/quantum_serverless/core/tracing.py
```diff
@@ -74,7 +74,7 @@ def get_tracer(
     provider = TracerProvider(resource=resource)
     if agent_host is not None and agent_port is not None:
         otel_exporter = BatchSpanProcessor(
-            OTLPSpanExporter(endpoint=f"{agent_host}:{agent_port}")
+            OTLPSpanExporter(endpoint=f"{agent_host}:{agent_port}", insecure=True)
```
Tansito, line 77: Is it possible to configure the `insecure` variable through a environment variable?

Maybe I'm wrong but I think we can have some warnings from the security team. What do you think @psschwei @akihikokuroda ? I suppose that can be `True` for development and `False` for production.

psschwei: Yeah, that should be doable as an envvar, similar to what we do for `SERVICE_NAME` just above

psschwei (approved):

IceKhan13 (approved):

PR #874, opened 2023-08-09
add OT tracing config in helm chart

akihikokuroda:
### Summary
A part of #841 

### Details and comments
This adds helm configurations for OT tracing


client/quantum_serverless/core/decorators.py
```diff
@@ -195,11 +197,14 @@ def _tracible_function(
     def decorator(func: Callable):
         @functools.wraps(func)
         def wraps(*args, **kwargs):
-            tracer = get_tracer(
-                func.__module__,
-                agent_host=os.environ.get(OT_JAEGER_HOST_KEY, None),
-                agent_port=int(os.environ.get(OT_JAEGER_PORT_KEY, 6831)),
-            )
+            if bool(int(os.environ.get(OT_RAY_TRACER, "0"))):
```
psschwei, line 200: Should we set this value outside of the `if` somehow, so that if the envvar isn't set we've already defaulted in a value? (that would also allow us to only set it one place, instead of both here and the tracing file)

psschwei (commented):
One minor nit, but otherwise LGTM

client/quantum_serverless/core/decorators.py
akihikokuroda: Thanks for comment.  I thought about refactoring the `get_tracer` function related to this.  I just wanted to make the smallest changes now and refactor together later.

PR #892, opened 2023-08-15
Drop program artifact after job scheduled

### LINKED ISSUE #871: Gateway: delete program artifact after job start

Delete program artifact after job is submitted.

psschwei:
### Summary

Fixes #871 
Removes artifact tar from remote storage after program is successfully submitted

### DIFF
--- a/gateway/api/schedule.py
+++ b/gateway/api/schedule.py
@@ -1,5 +1,6 @@
 """Scheduling related functions."""
 import logging
+import os
 import random
 import uuid
 from typing import List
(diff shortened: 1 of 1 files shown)


IceKhan13 (changes requested):
I was thinking to make it using `django signals`, but this is way more easier :) 

I think we only need to resolve conflicts.

We also can improve it a tiny bit by checking if file exists and (if I remember correctly) django stores full path to file if you call `<FILE_FIELD>.path`. Something like
```python
if os.path.exists(job.program.artifact.path):
    os.remove(job.program.artifact.path)
```

ref: https://docs.djangoproject.com/en/4.2/topics/files/#using-files-in-models

IceKhan13 (approved):

PR #893, opened 2023-08-15
pin redis to 4.x

akihikokuroda:
### Summary
Pin Redis to 4.x for now.

Client verify fails with redis 5.0.0

### DIFF
--- a/client/requirements.txt
+++ b/client/requirements.txt
@@ -3,7 +3,7 @@ requests>=2.31.0
 importlib-metadata>=5.2.0
 qiskit-terra==0.24.1
 qiskit-ibm-runtime>=0.11.2
-redis==4.6.0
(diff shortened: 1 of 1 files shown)


client/requirements.txt
```diff
@@ -3,7 +3,7 @@ requests>=2.31.0
 importlib-metadata>=5.2.0
 qiskit-terra==0.24.1
 qiskit-ibm-runtime>=0.11.2
-redis>=4.6.0
+redis==4.6.0
```
psschwei, line 6: Would it make sense to do something like this:

```suggestion
redis>=4.6.0, <5.0
```

just in case it takes us a while to get the update in for 5+?

psschwei (approved):

PR #896, opened 2023-08-21
Issue 843 | Files: download files saved during Program execution

### LINKED ISSUE #843: Client: allow users download files saved in `/data` directory

We need to provide users ability to download files they saved in `/data` folder. 

For users it would be ideal to have something like that:
```python
serverles = QuantumServerless(...)

path_to_downloaded_file = serverless.get_file(
   "my_awesome_data.tar"
) # similar to scp, which is users got used to
```

This will be a good UX, since users would only need client and would not need to go anywhere else

IceKhan13:
### Summary

Add feature to download save files to `/data` folder within programs.

Example:
```python
serverless.files()
serverless.download("my_awesome.tar")
```

> Limitations:
> - only `tar` files are supported
> - `tar` file should be saved in `/data` directory during your program execution to be visible by `.files()` method call
> - only `/data` directory is supported, `/data/other_folder` will not be visible

Todos:
- [x] client methods
- [x] gateway files api
- [x] gateway file api tests
- [x] doc/tutorial 

### Details and comments

Closes #843


client/quantum_serverless/quantum_serverless.py
```diff
@@ -156,17 +156,18 @@ def files(self):
         """
         return self._selected_provider.files()
 
-    def download(self, file: str):
+    def download(self, file: str, directory: str = "./"):
```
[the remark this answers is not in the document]
IceKhan13: oh, this is directory where to save downloaded files. I think I need to rename it to something more meaningful :)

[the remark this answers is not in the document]
psschwei: `download_location`?

Assuming `file` is the file to be downloaded, same question... could they do the same trick to break out?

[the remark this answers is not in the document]
IceKhan13: fixed name to `download_location`. 

For file path I use `os.path.basename` to extract only file name. https://github.com/Qiskit-Extensions/quantum-serverless/pull/896/commits/17a5d8b72f619a90da77a038101cc154739b57f7#diff-2025150bf47484f2582f9816af400b1336b7ca6adcbb4faa61bb2d3330df856dR233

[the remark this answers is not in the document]
psschwei: we ain't taking no guff from little bobby tables :sunglasses:

client/quantum_serverless/core/files.py
```diff
@@ -0,0 +1,92 @@
...
+            version: gateway version
+            token: authorization token
+        """
+        self.host = host
+        self.version = version
+        self._token = token
+
+    def download(self, file: str, download_location: str) -> Optional[str]:
+        """Downloads file."""
+        tracer = trace.get_tracer("client.tracer")
```
psschwei, line 57: This will let us view when someone is downloading a file in our monitoring stack, right? I could see us (or infra/security) wanting to set up some alerting around this down the line...

gateway/api/v1/views.py
```diff
@@ -39,3 +39,11 @@ class JobViewSet(views.JobViewSet):  # pylint: disable=too-many-ancestors
...
     def get_serializer_class(self):
         return v1_serializers.JobSerializer
+
+
+class FilesViewSet(views.FilesViewSet):
+    """
+    Files view set.
+    """
+
+    permission_classes = [permissions.IsAuthenticated, IsOwner]
```
psschwei, line 49: nice!

gateway/api/views.py
```diff
@@ -187,6 +190,67 @@ def stop(self, request, pk=None):  # pylint: disable=invalid-name,unused-argumen
...
 
+class FilesViewSet(viewsets.ViewSet):
+    """ViewSet for file operations handling.
+
+    Note: only tar files are available for list and download
+    """
+
+    BASE_NAME = "files"
+
+    # @action(methods=["GET"], detail=False)
```
psschwei, line 201:
```suggestion
    @action(methods=["GET"], detail=False)
```
?

gateway/api/views.py
```diff
@@ -187,6 +190,67 @@ def stop(self, request, pk=None):  # pylint: disable=invalid-name,unused-argumen
...
+                logger.warning(
+                    "Directory %s does not exist for %s.", user_dir, request.user
+                )
+
+        return Response({"results": files})
+
+    @action(methods=["GET"], detail=False)
+    def download(self, request):  # pylint: disable=invalid-name
+        """Download selected file."""
+        response = Response(
```
psschwei, line 223: nit
```suggestion
        # default response for file not found, overwritten if file is found
        response = Response(
```
or something like that

gateway/tests/api/test_files.py
```diff
@@ -0,0 +1,80 @@
...
+                auth, {"username": "test_user", "password": "123"}, format="json"
+            )
+            token = response.data.get("access")
+            self.client.credentials(HTTP_AUTHORIZATION="Bearer " + token)
+            url = reverse("v1:files-list")
+            response = self.client.get(url, format="json")
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            self.assertEqual(response.data, {"results": ["artifact.tar"]})
+
+    def test_non_existing_file_download(self):
```
psschwei, line 43: not sure how much value we'd actually get from this , but it might be good to add a test trying to escape the user's directory (more so that if it ever does become possible, we get an alert). a decent chance that it doesn't tell us anything we don't already know though...

client/quantum_serverless/core/files.py
```diff
@@ -0,0 +1,92 @@
...
+                timeout=REQUESTS_TIMEOUT,
+            ) as req:
+                req.raise_for_status()
+
+                total_size_in_bytes = int(req.headers.get("content-length", 0))
+                chunk_size = 8192
+                progress_bar = tqdm(
+                    total=total_size_in_bytes, unit="iB", unit_scale=True
+                )
+                file_name = f"downloaded_{str(uuid.uuid4())[:8]}_{file}"
```
akihikokuroda, line 73: Why does this download into an unique file name?  Avoiding override or appending?

IceKhan13: yes, to avoid overriding and we also return file name too, if user want to do something with it programmatically

gateway/api/views.py
IceKhan13: Forgot to remove it. `list` is a reserved django action, therefore it does not need decorator

gateway/tests/api/test_files.py
IceKhan13: Good suggestion! Added in [abf2247](https://github.com/Qiskit-Extensions/quantum-serverless/pull/896/commits/abf22471362ec7d559ce32fe816c47616672194f)

client/quantum_serverless/core/files.py
IceKhan13: yes, hopefully we should have traces of almost everything that happening

docs/development/guides/src/produce_files.py
```diff
@@ -0,0 +1,11 @@
+import os
+import tarfile
+from quantum_serverless import save_result
+
+with open("./my_file.txt", "w") as f:
+    f.write("Hello!")
+
+with tarfile.open("/data/my_file.tar", "w:gz") as tar:
```
akihikokuroda, line 8: The file in the data directory is not deleted after this execution.  The next execution of the same program (or generating the same name file) override the file. It it fails to generate the file, the old file may be down loaded.   It is user responsibility to generate an unique file name so that it would not be override or confused?

IceKhan13: yes, it is up to user to handle their files and names of them inside the program.

psschwei (commented):
LGTM
(since @akihikokuroda also had comments, will let him give final approval)

client/quantum_serverless/quantum_serverless.py
```diff
@@ -144,6 +144,31 @@ def get_jobs(self, **kwargs):
...
+            >>> serverless = QuantumServerless()
+            >>> serverless.files()
+
+        Returns:
+            list of available files
+        """
+        return self._selected_provider.files()
+
+    def download(self, file: str, download_location: str = "./"):
+        """Downloads file.
```
akihikokuroda, line 160: It may be better saying that the file down loaded in a different name and it's returned as the return value

IceKhan13: Agree! https://github.com/Qiskit-Extensions/quantum-serverless/pull/896/commits/84758f5a3c7f64a68fb1caad67c161ac1ecb8c7a

client/quantum_serverless/quantum_serverless.py
```diff
@@ -144,6 +144,31 @@ def get_jobs(self, **kwargs):
...
+        Example:
+            >>> serverless = QuantumServerless()
+            >>> serverless.files()
+
+        Returns:
+            list of available files
+        """
+        return self._selected_provider.files()
+
+    def download(self, file: str, download_location: str = "./"):
```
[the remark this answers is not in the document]
IceKhan13: I think if user downloads the file, he probably need it 😄 
Probably we do not need this flag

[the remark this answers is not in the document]
akihikokuroda: How he can delete it if he is done with it?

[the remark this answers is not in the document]
IceKhan13: ah, you mean delete from server. For now bucket policy will delete it or user can remove it in a job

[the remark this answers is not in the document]
IceKhan13: if we want to delete it it's probably worth createing follow up issue request to add `/delete` endpoint  for files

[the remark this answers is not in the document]
IceKhan13: and probably `/upload` would be great to add too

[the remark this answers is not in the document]
IceKhan13: but so far we only had download request :)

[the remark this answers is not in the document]
IceKhan13: https://github.com/Qiskit-Extensions/quantum-serverless/issues/934

akihikokuroda (approved):
LGTM Thanks!

PR #897, opened 2023-08-21
Jaeger local setup instructions

akihikokuroda:
### Summary
This PR adds Jaeger local setup instructions to the `qs-observability` README.md.

### DIFF
--- a/charts/qs-observability/README.md
+++ b/charts/qs-observability/README.md
@@ -53,3 +53,42 @@ If you are interested in more complex configurations, you have access to all the
 
 - For our promtail charts dependencies, we are using the default configuration created by Grafana project. To simplify the configuration we offered you with a straigh-forward initial parameters setup.
 But if you are interested in more complex configurations, you have access to all the parameters documented [here](https://github.com/grafana/helm-charts/blob/main/charts/promtail/README.md).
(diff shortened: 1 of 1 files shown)


charts/qs-observability/README.md
```diff
@@ -53,3 +53,42 @@ If you are interested in more complex configurations, you have access to all the
...
+```
+
+Install Jaeger Operator
+```shell
+kubectl create -f https://github.com/jaegertracing/jaeger-operator/releases/download/v1.47.0/jaeger-operator.yaml -n observability
+```
+
+Create Jaeger CRD instance
+```shell
+kubectl apply -n observability -f - <<EOF
```
Tansito, line 86: Not for now @akihikokuroda but should we create this CRD and apply it in the installation?

Tansito (approved):
Appart from my doubt LGTM 👍

charts/qs-observability/README.md
akihikokuroda: I couldn't add the cert-manager and Jaeger in the helm chart because their dependency and time delay for the cert-manager to be ready.  So the CRD instance may may not be a part of the helm chart soon.

PR #900, opened 2023-08-22
Relax Qiskit version constraint

caleb-johnson:
Let's unpin the Qiskit version from the client, so it can be used more flexibly.

May as well update to the new `qiskit` namespace while we're at it, I guess.


IceKhan13:
@akihikokuroda can you check of m1 when you have time?

akihikokuroda:
Works good in m1 :-)

PR #903, opened 2023-08-23
Version bump 0.4.0

Tansito:
### Summary

Release 0.4.0


Tansito:
After some investigation seems that the package 0.55 added an incompatible change in this new version that we require:
- https://github.com/pennersr/django-allauth/compare/0.54.0...0.55.0#diff-0d16353e7f1a8593229f982c23d4c0b092c7840b7a2d9bfd15c4088f70741ec1R56

Tansito:
Error solved in #904

PR #904, opened 2023-08-23
Pin django-allauth to 0.54

Tansito:
### Summary

Last day `django-allauth` released version `0.55` with an incompatibility change that it's not being supported by `dj-rest-auth`.

### Details and comments
- Changelog from `django-allauth`: https://github.com/pennersr/django-allauth/blob/main/ChangeLog.rst#backwards-incompatible-changes
- Issue related with this error: https://github.com/iMerica/dj-rest-auth/issues/534

### DIFF
--- a/gateway/requirements.txt
+++ b/gateway/requirements.txt
@@ -1,7 +1,7 @@
 django-filter>=23.2
 djangorestframework>=3.14.0
 Markdown>=3.4.4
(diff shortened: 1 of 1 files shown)


akihikokuroda (approved):
LGTM. We probably need an issue reviewing the pined dependencies in our release cycle.

PR #905, opened 2023-08-23
Gateway: symmetric encryption

IceKhan13:
### Summary

Adds symmetric encryption for tokens while program is in queue

### DIFF
--- a/charts/quantum-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/quantum-serverless/charts/gateway/templates/deployment.yaml
@@ -65,6 +65,8 @@ spec:
           env:
             - name: DEBUG
               value: {{ .Values.application.debug | quote }}
+            - name: DJANGO_SECRET_KEY
+              value: {{ .Values.application.secretKey | quote }}
             - name: SITE_HOST
               value: {{ .Values.application.siteHost | quote }}
             - name: RAY_HOST
@@ -182,6 +184,8 @@ spec:
           env:
             - name: DEBUG
               value: {{ .Values.application.debug | quote }}
+            - name: DJANGO_SECRET_KEY
+              value: {{ .Values.application.secretKey | quote }}
             - name: DATABASE_HOST
               value: {{ .Values.database.host }}
             - name: DATABASE_PORT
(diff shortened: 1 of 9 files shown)


charts/quantum-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -65,6 +65,8 @@ spec:
           env:
             - name: DEBUG
               value: {{ .Values.application.debug | quote }}
+            - name: DJANGO_SECRET_KEY
```
Tansito, line 68: I think we can consider that this value will be usually taken from a secret store as the [`DATABASE_PASSWORD`](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/charts/quantum-serverless/charts/gateway/templates/deployment.yaml#L118). I would change to something like:

```
            - name: DJANGO_SECRET_KEY
              valueFrom:
                secretKeyRef:
                  name: django
                  key: secret_key
```

what do you think?

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -8,6 +8,7 @@ useCertManager: false
 application:
 #  command: [ "gunicorn", "gateway.wsgi:application", "--bind", "0.0.0.0:8000", "--workers=3" ]
   debug: 0
+  secretKey: "django-insecure-&)i3b5aue*#-i6k9i-03qm(d!0h&662lbhj12on_*gimn3x8p7"
```
Tansito, line 11: If my previous comment apply we could remove this line letting only the default value in `gateway`:

https://github.com/Qiskit-Extensions/quantum-serverless/pull/905/files#diff-d3fab2ab912c6eb5c879651a1476f64f0551b747bf618b4ab13ddb4602dd45c1R26

Tansito (approved):
Thank you, @IceKhan13 ! In general I agree with implementation, just a minor comment that I think would be correct to add.

charts/quantum-serverless/charts/gateway/templates/deployment.yaml
IceKhan13: Could not agree more! Good catch!

Tansito (approved):
Nice! 🎉

PR #906, opened 2023-08-23
add program timeout

### LINKED ISSUE #902: Scheduler: program timeout

Implement program timeout of 14 days. 

If program in running state and nothing is happenning to it within 14 days, kill it.

akihikokuroda:
### Summary
Fix #902 

Add program timeout

### Details and comments
The timeout is configureable.  Default is 14 days.  

This PR also include 2 minor fixes
1. Wrong position of {{ -end }} in deployment.yaml
2. code change not to delete ray cluster when the program failed in the not delete option is true


gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -36,15 +38,35 @@ def handle(self, *args, **options):
...
 
+                if not success:
+                    if config.RAY_CLUSTER_NO_DELETE_ON_COMPLETE:
+                        logger.debug(
+                            "RAY_CLUSTER_NO_DELETE_ON_COMPLETE is enabled, "
+                            + "so cluster [%s] will not be removed",
+                            job.compute_resource.title,
+                        )
+                    else:
+                        kill_ray_cluster(job.compute_resource.title)
```
IceKhan13, line 64: Can we instead of killing cluster put job to `STOPPED` status and append log to job with something like `Job reached maximum runtime and stopped.`? 

Reason for this is one cluster can share multiple jobs, so we can incidentally kill other jobs too :) 

For killing clusters we have `free_resources` command, which will handle resource deallocation (killing clusters) where there is not actevely running jobs on it.  
https://github.com/Qiskit-Extensions/quantum-serverless/blob/913964586c65cbafba8b543cb3fb1b9135350fc1/gateway/api/management/commands/free_resources.py#L38

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -53,11 +75,11 @@ def handle(self, *args, **options):
                         job_status,
                     )
                     updated_jobs_counter += 1
-                # cleanup env vars
-                if job.in_terminal_state():
-                    job.env_vars = "{}"
-                job.status = job_status
-                job.save()
+                    # cleanup env vars
```
IceKhan13, line 78: oh, good suggestion!

IceKhan13 (changes requested):

gateway/api/management/commands/update_jobs_statuses.py
akihikokuroda: Oh, I see.  I didn't looked at the meaning of "success = False" correctly.

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -15,7 +17,7 @@ class Command(BaseCommand):
 
     help = "Update running job statuses and logs."
 
-    def handle(self, *args, **options):
+    def handle(self, *args, **options):  #  pylint: disable=too-many-branches
```
Tansito, line 20: Due to this I would say that this is a good time to create an issue to try to refactor this method.

akihikokuroda: New issue for it.  https://github.com/Qiskit-Extensions/quantum-serverless/issues/907

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -36,15 +38,34 @@ def handle(self, *args, **options):
...
+                now = datetime.now(tz=endtime.tzinfo)
+                if endtime < now:
+                    job_status = Job.STOPPED
+                    logger.warning(
+                        "Job [%s] reached maximum runtime [%s] days and stopped.",
+                        job.id,
+                        timeout,
                     )
 
+                if not success:
```
IceKhan13, line 55: we do not need to touch resources at all in this function. Just setting job to stop status would be enough. `free_resources` will handle cluster deletion automatically

IceKhan13 (changes requested):

gateway/api/management/commands/update_jobs_statuses.py
IceKhan13: a.k.a we can remove those additional lines :) 

and also it would be good to add
```python
job.logs = f"{job.logs}.\nMaximum job runtime reached. Stopping the job."
```

akihikokuroda: The kill_ray_cluster has been in this function.  I've just added code not to kill it when the  "RAY_CLUSTER_NO_DELETE_ON_COMPLETE" is true.  I can take out this block.

IceKhan13: oh, I see. Hmm, I will check it in a little bit. But overall looks great! :) We ill merge it

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -36,15 +38,37 @@ def handle(self, *args, **options):
...
 
-                if not success:
-                    kill_ray_cluster(job.compute_resource.title)
-                    job.compute_resource.delete()
-                    job.compute_resource = None
-                    job_status = Job.FAILED
+                timeout = config.PROGRAM_TIMEOUT
+                if job.updated:
+                    endtime = job.updated + timedelta(days=timeout)
+                else:
```
IceKhan13, line 44: I think we can remove this else statement and put logic of stopping job under `if job.updated` clause. Since technically job can sit for N days in a queue and not started.

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -36,15 +38,37 @@ def handle(self, *args, **options):
...
+                    endtime = job.created + timedelta(days=timeout)
+                now = datetime.now(tz=endtime.tzinfo)
+                if endtime < now:
+                    job_status = Job.STOPPED
                     job.logs = (
-                        f"{job.logs}\nSomething went wrong during updating job status."
+                        f"{job.logs}.\nMaximum job runtime reached. Stopping the job."
+                    )
```
IceKhan13, line 51: we probably need to run `job.save()` to commit changes to DB

IceKhan13 (approved):
Thank you @akihikokuroda !

PR #915, opened 2023-08-28
Disable helm version check

### LINKED ISSUE #914: Helm check force to increase a version

From #899 error in the helm action [here](https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/5997294623/job/16263408216?pr=899#step:3:433).

With the new helm check now that it's any difference in the chart we may need to increase the version. How we are not publishing with every change this is not needed and desired.

Disable version check.

Tansito:
### Summary

Fix #914 

### Details and comments

From #899 version check is not desired as we are not publishing the charts with every change.


akihikokuroda (approved):
LGTM

PR #916, opened 2023-08-28
Fix charts lint errors

Tansito:
### Summary

From #914 and #899 it seems that `chart-testing` only checks charts have changes. This pull request fixex all the lint errors found in those charts until now


IceKhan13 (approved):
Ah, lovely linter :)

PR #918, opened 2023-08-29
Unify yaml extension

### LINKED ISSUE #917: Unify files extension to .yaml or .yml

I would like to propose to unify YAML files extension to one of those two: `.yaml` or `.yml`.

I found problematic configurations in the GitHub Action due to the differences with these extensions.

I don't know if there is a more common extension or another: @IceKhan13 , @psschwei , @akihikokuroda

Tansito:
### Summary

Fix #917 , this pull request pretends to unify YAML extensions to `.yaml`.


docker-compose.yaml
```diff
@@ -161,7 +161,7 @@ services:
     profiles: [ "full" ]
     volumes:
       - /var/log:/var/log
-    command: -config.file=/etc/promtail/config.yml
```
Tansito, line 164: @akihikokuroda I had doubts about this configuration file. I didn't find it and I don't know if it's correct to change it. I tested everything and seems to be working but I would like to know your opinion.

psschwei: I _believe_ that file is in the container (and we don't touch it), so I don't think we want to update that one

Tansito: Solved in [f553791](https://github.com/Qiskit-Extensions/quantum-serverless/pull/918/commits/f5537916109f0e02752817d8909ad120f837fdca)

psschwei (approved):

psschwei:
@Tansito just need to resolve the conflicts and we can merge this one

Issue #907, opened 2023-08-24
Refactor gateway/api/management/commands/update_jobs_statuses.py


akihikokuroda:
### What is the expected enhancement?

The function in gateway/api/management/commands/update_jobs_statuses.py has a lint error and it is ignored by 
`#  pylint: disable=too-many-branches`. 

It should be refactored not to need the workaround. 

https://github.com/Qiskit-Extensions/quantum-serverless/pull/906#discussion_r1304377629

PR #919, opened 2023-08-29
refactor update_jobs_statuses

akihikokuroda:
### Summary
fix: #907 

### Details and comments

put some logic in separate functions

### DIFF
--- a/gateway/api/schedule.py
+++ b/gateway/api/schedule.py
@@ -180,7 +180,7 @@ def check_job_timeout(job: Job, job_status):
 
 
 def handle_job_status_not_available(job: Job, job_status):
-    """Process job status not available and update job """
+    """Process job status not available and update job"""
 
     if config.RAY_CLUSTER_NO_DELETE_ON_COMPLETE:
         logger.debug(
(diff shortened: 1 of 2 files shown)


gateway/api/schedule.py
```diff
@@ -153,3 +155,41 @@ def get_jobs_to_schedule_fair_share(slots: int) -> List[Job]:
         job_filter |= Q(author=entry["author"]) & Q(created=entry["job_date"])
 
     return Job.objects.filter(job_filter)
+
+
+def check_job_timeout(job: Job, job_status):
```
psschwei, line 160: I always get a bit mixed up on how Python handles passing variables, so just want to ask as a sanity check that the updates made to the `job` within these functions are seen outside of these functions, right?

akihikokuroda: That's correct.  I looked it up :-) and also tested.

psschwei (approved):

gateway/api/schedule.py
```diff
@@ -153,3 +155,41 @@ def get_jobs_to_schedule_fair_share(slots: int) -> List[Job]:
...
+    if job.updated and endtime < now:
+        job_status = Job.STOPPED
+        job.logs = f"{job.logs}.\nMaximum job runtime reached. Stopping the job."
+        job.save()
+        logger.warning(
+            "Job [%s] reached maximum runtime [%s] days and stopped.",
+            job.id,
+            timeout,
+        )
+    print(job_status)
```
IceKhan13, line 176: Remove this and we are good :)

IceKhan13 (approved):
Great! much cleaner!

PR #920, opened 2023-08-30
Increase max number of renovate updates

Tansito:
### Summary

As we did in #825 let's increase the number of updates that renovate can open meanwhile we solve the problems with those updates that are `on-hold`


akihikokuroda (approved):
LGTM

PR #921, opened 2023-08-30
Fix repository helm configuration

Tansito:
### Summary

From #899 we discovered that we had and error in our `repository` chart configuration. This pull request will fix it.

```
==> Linting charts/repository
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: template: repository/templates/deployment.yaml:59:31: executing "repository/templates/deployment.yaml" at <.Values.application.superuser.username>: nil pointer evaluating interface {}.username
```

### Details and comments
- Fixed a typo in the helm's GHA due to that it's not being fired when there was a change inside charts sub-folders.

### DIFF
--- a/charts/quantum-serverless/charts/repository/values.yaml
+++ b/charts/quantum-serverless/charts/repository/values.yaml
@@ -6,6 +6,10 @@ replicaCount: 1
 
 application:
   debug: 0
+  superuser:
+    username: "admin"
+    password: "passw0rd"
+    email: "admin@examplemail.io"
 
 database:
   host: postgresql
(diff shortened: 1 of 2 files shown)


charts/quantum-serverless/charts/repository/values.yaml
```diff
@@ -6,6 +6,10 @@ replicaCount: 1
 
 application:
   debug: 0
+  superuser:
```
akihikokuroda, line 9: We may need to address [this issue](https://github.com/Qiskit-Extensions/quantum-serverless/issues/371).   Are we going to bump up to high priority?

Tansito: I think we should, I'm going to start with the integration with the infrastructure team so I think it has a lot of sense to start working on this.

akihikokuroda (approved):
LGTM

IceKhan13 (approved):

Issue #926, opened 2023-08-30
Post message in slack channel when job completed 


akihikokuroda:
### What is the expected behavior?

It's difficult to monitor completion of the long running job.  It is useful getting a notification such as slack message when it completed.

PR #927, opened 2023-08-30
send notification

akihikokuroda:
### Summary
fix: #926

### Details and comments
This PR adds slack message sending option when a job completed.  This requires slack Web API app configured in the slack side.   The message in the slack channel is like this:
Screenshot 2023-08-30 at 12 32 21 PM

### DIFF
--- a/gateway/api/schedule.py
+++ b/gateway/api/schedule.py
@@ -1,5 +1,6 @@
 """Scheduling related functions."""
 import logging
+import os
 import random
 import uuid
 from typing import List
@@ -12,6 +13,7 @@ from django.db.models import Q
 from django.db.models.aggregates import Count, Min
 
 from opentelemetry import trace
+from slack_sdk import WebClient
 
 from api.models import Job, Program, ComputeResource
 from api.ray import submit_job, create_ray_cluster, kill_ray_cluster
@@ -192,3 +194,16 @@ def handle_job_status_not_available(job: Job, job_status):
         job_status = Job.FAILED
         job.logs = f"{job.logs}\nSomething went wrong during updating job status."
     return job_status
+
+
+def send_notifications(job: Job):
+    """Send job status change notifications to the configured targets"""
+    slack_token = os.environ.get("SLACK_TOKEN")
+    slack_channel = os.environ.get("SLACK_CHANNEL")
+    if slack_token and slack_channel:
+        slack_client = WebClient(token=slack_token)
+        if job.status in (Job.SUCCEEDED, Job.STOPPED, Job.FAILED):
+            slack_client.chat_postMessage(
+                channel=slack_channel,
+                text=f"Job: {job.id} completed.  Job status: {job.status}",
+            )
--- a/charts/quantum-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/quantum-serverless/charts/gateway/templates/deployment.yaml
@@ -231,6 +231,13 @@ spec:
               value: {{ .Values.application.ray.openTelemetryCollector.insecure | quote }}
             - name: PROGRAM_TIMEOUT
               value: {{ .Values.application.limits.programTimeoutDays | quote }}
+            - name: SLACK_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: gateway-slack
+                  key: token
+            - name: SLACK_CHANNEL
+              value: {{ .Values.application.slack.channel }}
       {{- with .Values.nodeSelector }}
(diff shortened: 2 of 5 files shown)


gateway/api/schedule.py
```diff
@@ -192,3 +194,15 @@ def handle_job_status_not_available(job: Job, job_status):
         job_status = Job.FAILED
         job.logs = f"{job.logs}\nSomething went wrong during updating job status."
     return job_status
+
+
+def send_notifications(job: Job):
+    slackToken = os.environ.get("SLACK_TOKEN")
+    slackChannel = os.environ.get("SLACK_CHANNEL")
+    if slackToken and slackChannel:
```
psschwei, line 202: Would it be possible to move this logic out of the function? Right now, we'd always call this function and then check if slack output was enabled. It might be better to check if slack output was enabled and if so then call this function...

psschwei (commented):
Thinking out loud here, but would this be the kind of alert that we'd want to emit a metric for and then let the user's monitoring tool pick it up and send the slack message? Essentially passing the alerting off to the monitoring stack rather than managing it directly ourselves?

gateway/api/schedule.py
akihikokuroda: I want to keep the caller(handle in update_jobs_statuses.py) of this function simple.  I've just refactor it to simplify to remove the too_many_branches.  This function is called only then the job status is changed.  I guess it's not too bad for the performance point of view.  WDYT?

psschwei: I was thinking something like this:

```python
def handle():
    ...
    if slack_has_been_enabled: # however that's best determined
        send_notifications(job)
```

mainly to help with readability: if we haven't enabled slack, then there's no reason to call the notifications method. 

That said, I'm not super-familiar with the too-many-branches error, would adding an sometimes call of the notifications function cause that problem here?

akihikokuroda: It's the total number of if, while, try and etc.. that cause the code branches. So adding `if` is not good from the point of view.

psschwei: That's a weird limitation of the framework, but I'm sure there's a reason for that :shrug: 

May also be another argument in favor of pushing alerts to the monitoring platform...

IceKhan13:
Oh, interesting idea about notifications!

I agree with Paul on comment that we should think of something more generic. With this implementation all notifications about all jobs of all users will be lending in same slack channel. 

How do end user will benefit from this? or it is for us? 

There are couple of way how we can make it more general.
1. Let's use [django signals](https://docs.djangoproject.com/en/4.2/topics/signals/) for that. With signals we can catch events of upserting models. And it will be living in a separate module and will be handled separately.
```python
# signals.py
from django.core.signals import post_save
from django.dispatch import receiver

@receiver(post_save, sender=Job)
def notification_haldling(sender, **kwargs):
    # handle notification sending at post job save
```
2.  We can implement some general notification backends. For example
```python
class JobNotificationHandler:
    def send(self, job: Job):
        raise NotImplementedError


class SlackJobNotificationHandler(JobNotificationHandler):
    def send(self, job: Job):
       # send to slack 
```
and then we can have it more generically configured, something like
```python
# settings.py

NOTIFICATION_BACKENDS = ["api.notifications.SlackJobNotificationHandler", ...]
```
so, your signal will be something like that
```python
# signals.py

@receiver(post_save, sender=Job)
def notification_haldling(sender, instance,  **kwargs):
    for notification_backend in settings.NOTIFICATION_BACKENDS:
       notification_handler =  notification_backend()
       notification_handler.send(instance)
```

gateway/api/schedule.py
akihikokuroda: Yes, this is one way to send the notification / alert.  This way is quite simple in our side.  So it' not bad to have the capability.

IceKhan13:
But again I would also delegate sending notification to monitoring systems :)

akihikokuroda:
@IceKhan13 @psschwei I agree. I don't want to make this slack specific.  It is one of the way to send the notification so I didn't put the slack in the function name.  @IceKhan13 I'll look into the way you put.

psschwei:
Since we're already logging in this block, we can probably just check if the log exists: https://cloud.ibm.com/docs/log-analysis?topic=log-analysis-alerts#alerts_presence

akihikokuroda:
> How do end user will benefit from this? or it is for us?

It can enable push notifications and it's useful.  Maybe the monitoring system has it and I just don't know.  I'll close this one.

PR #928, opened 2023-08-31
Improve secrets support

### LINKED ISSUE #371: Use secret for gateway superuser credentials
We currently define the gateway credentials for the Helm install in [`values.yaml`](https://github.com/Qiskit-Extensions/quantum-serverless/blob/9f02e40479e94128bee775d200cb33f010e79fd1/infrastructure/helm/quantumserverless/values.yaml#L30-L33). 

These are added as environmental variables for the container, but as a result anyone who can view pods in the `ray` namespace can get access to the superuser credentials:

```
$ k get po -n ray quantum-serverless-gateway-745bfd7b5f-tz4z4 -o json | jq '.spec.containers[].env'
[
  ...
  {
    "name": "DJANGO_SUPERUSER_USERNAME",
    "value": "admin"
  },
  {
    "name": "DJANGO_SUPERUSER_PASSWORD",
    "value": "passw0rd"
  },
  {
    "name": "DJANGO_SUPERUSER_EMAIL",
    "value": "admin@examplemail.io"
  }
]
```

Instead of using environmental variables, we should use a more secure method depending on the environment (for example, for Kubernetes we should use [secrets](https://kubernetes.io/docs/concepts/configuration/secret/))

Tansito:
### Summary

Fix #371 

### Details and comments

- Updated sensitive values to use secrets


akihikokuroda:
Necessary secrets are usually added in the template like this (https://github.com/Qiskit-Extensions/quantum-serverless/pull/930)

Tansito:
I see, I'm going to work on this for local, thanks @akihikokuroda 👍

charts/quantum-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -30,14 +30,35 @@ spec:
       initContainers:
         - name: waitpostresql
           image: actions/pg_isready
-          command: ['sh', '-c', 'until pg_isready -U {{ .Values.database.user | quote }} -d "dbname={{ .Values.database.name }}" -h {{ .Values.database.host }} -p {{ .Values.database.port | quote }}; do echo waiting for myservice; sleep 2; done']
+          command: ['sh', '-c', 'until pg_isready -U ${DATABASE_USER} -d "dbname=${DATABASE_NAME}" -h ${DATABASE_HOST} -p ${DATABASE_PORT}; do echo waiting for myservice; sleep 2; done']
```
IceKhan13, line 33: where this values will come from?

IceKhan13: oh, nevermind, it is from env vars :)

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -46,11 +41,41 @@ application:
...
+      port: database-port
+      userName: user-name
+    value:
+      databaseName: gateway_db
+      databasePassword: CHANGE_ME
+      host: postgresql
+      port: 5432
+      userName: bn_keycloak
+  superuser:
+    create: user
```
akihikokuroda, line 69: This is overridden so it may not matter but it should be true or false.

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -46,11 +41,41 @@ application:
...
+    value: "django-insecure-&)i3b5aue*#-i6k9i-03qm(d!0h&662lbhj12on_*gimn3x8p7"
+  servicePsql:
+    create: true
+    name: service-psql-binding
+    key:
+      databaseName: database-name
+      databasePassword: database-password
+      host: database-host
+      port: database-port
+      userName: user-name
```
akihikokuroda, line 61: These keys are not necessary in the values.yaml file.  They are hardcoded in the secret yaml file.

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -46,11 +41,41 @@ application:
...
+    value:
+      databaseName: gateway_db
+      databasePassword: CHANGE_ME
+      host: postgresql
+      port: 5432
+      userName: bn_keycloak
+  superuser:
+    create: user
+    name: gateway-superuser
+    key:
```
akihikokuroda, line 71: Same as above key.

akihikokuroda (changes requested):

charts/quantum-serverless/charts/gateway/values.yaml
Tansito: Ups, yep!

Tansito: Solved in [a2c898f](https://github.com/Qiskit-Extensions/quantum-serverless/pull/928/commits/a2c898fb2188c2bc4d916ea478cfb86d95cd3cf1)

charts/quantum-serverless/charts/gateway/values.yaml
Tansito: They are for us because in our environment we don't create all the secrets and we don't have control over the key names (that's the reason because I added the `create` field to create or not the secret when deploying).

charts/quantum-serverless/charts/gateway/values.yaml
Tansito: As I said above, in our environment we don't have control over all the secrets so we need to have a way to pass the key from where to obtain the value.

charts/quantum-serverless/charts/gateway/values.yaml
Tansito: To put an example this is the `key` for the `host` secret in our internal environment:

```
connection.postgres.hosts.0.hostname
```

Tansito: So my idea is, if you create the secrets with our scripts perfect. If not for whatever reason, you have a way to not install them and specify those custom secrets (as it is our case).

Tansito: I accept other approach if someone knows a better one, btw. This was just the easiest way to handle the problem.

akihikokuroda: I'm not sure how these are used for the other environment but if the secret is defined like this
```
stringData:
  {{ .Values.secrets.servicePsql.key.databaseName }}: {{ .Values.secrets.servicePsql.value.databaseName }}
  {{ .Values.secrets.servicePsql.key.databasePassword }}: {{ .Values.secrets.servicePsql.value.databasePassword }}
  {{ .Values.secrets.servicePsql.key.host }}: {{ .Values.secrets.servicePsql.value.host }}
  {{ .Values.secrets.servicePsql.key.port }}: {{ .Values.secrets.servicePsql.value.port }}
  {{ .Values.secrets.servicePsql.key.userName }}: {{ .Values.secrets.servicePsql.value.userName }}
```
instead of 
```
stringData:
  database-name: {{ .Values.secrets.servicePsql.value.databaseName }}
  database-password: {{ .Values.secrets.servicePsql.value.databasePassword }}
  database-host: {{ .Values.secrets.servicePsql.value.host }}
  database-port: {{ .Values.secrets.servicePsql.value.port }}
  user-name: {{ .Values.secrets.servicePsql.value.userName }}
```
then it makes sense.

Tansito: oh wait, is that possible? It gave me an error when I tried but maybe I did something wrong

Tansito: @akihikokuroda thank you for the feedback, it worked, i don't know what problem i had previously. Let me know if you see something more weird 🙏 

Solved in [4300cfb](https://github.com/Qiskit-Extensions/quantum-serverless/pull/928/commits/4300cfb41763f58ae58ee0bf72d8850cda71d901)

akihikokuroda (approved):
LGTM @Tansito Thanks!

PR #930, opened 2023-08-31
Add template for django secret

psschwei:
After #905 we now read the django secret_key from a kubernetes secret, so this adds the relevant helm stuff to create said secret.

### DIFF
--- a/charts/quantum-serverless/charts/gateway/templates/django-secret.yaml
+++ b/charts/quantum-serverless/charts/gateway/templates/django-secret.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: django
+stringData:
+  secret_key: {{ .Values.application.django.secret_key  }}
--- a/charts/quantum-serverless/charts/gateway/templates/django-secret.yaml
+++ b/charts/quantum-serverless/charts/gateway/templates/django-secret.yaml
@@ -1,6 +1,8 @@
+{{- if .Values.application.django.secret_key }}
 apiVersion: v1
 kind: Secret
 metadata:
   name: django
 stringData:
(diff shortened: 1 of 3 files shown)


akihikokuroda (approved):
LGTM Thanks!

IceKhan13 (approved):

charts/quantum-serverless/charts/gateway/templates/django-secret.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
IceKhan13, line 1: maybe add something like {{ if .Values.application.django.secret_key }} to check wether we need this template

psschwei: what kind of scenario are you thinking of?

IceKhan13: if secret already configured in cluster

IceKhan13: or it is still will work?

psschwei: no, it will fail if the secret was previously created:

`Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: Secret "django" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "qs"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"`

let me update that

IceKhan13 (approved):
sweet!

PR #933, opened 2023-09-01
Issue 932 | Client: drop state handler class

### LINKED ISSUE #932: Client: drop `StateHandler`

Drop `StateHandler` class as it is not used anywhere and we do not have plans to use it. 
This will lift a requirement to have redis and managing redis dependencies.

IceKhan13:
### Summary

Drop state handler class

### Details and comments

Closes #932


psschwei:
Does this mean we no longer need redis on the helm / docker installs? :pray:

IceKhan13:
Yes! I'm cutting it off now! 🕺

psschwei:
Can we also drop the docker compose files?

psschwei (approved):
:tada: :tada: :tada: 

Looks like we already dropped from docker compose a while ago (which I guess I missed)

Issue #326, opened 2023-03-22
Gateway chart improve configuration


IceKhan13:
### What is the expected enhancement?

Currently we have configuration for gateway through env variables.

Some of them can be derived from chart values, like ray host

https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98068912427586fa4162dec6734fe9ce072c507/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml#L57

We can use something like
```yaml
{{ .Release.Name }}-ray-head
```

Same thing is for keycloak

https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98068912427586fa4162dec6734fe9ce072c507/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml#L61

and site url (this is required for django + one internal function)

https://github.com/Qiskit-Extensions/quantum-serverless/blob/a98068912427586fa4162dec6734fe9ce072c507/infrastructure/helm/quantumserverless/charts/gateway/templates/deployment.yaml#L55

akihikokuroda:
The `rayHost` is no longer used.  It should be removed.

akihikokuroda:
The key cloak url setting depends on the setting in the keycloak section of the values.yaml file.
```
keycloak:
  fullnameOverride: "keycloak"
  service:
    ports:
      http: 31059
```

The siteHost must match the service name of the gateway.  It can be fixed to `gateway` by setting
```
gateway:
  fullnameOverride: "gateway"
```
in the value file.

IceKhan13:
closed in https://github.com/Qiskit-Extensions/quantum-serverless/pull/940

PR #940, opened 2023-09-06
simplify gateway config

akihikokuroda:
### Summary
fix: #326

### Details and comments
There are not a lot of things to simplify the gateway config so I moved most of configs that default values work for into the gateway/values.yaml to make the quantum-serverless/values.yaml and values-ibm.yaml simple.  I also add some comments in gateway/values.yam file.


Tansito:
is it possible to wait for #928 before close this? I will try to have it for today 🙏

Tansito:
@akihikokuroda #928 probably introduced some problems in the branch, ping me if you need to solve any problem! 🙏

charts/quantum-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -167,8 +165,8 @@ spec:
             - name: DATABASE_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ .Values.secrets.servicePsql.name }}
-                  key: {{ .Values.secrets.servicePsql.key.databasePassword }}
+                  name: postgresql
+                  key: password
```
Tansito, line 169: I don't know if this is due to the rebase @akihikokuroda but this change needs to remain as it is.

charts/quantum-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -280,8 +278,8 @@ spec:
...
```
Tansito, line 282: Same as my previous comment

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -3,44 +3,60 @@
...
+  # gateway site url - http://<fullnameOverride>:<service.port>
+  siteHost: "http://gateway:8000"
+  # authentification setting
   auth:
     token: {}
     keycloak: {}
-  superuser:
-    enable: true
+  # django secret key - any secret string
+  django:
```
Tansito, line 23: This key is not needed now, it goes in the secrets part.

charts/quantum-serverless/values-ibm.yaml
```diff
@@ -34,17 +34,7 @@ ingress:
```
Tansito, line 34: Now that we finished with the integration with the infrastructure team I think we could remove this file.

charts/quantum-serverless/values.yaml
```diff
@@ -53,51 +43,6 @@ gateway:
...
-      nodeImage: "icr.io/quantum-public/quantum-serverless-ray-node:0.6.0-py39"
-      opensslImage: registry.access.redhat.com/ubi8/openssl:8.8-7
-      kubectlImage: alpine/k8s:1.28.1@sha256:c37a54b738e4d26f88d4a4d28680bb74dfea1d0e756f039dad900306abc837a8
-    limits:
-      maxJobsPerUser: 2
-      maxComputeResources: 4
-  cos:
-    claimName: gateway-claim
-
-  secrets:
```
Tansito, line 69: I think you will need to promote this change to `gateway/values.yaml`.

Tansito (changes requested):

charts/quantum-serverless/values.yaml
akihikokuroda: Yes, they are already there except `service`.  The `service` here is in the wrong place.

charts/quantum-serverless/charts/gateway/templates/deployment.yaml
akihikokuroda: For the local install, the secret for the password is created by the database install.  So we need a little tweak.  I added another value `databaseSecretName` to separate the name of the secret and the secret name for the password.

charts/quantum-serverless/charts/gateway/values.yaml
akihikokuroda: I see. I take it out.

charts/quantum-serverless/values-ibm.yaml
akihikokuroda: OK.  I remove it.  Thanks!

Tansito:
@akihikokuroda what should we do with this pull request? Should we close it or do you want to solve the conflicts and that I take a look?

PR #944, opened 2023-09-07
Transform result to dict before returning from Job.result

### LINKED ISSUE #942: job.result() returns a string representation of dict rather than the dict itself

job.result() should translate the string into a dictionary for the user. We could probably catch serialization errors in the client and provide helpful error messages if the ones coming from `json` (or whatever) aren't informative enough.

caleb-johnson:
Fixes #942

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -381,9 +381,9 @@ class Job:
         if isinstance(results, str):
             try:
                 results = json.loads(results, cls=QiskitObjectsDecoder)
-            except json.JSONDecodeError:
-                pass
-
+            except json.JSONDecodeError as exception:
+                logging.warning(f"Error during results decoding. Details: {exception}")
+                
         return results
 
     def _in_terminal_state(self) -> bool:
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -381,7 +381,7 @@ class Job:
         if isinstance(results, str):
             try:
                 results = json.loads(
-                    self._job_client.result(self.job_id), cls=QiskitObjectsDecoder
+                    results, cls=QiskitObjectsDecoder
                 )
             except json.JSONDecodeError:
                 pass
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -382,7 +382,7 @@ class Job:
             try:
                 results = json.loads(results, cls=QiskitObjectsDecoder)
             except json.JSONDecodeError as exception:
-                logging.warning(f"Error during results decoding. Details: {exception}")
+                logging.warning("Error during results decoding. Details: %s", exception)
 
         return results
 
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -377,7 +377,9 @@ class Job:
                     logging.info(".")
         results = self._job_client.result(self.job_id)
         if isinstance(results, str):
-            results = json.loads(self._job_client.result(self.job_id))
+            results = json.loads(
+                self._job_client.result(self.job_id), cls=QiskitObjectsDecoder
+            )
 
         return results
 
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -383,7 +383,7 @@ class Job:
                 results = json.loads(results, cls=QiskitObjectsDecoder)
             except json.JSONDecodeError as exception:
                 logging.warning(f"Error during results decoding. Details: {exception}")
-                
+
         return results
 
     def _in_terminal_state(self) -> bool:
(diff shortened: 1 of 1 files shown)


client/quantum_serverless/core/job.py
```diff
@@ -375,7 +375,11 @@ def result(self, wait=True, cadence=5, verbose=False):
                 time.sleep(cadence)
                 if verbose:
                     logging.info(".")
-        return self._job_client.result(self.job_id)
+        results = self._job_client.result(self.job_id)
+        if isinstance(results, str):
```
psschwei, line 379: I could be overthinking things here, but would it be better to handle this within the job client (i.e. [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/c2a5e9d6d5adde09eb20590e9163298ada593154/client/quantum_serverless/core/job.py#L286))?

caleb-johnson: It looks like the return is already loaded into a dictionary in that line, but when I use `save_result` inside my remote program, `job.result()` returns me a string representation of a dictionary

caleb-johnson: OK I see what's happening. I think I am intended to use the client to retrieve results, but I was using the `job` object itself. I guess it's appropriate to do the decoding on both levels?

caleb-johnson: @IceKhan13 why would the return object in the deleted line ever be a string? `job_client.result()` returns a decoded object [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/c2a5e9d6d5adde09eb20590e9163298ada593154/client/quantum_serverless/core/job.py#L286)

IceKhan13: Hmm, good question. It should be decoded object (dict). I do not know why it's returning string. Worth digging into it

client/quantum_serverless/core/job.py
```diff
@@ -375,7 +375,18 @@ def result(self, wait=True, cadence=5, verbose=False):
...
                 if verbose:
                     logging.info(".")
-        return self._job_client.result(self.job_id)
+
+        # Retrieve the results. If they're string format, try to decode to a dictionary.
+        results = self._job_client.result(self.job_id)
+        if isinstance(results, str):
+            try:
+                results = json.loads(
+                    self._job_client.result(self.job_id), cls=QiskitObjectsDecoder
```
IceKhan13, line 384:
```suggestion
                    results, cls=QiskitObjectsDecoder
```

IceKhan13:
@caleb-johnson can you update code to pass lints and we can merge this :)

caleb-johnson:
> @caleb-johnson can you update code to pass lints and we can merge this :)

I can do that, but I am not really satisfied that this is the right fix. I do not understand why we need to decode the string here, when it is [already being done in the GatewayJobClient](https://github.com/Qiskit-Extensions/quantum-serverless/blob/97caeb6a09b116a2b5e58bb376f2eb93d084f567/client/quantum_serverless/core/job.py#L276).

Do you understand what is causing these JSON strings to make it through the `result` call?

psschwei:
Something looks a bit off with the commits...

caleb-johnson:
> Something looks a bit off with the commits...

*sigh* 

> Something looks a bit off with the commits...

Thanks, I cleaned it up. I made a mess by merging main into this branch :P

Going to take a closer look at this now that Aki solved the other problem with results

IceKhan13:
@caleb-johnson thing is we have DB schema in following format

```
Model: Job
Fields:
- id: str
- status: str
- logs: str
- results: str
```

so when it is returned from DB json looks like this
```json
{
  "id": "42",
  "status": "SUCCEEDED",
  "results": "{'bla-bla': 'bla-bla'}"
}
```

therefore we need to do another decoding on `results` field.

It can be fixed by changing https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/gateway/api/models.py#L94 to JsonField, probably. Or even better, by creating custom serializer which will decode results field before returning back to user https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/gateway/api/serializers.py#L23

client/quantum_serverless/core/job.py
```diff
@@ -375,7 +375,16 @@ def result(self, wait=True, cadence=5, verbose=False):
...
                     logging.info(".")
-        return self._job_client.result(self.job_id)
+
+        # Retrieve the results. If they're string format, try to decode to a dictionary.
+        results = self._job_client.result(self.job_id)
+        if isinstance(results, str):
+            try:
+                results = json.loads(results, cls=QiskitObjectsDecoder)
+            except json.JSONDecodeError:
+                pass
```
IceKhan13, line 385: can we log something here?

```
logger.warning("Error during results decoding. Details: %s", exception)
```

IceKhan13: and we are good to go :)

caleb-johnson:
@IceKhan13 , as discussed, we can just merge this patch and follow up with a better fix in a following PR.

Added a warning when the string isn't interpretable by json, as requested.

Ready for final review/approval

IceKhan13 (approved):

Issue #934, opened 2023-09-01
Files: add endpoints to upload and delete files


IceKhan13:
### What is the expected behavior?

Add ability to upload and delete files using client

IceKhan13:
Closed in #957 and #945

PR #945, opened 2023-09-08
add delete endpoint

akihikokuroda:
### Summary
Delete file part of #934


client/quantum_serverless/core/files.py
```diff
@@ -90,3 +90,17 @@ def list(self) -> List[str]:
...
             )
         return response_data.get("results", [])
+
+    def delete(self, file: str) -> Optional[str]:
+        """Deletes file uploaded or produced by the programs,"""
+        tracer = trace.get_tracer("client.tracer")
+        with tracer.start_as_current_span("files.delete"):
+            response_data = safe_json_request(
+                request=lambda: requests.get(
+                    f"{self.host}/api/{self.version}/files/delete/",
```
Tansito, line 100: I would like to propose here the next URI:

```
f"{self.host}/api/{self.version}/files/",
```

Similar to what we are doing with the `list` method. The difference would be the HTTP method and we would be doing an approximation to the API Rest URI nomenclature standard. We are not following it strictly because our API doesn't adjust to it but it's a good practice in general I would say.

gateway/api/views.py
```diff
@@ -1,4 +1,4 @@
-"""
+"""OA
```
Tansito, line 1: What it means `OA` here @akihikokuroda , just for curiosity.

Tansito (changes requested):

gateway/api/views.py
akihikokuroda: Thanks! It's nothing.  I put it by mistake. I take it out.

client/quantum_serverless/core/files.py
akihikokuroda: I tried it but it doesn't work because of the restriction of `viewsets.ModelViewSet` of `django`.

Tansito: Understood, no problem thanks @akihikokuroda 😄

Tansito (approved):
Good job here @akihikokuroda , from my part everything looks good 😄

IceKhan13 (approved):
Let's go! :)

PR #955, opened 2023-09-13
Renovate configuration for bitnami packages

### LINKED ISSUE #950: Update bitnami packages at the same time

From https://github.com/bitnami/charts/issues/18933 bitnami's team suggested us that all bitnami packages should be updated at the same time to avoid problems with the `bitnami/common` library that all charts make use of it.

Tansito:
### Summary

Fix #950 , configuration to try to update bitnami helm releases at the same time to avoid problems with `bitnami/common` lib. More info in #949 .

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -19,6 +19,13 @@
         "^qiskit"
       ],
       "groupName": "qiskit packages"
+    },
+    {
+      "matchPackagePatterns": [
+        "nginx-ingress-controller",
+        "keycloak"
+      ],
+      "groupName": "bitnami packages"
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


renovate.json
```diff
@@ -4,7 +4,7 @@
     "config:base",
     ":separateMultipleMajorReleases"
   ],
-  "prConcurrentLimit": 5,
+  "prConcurrentLimit": 4,
```
Tansito, line 7: Reduced the number or concurrent pull request opened by renovate now that we could clean-up some `on-hold`.

psschwei (approved):

Issue #946, opened 2023-09-08
job.result sometimes returns an empty dict when using save_result


caleb-johnson:
The following remote program sometimes returns an empty dictionary as the result

```
# program.py
from quantum_serverless import save_result

my_list = [1, 2, 3, 4]
save_result({"results": my_list})
```

<Job | b90b6513-4a93-451b-871f-39681b26cbb7> returned an empty dictionary

akihikokuroda:
This one is difficult to fix :-)

akihikokuroda:
This seems a concurrency issue between the gateway and the scheduler for the job record.

akihikokuroda:
I believe what's happening is:

1. the scheduler gets the job record for the job https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/management/commands/update_jobs_statuses.py#L22
2. the gateway gets the result request and update the job record with it https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/views.py#L145
3. the scheduler detects the job status change (PENDING -> SUCCEDED) https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/management/commands/update_jobs_statuses.py#L45
4. the scheduler updates the job record with the new status -- **this override the result field of the job record**. https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/management/commands/update_jobs_statuses.py#L57

caleb-johnson:
> I believe what's happening is:
> 
> 1. the scheduler gets the job record for the job https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/management/commands/update_jobs_statuses.py#L22
> 2. the gateway gets the result request and update the job record with it https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/views.py#L145
[...]

PR #956, opened 2023-09-13
prevent job record override and retry

akihikokuroda:
### Summary
Fix: #946 

### Details and comments
When the job record gets updated with the result by the gateway, the scheduler skips updating the job recode with the new status and lets the next cycle update the record.


IceKhan13:
Should we add retry or something like that here (save result API) too? 

https://github.com/Qiskit-Extensions/quantum-serverless/blob/db829650b2b9339d650326a84c30257e4848bcbc/gateway/api/views.py#L144

akihikokuroda:
@IceKhan13 The retry is done as a part of the next scheduler cycle.  So we don't need to add any extra code to do it.

akihikokuroda:
@IceKhan13 You are right.  I'll add retry there.

akihikokuroda:
@IceKhan13 The save result API could write the result anyway before.  By adding the `version = IntegerVersionField()` of 'concurrency` in the Job model, it can start raising exceptions.

Tansito (changes requested):
@akihikokuroda @IceKhan13 I was taking a look to the issue and the PR and I would like to propose another solution for this. 

My idea would be to use `updated_fields` to solve the problem:

https://docs.djangoproject.com/en/4.2/ref/models/instances/#specifying-which-fields-to-save

From what I understood the problem is that `save` overrides the entry because we are mostly saving the entire dataset. My proposal would be to just save only those required fields in each section to avoid that undesired override (in this case the `solution`field).

Something like:

```
# in update_jobs_statuses.py line 59
job.save(update_fields=["logs", "env_vars", "status"])

# in views.py line 149
job.save(update_fields=["result"])
```

I see several benefits with this solution versus the proposed one. The first one is that we don't need to play with a new `version` field in the model that I find confuse. And the use of `saved` and the `while` in the middle of the logic that difficult the readability in a code that from my point of view is getting quite complex to follow.

akihikokuroda:
@Tansito It seems good idea.  I can trash this PR and create a new one.  @IceKhan13 WDYT?

Tansito:
I tested this pull request and It's working. I can't say the same from my proposal 😂 

I would say that we can maintain this pull request open in case the another one doesn't solve the problem, Aki. But as you prefer, what it's more confortable for you.

IceKhan13:
@akihikokuroda let's leave this PR and create another one

IceKhan13:
I will just put it on hold for now. If other implementation will work, we can close this one :)

akihikokuroda:
#959 is used to fix the issue.  Closing this.

PR #959, opened 2023-09-18
save only selected Job fields

akihikokuroda:
### Summary
Fix: #946 

### Details and comments
This is another way to fix the concurrent issue.   This doesn't use the django-concurrency module (unlike #956).  There is still an issue with the "status" field.  There are some possibilities that the `Job.stop()` call is override.


IceKhan13 (approved):
It looks like everything is working!
Thank you Aki!

Issue #947, opened 2023-09-08
Filter out ray logs


Ibrahim-Shehzad:
### What is the expected enhancement?

Either ray logs should be filtered out by default or the user should be given the option to filter out them out. In particular, I was getting this in my logs Handshake failed with fatal error SSL_ERROR_SSL: error:1000009c:SSL routines:OPENSSL_internal:HTTP_REQUEST. and was not sure if I could ignore it. If this is something that can be safely ignored, not having it show up in the logs in the first place would be helpful.

akihikokuroda:
closing as done by #963

PR #963, opened 2023-09-19
add filter option to logs

akihikokuroda:
### Summary
fix: #947 

### Details and comments
This adds filteredLogs function that filters the log before return to the caller.  It uses rex expressing for include and exclude the log lines.

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -30,6 +30,7 @@ Quantum serverless job
 import json
 import logging
 import os
+import re
 import tarfile
 import time
 from pathlib import Path
@@ -92,6 +93,9 @@ class BaseJobClient:
         """Return logs."""
         raise NotImplementedError
 
+    def filteredLogs(self, job_id: str, **kwargs):
+        raise NotImplementedError
+
     def result(self, job_id: str):
         """Return results."""
         raise NotImplementedError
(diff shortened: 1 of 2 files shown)


IceKhan13:
I'm wondering, can we somehow disable those logs on a ray side? 🤔

akihikokuroda:
I was looking for a way to change the default logging level of the Ray but I couldn't find so far.  The default is `INFO`.  If we can find a way to change it, we can eliminate some log entries.

psschwei:
looks like we can configure the ray logger, but it requires a bit of config: https://docs.ray.io/en/releases-2.5.1/ray-observability/user-guides/configure-logging.html#using-rays-logger

IceKhan13:
> looks like we can configure the ray logger, but it requires a bit of config: https://docs.ray.io/en/releases-2.5.1/ray-observability/user-guides/configure-logging.html#using-rays-logger

Hm, maybe we should go with this option? 

If I'm understanding it correctly it require to specify logger for driver process and for tasks. But how can we hide it from user and do it automatically? 🤔 

We can hide all logs setup behind something like
```python
QuantumServerless.setup()
```

where behind the scenes we will do 
```python
def logger_setup():
     logger = logging.getLogger("ray")
    ...

class QuantumServerless:
     def setup(cls):
         logger_setup()
         ray.init({"setup_hook": logger_setup})

```

But it is still extra call by user. Hmmm 🤔

IceKhan13 (approved):
Let's merge this one and we can think how to move it to cluster configs later

PR #964, opened 2023-09-20
Issue 946 | Optimistic lock

### LINKED ISSUE #946: job.result sometimes returns an empty dict when using save_result

IceKhan13:
### Summary

Optimistic lock implementation based on https://github.com/Qiskit-Extensions/quantum-serverless/pull/956

### Details and comments

Looks like https://github.com/Qiskit-Extensions/quantum-serverless/pull/959 did not fixed an issue.
It looks like `update_fields` only changes query, but commit is still happening at the same time, therefore causing error. 

Closes #946 

Additionally  removed ray client logs call from `/logs` view, since logs are updated in management commands.


gateway/api/utils.py
```diff
@@ -165,3 +167,36 @@ def decrypt_env_vars(env_vars: Dict[str, str]) -> Dict[str, str]:
...
+            else:
+                model.save()
+            saved = True
+        except RecordModifiedError:
+            logger.warning(
+                "Model[%s] record has not been updated due to lock. Retrying. Attempts left %s",
+                model,
+                attempts_left,
+            )
+            continue
```
akihikokuroda, line 200: I believe that the model record must be refetched and updates re-applied to the fields to resolve `RecordModifiedError` exception.

IceKhan13: aaaah, good to know!

akihikokuroda: I think that the record need to be refetched in this loop, too.   

My understanding of the this concurrency is that it check if the version of the record in the db is same as the one to be saved.  The version is updated every time the record is saved.  If the version don't match, somebody has updated the the record while the recode is fetched and being saved.

akihikokuroda: I haven't verified but I was wondering if the `version` field need to be added for the `save()` call to make the `concurrency` work.

IceKhan13: I think I will remove this loop

IceKhan13: > I haven't verified but I was wondering if the version field need to be added for the save() call to make the concurrency work.

and remove update fields too, I think

IceKhan13: updated

akihikokuroda (approved):
LGTM  Thanks!

PR #969, opened 2023-09-22
Patch: lock exception

### LINKED ISSUE #946: job.result sometimes returns an empty dict when using save_result

IceKhan13:
### Summary

Yet another patch for lock.

### Details and comments

Fixes #946

### DIFF
--- a/gateway/api/management/commands/schedule_queued_jobs.py
+++ b/gateway/api/management/commands/schedule_queued_jobs.py
@@ -1,7 +1,10 @@
 """Cleanup resources command."""
 import json
 import logging
+import os
+import time
 
+from concurrency.exceptions import RecordModifiedError
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand
@@ -10,7 +13,7 @@ from django.db.models import Model
 from opentelemetry import trace
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 
-from api.models import ComputeResource
+from api.models import ComputeResource, Job
 from api.schedule import get_jobs_to_schedule_fair_share, execute_job
 
 User: Model = get_user_model()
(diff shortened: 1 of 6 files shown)


psschwei:
close / reopen to pull in latest changes for k8s test

IceKhan13:
hm ,notebooks cannot connect to server still 🤔

psschwei:
Strange, it was working fine in https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/6279180485 ... let's rerun and see (maybe it was something temporary)

psschwei:
ahh, should've looked further down... the gateway pod isn't ready:

```
--- default pods ---
NAME                                READY   STATUS             RESTARTS      AGE
gateway-59b44bcf6d-cdskc            0/1     CrashLoopBackOff   4 (19s ago)   3m49s
keycloak-0                          1/1     Running            0             3m49s
kuberay-operator-6c498d996f-ftgfb   1/1     Running            0             3m49s
postgresql-0                        1/1     Running            0             3m49s
prepuller-5fjrq                     1/1     Running            0             3m49s
scheduler-55dc447bdd-cdt5f          1/1     Running            0             3m49s
```

psschwei:
the restarts drop the logs, so going to work on debugging the failure in #972 ... but I wouldn't consider that blocking for this PR (i.e. feel free to override the k8s test)

akihikokuroda:
One question.  Is the 'version' field of the 'job' updated after successful `job.save()`?

IceKhan13:
@akihikokuroda yes, it basically [auto_update](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.DateField.auto_now)

akihikokuroda:
I see.  Did you remove some `job.save()` because they are redundant?

IceKhan13:
yes, they were redundant as we return object back. So, we only need to save in one place.

gateway/api/management/commands/schedule_queued_jobs.py
```diff
@@ -67,6 +70,38 @@ def handle(self, *args, **options):
...
+                        attempts -= 1
+
+                        try:
+                            job.save()
+                            # remove artifact after successful submission and save
+                            if os.path.exists(job.program.artifact.path):
+                                os.remove(job.program.artifact.path)
 
+                            succeed = True
+                        except RecordModifiedError:
```
akihikokuroda, line 92: Sorry for silly questions.  I'm still trying to understand.  Does this exception happen when the `update_job_statuses` update the `job'?

IceKhan13: yes, those exception might happen anywhere where we have `job.save()`

Tansito (approved):

gateway/api/management/commands/schedule_queued_jobs.py
Tansito: I don't see it ideal but by now I understand the fix at least. But yeah, I would like to see a TODO to improve it or something.

PR #973, opened 2023-09-25
Issue 316

scottmason88:
### Summary
Added coverage of execute_job condition where a ray cluster already exists. Since execute_job calls submit_job, coverage is added here as well. Coverage up to 79 (with recent commit we're at 81)%

### Details and comments
I have a couple questions:

1) When I use test_json in the GET, I'm required to have a ray_version. I used 2.6.1 because that's the one I had on my computer, but I'm not sure if we're okay with hardcoding a version here. If we are great, otherwise let me know.

2) For the third mocker.get, there is what seems to be a generated zip file name and I wasn't able to find where it's generated or how it's generated. I was also unable to find the path (api/packages/gcs...) when I did a search of the code. I assume I can mock the name of the zip if I know what's generating that, but I can't seem to find it.  Where would it be?

### DIFF
--- a/gateway/tests/api/test_schedule.py
+++ b/gateway/tests/api/test_schedule.py
@@ -48,7 +48,7 @@ class TestScheduleApi(APITestCase):
 
     @patch("tarfile.open")
     @patch("tarfile.os.path.basename")
-    @patch("api.schedule.os.remove", return_value=0)
+    @patch("os.remove", return_value=0)
     @patch.object(uuid, "uuid4", return_value="")
     def test_already_created_ray_cluster_execute_job(
         self, mock_open, mock_basename, mock_uuid, mock_remove


psschwei:
> When I use test_json in the GET, I'm required to have a ray_version. I used 2.6.1 because that's the one I had on my computer, but I'm not sure if we're okay with hardcoding a version here. If we are great, otherwise let me know.

we have v2.6.3 in the client requirements but are using v2.5.0 in the ray node dockerfile... I'd say go with v2.6.3, and maybe open an issue about standardizing the ray versions (if there isn't already one for it)

scottmason88:
> > When I use test_json in the GET, I'm required to have a ray_version. I used 2.6.1 because that's the one I had on my computer, but I'm not sure if we're okay with hardcoding a version here. If we are great, otherwise let me know.
> 
> we have v2.6.3 in the client requirements but are using v2.5.0 in the ray node dockerfile... I'd say go with v2.6.3, and maybe open an issue about standardizing the ray versions (if there isn't already one for it)

All set

psschwei (approved):

IceKhan13 (approved):
That is great! Thank you so much! 

Over 80% coverage is very nice!

Issue #895, opened 2023-08-21
Program proxy for Qiskit Runtime requests


IceKhan13:
### What is the expected behavior?

We need to have a sidecar that proxy all QiskitRutnime requests and associate runtime jobs to Program.

Benifits:
- associate primitives with programs
- when user kills program all associated primitives are also killed

#### Details

One way to implement this is to add list field `primitives_executions` to `Job` model and add primitives job ids to that list. Population of this field will be happening in sidecar (basically within sidecar we will be hitting `api/v1/jobs/<job_id>/primitives` endpoint).

When users run `job.stop()` from client we stop all associated primitives to that job. 

Epic https://github.com/Qiskit-Extensions/quantum-serverless/issues/1162

akihikokuroda:
How can the primitive execution be hacked in the sidecar?

IceKhan13:
Qiskit runtime is an Rest api service. It looks like client for runtime (`QiskitRUntimeService`) is using  `QISKIT_IBM_URL` env variable to detect where to send requests. 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/83091ee488113f68a0fe101e435d9628131dca44/qiskit_ibm_runtime/accounts/management.py#L247

We can use this env variable to send request not to runtime, but to our sidecar which will be sending requests to runtime service and preserve ids of executed services.

For example inside of sidecar container we will have simple http service, something like this pseudocode:
```python

@route("/")
def handle_runtime_request(request):
    # send request to runtime
    runtime_service_response = requests.post(url="RUNTIME_URL", data=request.data)

    # get primitive job id and save it to job 
    primitive_job_id = runtime_service_response.get("job_id")
    gateway_response = requests.post(
        "api/v1/jobs/<program_job_id>/primitives", 
        data={"primitive_id": primitive_job_id})
    return response
```

we still need to think how to pass program_job_id 🤔

IceKhan13:
probably this issue can be splitted into 2:
- gateway: add associated primitives ids to job
- implement sidecar service

akihikokuroda:
we may use [tracestate](https://www.w3.org/TR/trace-context/#tracestate-header) for `program_job_id`  propagation.

akihikokuroda:
I wonder if this feature should be in the middleware.  Should Qiskit or providers be responsible to the association between the program and primitives?

akihikokuroda:
[...]

PR #977, opened 2023-09-25
add primitive_job_id field in Job

akihikokuroda:
### Summary
A part of : #895 

### Details and comments
Add primitive job id field to Job model.  It keeps the list of primitive job ids in a string format.
Add "program job id" to ray runtime environment so that the primitive request can include it.

```
env = json.loads(os.environ.get("RAY_JOB_CONFIG_JSON_ENV_VAR"))
estimator.run(ansatz, hamiltonian, options={"job_id": env["runtime_env"]["quantum_middleware_job"]}, parameter_values=params).result().values[0]
```


gateway/tests/api/test_job.py
```diff
@@ -88,3 +88,40 @@ def test_stop_job(self):
...
+            reverse(
+                "v1:jobs-add-primitive", args=["1a7947f9-6ae8-4e3d-ac1e-e7d608deec82"]
+            ),
+            format="json",
+            data={"primitive": "primitive-id-2"},
+        )
+        self.assertEqual(jobs_response.status_code, status.HTTP_200_OK)
+        self.assertEqual(
+            jobs_response.data.get("primitive_job_id"),
+            '"primitive-id","primitive-id-2"',
```
IceKhan13, line 116: should we store it as json list? 
`["primitive_id_1", "primitive_id_2"]`

client/quantum_serverless/core/job.py
```diff
@@ -96,6 +96,14 @@ def result(self, job_id: str):
         """Return results."""
         raise NotImplementedError
 
+    def primitive(self, job_id: str):
```
IceKhan13, line 99: Why do we need this functions?

gateway/tests/api/test_job.py
```diff
@@ -88,3 +88,40 @@ def test_stop_job(self):
         ).first()
         self.assertEqual(job.status, Job.STOPPED)
         self.assertEqual(job_stop_response.data.get("message"), "Job has been stopped.")
+
+    def test_job_add_primitive(self):
```
IceKhan13, line 92: ah, I love tests 😇

gateway/tests/api/test_job.py
akihikokuroda: I'll change it.

client/quantum_serverless/core/job.py
akihikokuroda: Oh, yes, It is not necessary.  I'll take it out.

PR #1011, opened 2023-10-04
Change Provider class name to ServerlessProvider

### LINKED ISSUE #989: Rename Provider to ServerlessProvider
It has been suggested to renamed `Provider` to `ServerlessProvider` to avoid confusion with other providers in the Qiskit ecosystem. This issue is to discuss that topic.

caleb-johnson:
Fixes #989 

In this PR, we rename ``Provider`` to ``ServerlessProvider``, and we use the new naming convention in all the notebooks in the docs. We also deprecate ``Provider``.

### DIFF
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -320,6 +320,7 @@ class ServerlessProvider(BaseProvider):
         password: password
         token: authorization token
     """
+
     def __init__(
         self,
         name: Optional[str] = None,
@@ -362,13 +363,19 @@ class ServerlessProvider(BaseProvider):
         self._files_client = GatewayFilesClient(self.host, self._token, self.version)
 
     def get_compute_resources(self) -> List[ComputeResource]:
-        raise NotImplementedError(f"GatewayProvider does not support resources api yet.")
+        raise NotImplementedError(
+            f"GatewayProvider does not support resources api yet."
+        )
 
     def create_compute_resource(self, resource) -> int:
-        raise NotImplementedError(f"GatewayProvider does not support resources api yet.")
+        raise NotImplementedError(
+            f"GatewayProvider does not support resources api yet."
+        )
 
     def delete_compute_resource(self, resource) -> int:
-        raise NotImplementedError(f"GatewayProvider does not support resources api yet.")
+        raise NotImplementedError(
+            f"GatewayProvider does not support resources api yet."
+        )
 
     def get_job_by_id(self, job_id: str) -> Optional[Job]:
         return self._job_client.get(job_id)
@@ -428,6 +435,7 @@ class IBMServerlessProvider(ServerlessProvider):
     Args:
         token: IBM quantum token
     """
+
     def __init__(self, token: str):
         super().__init__(token=token, host=IBM_SERVERLESS_HOST_URL)
 
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -30,6 +30,7 @@ import logging
 import os.path
 from dataclasses import dataclass
 from typing import Optional, List, Dict, Any
+from abc import ABC, abstractmethod
 
 import ray
 import requests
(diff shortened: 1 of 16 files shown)


client/quantum_serverless/core/provider.py
```diff
@@ -147,7 +147,27 @@ def __repr__(self):
 
 
 class BaseProvider(JsonSerializable):
-    """Provider."""
+    """
+    The base class for Quantum Serverless providers.
+
+    Example:
+        >>> provider = ServerlessProvider(
```
psschwei, line 154: nit
```suggestion
        >>> provider = BaseProvider(
```

psschwei: also, since these args are for the init method, should we leave that docstring there? (I'm not 100% on the python best practices here, so not sure what the answer is)

caleb-johnson: Ya, I'm conflicted here, since this code won't run. Maybe I just move this example to ``ServerlessProvider``. I could duplicate it and remove ``host`` for ``IBMServerlessProvider``. It's a little redundant, but more correct than this.

caleb-johnson: Regarding the args, I took a quick glance at `qiskit.QuantumCircuit`, and they have their args there. It's probably better to just keep things consistent with the rest of the serverless repo. I can move the examples and docstrings back inside the init methods.

client/quantum_serverless/core/provider.py
```diff
@@ -308,8 +308,18 @@ def widget(self):
         return Widget(self).show()
 
 
-class Provider(BaseProvider):
-    """Provider."""
+class ServerlessProvider(BaseProvider):
+    """
+    A provider class for connecting to any specified host.
```
caleb-johnson, line 313: remove "class"

client/quantum_serverless/core/provider.py
psschwei: > Maybe I just move this example to ServerlessProvider

I think that makes sense, especially as it'll show on the site in the API reference: https://qiskit-extensions.github.io/quantum-serverless/stubs/quantum_serverless.core.BaseProvider.html

though, looking at that link, it looks like the current version has the same issue...

psschwei: maybe we update the docstring for `BaseProvider` to somehow note that it isn't implemented (?)

client/quantum_serverless/core/provider.py
```diff
@@ -308,8 +308,18 @@ def widget(self):
         return Widget(self).show()
 
 
-class Provider(BaseProvider):
-    """Provider."""
+class ServerlessProvider(BaseProvider):
```
IceKhan13, line 311: We need to deprecate it first before removing, since it is one of the most used class.

Something like
```python
class ServerlessProvider(BaseProvider):
    ...

class Provider(ServerlessProvider):
    def __init__(....):
         warnings.warn(
              "`Provider` class has been deprecated and will be removed in release 0.8.0. Please, use `ServerlessProvider` instead.", 
              DeprecationWarning)
         self.super(...)

```

caleb-johnson: Will do

client/quantum_serverless/core/provider.py
caleb-johnson: @psschwei , I've turned `BaseProvider` into a proper abstract base class and flagged the class methods which should be implemented by subclasses. I've moved args down into constructors with a generic constructor description and left the examples and class descriptions up at the class level. Let me know if this looks good to you

caleb-johnson: @IceKhan13 , any reason `BaseProvider` is used directly all throughout the [quantum_serverless](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/quantum_serverless/quantum_serverless.py) module? I feel like this should be an abstract base class, and that `ServerlessProvider` should be the most general, usable class.

This would be a breaking change, since you're already using this within `quantum_serverless`, so I hesitate to do it. I could either raise a deprecation warning inside the `BaseProvider` constructor and we could make it an abstract class later, or I could do a breaking change now to make it abstract. I only suggest the breaking change because this class isn't featured in any of our docs, and it isn't clear to me you ever intended for users to instantiate this class. It seems `ServerlessProvider`/`Provider` give you everything you need.

caleb-johnson: Ahh, I can't make it abstract. It has some extra functionality for specifying a compute resource that `ServerlessProvider` doesn't provide. Standby...

IceKhan13: `ServerlessProvider/Provider` is to connect to gateway service. 
If somebody will want to implement Azure or AWS provider, they can do that by extending base provider. 

`BaseProvider` is only used for type hints and as base class for `Provider`.

Code is using base provider in some shared cases, I think. But I agree it might need some love. I would argue though it is out of the scope of renaming pr :)

IceKhan13: this PR should be rename/reprecate. For refactor we will need separate PR which will need more test time

IceKhan13: or, keep this refactor PR and open `just rename PR` :)

IceKhan13: Yeah, we will do it in separate PR abstract and that is it :)

IceKhan13 (approved):
Let's go! :)

PR #1016, opened 2023-10-05
Add metadata to job client, including job runtime

### LINKED ISSUE #851: Gateway: add runtime metadata to Jobs

Multiple users requested for simple metadata about jobs, like runtime. 

We can add signal to `Job` model to report `runtime` when job is finished.

caleb-johnson:
Fixes #851 

This PR establishes a metadata field in the client job class. We will start by including the job runtime in the job model and using that field to populate a client job metadata field. We could continue adding items to the metadata in this way as needed

I think the correct way to do this is to expose the relevant fields in the job model (e.g. `Model.Job.runtime`) and add it to the client job as metadata (e.g. `QuantumServerless.core.Job.metadata['runtime'] = Model.Job.runtime`).

### DIFF
--- a/gateway/api/migrations/0001_initial.py
+++ b/gateway/api/migrations/0001_initial.py
@@ -90,6 +90,7 @@ class Migration(migrations.Migration):
                 ),
                 ("created", models.DateTimeField(auto_now_add=True)),
                 ("result", models.TextField(blank=True, null=True)),
+                ("runtime", models.DurationField(null=True)),
                 (
                     "status",
                     models.CharField(
--- a/gateway/api/migrations/0001_initial.py
+++ b/gateway/api/migrations/0001_initial.py
@@ -90,7 +90,8 @@ class Migration(migrations.Migration):
                 ),
                 ("created", models.DateTimeField(auto_now_add=True)),
                 ("result", models.TextField(blank=True, null=True)),
-                ("runtime", models.DurationField(null=True)),
+                ("start_time", models.DateTimeField(null=True)),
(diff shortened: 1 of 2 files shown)


caleb-johnson:
@IceKhan13 , open to specific implementation ideas.

Like I wrote in the PR explanation, I figure we expose the fields we need in the job model, and we create a metadata field in the client job class and fill it out with the fields we create in the job model. Does that seem right?

psschwei:
Thinking out loud: could we add just a single `metadata` field to the class and use it as a key/value store for all the various metadata fields? If we put them all at the top level, that might get a little unwieldy (depending on how many we end up with)

caleb-johnson:
> Thinking out loud: could we add just a single `metadata` field to the class and use it as a key/value store for all the various metadata fields? If we put them all at the top level, that might get a little unwieldy (depending on how many we end up with)

The way I see it is that we can either continuously poll the job somewhere in order to find out when it stops, or we can leverage some kind of push notification of when a job stopped. I don't really have a good idea of how to do either of those things very elegantly right now.

For now, I am just trying to figure out if I can write a date/time to the job model when a job is submitted to the backend (I think I can). I'm also trying to figure out if I can do the same when the job hits a terminal state (not sure how to do this without either continuous polling or some sort of push notification)

psschwei:
I think we added traces to the jobs, which should signal when a job starts / stops.

IceKhan13:
ref: https://docs.djangoproject.com/en/4.2/topics/signals/

IceKhan13:
Closing, since original issue is closed and we do not have any follow ups on this.

PR #1021, opened 2023-10-06
Issue 1013 | Client: Ray provider

### LINKED ISSUE #1013: Client: ray provider

Implement ray provider that directly connects to Ray.

IceKhan13:
### Summary

Implementation of Ray provider to connect directly to ray head node

### Details and comments

Closes #1013

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -131,21 +131,13 @@ class RayJobClient(BaseJobClient):
 
     def run(self, program: Program, arguments: Optional[Dict[str, Any]] = None):
         arguments = arguments or {}
-        arguments_string = ""
-        if arguments is not None:
-            arg_list = []
-            for key, value in arguments.items():
-                if isinstance(value, dict):
-                    arg_list.append(f"--{key}='{json.dumps(value)}'")
-                else:
-                    arg_list.append(f"--{key}={value}")
-            arguments_string = " ".join(arg_list)
-        entrypoint = f"python {program.entrypoint} {arguments_string}"
+        entrypoint = f"python {program.entrypoint}"
 
         # set program name so OT can use it as parent span name
         env_vars = {
             **(program.env_vars or {}),
             **{OT_PROGRAM_NAME: program.title},
+            **{"ENV_JOB_ARGUMENTS": json.dumps(arguments, cls=QiskitObjectsEncoder)},
         }
 
         job_id = self._job_client.submit_job(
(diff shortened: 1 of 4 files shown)


client/quantum_serverless/core/job.py
```diff
@@ -131,21 +132,13 @@ def list(self, **kwargs) -> List["Job"]:
...
-            for key, value in arguments.items():
-                if isinstance(value, dict):
-                    arg_list.append(f"--{key}='{json.dumps(value)}'")
-                else:
-                    arg_list.append(f"--{key}={value}")
-            arguments_string = " ".join(arg_list)
-        entrypoint = f"python {program.entrypoint} {arguments_string}"
+        entrypoint = f"python {program.entrypoint}"
 
         # set program name so OT can use it as parent span name
```
psschwei, line 137: nit: update comment as we're not just updating program name anymore (?)

client/quantum_serverless/core/job.py
```diff
@@ -131,21 +132,13 @@ def list(self, **kwargs) -> List["Job"]:
...
         env_vars = {
             **(program.env_vars or {}),
             **{OT_PROGRAM_NAME: program.title},
+            **{ENV_JOB_ARGUMENTS: json.dumps(arguments, cls=QiskitObjectsEncoder)},
```
psschwei, line 141: non-blocking question: is this envvar currently being used? (the github search wasn't super helpful...)

psschwei (approved):

client/quantum_serverless/core/job.py
IceKhan13: yes, in `get_arguments` function

client/quantum_serverless/core/job.py
IceKhan13: what do you mean? :)

psschwei: github UI :anger: 

that referred to the comment on L137: it only refers to setting the program name, but now we're also setting job arguments...

client/quantum_serverless/core/job.py
psschwei: didn't phrase this well, but what I was wondering about if it's being used to set how the program runs ( I probably should've put this on the line where `entrypoint` was defined... ), but looking at it a bit more, it seems like this is just moving the ray client more in line with how the other clients take arguments

IceKhan13: exactly :)

client/quantum_serverless/core/job.py
IceKhan13: ah! :) now I understand. You are right, comment needs to be updated/removed. Will update it in any of next PRs :)

PR #1025, opened 2023-10-11
expose number of workers

akihikokuroda:
### Summary
expose number of workers option

### Details and comments
The number of worker can be specified in the config dictionary argument for QuantumServerless.run function.
```
)
program = Program(
)

job = serverless.run(program, config={"worker": 2})
```

The contents of the `config` are probably provider dependent.  So this uses dictionary instead of defining the `Configuration` class

Right now, the kuberay autoscaling is not enabled.   It is possible put the autoscaling option with min/max worker in the config dictionary.


psschwei (commented):
Should we put this setting as part of the `distributed_program()` config? 

ex: https://qiskit-extensions.github.io/quantum-serverless/getting_started/experimental/running_programs_using_decorators.html#Program-with-distributed-tasks

Since we currently let users set CPU/memory there, it might be nice to also configure the number of workers in the same place. That said, does `distributed_program` actually spin up a new ray cluster, or is it too late once someone has already started running their program? 

(as you may be able to tell, this is more me thinking out loud than anything else, so take it for what it's worth :smile:  )

akihikokuroda:
@psschwei Yes, it's too late.  The cluster must be there to run the program.  We probably should expose the autoscaling option so that the cluster scales based on the resources requested in the program.

gateway/api/models.py
```diff
@@ -35,6 +35,7 @@ class Program(ExportModelOperationsMixin("program"), models.Model):
     arguments = models.TextField(null=False, blank=True, default="{}")
     env_vars = models.TextField(null=False, blank=True, default="{}")
     dependencies = models.TextField(null=False, blank=True, default="[]")
+    config = models.TextField(null=False, blank=True, default="[]")
```
IceKhan13, line 38: Can we make this config option as another model that will be connected to `Program`. Something like
```python
class ProgramConfig(models.Model):
     PY_39: str = "PY_39"
     python_version = models.ChoiceField([PY_39, ...])
     min_workers=model.IntegerField(...)
     ...

class Program(...):
    ...
    config = models.ForeignField(ProgramConfig, on_delete=CASCADE, null=True)
```

This will enable schema verification out of the box. And we will avoid unstructured jsons

akihikokuroda: @IceKhan13 Thanks for review.  I could make a model but this configuration is provider dependent so it may not be good to define a model if it’s not common among providers. Does it make sense?  WDYT?

IceKhan13: on client side each provider can have different configuration, but for gateway (backend) and gateway provider we need to enforce this, I think.

akihikokuroda:
@IceKhan13 Should the `ProgramConfig` in the `Program` or the argument of `serverless.run`?


```
program = Program(
    config=ProgramConfig(num_workers=2)
)

job = serverless.run(program)
```
or 
```
program = Program(
)

job = serverless.run(program, config=ProgramConfig(num_workers=2))
```

akihikokuroda:
rewoek in #1051

PR #1027, opened 2023-10-12
Add python version number config option

akihikokuroda:
### Summary
Add python version number config option

### Details and comments
This PR depends on #1025 

The python version can be specified in the config dictionary argument for QuantumServerless.run function.

)
program = Program(
)

job = serverless.run(program, config={"pythonversion": "py38"})


Tansito:
Is it possible to use "python_version" following the rules that we use for the rest of the variables?

PR #1043, opened 2023-10-19
Issue 1040 | Gateway: update logs

### LINKED ISSUE #1040: Job logs should be updated as the code is executed

Currently the the job logs do not update until a program is finished.  This means one cannot track the progress of a function as it goes through its steps.  As functions can get arbitrarily long, it would be nice to be able to rack the status of them as they go through.

IceKhan13:
### Summary

Bugfix.
Update logs regardless of status.

### Details and comments

Closes #1040


akihikokuroda (approved):
LGTM.  Thanks!

psschwei (approved):
nit: the comment in L21 probably should be tweaked as we're returning logs independent of job status changes now (at some point, we may potentially possibly want to see about splitting these two tasks out, but not now...)

PR #1048, opened 2023-10-20
Allow IBMServerlessProvider users to save their credentials to disk

### LINKED ISSUE #1042: Add a Provider.save_account method
Users should be able to save/load their provider tokens to/from their local environment. This allows them to more easily instantiate their serverless provider in their code. The interface should likely be similar to other IBM interfaces, such as Qiskit Runtime [IBMProvider.save_account](https://qiskit.org/ecosystem/ibm-provider/stubs/qiskit_ibm_provider.IBMProvider.save_account.html#qiskit_ibm_provider.IBMProvider.save_account).

Outstanding questions:
- Should this method be included in the `BaseProvider` class, or should only some providers implement it?
    - Only IBMServerlessProvider

caleb-johnson:
Fixes #1042

### DIFF
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -502,9 +502,10 @@ class IBMServerlessProvider(ServerlessProvider):
         # If no env variables contain the token, try to get it from the Qiskit IBM config file
         if token is None:
             name = name or "qiskit-ibm-default"
-            _ensure_file_exists(self._default_account_config_json_file)
+            _ensure_file_exists(IBMServerlessProvider._default_account_config_json_file)
             with open(
-                self._default_account_config_json_file, encoding="utf-8"
+                IBMServerlessProvider._default_account_config_json_file,
+                encoding="utf-8",
             ) as json_file:
                 data = json.load(json_file)
             if name in data:
--- a/client/quantum_serverless/core/provider.py
+++ b/client/quantum_serverless/core/provider.py
@@ -480,31 +480,32 @@ class IBMServerlessProvider(ServerlessProvider):
         provider = IBMServerlessProvider(token=<INSERT_IBM_QUANTUM_TOKEN>)
     """
 
-    def __init__(self, token: Optional[str] = None):
+    def __init__(self, token: Optional[str] = None, name: Optional[str]):
         """
         Initialize a provider with access to an IBMQ-provided remote cluster.
 
         Args:
             token: IBM quantum token
+            name: Name of the account to load
         """
         self._default_account_name_ibm_quantum = "default-ibm-quantum"
         self._default_account_config_json_file = os.path.join(
             os.path.expanduser("~"), ".qiskit", "qiskit-ibm.json"
         )
-        if token is None:
-            token = os.environ.get(ENV_GATEWAY_PROVIDER_TOKEN)
 
-        if token is None:
-            token = os.getenv("QISKIT_IBM_TOKEN")
+        # Try to read token from env variables
+        token = os.environ.get(ENV_GATEWAY_PROVIDER_TOKEN) or os.getenv("QISKIT_IBM_TOKEN")
 
+        # If no env variables contain the token, try to get it from the Qiskit IBM config file
         if token is None:
+            name = name or self._default_account_name_ibm_quantum
             self._ensure_file_exists(self._default_account_config_json_file)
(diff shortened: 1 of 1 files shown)


IceKhan13 (commented):
Should we use what QiskitRuntimeService already have? 🤔 

https://github.com/Qiskit/qiskit-ibm-runtime/blob/dc3b1d0d4a8f9426726e5dd9cc74d6373fd68709/qiskit_ibm_runtime/qiskit_runtime_service.py#L683

then implementation would be something like this:
```python
class IBMServerlessProvider(...):
    def __init__(self, token: Optional[str] = None, name: Optional[str] = None):
         if token is None:
              self.token = QiskitRuntimeService(name=name).token
         else:
              self.token = token

    @classmethod
    def save_account(cls, token: Union[str, QiskitRuntimeService], overwrite: bool = True):
           if isinstance(token, QiskitRuntimeService):
               QiskitRuntimeService.save_acoount(**token.get_account())
           else:
               # token string
               QiskitRuntimeService.save_account(token=token, channel=..., ...)
```

In that case it will be same file as runtime is using. 

But in general I'm ok with any option :)

caleb-johnson:
> Should we use what QiskitRuntimeService already have? 🤔
> 
> https://github.com/Qiskit/qiskit-ibm-runtime/blob/dc3b1d0d4a8f9426726e5dd9cc74d6373fd68709/qiskit_ibm_runtime/qiskit_runtime_service.py#L683
> 
> then implementation would be something like this:
> 
> ```python
> class IBMServerlessProvider(...):
>     def __init__(self, token: Optional[str] = None, name: Optional[str] = None):
>          if token is None:
>               self.token = QiskitRuntimeService(name=name).token
>          else:
>               self.token = token
> 
>     @classmethod
>     def save_account(cls, token: Union[str, QiskitRuntimeService], overwrite: bool = True):
>            if isinstance(token, QiskitRuntimeService):
>                QiskitRuntimeService.save_acoount(**token.get_account())
>            else:
>                # token string
>                QiskitRuntimeService.save_account(token=token, channel=..., ...)
> ```
> 
> In that case it will be same file as runtime is using.
> 
> But in general I'm ok with any option :)

Yes, I think this is a better solution. Let me take a look..

client/quantum_serverless/core/provider.py
```diff
@@ -455,19 +455,58 @@ class IBMServerlessProvider(ServerlessProvider):
...
-    def __init__(self, token: str):
+    def __init__(self, token: Optional[str] = None, name: Optional[str] = None):
         """
         Initialize a provider with access to an IBMQ-provided remote cluster.
 
         Args:
             token: IBM quantum token
+            name: Name of the account to load
         """
+        token = token or IBMProvider(name=name).active_account().get("token")
```
caleb-johnson, line 490: If the user passes a token, everything is normal and `name` is ignored.

If a user only passes a name, it will retrieve the token from that account.  I don't believe a user can save an account without passing a token, so it should be "guaranteed" to be there.

If a user passes neither argument, `IBMProvider` will check env variables and the default IBM Quantum json file location for credentials. An error is raised if it can't find a token in either location.

If a user passes a name of an account which doesn't exist, `IBMProvider` will raise an error.

client/quantum_serverless/core/provider.py
```diff
@@ -455,19 +455,58 @@ class IBMServerlessProvider(ServerlessProvider):
...
+    ) -> None:
+        """
+        Save the account to disk for future use.
+
+        Args:
+            token: IBM Quantum API token
+            name: Name of the account to save
+            overwrite: ``True`` if the existing account is to be overwritten
+        """
+        IBMProvider.save_account(token=token, name=name, overwrite=overwrite)
```
caleb-johnson, line 508: Here we have a subset of input arguments used in `IBMProvider.save_account`. The other arguments did not seem relevant to our software, so I left them out. We will just pass the inputs from this function "blindly" into `IBMProvider.save_account` and let `IBMProvider` handle the error messaging.

client/quantum_serverless/core/provider.py
IceKhan13: So much better! :) if we put it in the doc that would be it :)

IceKhan13 (approved):
Looks clean! :) 

Feel free to merge after docstring update

PR #1051, opened 2023-10-23
Add worker config options

### LINKED ISSUE #1031: Expose ability to set min/max number of workers per cluster

Expose ability to set min/max number of workers per cluster.

Requirements:
- Defaults should be configurable in gateway. 
- Min and max should have verification rules in gateway, so user would not be able to set crazy numbers (positive integers within boundaries).

From client it should look something along those lines:
```python
job = serverless.run(
  program,
  config=Configuration(
    workers=[Worker(2, 4)] # 2 cpu, 4 GB mem
  )
)
```

akihikokuroda:
### Summary
Fix #1031 

This PR is rework of PR #1025 

### Details and comments
Summary
expose number of workers option (number of workers, min workers and max workers when auto scaling is enabled)

Details and comments
The number of worker can be specified in the Configuration object for QuantumServerless.run function.

)
program = Program(
)

job = serverless.run(program, config=Configuration(workers=2, min_workers=1, max_workers=5, auto_scaling=True))


akihikokuroda:
I still need to work with tests for this but please take a look.  Thanks!

client/quantum_serverless/core/program.py
```diff
@@ -76,6 +76,29 @@ def from_json(cls, data: Dict[str, Any]):
...
+
+    Args:
+        title: program name
+        entrypoint: is a script that will be executed as a job
+            ex: job.py
+        env_vars: env vars
+        dependencies: list of python dependencies to execute a program
+        working_dir: directory where entrypoint file is located (max size 50MB)
+        description: description of a program
+        version: version of a program
```
psschwei, line 91: should the args here be workers, cpu, etc.?

psschwei (commented):
general question on workers vs. min/max workers : does one ever override the other (i.e. if we set workers, does that invalidate min/max scaling)? 

I didn't check if there's a ray doc that covers this, so if there is feel free to tell me to go read the #$%#$%# docs :smile:

akihikokuroda:
@psschwei The `workers` is used when the auto scaling is not enabled.  The min/max workers are used when the auto scaling is enabled.

Tansito:
I was a bit out these last weeks so maybe this question is a bit dump so sorry in advance (and maybe this may be discussed in the issue). Is it has sense to let the user to modify the min/max number of workers? And if so, I see valuable to have an option that prevent to do that in case the user doesn't want to allow that kind of behavior from external users. Opinions: @IceKhan13 , @psschwei , @akihikokuroda ? (maybe a dump question, as I said 😅 ).

client/quantum_serverless/core/program.py
akihikokuroda: The configuration is specified in `run` method of `Quantum_serverless` 
```
serverless.run(program, config=Configuration(max_workers=12, min_workers=5)
```

I'll update the doc string of the function

akihikokuroda:
@Tansito The max allowed values for the workers are specified in the settings.py (read from env variables).  So users are not allowed to set values bigger than those.

psschwei:
>  let the user to modify the min/max number of workers

it came up during the internal tests

Tansito:
> @Tansito The max allowed values for the workers are specified in the settings.py (read from env variables). So users are not allowed to set values bigger than those.

I see so we are going to allow only modifications inside a threshold. Thanks @akihikokuroda ❤️ 

Happy to return to the work, I will try to catch-up and return 100% as soon as I can haha. Thanks team!

psschwei:
> @psschwei The `workers` is used when the auto scaling is not enabled. The min/max workers are used when the auto scaling is enabled.

so, for example, something like this

```python
serverless.run(program, config=Configuration(workers=2, min_workers=1, max_workers=5, auto_scaling=True))
```

would ignore the `workers=2` since auto_scaling is set to true, right?   

We should probably add a doc on program configuration (to go along with the one for client configuration) to spell out all the options.

gateway/api/ray.py
```diff
@@ -186,6 +189,24 @@ def create_ray_cluster(
     raycluster_client = dyn_client.resources.get(
         api_version="v1alpha1", kind="RayCluster"
     )
+    if program_config:
```
IceKhan13, line 192: can we do this in jinja template? 

it will be something like
```python
manifest = cluster.render(
  {
    "cluster_name": cluster_name, 
    "user_id": user.username,
    "workers": config.workers,
    ...
  }
)
```

gateway/api/schedule.py
```diff
@@ -13,26 +14,60 @@
...
 
     Args:
-        serializer: program serializer with data attached.
+        request: request data.
 
     Returns:
         saved program
     """
-    program = Program(**serializer.data)
+    if request.data.get("config"):
```
IceKhan13, line 34: model serializer should be able to handle this. I do not think you need any of this. 
You already implemented serializer, which is doing mapping between request data and Program object and all related object (which is config)

gateway/api/views.py
```diff
@@ -86,11 +87,10 @@ def run(self, request):
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.program.run", context=ctx):
-            serializer = self.get_serializer(data=request.data)
```
IceKhan13, line 89: serializer should handle all related entities, so you do not need to created them.

gateway/main/settings.py
```diff
@@ -287,12 +287,24 @@
     ),
 }
 RAY_CLUSTER_WORKER_REPLICAS = int(os.environ.get("RAY_CLUSTER_WORKER_REPLICAS", "0"))
+RAY_CLUSTER_WORKER_REPLICAS_MAX = int(
```
IceKhan13, line 290: I remember we had a `readme` doc with all settings. I cannot find it for some reason 🤔 
If would be great to have all this settings specified in a document

gateway/api/schedule.py
akihikokuroda: I tried but I couldn't make it work with `models.ForeignKey` field.  I'll try it again.

client/quantum_serverless/core/job.py
```diff
@@ -203,21 +215,34 @@ def run(  # pylint: disable=too-many-locals
...
+                data = {
+                    "title": program.title,
+                    "entrypoint": program.entrypoint,
+                    "arguments": json.dumps(arguments or {}, cls=QiskitObjectsEncoder),
+                    "dependencies": json.dumps(program.dependencies or []),
+                }
+                if config:
+                    config_string = json.dumps(
+                        {
+                            "workers": config.workers,
```
IceKhan13, line 227: dataclasses support `asdict` function.
you can do something like `json.dumps(asdict(config))

gateway/main/settings.py
akihikokuroda: I'll look for it.

gateway/api/schedule.py
IceKhan13: this might help https://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers

gateway/api/ray.py
akihikokuroda: I remember I tried jinja before and it didn't work.  I don't remember the reason.  I'll try if I can make it work here.

akihikokuroda: I found that I found the solution of my issue before.  I change this to use ninja template.

gateway/api/views.py
akihikokuroda: I restore the use of the ProgramSerializer.  Thanks!

client/quantum_serverless/core/job.py
```diff
@@ -1,4 +1,4 @@
-# This code is a Qiskit project.
+#  This code is a Qiskit project.
```
psschwei, line 1: nit
```suggestion
# This code is a Qiskit project.
```

gateway/api/ray.py
```diff
@@ -176,7 +179,14 @@ def create_ray_cluster(
...
         cluster = get_template("rayclustertemplate.yaml")
         manifest = cluster.render(
-            {"cluster_name": cluster_name, "user_id": user.username}
+            {
+                "cluster_name": cluster_name,
+                "user_id": user.username,
+                "workers": program_config.workers,
+                "min_workers": program_config.min_workers,
+                "max_workers": program_config.max_workers,
+                "auto_scaling": program_config.auto_scaling,
```
psschwei, line 188: What happens if bad values are passed? For example, a user sets min=5 and max=2... what kind of error message would they see? Is that caught by Ray or do we need to catch it? (I haven't tried yet, but based on past experience users will pass all kinds of combinations of bad values when configuring...)

akihikokuroda: I tried it with max=1, min=2.  Ray didn't care, it just created 1 worker.

akihikokuroda:
I have to add the config to the existing program path.

psschwei:
LGTM

IceKhan13:
Looks great! I will just compile and test it once more and it will be a candidate for 0.7! :)

IceKhan13:
@akihikokuroda sorry, I finally reviewing this PR :) 

I have a discussion suggestion.
In this PR we are connecting `Config` to a `Program` model, but when user is running it with client it is connected to job. 
Should we connect `Config` to `Job` model and then `job = serverless.run("...", config=...)` will save config in db for this particular job and will not overwrite `Program` field? Or we should leave `Config` connected to `Program`, which will mean that this program will require this amount of resources `program = Program("...", config=Config(....))`?

WDYT @akihikokuroda , @psschwei , @Tansito ?

akihikokuroda:
It probably makes sense attaching the config to the `Job` instead of `Program` for the current items(workers, min/max_workers, etc..).  We also need to add python version which may make sense to attach to the `Program`, though.

IceKhan13:
I think you are right @akihikokuroda .

JobConfig will be something that job is using during runtime. I would remove python version from this PR for now and only focus on worker config here.

psschwei:
+1 for tying the config to the job... folks may want to run their program with different configs at different times, and that should help make the program (more) reusable...

Tansito:
Agree to attach the configuration to a Job!

IceKhan13:
Woohoo! `Config to Job connection` flawless victory!

akihikokuroda:
ProgramConfig has been changed to JobConfig and it is related to Job.  Please review.  Thanks!

IceKhan13:
I'm on it! Thank you Aki!

gateway/api/models.py
```diff
@@ -13,6 +17,39 @@ def get_upload_path(instance, filename):
...
+
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    created = models.DateTimeField(auto_now_add=True)
+
+    auto_scaling = models.BooleanField(default=False, null=True)
+    workers = models.IntegerField(
+        null=True,
+        validators=[
+            MinValueValidator(0),
+            MaxValueValidator(settings.RAY_CLUSTER_WORKER_REPLICAS_MAX),
```
IceKhan13, line 31: configuration values in models will create conflict. If I change my settings now it will generate new migration diff and django will be complaining about migrations.

gateway/api/ray.py
```diff
@@ -157,7 +157,10 @@ def submit_job(job: Job) -> Job:
 
 
 def create_ray_cluster(
-    user: Any, cluster_name: Optional[str] = None, cluster_data: Optional[str] = None
+    user: Any,
+    cluster_name: Optional[str] = None,
+    cluster_data: Optional[str] = None,
+    job_config: Optional[JobConfig] = None,
```
IceKhan13, line 163: we accept optional job_config here, but later on we always act like it is not None.

Here we should check for None and fall back to default from settings.

gateway/api/views.py
```diff
@@ -150,12 +157,18 @@ def run(self, request):
             if not serializer.is_valid():
                 return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+            try:
```
IceKhan13, line 160: If we replace it with

```python
jobconfig = None
config_data = request.data.get("config")
if config_data:
    config_serializer = JobConfigSerializer(data=json.loads(config_data))
    if not config_serializer.is_valid():
        return Response(config_serializer.errors, status=status.HTTP_400_BAD_REQUEST)

    jobconfig = config_serializer.save()
```

then we resolve validation + errors will be in a nice understandable format 

```
QuantumServerlessException: 
| Message: Http bad request.
| Code: 400
| Details: {"workers":["Ensure this value is less than or equal to 5."]}
```

and we will not longer need `save_jobconfig` function.

IceKhan13 (changes requested):
Almost there :) 

Remaining things to fix:
- remove settings variables from models and migrations
- move validation to serializer [ref](https://www.django-rest-framework.org/api-guide/validators/#validation-in-rest-framework)
- write tests for serializer and api view with configuration

akihikokuroda:
@IceKhan13 Thanks for review.  I'll make chages.

IceKhan13:
No, thank you for handling such a big change!

akihikokuroda:
There remaining fixes have done.  Please take a look!  Thanks!

IceKhan13 (approved):
LGTM!

Thank you @akihikokuroda for working on it!

akihikokuroda:
Thanks!

IceKhan13:
Reopened https://github.com/Qiskit-Extensions/quantum-serverless/pull/1085

PR #1052, opened 2023-10-25
Issue 1035 | Deprecate QuantumServerless class

### LINKED ISSUE #1035: Deprecate `QuantumServerless` class

Deprecate `QuantumServerless` class. 

Basically `QuantumServerless` class is a complete copy in terms of methods of `ServerlessProvider`. We can just leave providers and have providers run jobs.

```python
from quantum_serverless import ServerlessProvider

provider = ServerlessProvider(...)
provider.run(program, ...)
```

IceKhan13:
### Summary

Deprecate `QuantumServerless` class and update documentation with `build` -> `deploy` -> `run` steps

### Details and comments

Closes #1035


psschwei (approved):
Is the upload step now required to run a program? Or does the old workflow still work (declare program and pass it as an argument to the run function)?

IceKhan13:
@psschwei old way still works. There is actually not that much of a change happened. We just exposed upload as a separate step explicitly.

PR #1057, opened 2023-10-27
Client + docs: rename `Program`

IceKhan13:
### Summary

Client + docs: rename `Program`


client/quantum_serverless/core/program.py
```diff
@@ -72,21 +74,58 @@ class Program:  # pylint: disable=too-many-instance-attributes
...
+            version,
+            tags,
+            raw_data,
+        )
+
+    def __repr__(self):
+        return f"Program({self.title})"
+
+
 class ProgramStorage(ABC):
```
psschwei, line 125: Do we need to `s/Program/Pattern/` here (and elsewhere in this class, i.e. `save_program` and the like)?
Or :shudder: change the filename itself from program to pattern?

IceKhan13: good point!

psschwei (approved):
Should this also get tagged as a feature branch? there's a new class being added / an old one being deprecated, so _technically_ ... (?)

docs/getting_started/basic/02_arguments_and_results.ipynb
```diff
@@ -186,20 +186,9 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 5,
+   "execution_count": null,
```
psschwei: do you want to execute this cell? (looks like the rest of them were execute...)

docs/getting_started/experimental/running_programs_using_decorators.ipynb
```diff
@@ -216,65 +216,43 @@
...
     "\n",
-    "job = my_program_with_modules()\n",
+    "job = my_pattern_with_modules()\n",
     "job"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 7,
+   "execution_count": null,
```
psschwei, line 252: do you want to execute this cell? (think there might also one other in this notebook that wasn't executed...)

client/quantum_serverless/core/job.py
```diff
@@ -166,11 +168,13 @@ def run(self, program: Program, arguments: Optional[Dict[str, Any]] = None):
         )
         return Job(job_id=job_id, job_client=self)
 
-    def upload(self, program: Program):
+    def upload(self, program: QiskitPattern):
```
Tansito, line 171: This can a breaking change so it's a bit tricky but should we call it `pattern`?

Tansito (approved):
I left a couple of comments and I found a couple of references in the documentation that maybe we can review but I would merge it and do next week another pass, what do you think, @IceKhan13 ?

IceKhan13:
> Should this also get tagged as a feature branch? there's a new class being added / an old one being deprecated, so technically ... (?)

This is more of a docs change + name alias. It does not change any functionality

client/quantum_serverless/core/job.py
psschwei: can you overload functions in python? (i.e. have one that takes a program and mark it deprecated and another that takes a pattern)

Tansito: good question!

IceKhan13:
We will follow with updates in next PR :)

client/quantum_serverless/core/job.py
psschwei: seems like answer is no (though there are a few workarounds if you search around on the topic...)

PR #1059, opened 2023-10-30
Issue 1058 | Gateway: allow nested fields verification

### LINKED ISSUE #1058: Custom auth: allow multiple fields for verification

Currently we only support one field for verification. We need to a allow multiple fields.

https://github.com/Qiskit-Extensions/quantum-serverless/blob/e60d0eec6eb384ce81bb97414e66b479f15fea1d/gateway/api/authentication.py#L75

IceKhan13:
### Summary

Allow nested fields verification from auth API

### Details and comments

Now if verification API has structure something like that
```json
{
  "is_valid": true,
  "some": {
    "nested": {
      "field": true
    },
    "other": "bla"
  }
}
```

We can verify agains it with following settings:
```python
"SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD", "is_valid;some,nested,field"
```
---

Back compatible PR

Closes #1058

### DIFF
--- a/gateway/main/settings.py
+++ b/gateway/main/settings.py
@@ -272,7 +272,18 @@ SETTINGS_TOKEN_AUTH_VERIFICATION_URL = os.environ.get(
 )
 # verification fields to check when returned from auth api
 # Example of checking multiple fields:
-#   "SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD", "is_valid;some,nested,field"
+#    For following verification data
+#    {
+#       "is_valid": true,
+#       "some": {
+#         "nested": {
+#           "field": true
+#         },
+#         "other": "bla"
+#       }
+#    }
+#   setting string will be:
+#    "SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD", "is_valid;some,nested,field"
 SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD = os.environ.get(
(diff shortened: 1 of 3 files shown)


IceKhan13:
If we need to check for `other` too
```
"SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD", "is_valid;some,nested,field;some,nested,other"
```

akihikokuroda (approved):
LGTM

gateway/main/settings.py
```diff
@@ -270,6 +270,9 @@
 SETTINGS_TOKEN_AUTH_VERIFICATION_URL = os.environ.get(
     "SETTINGS_TOKEN_AUTH_VERIFICATION_URL", None
 )
+# verification fields to check when returned from auth api
+# Example of checking multiple fields:
+#   "SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD", "is_valid;some,nested,field"
```
psschwei, line 275: Could we add the example JSON to the comment here too? I found it helpful when trying to understand how to properly format the string for this field.

psschwei (approved):

gateway/main/settings.py
IceKhan13: Good suggestion!

Added in https://github.com/Qiskit-Extensions/quantum-serverless/pull/1059/commits/a720041c2280b64bd39227f60ff86cf084754678

PR #1062, opened 2023-10-30
Translate all job statuses to Qiskit terminology in client

### LINKED ISSUE #1039: Job completed status should match that of Qiskit
Currently a completed job has the status `'SUCCEEDED'` where as the rest of Qiskit uses `'DONE'`: https://qiskit.org/documentation/stable/0.37/stubs/qiskit.providers.JobStatus.html

It would be nice to be uniform here.

caleb-johnson:
Fixes #1039 

### Summary
In this PR, we translate job statuses from all job clients to Qiskit terminology. Rather than changing the DB structure (currently uses ray status names), we will just do a translation in the serverless client when the user requests the status.

**Serverless --> Qiskit nomenclature changes**

Pending --> Initializing
Stopped --> Canceled
Succeeded --> Done
Failed --> Error


gateway/api/migrations/0001_initial.py
```diff
@@ -94,14 +94,14 @@ class Migration(migrations.Migration):
                     "status",
                     models.CharField(
                         choices=[
-                            ("PENDING", "Pending"),
+                            ("INITIALIZING", "Initializing"),
```
IceKhan13, line 97: Migrations should not be changed. They are append only.
Since they are changing structure of schema of database, they can only modify state of db that was there before. 

You need to create new migration that will change schema from this types of statuses to new ones.

IceKhan13 (changes requested):
Also, when changing records in existing db you need to take care of entries in DB that are already there and convert them too. 

---
Other hotfix options would be just to do mapping in client, so you will not need to change anything on a backend and just show in client mapped statuses. Something like
```python
def gateway_to_runtime_statuses(gateway_status):
     mapping = {
         "SUCCEEDED": "DONE"
     }
     return mapping.get(gateway_status, f"APIStatus[{gateway_status}]")
```

IceKhan13:
Basically what migrations are doing:
- they are checking in which state db is (which migration version is applied to DB recenty)
- if there are upapplied migrations, they are applied
- if any of previous migrations changed, migration will fail

gateway/api/migrations/0001_initial.py
caleb-johnson: I think handling the migration on client side, like you suggested, is a nice solution here. I don't even think it's really that hacky. We leave everything on the backend in Ray language for clarity, and we just convert to our terminology on client side. Makes a lot of sense to me.

I can also generate a new migration file if you like that solution better.

client/quantum_serverless/core/job.py
```diff
@@ -127,7 +127,7 @@ def __init__(self, client: JobSubmissionClient):
         self._job_client = client
 
     def status(self, job_id: str):
-        return self._job_client.get_job_status(job_id)
+        return self._job_client.get_job_status(job_id).value
```
caleb-johnson, line 131: Since `status` is an unimplemented `BaseJobClient` method, I think both the ray and gateway job clients should return a string indicating the status. I updated this in this PR because it affects how I do the status translation in the user-facing job client

client/quantum_serverless/core/job.py
```diff
@@ -473,7 +473,7 @@ def __init__(
 
     def status(self):
         """Returns status of the job."""
-        return self._job_client.status(self.job_id)
+        return _map_status_to_serverless(self._job_client.status(self.job_id))
```
caleb-johnson, line 476: We will do the translation from any supported job client status to Qiskit terminology here in the user-facing `Job` client. This is the only place we'll do the translation. Everything underneath this will be Ray terminology.

client/quantum_serverless/core/job.py
```diff
@@ -496,7 +496,7 @@ def result(self, wait=True, cadence=5, verbose=False):
         if wait:
             if verbose:
                 logging.info("Waiting for job result.")
-            while not self._in_terminal_state():
+            while not self.in_terminal_state():
```
caleb-johnson, line 499: We don't have a `JobStatus` object like ray that we can query like Ray's `JobStatus.is_terminal`, so we will make this `Job.in_terminal_state` method public, so clients have a way to query whether the job has finished.

caleb-johnson:
@IceKhan13 , I've fixed this up as you suggested. I don't currently have an idea about this strange sphinx error, but Im looking into it. Doesn't seem to be related to the work in this PR, so it's a little confusing

caleb-johnson:
> @IceKhan13 , I've fixed this up as you suggested. I don't currently have an idea about this strange sphinx error, but Im looking into it. Doesn't seem to be related to the work in this PR, so it's a little confusing

Hmm, I fixed it by removing `Program` from the init file, but it is still not clear to me why that error occured in this PR.

client/quantum_serverless/core/__init__.py
```diff
@@ -39,7 +39,6 @@
     RayJobClient
     save_result
     QiskitPattern
-    Program
```
caleb-johnson, line 42: This is causing Sphinx errors for me since it is no longer imported in this file, not sure how main branch is passing sphinx build with no warnings?

client/quantum_serverless/core/job.py
```diff
@@ -723,3 +723,20 @@ def save_result(result: Dict[str, Any]):
...
+        "STOPPED": "CANCELED",
+        "SUCCEEDED": "DONE",
+        "FAILED": "ERROR",
+        "QUEUED": "QUEUED",
+    }
+
+    try:
+        return status_map[status]
+    except KeyError as exc:
+        raise KeyError(f"Cannot interpret unknown job status: {status}") from exc
```
caleb-johnson, line 742: Maybe we shouldn't error, maybe we should just print the status like you suggested? Not really sure

IceKhan13:
> Hmm, I fixed it locally by removing Program from the docstring in the init file, but the error is still happening in CI. Can't reproduce this locally after removing Program from the docstring.

merge main and you will be good :)

client/quantum_serverless/core/job.py
IceKhan13: I think printing is less invasive :)

client/quantum_serverless/core/job.py
IceKhan13: so, you basically do not want to return status class? and return status string?
I have no objections, just to confirm :)

IceKhan13:
PR looks good! simple and clean like all great things 😃

client/quantum_serverless/core/job.py
caleb-johnson: I figured since the gateway job client returns a string, we may as well return a string from the ray job client as well, since they both derive from the same base class. Not a big deal, but I thought it was a clean simplification.

IceKhan13 (approved):
LGTM

PR #1067, opened 2023-11-01
Issue 1028 | Docs: make sphinx build local docs + add save result notes

### LINKED ISSUE #1028: Docs: add direction on `save_result` types support

Docs: add direction on `save_result` types support.

`save_result` only supports python dict with specific types.

IceKhan13:
### Summary

Docs: add save result notes

### Details and comments

Closes #1028

### DIFF
--- a/client/quantum_serverless/core/__init__.py
+++ b/client/quantum_serverless/core/__init__.py
@@ -39,7 +39,6 @@ Core abstractions
     RayJobClient
     save_result
     QiskitPattern
-    Program
     ProgramStorage
     ProgramRepository
     download_and_unpack_artifact
@@ -75,6 +74,7 @@ from .job import (
 )
 from .pattern import (
     QiskitPattern,
+    Program,
     ProgramStorage,
     ProgramRepository,
     download_and_unpack_artifact,
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,9 @@ skip_missing_interpreters = true
 
 [testenv:docs]
 skip_install = false
+install_command =
+  pip install -U {opts} {packages}
+  pip install -e ./client
 setenv =
     PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python
 deps =
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -672,7 +672,7 @@ def save_result(result: Dict[str, Any]):
 
     Links:
         Source of serializer:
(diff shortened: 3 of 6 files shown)


IceKhan13:
Screenshot 2023-11-01 at 5 04 19 PM

IceKhan13:
Screenshot 2023-11-01 at 5 08 33 PM

psschwei (approved):

tox.ini
```diff
@@ -8,6 +8,9 @@ skip_missing_interpreters = true
 
 [testenv:docs]
 skip_install = false
+install_command =
```
caleb-johnson, line 11: Adding `Program` back as importable from the `core` package is right, but do we want the hosted docs to reflect the status of `main` branch? I thought we decided a while back it was better for docs to reflect the status of the latest release?

client/quantum_serverless/core/__init__.py
```diff
@@ -75,6 +74,7 @@
 )
 from .pattern import (
     QiskitPattern,
+    Program,
```
caleb-johnson, line 77: Thank you for fixing this, I think it is causing weird behavior in sphinx builds since we build from latest release.

tox.ini
IceKhan13: we just had an interesting problem because of this. Basically with older docs approach it was impossible to change documentation without a Pypi release. Because sphinx was building docs from sourcecode that is in PyPI.

client/quantum_serverless/core/__init__.py
IceKhan13: This weird behavior was due to we were building docs from release

tox.ini
caleb-johnson: Yes, but users will generally be installed from `PyPI`, and when they visit our docs, we want them to reflect that, right? `main` branch is more likely to change day-to-day, and I feel like users will be more likely to be working off a release.

The only reason there was an error was because the `Program` was removed as importable but remained in the docstring. It shouldve been cleaned up in both or remained in both.

It's up to you, but I do think building docs from PyPI is better than main.

IceKhan13: on top of that CI was doing basically nothing with old approach and was checking not sourcecode of repository, but what was on Pypi 😓

caleb-johnson: Another option might be to run CI on the `main` branch docs build, but continue building our `qiskit.org` docs from the release. If we did this, it would have caught the error caused by removing `Program` in one place but not the other. Just a thought :)

PR #1029, opened 2023-10-16
LocalProvider

akihikokuroda:
### Summary
Add LocalProvider

PR #1068, opened 2023-11-01
Re-add Program as importable

caleb-johnson:
#1029 accidentally removed `Program` from being importable at the `core` level, when it is still in its deprecation period.

This PR re-adds that module as importable from the `core` package until its deprecation period is over. If we remove it, we've broken users' interface, which defeats the purpose of deprecating.

### Details and comments

I believe this is manifesting in strange Sphinx errors in `tox` builds (like the one happening in this PR).


caleb-johnson:
Closing, as this will be resolved by #1067

PR #1075, opened 2023-11-03
Fix update version action

psschwei:

default checkout does not get all tags, which was causing the action to fail


IceKhan13 (approved):
let's go :)

PR #1077, opened 2023-11-06
Add python_version to the Configuration 

akihikokuroda:
### Summary

Fix: https://github.com/Qiskit-Extensions/quantum-serverless/issues/1034

### Details and comments

Depend on: https://github.com/Qiskit-Extensions/quantum-serverless/pull/1051
Override: https://github.com/Qiskit-Extensions/quantum-serverless/pull/1027

The python version is added into the JobConfig model.


gateway/api/models.py
```diff
@@ -13,6 +15,28 @@ def get_upload_path(instance, filename):
...
+    workers = models.IntegerField(
+        null=True,
+    )
+    min_workers = models.IntegerField(
+        null=True,
+    )
+    max_workers = models.IntegerField(
+        null=True,
+    )
+    python_version = models.TextField(null=True, blank=True)
```
IceKhan13, line 34: it's better to have `ChoiceField` with hardcoded versions. It will enable validation + migration out of the box. 

https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.Field.choices

gateway/api/ray.py
```diff
@@ -174,9 +177,36 @@ def create_ray_cluster(
...
+        if not job_config.max_workers:
+            job_config.max_workers = settings.RAY_CLUSTER_WORKER_MAX_REPLICAS
+        if not job_config.auto_scaling:
+            job_config.auto_scaling = settings.RAY_CLUSTER_WORKER_AUTO_SCALING
+        if not job_config.python_version:
+            job_config.python_version = settings.RAY_PYTHON_VERSION
+
+        py_version = job_config.python_version
+        node_image = (
+            settings.RAY_NODE_IMAGE[: settings.RAY_NODE_IMAGE.rindex("-")]
```
IceKhan13, line 195: I'm a bit worried about this. If we do not have `asdaskljdlaksjd-py39` naming convention in our images it will fail, right? 

Maybe in settings we should have a map of all available versions?
```python
# settings
...
RAY_NODE_IMAGE = os.environ.get("RAY_NODE_IMAGE", ...) # this is default version
RAY_NODE_IMAGES_MAP = {
  "PY_39": "supercool/image-py39",
  "PY_310": "...",
  ...
}
...

# api/ray.py
...
if job_config.python_version in  settings.RAY_NODE_IMAGES_MAP:
  node_image = settings.RAY_NODE_IMAGES_MAP[job_config.python_version]
else:
  logger.warn(f"Specified python version {job_config.python_version} "
    "not in a list of supported python versions {list(settings.RAY_NODE_IMAGES_MAP.keys())}. "
    "Default image will be used instead."
  )
  node_image = settings.RAY_NODE_IMAGE
...

```

What do you think?

akihikokuroda: > I'm a bit worried about this. If we do not have asdaskljdlaksjd-py39 naming convention in our images it will fail, right?

Right.  This assumes that the image tag has "-pyxxx" postfix.  I like the idea making the map.  The image names must be configureable in the values file.   I'll work on it.  Thanks!

akihikokuroda:
@IceKhan13 I made changes for both comments.  Thanks!

akihikokuroda:
It's merged now.

gateway/api/serializers.py
```diff
@@ -43,6 +44,9 @@ class Meta:
     auto_scaling = serializers.BooleanField(
         default=False, required=False, allow_null=True
     )
+    python_version = serializers.CharField(
```
IceKhan13, line 47: Should we use ChoiceField? https://www.django-rest-framework.org/api-guide/fields/#choicefield

gateway/api/migrations/0011_jobconfig_job_config.py
```diff
@@ -27,21 +26,19 @@ class Migration(migrations.Migration):
...
-                (
-                    "min_workers",
-                    models.IntegerField(
-                        null=True,
-                    ),
-                ),
-                (
-                    "max_workers",
-                    models.IntegerField(
+                    "python_version",
```
IceKhan13, line 33: I think, this should be a separate migration file, rather than changing existing one

gateway/main/settings.py
```diff
@@ -300,6 +300,14 @@
         "RAY_CLUSTER_MODE_LOCAL_HOST", "http://localhost:8265"
     ),
 }
+RAY_NODE_IMAGE = os.environ.get(
+    "RAY_NODE_IMAGE", "icr.io/quantum-public/quantum-serverless-ray-node:0.6.6-py39"
```
IceKhan13, line 304:
```suggestion
    "RAY_NODE_IMAGE", "icr.io/quantum-public/quantum-serverless-ray-node:0.7.1-py39"
```

gateway/api/migrations/0011_jobconfig_job_config.py
akihikokuroda: OK.  I thought it's better changing the file than adding another file because it's in the same model.  It doesn't matter.  I'll add another migration file.

IceKhan13: You are right, in this case it is not matter as we will introduce those changes together. But if we were introducing those changes separately, we might bump into a situation when previous migration file is applied and we introduced new changes in this file, that will cause migration conflict.

gateway/api/serializers.py
akihikokuroda: I realized that the `choicefield` is one of the serializer fields.  These is no choicefield in the model fields. https://docs.djangoproject.com/en/4.2/ref/models/fields/. I guess that it's better leaving it as the charfield.

IceKhan13: There is a choicefield in model fields and in serializer fields :)

IceKhan13: Here in models https://docs.djangoproject.com/en/4.2/ref/models/fields/#choices
here in serializers https://www.django-rest-framework.org/api-guide/fields/#choicefield

akihikokuroda: There is a `choices` field **option** that this change is using in the CharField.  But there is no 'ChoiceField' in the Models.  Am I missing?

IceKhan13: oh, sorry! you are right, I misinterpret that :)

IceKhan13: There is `CharField` with `choices` option in models (which you are using now) and `ChoicesField` in serializers.
I think `ChoicesField` in serializers is just to check on django side that value passed is correct before sending database query. I think we can can `ChoicesField` here as it is just for verification and value that will be serialized is string, which will be compatible with `CharField` in models.

akihikokuroda: I finally got it.  Thanks!

IceKhan13 (approved):
Great work, Aki! Thank you!

PR #1084, opened 2023-11-11
gateway scheduler refactor prototype

akihikokuroda:
### Summary
Prototype for https://github.com/Qiskit-Extensions/quantum-serverless/issues/1033

Recreate on top of current main branch.

### Details and comments
The changes are:

1. wrap the entry point sent to Ray. it sends a notification to the gateway when the entry point completes execution. It eliminates the polling of the ray job status.
2. put a signal listener for Job db changes. It can check if a Job to be scheduled and it some resources to be cleaned up.


akihikokuroda:
Should we go this direction or is there something better way? @IceKhan13 @Tansito @psschwei WDYT?

Tansito:
Thank you @akihikokuroda ! My idea with the proposal is to create a real queue and separate it from our database / application. I will try to have a diagram along this week to discuss it with you 🙏

psschwei:
> real queue

Do you mean setting up some application like RabbitMQ / Kafka ?

Tansito:
> Do you mean setting up some application like RabbitMQ / Kafka ?

If I can be honest I don't have it clear yet but definitely I think Kafka would be too much, probably. I was thinking in something nearer to Celery maybe. My core idea would be what I commented in the discussion basically, move us to a event/task/message pattern and reduce the pressure in the database to manage the queue to use RabbitMQ or Redis. It's a good effort but needed for the next year for production. That's the mean reason because I want to start conversations around this.

akihikokuroda:
I feel the real Q is too much at least for now.  This prototype is to remove the necessity of polling loop.

psschwei:
@akihikokuroda @Tansito is this one we're still looking to land?

Tansito:
Definitely not a priority right now with all the things that we have 😅 , It can be closed by now.

akihikokuroda:
I close.

PR #1096, opened 2023-11-15
Setup log level depending of the environment

### LINKED ISSUE #727: Responsive logging

From #719 and #723 we should be able to configure responsive logging depending of the environment where we are. `DEBUG` for `development`, `WARNING` for `production` and so on. 

More info here: https://docs.djangoproject.com/en/4.2/howto/logging/

Little example provided:
```python
"level": os.getenv("DJANGO_LOG_LEVEL", "WARNING")
```

Tansito:
### Summary

Fix #727 . This PR setup different levels of logging depending of the `DEBUG` mode.

### Details and comments

- For `DEBUG=1` the gateway will print all the logs available
- For `DEBUG=0` the gateway will print the logs from **"WARNING"** and above (ref: [here](https://docs.djangoproject.com/en/4.2/topics/logging/#loggers))

In this case instead to add a new environment variable to manage the logs independently I opted for a way where the log level depends on the `DEBUG` mode to avoid to add more environment variables.

### DIFF
--- a/gateway/main/settings.py
+++ b/gateway/main/settings.py
@@ -31,6 +31,9 @@ SECRET_KEY = os.environ.get(
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = int(os.environ.get("DEBUG", 1))
 
+# SECURITY WARNING: don't run with debug turned on in production!
+LOG_LEVEL = "DEBUG" if int(os.environ.get("DEBUG", 1)) else "WARNING"
+
 ALLOWED_HOSTS = os.environ.get("ALLOWED_HOSTS", "*").split(",")
 
 # allow connections from any kubernetes pod within the cluster
@@ -111,17 +114,17 @@ LOGGING = {
     },
     "root": {
         "handlers": ["console"],
-        "level": "DEBUG",
+        "level": LOG_LEVEL,
     },
     "loggers": {
         "commands": {
             "handlers": ["console"],
-            "level": "DEBUG",
+            "level": LOG_LEVEL,
             "propagate": False,
         },
         "gateway": {
             "handlers": ["console"],
-            "level": "DEBUG",
+            "level": LOG_LEVEL,
             "propagate": False,
         },
     },


psschwei:
Tangential question on logging that seeing "don't run debug in prod" made me think of: if we _did_ need to run debug in prod (say there's a major incident that we need to resolve), what would it take to enable debug logs? Would we have to restart the gateway (and if so, what are the consequences of doing that)?

psschwei (approved):
Is there ever a case where we'd want to run with log level other than DEBUG or WARNING? I'm going to assume not (you opened the issue, so you would know :smile: ), and if we did need to run in INFO it wouldn't be too difficult to update later on.

akihikokuroda:
It may be easier to understand to expose all log levels (instead of DEBUG 0 or 1) but this may be good for now because it requires minimum changes.

akihikokuroda (approved):

Tansito:
> Is there ever a case where we'd want to run with log level other than DEBUG or WARNING? I'm going to assume not (you opened the issue, so you would know 😄 ), and if we did need to run in INFO it wouldn't be too difficult to update later on.

Initially no, to be honest. Or I don't see an use case where we are going to need to see all the logs without activate the `DEBUG` mode. Other thing is what you commented above, it would be nice if we could activate or deactivate the `DEBUG` mode without restarting the application. But I think that's another conversation haha.

psschwei:
Assuming restarting gateway / scheduler is totally fine and wouldn't impact running jobs?

Tansito:
> Assuming restarting gateway / scheduler is totally fine and wouldn't impact running jobs?

Thinking in loud right now. So let's take my words carefully but the problematic points would be:
- If `RUNNING` job wants to `save` something. If the `gateway` is down the job is not going to be able to store it
- New ray-cluster of the user takes some time and we can get a Job in the middle of the process
- We interrupt in the middle of a logic the queue

We could find bad situations but it could affect to one job or so. It should not have bigger implications. The jobs running should not be affected by the restart of the gateway or the scheduler if they are just running in the ray cluster.

PR #1112, opened 2023-11-20
Program upload refactorization

Tansito:
### Summary

The idea of this PR is to propose a structure to start simplifying our `views`. We remove logic from them to `services` to improve readability and separation of concerns. In this case the implementation is the next one:
- `ProgramService` creation where we are going to add the business logic around `Programs`
- Versioning for `ProgramService`, this way we complement the `serializers` and the `views`
- Included a test for the new service and its method
- Included a set of custom exceptions to help us to manage tricky situations

Now the view it's only responsible to manage information with `Requests` & `Responses` delegating the rest of the logic into the `services`.

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -73,6 +73,14 @@ class ProgramViewSet(viewsets.ModelViewSet):  # pylint: disable=too-many-ancesto
 
         return JobSerializer
 
+    @staticmethod
+    def get_service_program_class():
+        """
+        This method returns Program service to be used in Program ViewSet.
+        """
+
+        return ProgramService
+
     def get_serializer_class(self):
         return self.serializer_class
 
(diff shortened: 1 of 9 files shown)


akihikokuroda (approved):
LGTM

gateway/api/views.py
```diff
@@ -173,7 +193,15 @@ def run(self, request):
...
-            program = save_program(serializer=serializer, request=request)
+            program_service = self.get_service_program_class()
+            try:
+                program = program_service.save(
+                    serializer=serializer,
+                    author=request.user,
+                    artifact=request.FILES.get("artifact"),
+                )
+            except InternalServerErrorException as exception:
+                return Response(exception, exception.http_code)
```
IceKhan13, line 204: Should we log exception and return string something like `Internal server error` instead of stack trace?

Tansito: Yeah, that's exactly what I'm doing there. I created a set of custom exceptions and `InternalServerErrorException` is one of them. [Here](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1112/files#diff-75b075598929b18be555b0fd14f9a1ea4472aad00cf2ab36d85044d3b20e179eR61) what I'm doing is catching the exception, logging it internally as an error and raising our custom exception to hide information to the user.

IceKhan13: OOOOH! looks great!

IceKhan13 (approved):
Looks clean! ✨

PR #1116, opened 2023-11-22
Gateway: fix note image selection

IceKhan13:
### Summary

Gateway: fix note image selection

### Details and comments

Currently if python version is not specified scheduler will try use `default` image in image map, which is not specified. This will lead to fall back to default behavior and log warning.


akihikokuroda (approved):
LGTM.  Thanks!

PR #1135, opened 2023-12-06
add h5 extension

### LINKED ISSUE #1113: Files download: support for h5 and other popular formats.

Currently we only support `tar` files download. Users requested for more file types support. 

We at least need to support `h5` and maybe other formats.

akihikokuroda:
### Summary
Fix #1113 

### Details and comments

This just adds `h5` file extension in the `files` api out.

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -328,7 +328,9 @@ class FilesViewSet(viewsets.ViewSet):
             user_dir = os.path.join(settings.MEDIA_ROOT, request.user.username)
             if os.path.exists(user_dir):
                 files = [
-                    os.path.basename(path) for path in glob.glob(f"{user_dir}/*.tar")
+                    os.path.basename(path)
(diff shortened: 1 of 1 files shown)


Tansito:
@akihikokuroda just to confirm, I'm not 100% sure. Did you try to do a migration?

```
python manage.py makemigrations
```

FYI: https://docs.djangoproject.com/en/4.2/topics/migrations/

I don't know if it applies when we add a new end of file but just to be sure.

akihikokuroda:
@Tansito I didn't do it.  This change doesn't affect the models so it should be fine.

Tansito:
Well, correct me if I'm wrong @akihikokuroda . Maybe this is not needed? We are not storing a QiskitPattern, this is just to upload files as resources, isn't it?

akihikokuroda:
The model attribute is for the tar file that has the program working directory contents created and sent to ray node internally.  It is not the file for the download.

PR #1140, opened 2023-12-07
adjust username included in cluster name

### LINKED ISSUE #1133: Gateway: invalid subdomain for user cluster if name is not valid

Username with special characters (like `_`).

 it is not valid name for subdomain and cluster creation will fail. 

Cluster should be created and username should not affect that

akihikokuroda:
### Summary
Fix #1133 

### Details and comments
The user name is used as a part of the cluster name.  This PR hashes the user name and uses the hex value of it as a part of the cluster name.

### DIFF
--- a/charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
+++ b/charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
@@ -13,6 +13,8 @@ data:
       labels:
         nodelete: "true"
 {{- end }}
+      annotations: 
+        user: {{`{{ user }}`}}
     spec:
 {{- if .Values.application.ray.scrapeWithPrometheus }}      
       headServiceAnnotations:
--- a/gateway/api/ray.py
+++ b/gateway/api/ray.py
@@ -1,4 +1,5 @@
 """Ray cluster related functions."""
+import hashlib
 import json
 import logging
 import os
@@ -175,7 +176,10 @@ def create_ray_cluster(
         or None if something went wrong with cluster creation.
     """
     namespace = settings.RAY_KUBERAY_NAMESPACE
-    cluster_name = cluster_name or f"{user.username}-{str(uuid.uuid4())[:8]}"
+    hashed_username = hashlib.shake_128(user.username.encode("utf-8"))
+    cluster_name = (
+        cluster_name or f"c-{hashed_username.hexdigest(6)}-{str(uuid.uuid4())[:8]}"
+    )
     if not cluster_data:
         if not job_config:
             job_config = JobConfig()
--- a/gateway/api/ray.py
+++ b/gateway/api/ray.py
@@ -1,5 +1,4 @@
 """Ray cluster related functions."""
-import hashlib
 import json
 import logging
 import os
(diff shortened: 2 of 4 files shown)


Tansito:
@akihikokuroda I see it perfect the logic, could you mind to move the logic into a function in utils or something like that to have it more centralized?

akihikokuroda:
@psschwei  The hash logic generates always same value for a user.   So we can map the user to the the cluster.

psschwei:
My only worry about mapping is that it'll add an extra step when trying to view logs / monitoring... for example, if there's a user having a problem, someone would need to hash their ID and then search for that. Or if some pod was causing a problem, then we'd have to de-hash it to figure out who it is. Not a deal breaker, but would be nice not to have to do that. 

(thinking out loud) would it be possible to add the username as an additional label on the pod? Or to strip out any illegal characters from the username?

akihikokuroda:
@psschwei I added 
```
  kind: RayCluster
  metadata:
    annotations:
      user: mockuser
```
in the raycluster CRD instance.

psschwei:
One other thought that just came to me now: what if we check to see if the username is valid, and then only hash if not? In other words, only hashing if we have to, to catch invalid names. Or just dropping invalid characters? that might let us keep the username in the pod but not have its name cause a problem...

________________________________________
From: Akihiko (Aki) Kuroda ***@***.***>
Sent: Thursday, December 7, 2023 5:42 PM
To: Qiskit-Extensions/quantum-serverless
Cc: Paul Schweigert; Mention
Subject: [EXTERNAL] Re: [Qiskit-Extensions/quantum-serverless] hash username in cluster name (PR #1140)

@psschwei I added kind: RayCluster metadata: annotations: user: mockuser in the raycluster CRD instance. — Reply to this email directly, view it on GitHub, or unsubscribe. You are receiving this because you were mentioned. Message ID: 
ZjQcmQRYFpfptBannerStart
This Message Is From an External Sender
This message came from outside your organization.
<https://us-phishalarm-ewt.proofpoint.com/EWT/v1/PjiDSg!1w-t6ZYUIZEZFIyzUiqN5b6dYFXTRi7ZNPt3dETcADeBIYb7qypLWpypMGAP-NOP4gSk_GBeEe1OkdKKk_OlsOyW4hBLdl3KLEPWUqb2BPqOFsSAAfGfJ-tU8uCSr_A$>
Report Suspicious

ZjQcmQRYFpfptBannerEnd

@psschwei<https://github.com/psschwei> I added

  kind: RayCluster
  metadata:
    annotations:
      user: mockuser


in the raycluster CRD instance.

Tansito:
I hadn't read the messages yet.

> what if we check to see if the username is valid, and then only hash if not? Or just dropping invalid characters?

We are not storing usernames, e-mails or sensitive data. That should be only the id from IQP (or the custom token user that it is a very specific use-case). My two cents are for validate and drop or replace invalid characters. Agree.

akihikokuroda:
I changed to replace all not allowed characters in the user name to "-".

psschwei (approved):
LGTM
Sorry for all the back and forth on this!

Tansito (approved):
Thank you so much, Aki 😄

PR #1146, opened 2023-12-11
pick up some client config from env var

### LINKED ISSUE #1064: Config values should be configurable without having to reinstall

We have some values in constants.py that admins may want to update after deployment, but currently that would require reinstalling. For example:

* https://github.com/Qiskit-Extensions/quantum-serverless/blob/v0.7.0/client/quantum_serverless/core/constants.py#L34
* https://github.com/Qiskit-Extensions/quantum-serverless/blob/v0.7.0/client/quantum_serverless/core/constants.py#L49
* https://github.com/Qiskit-Extensions/quantum-serverless/blob/v0.7.0/client/quantum_serverless/core/constants.py#L52

Ideally, these types of configs should be picked up from something like a configmap, so that they can be changed without needing to reinstall to update the values.

akihikokuroda:
### Summary
Fix #1064 

### Details and comments

These config values can be overridden by setting the following environment variables in the client environment.

```
REQUESTS_TIMEOUT_OVERRIDE
MAX_ARTIFACT_FILE_SIZE_MB_OVERRIDE
IBM_SERVERLESS_HOST_URL_OVERRIDE
```


IceKhan13 (approved):
LGTM

Tansito (approved):

PR #1147, opened 2023-12-11
Program run methods refactorization

Tansito:
### Summary

This pull request continues the effort from #1112 . In this case we are going to refactorize `run_existing` and `run` end-points due to the similitudes between them.

### Details and comments

- A new 404 exception created: `ResourceNotFoundException`
- Created services:
  - `JobConfigService`
  - `JobService`
- Standardized `JobConfig` serializer to be able to have different versions of it
- Created tests for each of the new services and their logic


gateway/api/services.py
```diff
@@ -65,3 +75,135 @@ def save(serializer, author, artifact) -> Program:
...
+        job = None
+        try:
+            job = Job(
+                program=program,
+                arguments=arguments,
+                author=author,
+                status=status,
+                config=jobconfig,
+            )
+            job.save()
```
akihikokuroda, line 180: This is not related to this refactoring but this `save()` is not good.  The job is created here and updated with the environment later may cause an issue.  The scheduler may find the job between these 2 save() and execute without the environment information.

IceKhan13 (approved):
LGTM! Way-way cleaner that it was before!

And tested on rancher.

If we can incorporate Aki's suggestion on saving once, it would be great, but not necessary in this PR.

akihikokuroda (approved):
I'm OK to merge this, too.

gateway/api/services.py
Tansito: Mmm... You are totally right. I tried to use one unique `save` but I couldn't figure out how. Any ideas @IceKhan13 ?

Tansito: Two options that come to my mind:
- Have a new state before `QUEUED` that reflects some kind of an `under state preparation`.
- Remove job.id from [environment variables](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1147/files#diff-b1f12a07f91db92d539b60433b3f9831281e139afeee593565c3c25563f3c206L131) but I think we need to have it

Tansito: First solution could be implemented in a new pull request and the good thing is that is not going to require any migration or breaking change.

PR #1149, opened 2023-12-12
Improve doc

### LINKED ISSUE #1114: Docs: round of improvements

Improve docs on following topics:

1. we moved to new `build` -> `upload` -> `run` type of code and if user did not upload new program it will not be executed. What title of a program user provide that would be executed. This allows user to upload once and run many. IF user want to introduce new code, they need to reupload program. Probably we need to make it extra clear, because users change code and do not reupload and then getting old results.
2. `/data` folder is persistent and shared between all jobs. This allows user to have shared storage and use it in between jobs. We have it in docs, but probably we need to make it extra clear and provide reasoning. Explain that `files` api is querying all files in shared `/data` folder produced by all jobs.

akihikokuroda:
### Summary
Fix #1114 

### Details and comments

Add one example for /data directory usage.
Minor update and fix.

### DIFF
--- a/docs/getting_started/experimental/source_files/consume_files.py
+++ b/docs/getting_started/experimental/source_files/consume_files.py
@@ -1,0 +1,10 @@
+import os
+import tarfile
+from quantum_serverless import save_result
+
+with tarfile.open("/data/my_file.tar", "r:gz") as tar:
+    with tar.extractfile("./my_file.txt") as f:
+        text =f.read()
+
+print(text)
+save_result({"Message": str(text)})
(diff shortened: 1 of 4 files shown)


docs/getting_started/experimental/source_files/consume_files.py
```diff
@@ -0,0 +1,10 @@
+import os
```
IceKhan13, line 1: is it used in any notebook?

IceKhan13 (commented):
Should we rename notebook with file download to something like `File handling` where we will describe how to download files and explain that `/data` folder is persistent and shared, which enable data sharing between jobs ?

docs/getting_started/experimental/source_files/consume_files.py
akihikokuroda: I missed one file to put.  I'm adding manage_data_directory.ipynb.

IceKhan13 (approved):

PR #1154, opened 2024-01-04
Create lock files

### LINKED ISSUE #1153: Docker build is taking too much due to dependency resolver

Specifically with python 3.9 but it happens with less impact using 3.8 and 3.10 too. Our docker images are taking a lot of time to build and being pushed.

Example here: https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/7411115942

The short solution that we found by now is create a `lock` file with pinned versions until we can address #901

Tansito:
### Summary

Fix #1153 

This PR creates a lock file for:
- `ray-node`
- `notebook`
- `gateway`
- `repository`

This way we reduce build times that were starting to be problematic until we close a more definitive version. This dependencies are being ignored also through renovate.


Dockerfile-ray-node
```diff
@@ -11,7 +11,7 @@ COPY --chown=$RAY_UID:$RAY_UID ./client ./qs
 WORKDIR /qs
 # TODO: certifi version pinned to 2023.7.22. More info in #870
 # TODO: pyarrow version pinned to 14.0.1
-RUN pip install . --no-cache-dir &&\
+RUN pip install -r requirements-lock.txt --no-cache-dir &&\
     pip install --no-cache-dir pyarrow==14.0.1 &&\
```
psschwei, line 15: I think you also need to install quantum-serverless from source:

```suggestion
RUN pip install -r requirements-lock.txt --no-cache-dir &&\
    pip install --no-cache . &&\
    pip install --no-cache-dir pyarrow==14.0.1 &&\
```

Tansito: Maybe you are right, let's see!

psschwei:
For python 3.8, getting this error: `Could not find a version that satisfies the requirement ipython==8.18.1`

Tansito:
My two cents @psschwei , could be that how I created this `requirements-lock` with python `3.10` the libraries are not compatible with python `3.8`? That would mean that we will need to have a lock per python version, right?

Tansito:
Btw, yes, 8.18.1 is not compatible:
https://github.com/Qiskit-Extensions/quantum-serverless/pull/1118

psschwei:
That makes sense... I pulled the versions from the 3.9 image... let's pin to 8.13.0 (?)

Tansito:
Closing this in favour of #1155

PR #1157, opened 2024-01-05
Run image builds as separate jobs

psschwei:

We found that running all image builds as a single job could create a bottleneck, so we decided to split them into separate jobs. This should also make reviewing logs easier, as there should be less output for the web UI to have to render.


IceKhan13:
Should we give it a go?

psschwei:
yeah, let me rebase...

IceKhan13 (approved):
let's go!

PR #1159, opened 2024-01-14
PathTraversal fix

akihikokuroda:
### Summary
This PR address the PathTraversal issues


client/quantum_serverless/utils/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def sanitize_file_path(path: str):
+    """sanitize file path.
+
+    Args:
+        path: file path
+
+    Returns:
+        saanitized filepath
```
psschwei, line 40:
```suggestion
        sanitized filepath
```

client/quantum_serverless/utils/utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+    utility functions
+"""
+
+import os
+import re
+
+
+def sanitize_file_path(path: str):
+    """sanitize file path.
```
psschwei, line 34: nit: let's document clearly what type of sanitizing we're doing, either here in the doc string or with comments in the code on each thing we clean

gateway/api/utils.py
```diff
@@ -182,3 +183,18 @@ def generate_cluster_name(username: str) -> str:
     pattern = re.compile("[^a-zA-Z0-9-.]")
     cluster_name = f"c-{re.sub(pattern,'-',username)}-{str(uuid.uuid4())[:8]}"
     return cluster_name
+
+
+def sanitize_file_path(path: str):
```
psschwei, line 188: is there a way we can use the one from the client utils rather than duplicating the code here? (if not, maybe we need a high-level utils package?)

gateway/main/settings.py
```diff
@@ -11,10 +11,27 @@
 """
 import os
 import os.path
+import re
 import sys
 from datetime import timedelta
 from pathlib import Path
 
+# pylint: disable=duplicate-code
+def sanitize_file_path(path: str):
```
psschwei, line 20: same comment re: client utils

gateway/api/utils.py
akihikokuroda: I took out one duplicate function in the gateway.  There still one in each client and gateway.  The gateway doesn't import the client.  I want to keep that way.

psschwei: Agree on not importing client in gateway. 
I'll open an issue about utils common across gateway and client. In an ideal world we wouldn't duplicate code, but for just one utility this may be the easy path.

psschwei (approved):

gateway/api/utils.py
IceKhan13: If some code needs to be shared usually it is formed into a separate library. It is starting to make sense when we have decent amount of duplicate code. Couple of functions is fine :)

IceKhan13:
So, sanitize will remove all `..` (and other symbols) from gives paths. 
Does that mean that user will not be able to do this:
```python
pattern = QiskitPattern(
  entrypoint="entrypoint.py",
  working_dir="../../"
)
```

akihikokuroda:
@IceKhan13 No, user can not do it.  Is it too inconvenient to the user?

IceKhan13:
Maybe we should get abs path during upload on client side and then sanitize it?

```
path = sanizite_path(
  os.path.abspath("../src/")
)
```

psschwei:
So that would be a directory structure something like this:

```
├── scr
    └── library.py
└── app
    └── entrypoint.py
```

Is that a common pattern? Or do we want to have something like a `function_root_dir` that limits how high up the tree folks can go?

Or should we ask folks to declare `working_dir="src/", entrypoint=app/entrypoint.py`?

psschwei:
Hmmm... thinking about it, would we need to worry about that? Going up a directory is usually only an issue on a server... on your own system, there shouldn't be any security risk in using `..` ... so in other words, is this scenario a false positive in the scanner?

IceKhan13:
> so in other words, is this scenario a false positive in the scanner?

I think so. We should worry about it only on server. Client is local to user.

psschwei:
If we limit the fix to the gateway, that also solves the code duplication issue... 
:rock: --> :bird: :bird:

akihikokuroda:
OK.  I'll take out all changes for the client side and mark the issues as `noise`.

IceKhan13 (approved):
Thank you @akihikokuroda !

PR #1172, opened 2024-01-18
Add catalog entry model

### LINKED ISSUE #1169: Gateway / repository: add catalog entry model

Add catalog entry to gateway or repository.

Should be something like:
```python
# api/models.py

class CatalogEntry(models.Model):
     title = ...
     description = models.TextField(...)
     tags = ...
     created_at = ...
     updated_at = ...
     program = models.ForeginField(model=Program, ...)
```

Epic https://github.com/Qiskit-Extensions/quantum-serverless/issues/1168

psschwei:
### Summary

Fixes #1169 

### Details and comments

Add catalog table to database


gateway/api/models.py
```diff
@@ -161,3 +161,14 @@ def __str__(self):
...
         """Returns true if job is in terminal state."""
         return self.status in self.TERMINAL_STATES
+
+
+class CatalogEntry(models.Model):
+    """Catalog Entry model."""
+
+    title = models.CharField(max_length=255)
+    description = models.TextField()
+    tags = models.TextField(null=False, blank=True, default="[]")
```
IceKhan13, line 171: thinking outloud: if in the future (not that distant) we want to filter / search by tag would that be better if we have `Tag` as separate model and in `ManyTomany` relationship? then all search queries will be something like `CatalogEntry.objects.filter(tags__in=["chemistry", "finance", ...])`

IceKhan13: I'm not sure about performance, though 🤔

psschwei: Model = table, right? (still learning what the django abstractions are)
If so, I think for performance we might be better with multiple rows in the catalog (one for each tag), as usually its quicker to search the DB than do joins... but it also somewhat depends on how what type of queries we expect to be doing most frequently ... 
All that said, I also don't expect the catalog to be so large that changing the data model a few months down the line ends up being prohibitive either...

IceKhan13 (approved):
Let's go! first q2 item is about to merge :)

PR #1181, opened 2024-01-21
Add `public` flag in Program

### LINKED ISSUE #1167: Gateway: add ability to share functions / programs

Add ability to share serverless programs. 

At this point adding `public` flag would be enough. But once fined-grained access will be added, we can consider sharing with specific groups only.

akihikokuroda:
### Summary
fix #1167 

### Details and comments

- `public` field is added in `program` model.
- It is `False` by default.
- `set_public` api is added to set / reset the field.  It is allowed to the author of the program only.


gateway/api/migrations/0014_runtimejob.py
```diff
@@ -0,0 +1,33 @@
+# Generated by Django 4.2.2 on 2024-01-20 01:44
+
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+class Migration(migrations.Migration):
```
IceKhan13, line 7: Do we need runtime job in this PR?

akihikokuroda: @IceKhan13 No, I can take it out.

gateway/api/views.py
```diff
@@ -255,6 +268,31 @@ def run(self, request):
             job_serializer = self.get_serializer_job_class()(job)
         return Response(job_serializer.data)
 
+    @action(methods=["POST"], detail=False)
+    def set_public(self, request):
```
IceKhan13, line 272: I think django out of the box is giving update API on every speicified field in serializer🤔 

Can we check that? Not to create redundant enpoinds.

gateway/api/migrations/0014_runtimejob.py
IceKhan13: We can leave it, if you think it is ready to be reviewed.
I only saw public in title and description of this PR that is why I want to make sure it is intended :)

IceKhan13: if we want to leave it. We can just add it into description and crosslink issues.

IceKhan13: Ah, we have runtime model here https://github.com/Qiskit-Extensions/quantum-serverless/pull/1177

IceKhan13: Interesting that we still have this file even though it should be in while we are looking for best configuration

IceKhan13: ah, files called differently. We should take this one out

IceKhan13: hmm, nope files called the same way 🤔

IceKhan13:
Can you rebase from main. I see file changes in PR which already should be in main 🤔

gateway/api/migrations/0015_program_public.py
```diff
@@ -0,0 +1,18 @@
...
+    dependencies = [
+        ("api", "0014_runtimejob"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="program",
+            name="public",
+            field=models.BooleanField(default=False),
+        ),
```
psschwei, line 17: Assuming we may add other groups in the future, should we maybe make this a bit more generic? i.e. instead of a boolean using a text field named `group`, which we could set to `public`

akihikokuroda: I thought about it.  I think that we can use the `public` as the master switch for the sharing.  The `group` can be added later for fine access control.

IceKhan13 (approved):

PR #1186, opened 2024-01-22
Add vim to images

psschwei:

We've found that having vim available on images is useful when needing to debug. So this PR adds it to the ray-node, gateway, and repository images.


akihikokuroda (approved):
LGTM

PR #1191, opened 2024-01-26
Runtimejob API ModelView 

akihikokuroda:
### Summary
This PR adds ModelView API for RuntimeJob model

### Details and comments

With this, the get, list and delete operations are easy and useful but the create operation is quite difficult to issue.  Form the usage of the RuntimeJob view, it may be better to have the following additional API in the `Job`.

`/api/v1/jobs/{job id}/add_runtimejob`
and
`/api/v1/jobs/{job id}/list_runtimejob`

### DIFF
--- a/gateway/api/v1/views.py
+++ b/gateway/api/v1/views.py
@@ -75,9 +75,11 @@ class RuntimeJobViewSet(views.RuntimeJobViewSet):  # pylint: disable=too-many-an
     RuntimeJob view set first version. Use RuntomeJobSerializer V1.
     """
 
-    queryset = RuntimeJob.objects.all()
     serializer_class = v1_serializers.RuntimeJobSerializer
-    permission_classes = [permissions.IsAuthenticated]
+    permission_classes = [permissions.IsAuthenticated, IsOwner]
 
     def get_serializer_class(self):
         return v1_serializers.RuntimeJobSerializer
+
+    def get_queryset(self):
+        return RuntimeJob.objects.all().filter(job__author=self.request.user)
--- a/gateway/api/v1/views.py
+++ b/gateway/api/v1/views.py
@@ -6,7 +6,7 @@ from rest_framework import permissions
 
 
 from api import views
-from api.models import Program, Job
+from api.models import Program, Job, RuntimeJob
 from api.permissions import IsOwner
 from . import serializers as v1_serializers
 from . import services as v1_services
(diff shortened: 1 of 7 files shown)


psschwei:
@akihikokuroda would adding the additional API change this PR? or would it be done separately?

akihikokuroda:
@psschwei I'll do it separately.

psschwei (commented):
LGTM
I still feel a bit iffy on the django framework, so I'd like either @IceKhan13 or @Tansito to review also

akihikokuroda:
@psschwei Thanks for review!

gateway/api/v1/views.py
```diff
@@ -68,3 +68,16 @@ class FilesViewSet(views.FilesViewSet):
...
 
     permission_classes = [permissions.IsAuthenticated, IsOwner]
+
+
+class RuntimeJobViewSet(views.RuntimeJobViewSet):  # pylint: disable=too-many-ancestors
+    """
+    RuntimeJob view set first version. Use RuntomeJobSerializer V1.
+    """
+
+    queryset = RuntimeJob.objects.all()
```
IceKhan13, line 78: it will select all runtime jobs, even not authored by user :)

gateway/api/v1/views.py
```diff
@@ -68,3 +68,16 @@ class FilesViewSet(views.FilesViewSet):
...
+    serializer_class = v1_serializers.RuntimeJobSerializer
+    permission_classes = [permissions.IsAuthenticated]
```
IceKhan13, line 80: Since `IsOwner` changed, we can add it now here

gateway/api/views.py
```diff
@@ -468,3 +468,17 @@ def upload(self, request):  # pylint: disable=invalid-name
...
+    RuntimeJob ViewSet configuration using ModelViewSet.
+    """
+
+    BASE_NAME = "runtime_jobs"
+
+    def get_serializer_class(self):
+        return self.serializer_class
+
+    def get_queryset(self):
+        return RuntimeJob.objects.all()
```
IceKhan13, line 484: same as above. We need to select only authored runtime jobs.

IceKhan13 (commented):
Looks good. Couple of tweaks and we are good to go!

akihikokuroda:
@IceKhan13 Thanks for review.  I updated the code and also updated the test to make sure it works.

akihikokuroda:
@IceKhan13 WDYT about additional APIs

Adding runtime job to the job:
```
/api/v1/jobs/{job id}/add_runtimejob
```

list the runtime jobs created in the job:
```
/api/v1/jobs/{job id}/list_runtimejob
```

I can create another PR if they are good.

IceKhan13:
> WDYT about additional APIs

sounds like a good idea! Let's create a separate PR for this APIs

Thank you!

IceKhan13 (approved):
Looks great! Thank you @akihikokuroda !

PR #1195, opened 2024-01-26
Bump ipython to 8.19.0

psschwei:


psschwei:
forgot we had pinned this due to other requirements... will close this and let dependabot handle this...

PR #1212, opened 2024-02-06
Use LocalProvider for test

### LINKED ISSUE #1141: Run tests using LocalProvider

Now that we have `LocalProvider`, we should make that the default environment for running tests in the CI. Specifically, https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/.github/workflows/docker-build.yaml could be updated to just run the notebooks (rather than docker build, etc.)

In theory, this means we could also drop docker compose. Longer-term though, we still need full e2e testing on Kubernetes (xref https://github.com/Qiskit-Extensions/quantum-serverless/issues/117#issuecomment-1594718083).

akihikokuroda:
### Summary
fix #1141


psschwei (commented):
Thanks Aki!

Would it be possible to make this a separate test? I'd lean towards keeping the docker compose test for testing gateway / scheduler until we can get an e2e test on kubernetes working (I have some ideas for how we can do that which I'll try to tackle new week).

.github/workflows/notebook-local-verify.yaml
```diff
@@ -0,0 +1,38 @@
+name: Docker notebook tests
```
psschwei: nit:

```suggestion
name: Test notebooks with LocalProvider
```

or soemthing like that

psschwei: was already done before I hit "comment" :smile:

akihikokuroda: I've just noticed.

psschwei (approved):
LGTM!

PR #1214, opened 2024-02-07
Update contribution guidelines

karlaspuldaro:
### Summary
This PR updates the contributing documentation by clarifying a few items.

### Details and comments
These suggestions are based on my first contribution experience following these guidelines, whenever I needed more context to overcome build failures following these steps.


CONTRIBUTING.md
```diff
@@ -69,11 +69,12 @@ if needed.
 
 You'll need to install these tools on your development environment:
 
-1. [python](https://www.python.org/): the language quantum-serverless is written in
+1. [python](https://www.python.org/): the language quantum-serverless is written in (Note that we currently support Python >=3.6,<3.9).
```
psschwei, line 72: I believe we're only supporting 3.8, 3.9, and 3.10 currently (with 3.8 support dropping in the near future)

karlaspuldaro: When installing dependencies in gateway and clients  with `pip install -r requirements.txt requirements-dev.txt`
using python 3.10 on my env, I got some errors related to python version incompatibility that mentioned `Requires-Python >=3.6,<3.9`. The error went away when I switched to py3.8, as indicated. 
For some reason, I can't recreate that error. I'll update the supported versions then.

karlaspuldaro: > The Python Software Foundation (PSF) will no longer support Python 3.7 on June 27th, 2023.

If we're not supporting py3.7 as well, I guess we need to clean up references to it in READMEs and client [setup.py](https://github.com/Qiskit-Extensions/quantum-serverless/blob/4afc634f3aed53b50cc06775bff77b9583781770/client/setup.py#L26)

psschwei: Yeah, let's clean those up too

karlaspuldaro:
@psschwei I addressed your comments and updated the PR title and description.

psschwei (approved):
Thanks @karlaspuldaro !

PR #1215, opened 2024-02-08
Remove default login view

### LINKED ISSUE #1209: Gateway: remove django login views

I think we do not use default django login views. Maybe we should remove them.

akihikokuroda:
### Summary
fix #1209

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -201,7 +201,9 @@ class ProgramViewSet(viewsets.ModelViewSet):  # pylint: disable=too-many-ancesto
             carrier = {}
             TraceContextTextMapPropagator().inject(carrier)
             arguments = serializer.data.get("arguments")
-            token = request.auth.token.decode()
+            token = ""
+            if request.auth:
+                token = request.auth.token.decode()
             try:
                 job = self.get_service_job_class().save(
                     program=program,
--- a/gateway/tests/api/test_files.py
+++ b/gateway/tests/api/test_files.py
@@ -5,6 +5,7 @@ from urllib.parse import quote_plus
 from django.urls import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
+from django.contrib.auth import models
 
 
 class TestFilesApi(APITestCase):
@@ -30,12 +31,8 @@ class TestFilesApi(APITestCase):
         media_root = os.path.normpath(os.path.join(os.getcwd(), media_root))
 
         with self.settings(MEDIA_ROOT=media_root):
-            auth = reverse("rest_login")
-            response = self.client.post(
-                auth, {"username": "test_user", "password": "123"}, format="json"
-            )
-            token = response.data.get("access")
-            self.client.credentials(HTTP_AUTHORIZATION="Bearer " + token)
+            user = models.User.objects.get(username="test_user")
+            self.client.force_authenticate(user=user)
             url = reverse("v1:files-list")
             response = self.client.get(url, format="json")
             self.assertEqual(response.status_code, status.HTTP_200_OK)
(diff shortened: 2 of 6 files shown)


Tansito:
Hey @akihikokuroda I just saw your PR. The only requirement is remove the end-points in the view, we continue using custom logic through the library to manage access in the rest of the end-points with iqp.

akihikokuroda:
@Tansito I can simply take out the urls from the `main/url.py` but it may be better to take out unnecessary middlewares and modules.  WDYT?  I'm trying but it's messing up the tests.

Tansito:
Yes! I think it has sense. But I think there is no a lot of things to remove there because we are using a custom rule for IQP using most of the modules for the authentication (If I remember correctly).

Tansito:
But yeah, definitely it will deserve the effort

psschwei (approved):
:rocket:

gateway/api/views.py
```diff
@@ -201,7 +201,9 @@ def run_existing(self, request):
             carrier = {}
             TraceContextTextMapPropagator().inject(carrier)
             arguments = serializer.data.get("arguments")
-            token = request.auth.token.decode()
+            token = ""
+            if request.auth:
```
Tansito, line 205: Maybe it's needed I would need to take a look more in-depth but why are we adding now this `if`?

gateway/tests/api/test_files.py
```diff
@@ -30,25 +31,17 @@ def test_files_list(self):
         media_root = os.path.normpath(os.path.join(os.getcwd(), media_root))
 
         with self.settings(MEDIA_ROOT=media_root):
-            auth = reverse("rest_login")
-            response = self.client.post(
-                auth, {"username": "test_user", "password": "123"}, format="json"
-            )
-            token = response.data.get("access")
-            self.client.credentials(HTTP_AUTHORIZATION="Bearer " + token)
+            user = models.User.objects.get(username="test_user")
```
Tansito, line 34: I would like to understand more in-depth this change because until now we had two authentication mechanisms: `mock_token` & `custom_token` (ref [here](https://github.com/Qiskit-Extensions/quantum-serverless/blob/28cf8112133ed2d66ea4e88829dace0a63d42996/gateway/README.md?plain=1#L25)).

with this change are we going to loose this functionality? (Sorry, I couldn't test it yet I can take a look more in-depth next week)

Tansito:
Apart from my comments @akihikokuroda but removing `dj_rest` from `INSTALLED_APPS` could mean that we could removed the dependency from the `requirements.txt` or are we using it in another configuration that I'm not being to see in the pull request?

gateway/api/views.py
akihikokuroda: I couldn't make the test to fake the `request.auth`.  It comes here without `auth` object here.  So I added this check.  I don't believe that this causes any bad effects in the runtime.

gateway/tests/api/test_files.py
akihikokuroda: They should not be affected these changes.  This test is using `/dj-rest-auth/login` endpoint that is taken out in this PR.  So I had to use another way to access the endpoint that this test is testing.  Would you check if the `custom_token` authentication is OK?

gateway/api/views.py
Tansito: Oh, I see it's due to the changes in the tests, then.

akihikokuroda:
@Tansito I tried to takeout some modules in the requirements.txt but some part of them are still used.  So I couldn't take any out.

gateway/tests/api/test_files.py
Tansito: Sure, I will take a look monday on this and let you know @akihikokuroda , thank you!

akihikokuroda: Thanks!

Tansito: Everything working in my side 🎉

Tansito (approved):
`mock_token` continues working, changes LGTM, thank you @akihikokuroda 👍 

(maybe we should use the mock token `awesome_token` for authenticate but it's another conversation)

akihikokuroda:
@Tansito Thanks for checking!

PR #1216, opened 2024-02-09
Test notebooks against k8s qs deployment

psschwei:
### Summary

Let's try this again...

### DIFF
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -1,0 +1,102 @@
+name: Test QS on Kubernetes
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Kind
+        run: |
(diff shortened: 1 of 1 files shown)


akihikokuroda (approved):
LGTM!

Tansito:
@psschwei should we review it or do we need to wait? (how it has the WIP label)

psschwei:
> @psschwei should we review it or do we need to wait? (how it has the WIP label)

Whoops -- I put that there while I was getting the test to pass... it's ready for review now

Tansito (approved):
It looks good to me @psschwei thank you! ❤️ 

Is there a reason because you preferred to download kind directly instead to use the GHA? https://github.com/helm/kind-action , thinking in it would be easier to maintain.

psschwei:
> Is there a reason because you preferred to download kind directly instead to use the GHA? https://github.com/helm/kind-action , thinking in it would be easier to maintain.

Mostly because we need a [load balancer](https://kind.sigs.k8s.io/docs/user/loadbalancer/) to make the service available outside the cluster without using port-forwarding (which caused all kinds of problems when I tried that last time). The kind-action you linked doesn't install a load balancer. whereas the one from Chainguard [that I used](https://github.com/chainguard-dev/actions/tree/main/setup-kind) does.

Tansito:
Ooooh, I see. Thank you @psschwei !

PR #1232, opened 2024-02-26
Sidecar proxy prototype

akihikokuroda:
### Summary

Prototype for #895 

### Details and comments

This PR is to share the prototype code for the issue within the team.


akihikokuroda:
@Tansito @IceKhan13 @psschwei Hi! It is not working yet but here the code I have so far.  Please take a look.

akihikokuroda:
I put the logs of proxy side car in `gateway/proxy/proxy.log`
Looks like the response header attribute `correlation-id: cnfuhom6ll0enki0tmng` for  `GET /runtime/backends?provider=ibm-q%2Fopen%2F` request is the one we need.

akihikokuroda:
The middleware job id is necessary to issue the API call.

akihikokuroda:
It is actually  `"id":"cng91pej1613i2k2c8p0"` in the response data for `POST /runtime/jobs HTTP/1.1\r\nHost: api.quantum.ibm.com` request.

```
data received from program
Received: b'POST /runtime/jobs HTTP/1.1\r\nHost: api.quantum.ibm.com\r\nUser-Agent: python-requests/2.31.0\r\nAccept-Encoding: gzip, deflate, br\r\nAccept: */*\r\nConnection: keep-alive\r\nX-Qx-Client-Application: qiskit-version-2/qiskit_ibm_runtime-0.18.0,qiskit_terra-0.45.2/qiskit_ibm_runtime~estimator.py\r\nContent-Length: 1965\r\nX-Access-Token: 4Y34kb1f5dkPrd5zohVOKzoF9EtgCSfqpwb7UAdnx6ztqO5zG1ZaO9D8lrEwSQg9\r\n\r\n'
data received from program
Received: b'{"program_id": "estimator", "params": {"circuits": [{"__type__": "QuantumCircuit", "__value__": "eJwL9Az29gzhYtBlYoAAxoJCBpY0Bg4gk5UBAUDSjFC2eKCjvyNMorq2kBGslpGxkAEVMCLpBQFmKM0CV8EGFWVkwA5Cjd0TS1LB5qZBhTgkdF1Cfiv+tM+ECcAUo6ngdCDBeEZyjGeHGs3EgB0ERUXBnQ+2JBUqkQelYe4y8C3N0XDLyU8s0VA30jNQ11EoKEpNzizOzM+zNTXW1FEIrsxNys/RUD+3OdogVl1Ts6wMbgnMciadfe8k3rE6XVt2YNkyPct+mB0M5zaDw4ERTzgERcDDgXInbsLtxL+S5nriio55Ew1O7Ci+w6AKd+ImEqKKiR5RxTT4o4oKTqR1VDHTI6qYB39UUcGJtI4qFnpEFcvgjyqGgYqqAalMDIkMy3yYHcCwHJCilB4OpUpBQo5D6Vw943QiWupUgztxE72rJdo6kSrFMW2dSIWSkj7FEFWKdNpmGqoU6UMguslxIgPDfyQA0wkAE2hD3Q=="}], "observables": [{"__type__": "settings", "__module__": "qiskit.quantum_info.operators.symplectic.sparse_pauli_op", "__class__": "SparsePauliOp", "__value__": {"data": {"__type__": "settings", "__module__": "qiskit.quantum_info.operators.symplectic.pauli_list", "__class__": "PauliList", "__value__": {"data": ["IIIZZ", "IIZIZ", "IZIIZ", "ZIIIZ"]}}, "coeffs": {"__type__": "ndarray", "__value__": "eJyb7BfqGxDJyFDGUK2eklqcXKRupaBuk2xop'
data received from program
Received: b'q6joJ6WX1RSlJgXn1+UkgqScEvMKU4FihdnJBakAvkaJjqaOgq1CuQDLgYw+GDPgAKI5wMAzRsgJw=="}}}], "parameters": [[{"__type__": "Parameter", "__value__": "AAX9GTcuFyFBbpEwyLhz3AAlzrJbMF0="}, {"__type__": "Parameter", "__value__": "AAX9GTcuFyFBbpEwyLhz3AAmzrJbMV0="}, {"__type__": "Parameter", "__value__": "AAUsvu4Y7gVC1qbApqYuOY9uzrNbMF0="}, {"__type__": "Parameter", "__value__": "AAUsvu4Y7gVC1qbApqYuOY9vzrNbMV0="}]], "parameter_values": [[1.2736114643252021, 1.7699802699954126, 5.0730499437842544, 3.6970486879700917]], "transpilation_settings": {"skip_transpilation": false, "optimization_settings": {"level": 3}, "coupling_map": null, "basis_gates": null}, "resilience_settings": {"level": 0}, "run_options": {"shots": 4000, "init_qubits": true, "noise_model": null, "seed_simulator": null}}, "log_level": "WARNING", "backend": "ibmq_qasm_simulator", "start_session": true, "session_time": null, "hub": "ibm-q", "group": "open", "project": "main"}'
data received from backend
Received header from backend: Date: Thu, 29 Feb 2024 14:12:53 GMT
Content-Type: application/json; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
content-security-policy: default-src 'self' https://cdnjs.cloudflare.com data:; frame-ancestors 'none'; form-action 'self'
cross-origin-embedder-policy: require-corp
cross-origin-opener-policy: same-origin
cross-origin-resource-policy: same-origin
x-dns-prefetch-control: off
x-frame-options: SAMEORIGIN
strict-transport-security: max-age=15724800; includeSubDomains
x-download-options: noopen
x-content-type-options: nosniff
origin-agent-cluster: ?1
x-permitted-cross-domain-policies: none
referrer-policy: no-referrer
x-xss-protection: 1
access-control-allow-origin: *
correlation-id: cng91pej1613i2k2c8og
x-envoy-upstream-service-time: 275
cf-cache-status: DYNAMIC
content-encoding: gzip
Server: cloudflare
CF-RAY: 85d1813acdeb054a-IAD
data received from backend
end of chunked data received from backend
b'{"id":"cng91pej1613i2k2c8p0","backend":"ibmq_qasm_simulator","session_id":"cng91pej1613i2k2c8p0"}'
```

akihikokuroda:
The proxy still getting the following error after sending the first request to the backend.
```
Traceback (most recent call last):
  File "/usr/src/app/server.py", line 110, in <module>
    connection, client_address = server.accept()
  File "/usr/local/lib/python3.9/ssl.py", line 1356, in accept
    newsock = self.context.wrap_socket(newsock,
  File "/usr/local/lib/python3.9/ssl.py", line 501, in wrap_socket
    return self.sslsocket_class._create(
  File "/usr/local/lib/python3.9/ssl.py", line 1041, in _create
    self.do_handshake()
  File "/usr/local/lib/python3.9/ssl.py", line 1310, in do_handshake
    self._sslobj.do_handshake()
ssl.SSLError: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:1129)
```

psschwei:
I tried running your notebook (using my own token), but the job just hung (pods were running, but never got any results).

Will post logs in a bit (need to remove tokens, etc. from them)

akihikokuroda:
@psschwei Thanks for looking at this.  It is still getting an certificate exception in the proxy.  So it stack in the running state.

akihikokuroda:
It's in the proxy container running in the ray head node pod.

akihikokuroda:
It seems like the **WebSocket** client created in `qiskit-ibm-provider/qiskit_ibm_provider/job/ibm_circuit_job.py` is verifying the server certificate.  I don't see any way to turn it off in the Qiskit API level. :-(

Tansito:
Awesome job anyway, @akihikokuroda . Thank you so much for this exploration, we can analyze it all of us next week

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,84 @@ data:
...
-            - image: fluent/fluent-bit:1.9.10
-              name: ray-head-logs
+            - image: proxy:local
+              #imagePullPolicy: IfNotPresent
+              #name: tcpdump-sidecar
+              #command: ["/bin/bash"]
+              #args: ["-c", "while true; do ping localhost; sleep 60;done"]
+              securityContext:
+                runAsGroup: 123
+                privileged: true
```
psschwei, line 151: Is it possible to not run the proxy as privileged?

akihikokuroda: Yes, I think so.  The setting of the iptables must be separated out into an init container then the proxy can run in none-privileged.  That is one of TODOs.

akihikokuroda:
For the certification error,
`run_forever` function must be called with `sslopt` based on the `verify` option of `QiskitRuntimeService` object creation.
```
                self._ws.run_forever(

                    ping_interval=60, ping_timeout=10, **self._proxy_params

                )
``` 
it's in
```
qiskit-ibm-provider/qiskit_ibm_provider/api/clients/base.py
```

To pass the verify option of `QiskitRuntimeService`, `qiskit_runtime_service.py`, `runtime_job.py`, `runtime_ws.py` and `base.py` probably need to be updated

akihikokuroda:
New proxy code httpserver.py is almost ready now.  Please take a look.

akihikokuroda:
I still see some error.  I'll continue debugging.

akihikokuroda:
Finally, the `httpserver.py` is working. :-).   I'll remove `server.py` later.  There are 2 issues in qiskit_ibm_runtime.  Here is one I've just opened. https://github.com/Qiskit/qiskit-ibm-runtime/issues/1508
I'll open another one soon.

docs/examples/02_qaoa.ipynb
```diff
@@ -220,15 +260,17 @@
    "source": [
     "serverless = ServerlessProvider(\n",
     "    token=os.environ.get(\"GATEWAY_TOKEN\", \"awesome_token\"),\n",
-    "    host=os.environ.get(\"GATEWAY_HOST\", \"http://localhost:8000\"),\n",
+    "    host=os.environ.get(\"GATEWAY_HOST\", \"http://localhost:59994\"),\n",
```
psschwei, line 263: this looks like the only code change needed to run the notebook? if so, I'd just make the update in the notebook but not rerun all the cells

gateway/proxy/proxy.log
```diff
@@ -0,0 +1,173 @@
+entrypoint.sh
```
psschwei, line 1: this file doesn't need to be committed, right?

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,84 @@ data:
                 - name: OT_ENABLED
                   value: {{ .Values.application.ray.openTelemetryCollector.enabled | quote }}
 {{- end }}
-            - image: fluent/fluent-bit:1.9.10
-              name: ray-head-logs
+            - image: proxy:local
```
psschwei, line 144: a few weeks back we had discussed possibly running this as a separate service rather than a sidecar on each ray cluster, but I don't remember where we landed on that

gateway/proxy/proxy.log
akihikokuroda: Yes, it was for reference.  I take out.

docs/examples/02_qaoa.ipynb
akihikokuroda: The only necessary change is `verify=False` in 
```
service = QiskitRuntimeService(
    channel='ibm_quantum',
    instance='ibm-q/open/main',
    token='<token>',
    verify=False,
)
```
I'll make it the only change and recheck in.
Thanks!

gateway/proxy/proxy.log
akihikokuroda: Right.  I put it just for reference.  I take it out.  Thanks!

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: We talked about possibilities.  I think that it may be better as a sidecar because the communication between the ray worker container and the proxy may not be safe so it is better to be an internal communication in the pod instead of between pods. WDYT?

akihikokuroda:
There still is a test error but this is ready for review.  The iptables can be removed when we know how to change the backend api url.

gateway/proxy/proxy/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+echo "entrypoint.sh"
+
+iptables -t nat -N proxy_redirect
+iptables -t nat -A proxy_redirect -p tcp -d 104.18.29.94 -j REDIRECT --to-ports 8443
+iptables -t nat -A proxy_redirect -p tcp -d 104.18.28.94 -j REDIRECT --to-ports 8443
```
psschwei, line 7: assuming this is the IP for runtime? is this something we want to hardcode, or should it be configurable?

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -313,6 +377,87 @@ data:
                   cpu: {{ .Values.application.ray.cpu }}
                   memory: {{ .Values.application.ray.memory }}Gi
               securityContext:
+
+
+            - resources:
```
psschwei, line 382:
```suggestion
              securityContext:
            - resources:
```

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -313,6 +377,87 @@ data:
...
+              securityContext:
+                runAsGroup: 123
+                privileged: true
+                capabilities:
+                  add: ["NET_ADMIN"]
+
+
+
+
             imagePullSecrets: []
```
psschwei, line 461:
```suggestion
                  add: ["NET_ADMIN"]
            imagePullSecrets: []
```

psschwei (commented):
Will give a full review tomorrow, but did have one quick question

gateway/proxy/proxy/entrypoint.sh
akihikokuroda: @psschwei Thanks for review.
I hope that we can take put the iptables completely by setting the runtime url to the proxy.
Yes, I can change it configurable.  I have to try the dns name, too.

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: I'll update.  Thanks!

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: I'll update.  Thanks!

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,85 @@ data:
...
 {{- end }}
-            - image: fluent/fluent-bit:1.9.10
-              name: ray-head-logs
+            - image: {{ .Values.application.ray.proxyImage }}
+              securityContext:
+                runAsGroup: 123
+                privileged: true
+                capabilities:
+                  add: ["NET_ADMIN"]
+              imagePullPolicy: Never
```
IceKhan13, line 150: Shouldn't it be `IfNotPresented` or something liek that? :)

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,85 @@ data:
...
+              name: proxy
+              ports:
+              - containerPort: 6379
```
IceKhan13, line 153: Proxy image needs the same ports as ray node?

gateway/proxy/proxy/entrypoint.sh
IceKhan13: It looks like runtime client has `proxies` field. https://github.com/Qiskit/qiskit-ibm-runtime/blob/c582e5789797861174bafcb96e6470f7b637d3b7/qiskit_ibm_runtime/qiskit_runtime_service.py#L77

Can we test our proxy with this field? 

If it will work we can ask runtime team to add ability to configure this field using env variables.

IceKhan13: Ah, disregard comment above. `proxies` parameter is for mapping between protocol and url

IceKhan13: Can we try to use `url` parameter and test if it will work? If it will work we can ask runtime team to add ability to configure this field using env variables.

https://github.com/Qiskit/qiskit-ibm-runtime/blob/c582e5789797861174bafcb96e6470f7b637d3b7/qiskit_ibm_runtime/qiskit_runtime_service.py#L73

gateway/proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,159 @@
...
+    server_context.load_cert_chain(
+        certfile="/etc/ray/tls/tls.crt", keyfile="/etc/ray/tls/tls.key"
+    )
+    server_context.load_verify_locations(cafile="/etc/ca/tls/ca.crt")
+    server_context.verify_mode = ssl.CERT_NONE
+
+    httpd = ThreadingHTTPServer((HOST, PORT), ProxyRequestHandler)
+    httpd.socket = server_context.wrap_socket(httpd.socket, server_hostname=HOST)
+
+    print("forever")
```
IceKhan13, line 158:
```suggestion
    print("Running proxy server...")
```

gateway/tests/proxy/test_httpserver.py
```diff
@@ -0,0 +1,144 @@
...
+import threading
+from http.server import HTTPServer
+import http, http.client, urllib.parse
+import unittest
+import operator
+from proxy.proxy.httpserver import ProxyRequestHandler
+from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
+
+
+class TestServerThread(threading.Thread):
```
IceKhan13, line 11: Interesting, why we are not getting lint errors 🤔 

Linter should check for docstrings for tests as well 🤔

gateway/proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,159 @@
+import ssl
```
Tansito, line 1: Taking into account that Gateway is a Django project I would maintain `proxy` project outside `gateway` folder with it's own pipeline and so on if we are not using anything from Django.

gateway/api/ray.py
```diff
@@ -229,6 +233,10 @@ def create_ray_cluster(
     dyn_client = DynamicClient(k8s_client)
     raycluster_client = dyn_client.resources.get(api_version="v1", kind="RayCluster")
     response = raycluster_client.create(body=cluster_data, namespace=namespace)
+    print(cluster_data)
```
Tansito: Right now I'm just taking an overview of the project, @akihikokuroda . Let me know if you want that I enter to see specific things from the code like this one.

gateway/proxy/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ENV PYTHONUNBUFFERED 1
+
+USER 0:123
+RUN apt-get -y update
+RUN apt-get install -y iptables
+RUN apt-get install -y tcpdump
+RUN pip install headerparser
+RUN pip install certifi
+RUN pip install requests
+COPY gateway/proxy/proxy .
```
Tansito, line 16: Similar to my previous comment. I would separate this project from `gateway`.

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -72,17 +78,12 @@ jobs:
           echo $GATEWAY_HOST
           # basic tests
           cd /home/runner/work/quantum-serverless/quantum-serverless/tests/basic
-          for f in *.py; do echo "$f" && python "$f"; done
-          # experimental tests
-          cd /home/runner/work/quantum-serverless/quantum-serverless/tests/experimental
-          for f in *.py; do echo "$f" && python "$f"; done
+          for f in *.py; do echo "$f" && python "$f" && kubectl get raycluster -o yaml; done
```
Tansito, line 81: What does this piece of code do exactly, @akihikokuroda ?

.github/workflows/icr-image-build-and-push.yaml
```diff
@@ -31,6 +31,10 @@ jobs:
             pythonversion: ''
             dockerfile: ./gateway/Dockerfile
             platforms: linux/amd64,linux/arm64
+          - imagename: quantum-serverless-proxy
+            pythonversion: ''
+            dockerfile: ./gateway/proxy/Dockerfile
```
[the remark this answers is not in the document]
akihikokuroda: OK.  I'll move to ./proxy

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: You are right.  It's for my development environment.  I'll change it `IfNotPresented`.

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,85 @@ data:
...
+                  cpu: {{ .Values.application.ray.cpu }}
+                  memory: {{ .Values.application.ray.memory }}Gi
               volumeMounts:
               - mountPath: /tmp/ray
                 name: log-volume
-              - mountPath: /fluent-bit/etc/fluent-bit.conf
-                name: fluentbit-config
-                subPath: fluent-bit.conf
+              - mountPath: /etc/ca/tls
+                name: ca-tls
```
Tansito, line 173: I need to take a look more in-depth on this @akihikokuroda but is it possible to refactorize this variables in some way. I have the feeling that we are repeating them in different parts.

gateway/proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,159 @@
...
+                )
+            except:
+                None
+        self.handle_response(resp)
+        self.wfile.flush()  # actually send the response if not already done.
+        if job_request and id:
+            jsondata = json.loads(resp.content.decode("utf-8"))
+            logging.debug("job id: %s", jsondata["id"])
+
+            token = "awesome_token"
```
Tansito, line 133: Mmm... We are going to need to take a look on how to manage authentication between `gateway` and `proxy`.

Tansito (approved):

docs/examples/02_qaoa.ipynb
```diff
@@ -179,7 +179,7 @@
     "service = None\n",
     "backend = \"ibmq_qasm_simulator\"\n",
     "if USE_RUNTIME_SERVICE:\n",
-    "    service = QiskitRuntimeService()\n",
+    "    service = QiskitRuntimeService(verify=False)\n",
```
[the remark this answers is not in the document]
akihikokuroda: The certificate that the proxy uses is not the certificate for RuntimeService.  So the certificate verification fails when the client gets the certificate of the proxy.  we probably need to add an environment variable to turn off the verification in the runtime service.

gateway/proxy/proxy/httpserver.py
akihikokuroda: I'll update.  Thanks.

gateway/tests/proxy/test_httpserver.py
akihikokuroda: I'll add the docstrings.

gateway/proxy/proxy/httpserver.py
akihikokuroda: I'll move it to ./proxy.

gateway/api/ray.py
akihikokuroda: I'm debugging the CI test error now.  This is added just for it.  I'll take out when it's done.  Thanks!

gateway/proxy/Dockerfile
akihikokuroda: OK

.github/workflows/kubernetes-deploy.yaml
akihikokuroda: I'm debugging this test failure.  I'll take it out when it's done.  
Thanks!

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: The init container generate the certificate and the other containers use the certificate.  So the same volumes are mounted by multiple containers.   I might made some extra copy of the mount in the yaml file.  Please let me know.  Thanks!

gateway/proxy/proxy/httpserver.py
akihikokuroda: Oh, I forgot that.  Yes, we have to decide how to do it.  Thanks!

psschwei:
One thing I noticed when trying this out locally: we log the circuit being sent to runtime. Here's a snippet from the logs:

```
DEBUG:root:New request: b'GET /runtime/backends/ibmq_qasm_simulator/configuration HTTP/1.1\r\n'
DEBUG:root:Passthrough none POST request: https://api.quantum.ibm.com/runtime/backends/ibmq_qasm_simulator/configuration
DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): api.quantum.ibm.com:443
DEBUG:urllib3.connectionpool:https://api.quantum.ibm.com:443 "GET /runtime/backends/ibmq_qasm_simulator/configuration HTTP/1.1" 200 None
10.244.0.16 - - [29/Mar/2024 13:30:21] "GET /runtime/backends/ibmq_qasm_simulator/configuration HTTP/1.1" 200 -
DEBUG:root:Recieved response content: {"backend_name":"ibmq_qasm_simulator","coupling_map":null,"backend_version":"0.1.547","n_qubits":32,"basis_gates":["u1","u2","u3","u","p","r","rx","ry","rz","id","x","y","z","h","s","sdg","sx","t","tdg","swap","cx","cy","cz","csx","cp","cu1","cu2","cu3","rxx","ryy","rzz","rzx","ccx","cswap","mcx","mcy","mcz","mcsx","mcp","mcu1","mcu2","mcu3","mcrx","mcry","mcrz","mcr","mcswap","unitary","diagonal","multiplexer","initialize","kraus","roerror","delay"],"gates":[{"name":"u1","parameters":["lam"],"conditional":true,"description":"Single-qubit gate [[1, 0], [0, exp(1j*lam)]]","qasm_def":"gate u1(lam) q { U(0,0,lam) q; }"},{"name":"u2","parameters":["phi","lam"],"conditional":true,"description":"Single-qubit gate [  
<snip>
```

We should probably not log that info (though not sure exactly what the best way to go about that is either)

Another very minor thing... proxy logs don't have timestamps. I don't think this is a super big deal, as in LogDNA that will be appended, but if there's an easy switch to flip to add them might not be a bad idea.

Also, looks like the qaoa notebook launched about 50 jobs on IQP before reaching a result... assume that's expected, but noting just in case.

akihikokuroda:
@psschwei Thanks for review.

> We should probably not log that info (though not sure exactly what the best way to go about that is either)

I can take out some log statements not to log the circuits.

>Another very minor thing... proxy logs don't have timestamps. I don't think this is a super big deal, as in LogDNA that will be appended, but if there's an easy switch to flip to add them might not be a bad idea.

I'll look into configuration of the `logging`

> Also, looks like the qaoa notebook launched about 50 jobs on IQP before reaching a result... assume that's expected, but noting just in case.

I believe that expected.

Makefile
```diff
@@ -2,7 +2,7 @@
 # Constants
 # =========
 
-version=latest
+version=local
```
psschwei, line 5: was this part of local testing?

Makefile
```diff
@@ -12,16 +12,19 @@ endif
 
 notebookImageName=$(repository)/quantum-serverless-notebook
 rayNodeImageName=$(repository)/quantum-serverless-ray-node
-gatewayImageName=$(repository)/quantum-serverless-gateway
+#gatewayImageName=$(repository)/quantum-serverless-gateway
+gatewayImageName=qiskit/quantum-serverless-gateway
+#proxyImageName=$(repository)/quantum-serverless-proxy
+proxyImageName=qiskit/quantum-serverless-proxy
```
psschwei, line 18: assuming these were local testing too... though also wondering if anyone uses the makefile to build images?

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,21 +141,75 @@ data:
...
+                capabilities:
+                  add: ["NET_ADMIN"]
+              imagePullPolicy: IfNotPresent
+              name: proxy
               resources:
                 limits:
-                  cpu: 100m
-                  memory: 128Mi
+                  cpu: {{ .Values.application.ray.cpu }}
+                  memory: {{ .Values.application.ray.memory }}Gi
```
psschwei, line 155: does the proxy need the same amount of resources as the ray container(s)? if so that would essentially double our compute requirements

proxy/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+WORKDIR /usr/src/app
+
+# set environment variables
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+USER 0:123
+RUN apt-get -y update
+RUN apt-get install -y iptables
+RUN apt-get install -y tcpdump
```
psschwei, line 12: nit: I think this will use fewer layers / make for a slightly more optimal build
```suggestion
RUN apt-get -y update && apt-get install -y iptables tcpdump
```

proxy/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN pip install headerparser
+RUN pip install certifi
+RUN pip install requests
```
psschwei, line 15: could we set these in a requirements.txt file? I think that'll also allow for dependabot version updates

psschwei (commented):
still have to go through the httpserver code / tests (and will do early this afternoon), but here's some comments so far

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: Probably not.  I'll add separate variables

gateway/tests/proxy/test_httpserver.py
akihikokuroda: I moved the proxy out of the gateway and add `tox.ini` for lint and test

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -31,6 +31,11 @@ jobs:
           docker build -t ray:test --build-arg TARGETARCH="amd64" -f ./Dockerfile-ray-node .
           kind load docker-image ray:test
           docker image rm ray:test
+      - name: Build and load proxy
+        run: |
+          docker build -t proxy:test --build-arg TARGETARCH="amd64" -f ./proxy/Dockerfile .
+          kind load docker-image proxy:test
```
Tansito, line 37: are using kind just for testing here, Aki?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+from http import HTTPStatus
+from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
+import json
+import logging
+import os
+import ssl
+import zlib
+import requests
+
+logging.basicConfig(level=logging.DEBUG)
```
Tansito, line 11: We will need to configure the level of the logs by and environment variable

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+
+HOST = "127.0.0.1"
+PORT = 8443
```
Tansito, line 14: I see that we are configuring the port through several places. I assume we will need to use environment variables here too.

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+        """handle response."""
+        self.send_response(resp.status_code)
+        if resp.content and len(resp.content) != 0:
+            content = resp.content
+            logging.debug("Recieved response content: %s", content.decode("utf-8"))
+            if (
+                "content-encoding" in resp.headers
+                and resp.headers["content-encoding"] == "gzip"
+            ):
+                content = self.gzip_encode(content)
```
Tansito, line 42: Curious about this @akihikokuroda . Can we assume that responses are always being `gzipped` and `chunked` or did you check situations where that was not happen?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+            if (
+                "Transfer-Encoding" in resp.headers
+                and resp.headers["Transfer-Encoding"] == "chunked"
+            ):
+                content = (
+                    bytes(f"{len(content):x}", "utf-8")
```
Tansito, line 48: Thinking in a possible unit-test I would move this logic to another function like you did with `gzip_encode`.

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
+"""proxy server"""
+from http import HTTPStatus
+from http.server import ThreadingHTTPServer, BaseHTTPRequestHandler
```
Tansito, line 3: From what I know `http.server` is not recommended for security reasons. Would it be possible to use another approach?

https://docs.python.org/3/library/http.server.html

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+            self.wfile.write(content)
+            logging.debug("Sending response content: %s", content)
+
+    def handle_one_request(self):
+        """Handle a single HTTP request."""
+
+        try:
+            logging.debug("New request")
+            self.raw_requestline = (  # pylint: disable=attribute-defined-outside-init
+                self.rfile.readline(65537)
```
Tansito, line 67: What are we doing here, @akihikokuroda ?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+                and resp.headers["Transfer-Encoding"] == "chunked"
+            ):
+                content = (
+                    bytes(f"{len(content):x}", "utf-8")
+                    + b"\x0d\x0a"
+                    + content
+                    + b"\x0d\x0a0\x0d\x0a\x0d\x0a"
+                )
+        for header in resp.headers:
+            self.send_header(header, resp.headers[header])
```
Tansito, line 54: I don't know if this can be used here but it seems `http.server` has methods to modify `send_headers` and `end_headers`:

https://stackoverflow.com/questions/12499171/can-i-set-a-header-with-pythons-simplehttpserver

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+        url = f"{dst_protocol}://" + self.headers["Host"] + self.path
+        logging.debug("Passthrough POST request: %s", url)
+        content_length = int(self.headers["Content-Length"])
+        data = self.rfile.read(content_length)
+        job_request = self.path.find("/runtime/jobs") != -1
+        middleware_job_id = None
+        if job_request:
+            if "X-Qx-Client-Application" in self.headers:
+                logging.debug(
+                    "X-Qx-Client-Application: %s",
```
Tansito, line 116: I didn't know this header, is a custom header that we just define, @akihikokuroda ?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+                    self.headers["X-Qx-Client-Application"],
+                )
+                pos = self.headers["X-Qx-Client-Application"].find("middleware_job_id")
+                if pos != -1:
+                    middleware_job_id = self.headers["X-Qx-Client-Application"][
+                        pos
+                        + len("middleware_job_id/") : pos
+                        + len("middleware_job_id/")
+                        + 36
```
Tansito, line 125: For what is this 36 here, @akihikokuroda ?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+                    ]
+                    logging.debug("Middleware Job ID: %s", middleware_job_id)
+
+        resp = None
+        while resp is None:
```
Tansito, line 130: Curious about this, why do we need to do a retry here? Did you find problems running the POST?

proxy/proxy/httpserver.py
```diff
@@ -0,0 +1,171 @@
...
+        logging.debug("response from backend: %s", resp.status_code)
+
+
+if __name__ == "__main__":
+    server_context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
+    server_context.load_cert_chain(
+        certfile="/etc/ray/tls/tls.crt", keyfile="/etc/ray/tls/tls.key"
+    )
+    server_context.load_verify_locations(cafile="/etc/ca/tls/ca.crt")
+    server_context.verify_mode = ssl.CERT_NONE
```
Tansito, line 165: This is totally outside my knowledge so just asking. We are configuring certificates for the proxy but configuring the mode as: `ssl.CERT_NONE`. Should not we use another mode?

proxy/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM python:3.9.16-slim-buster
```
Tansito, line 1: Can we setup docker-verify workflow with this Dockerfile?

https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/.github/workflows/docker-verify.yaml

gateway/proxy/proxy/entrypoint.sh
akihikokuroda: @IceKhan13 The `url` doesn't work.  It is used in the constructor of the `QiskitRuntimeService` to access the auth server.  The  `QiskitRuntimeService` is created at the client system instead of the ray node so if the url is pointing to the proxy, it fails in the constructor.

.github/workflows/kubernetes-deploy.yaml
akihikokuroda: Yes, the proxy image must be loaded into kind for this test.

proxy/proxy/httpserver.py
akihikokuroda: No, we can not assume them.  We have to check the header for them.

proxy/proxy/httpserver.py
akihikokuroda: I believe that It has `end_headers` but doesn't have `send_headers`. https://github.com/python/cpython/blob/3.12/Lib/http/server.py

proxy/proxy/httpserver.py
akihikokuroda: No, this is defined by `qiskit_ibm_runtime`.

proxy/proxy/httpserver.py
akihikokuroda: It's the length of the middleware jib id.  I'll make a constant for it .

proxy/proxy/httpserver.py
akihikokuroda: It sometimes causes `ChunkedEncodingError`.  People seem retrying to workaround it.

proxy/proxy/httpserver.py
akihikokuroda: It may not have any effects because it is set at the server side.  I'll try if it works without setting it.

proxy/Dockerfile
akihikokuroda: OK.  I'll try.  I still have to create the proxy-verify workflow, too.

proxy/proxy/httpserver.py
Tansito: Thanks for the link, Aki. I will take a look 🙏

proxy/proxy/httpserver.py
akihikokuroda: It's reading the request contents.

proxy/proxy/httpserver.py
akihikokuroda: I looked into some other options but I couldn't find a good one to handle low level data for the proxy.

proxy/proxy/httpserver.py
Tansito: How do you see to add a limit of retries + a delay between retries? Thinking in:
- If the service had a problem not send requests indefinitely
- Letting some space between retries.

proxy/proxy/httpserver.py
Tansito: Or if another mode has more sense. I don't really know here 😂

proxy/proxy/httpserver.py
Tansito: for what is this number? Can this number change?

akihikokuroda: I don't remember where I picked this value.  It's just a big enough number for the request data size.

Tansito: This data size can change depending of the input of the user? Because something that we are discussing is that for the uses cases that we are going to need to handle we can expect a couple of hundred of megabytes in some user inputs.

akihikokuroda: I looked this line again.  It is reading the request uri and headers.  It doesn't grow  by the user input.  The user data in the POST requests are read in the `do_POST` with the `Content-Length` header value.

akihikokuroda:
This https://github.com/Qiskit/qiskit-ibm-runtime/issues/1508 must be fixed to make this work correctly.

proxy/proxy/httpserver.py
psschwei: what about a simple flask app, something like:

```python
from flask import Flask, request, Response
import requests

app = Flask(__name__)


@app.route('/', methods=['POST'])
def proxy_request():
    """Append a header and pass request on."""
    runtime_url = "https://runtime.quantum.ibm.com"
    gateway_url = "https://middleware.quantum.ibm.com"

    headers = dict(request.headers)
    data = dict(request.data)
    # add our header(s)
    headers['serverless-id'] = "12345"
    
    # Send request to runtime
    r = requests.post(runtime_url, headers=headers, data=data)
    # update our DB
    r2 = requests.post(gateway_url+"/update/jobconfig", ....)

    # return response to user
    resp = Response("request forwarded")
    resp.headers = r.headers
    resp.set_data(r.text)
    return resp


if __name__ == "__main__":
    app.run(debug=True, host='0.0.0.0', port=int(8080))
```

and since we need to use Istio anyways, in theory we could lean on [Istio's traffic management](https://istio.io/latest/docs/concepts/traffic-management/) to redirect user calls to runtime to our proxy service...

Not sure how practical this would be, just trying to help think of other options if `http.server` won't work...

proxy/proxy/httpserver.py
Tansito: Thank you so much, Aki. It would be nice if we could clarify from where this number comes from 🙏

akihikokuroda: I found the RFC that has the recommended value for this.  I made a constant for it and put the reference in the comment.

Tansito: Thank you @akihikokuroda

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
psschwei: sorry, missed this one with all the other threads...

assuming we use Istio to encrypt intracluster traffic, would that be sufficient to mitigate the risk? 

The other thought I had was if some of the complexity (dipping into iptables for example) could be avoided by running as a separate service, something like [this](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1232#discussion_r1544833488) maybe... also leveraging Istio to handle the redirect... just thinking out loud, don't have a strong opinion here (other than the simpler the better :) )

psschwei:
@akihikokuroda is this ready for review?

akihikokuroda:
@psschwei Yes, it is.  Thanks!

proxy/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
```
psschwei, line 1: Now that we've bumped to python 3.10 as default

```suggestion
FROM python:3.10.12-slim-buster
```

psschwei (commented):
Running locally and not using the runtime service, the proxy container failed on my ray worker with a permission error when trying to load the certs.

Full logs from proxy container:
```
entrypoint.sh
iptables v1.8.2 (nf_tables): Chain already exists
iptables v1.8.2 (nf_tables): Chain already exists
Traceback (most recent call last):
  File "/home/worker/httpserver.py", line 172, in <module>
    server_context.load_cert_chain(
PermissionError: [Errno 13] Permission denied
```

No issues on the head node though...

psschwei:
weird, looks like the head and worker nodes have different permissions on the certs:
```
$ k exec -it c-mockuser-ed758c49-head-7xn7g -- bash
Defaulted container "ray-head" out of: ray-head, proxy, ray-head-tls (init)
(base) root@c-mockuser-ed758c49-head-7xn7g:/# ll /etc/ray/tls/
total 36
drwxrwsrwx. 1 root  123  186 Apr 11 09:41 ./
drwxr-xr-x. 1 root root    6 Apr 11 09:41 ../
drwxr-sr-x. 1 root  123   24 Apr 11 09:41 ..2024_04_11_16_41_07.3713608326/
lrwxrwxrwx. 1 root  123   32 Apr 11 09:41 ..data -> ..2024_04_11_16_41_07.3713608326/
lrwxrwxrwx. 1 root  123   13 Apr 11 09:41 ca.crt -> ..data/ca.crt
-rw-r--r--. 1 root  123 1159 Apr 11 09:41 ca.csr
lrwxrwxrwx. 1 root  123   13 Apr 11 09:41 ca.key -> ..data/ca.key
-rw-r--r--. 1 root  123   41 Apr 11 09:41 ca.srl
-rw-r--r--. 1 root  123  301 Apr 11 09:41 cert.conf
-rw-r--r--. 1 root  123  370 Apr 11 09:41 csr.conf
-rw-r--r--. 1 root  123 1387 Apr 11 09:41 tls.crt
-rw-r--r--. 1 root  123 1679 Apr 11 09:41 tls.key
(base) root@c-mockuser-ed758c49-head-7xn7g:/# exit
exit

$ k exec -it c-mockuser-ed758c49-worker-g-jk6dz -- bash
Defaulted container "ray-worker" out of: ray-worker, proxy, ray-worker-tls (init), wait-gcs-ready (init)
(base) root@c-mockuser-ed758c49-worker-g-jk6dz:/# ll /etc/ray/tls
total 36
drwxrwsrwx. 1 root  123  186 Apr 11 09:41 ./
drwxr-xr-x. 1 root root    6 Apr 11 09:41 ../
drwxr-sr-x. 1 root  123   24 Apr 11 09:41 ..2024_04_11_16_41_07.2639855063/
lrwxrwxrwx. 1 root  123   32 Apr 11 09:41 ..data -> ..2024_04_11_16_41_07.2639855063/
lrwxrwxrwx. 1 root  123   13 Apr 11 09:41 ca.crt -> ..data/ca.crt
-rw-r--r--. 1 root  123 1086 Apr 11 09:41 ca.csr
lrwxrwxrwx. 1 root  123   13 Apr 11 09:41 ca.key -> ..data/ca.key
-rw-r--r--. 1 root  123   41 Apr 11 09:41 ca.srl
-rw-r--r--. 1 root  123  238 Apr 11 09:41 cert.conf
-rw-r--r--. 1 root  123  307 Apr 11 09:41 csr.conf
-rw-r--r--. 1 root  123 1306 Apr 11 09:41 tls.crt
-rw-------. 1 root  123 1679 Apr 11 09:41 tls.key
```

And since the proxy runs as nonroot (or at least it does on the head node), that would explain the error...

proxy/proxy/httpserver.py
psschwei: > couldn't find a good one to handle low level data for the proxy

What kind of low-level data was causing problems? While it seems like the security issues mentioned in the python docs don't apply in our case, the "`http.server` is not recommended for production" warning makes me a little uncomfortable...

akihikokuroda:
@psschwei I remember that I tweaked some file permission.  I'll look into it.  Thanks!

proxy/proxy/httpserver.py
akihikokuroda: I believe that our use of this http.server is safe.  We handle all requests by ourself. So I guess it's OK.  The proxy implementation may not requires very low level but it requires header handling and raw data manipulation. Thanks!

psschwei: I think it's more likely than not that we're safe too, but I'm not as confident as I'd like...
@Tansito @IceKhan13 wdyt?

Tansito: Yeah, I agree with @psschwei here. If there was the only option to manage header handling and raw data manipulation I agree but I would like to take a look to other more "production ready" options.

akihikokuroda:
This is ready for review.  It's using `gunicorn` as the webserver (wsgi server) now.

akihikokuroda:
I tried to remove the iptables but I couldn't find any other way without them.

Tansito:
Maybe has it sense to use istio for this?

akihikokuroda:
@Tansito I'm not sure if istio can help for this.

charts/quantum-serverless/charts/gateway/values.yaml
```diff
@@ -27,6 +27,7 @@ application:
     maxReplicas: 4
     opensslImage: registry.access.redhat.com/ubi8/openssl:8.8-9
     kubectlImage: alpine/k8s:1.29.2@sha256:a51aa37f0a34ff827c7f2f9cb7f6fbb8f0e290fa625341be14c2fcc4b1880f60
+    proxyImage: "icr.io/quantum-public/quantum-serverless-proxy:0.9.0"
```
psschwei, line 30: Should this be added to the update components action?

proxy/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN pip install --upgrade pip==24.0 --no-cache-dir
+
+RUN adduser --disabled-password --gecos '' worker
+RUN adduser worker sudo
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
+USER worker
+WORKDIR /home/worker
+
+RUN pip install gunicorn==22.0.0 flask==3.0.3 certifi==2024.2.2 requests==2.31.0 --no-cache-dir
```
psschwei, line 20: Can we add these versions to the requirements file and then just do a `pip install -r requirements.txt`?

proxy/proxy/entrypoint.sh
```diff
@@ -0,0 +1,16 @@
...
+sudo iptables -t nat -N proxy_redirect
+sudo iptables -t nat -A proxy_redirect -p tcp -d 104.18.29.94 -j REDIRECT --to-ports 8443
+sudo iptables -t nat -A proxy_redirect -p tcp -d 104.18.28.94 -j REDIRECT --to-ports 8443
+
+sudo iptables -t nat -N proxy_output
+sudo iptables -t nat -A proxy_output -p tcp --dport 6379 -j RETURN
+sudo iptables -t nat -A proxy_output -p tcp -m owner --gid-owner 123 -j RETURN
+sudo iptables -t nat -A proxy_output -j proxy_redirect
+
+sudo iptables -t nat -A OUTPUT -p tcp -j proxy_output
```
psschwei, line 14: Would it make sense to do this setup in an init container rather than in the entrypoint script? (thinking that might also remove the need for running the proxy as privileged...)

charts/quantum-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -141,28 +141,80 @@ data:
                 - name: OT_ENABLED
                   value: {{ .Values.application.ray.openTelemetryCollector.enabled | quote }}
 {{- end }}
-            - image: fluent/fluent-bit:1.9.10
-              name: ray-head-logs
+            - image: {{ .Values.application.ray.proxyImage }}
```
psschwei, line 144: I think it might be a good idea to add a Helm config option to enable / disable the proxy. I'd also make it default to disabled.

psschwei (commented):
I think we can provisionally merge this after the last couple of comments are addressed, but before we flip it on in production I think we'll need to revisit cert verification.

docs/examples/source_files/qaoa/qaoa.py
```diff
@@ -66,11 +66,11 @@ def run_qaoa(
...
         backend = arguments.get("backend", "ibmq_qasm_simulator")
-        with Session(service=service, backend=backend) as session:
-            options = Options()
-            options.optimization_level = 3
+        session = Session(service=service, backend=backend)
+        options = Options()
+        options.optimization_level = 3
 
-            estimator = Estimator(session=session, options=options)
+        estimator = Estimator(session=session, options=options)
```
psschwei, line 72: I can't comment earlier in this file, but when trying to run the QAOA notebook I had to delete the import of `QuantumServerless` in https://github.com/akihikokuroda/quantum-serverless/blob/sidecarproxy/docs/examples/source_files/qaoa/qaoa.py#L14

proxy/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM python:3.9.16-slim-buster
```
psschwei, line 1:
```suggestion
FROM python:3.10.12-slim-buster
```

psschwei (commented):
Couple of nits, otherwise LGTM

docs/examples/source_files/qaoa/qaoa.py
psschwei: fixed while I was writing my review :smile:

proxy/proxy/entrypoint.sh
akihikokuroda: Thanks!  It was in my to-do list.  I change it now.

psschwei (approved):
Nice work @akihikokuroda !

akihikokuroda:
Thanks!  I'll merge this finally!

Tansito:
Hahaha really good work @akihikokuroda , thank you ❤️

Issue #1238, opened 2024-03-06
Errors when executing running_programs_using_decorators notebook


psschwei:
### Steps to reproduce the problem

Run the [decorators notebook](https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/docs/getting_started/experimental/running_programs_using_decorators.ipynb)

Check job status and logs

### What is the current behavior?

```python
job.result()
# {}

job.status()
# 'ERROR'

job.logs()
# 'Traceback (most recent call last):\n  File "/tmp/ray/session_2024-03-06_10-13-47_055610_14/runtime_resources/working_dir_files/_ray_pkg_687e472019169c40/entrypoint_60573991.py", line 10, in <module>\n    result = function(**arguments)\nTypeError: __main__.hello_qiskit() argument after ** must be a mapping, not str\n'
```

This is for the Hello, Qiskit section, but you get similar results in the other sections

### What is the expected behavior?

Successful job completion and results, no error in logs

Tansito:
I'm going to take a look into this that is blocking us some developments.

Tansito:
Ok, I don't know when we introduced this _bug_ but it seems `get_arguments` is returning a `str` instead of a `dict` even for our standard executions. 

The problem is that in the decorator we use `function(**arguments)` and it's there where the problem popped up. If we used that approach in our standard flow we would have the same error. I'm going to review our flow and check where the problem it is.

What I don't know is why this problem didn't appear before because we were not having too many changes lately, any idea @psschwei ?

psschwei:
> What I don't know is why this problem didn't appear before. We were not having too many changes lately, any idea @psschwei ?

I think it was being masked by using `pytest` to run the notebooks, similar to what Aki found in #1094 ... basically, it was running all the cells but only checking _that_ they ran, not whether they were _successful_ or not

Issue #1166, opened 2024-01-17
Docs: move docs to IBM docs


IceKhan13:
### What is the expected enhancement?

Move documentation from gh to ibm docs.

Epic https://github.com/Qiskit-Extensions/quantum-serverless/issues/1164

IceKhan13:
I think you are right about default provider. But I think it is still worth it to have local provider for tests. 

What I think will not be needed anymore is https://github.com/Qiskit-Extensions/quantum-serverless/issues/812

psschwei:
Let me see if I can update the notebooks to run using local provider...

psschwei:
Think we will also want to purge some of the docs on the move (for example, installation and deployment may no longer be needed)

Tansito:
Our documentation is now leaving in IBM Quantum and we will need to deprecate our GitHub pages workflow. Closing this issue in favor of new issues to handle the current status of the project.

Issue #1094, opened 2023-11-15
Docker notebook test doesn't detect notebook execution failure


akihikokuroda:
### Steps to reproduce the problem
Run CI build with some notebooks that have failures in program execution

### What is the current behavior?
Docker notebook test completes successfully.

### What is the expected behavior?
Docker notebook test detests failure of the program execution and fails test.

I simulated the test locally

test output:

```
(base) jovyan@jupyter-7ffdf5486b-htxb9:~$ pytest --nbmake --ignore=/home/jovyan/serverless/guides/07_working_with_datasets.ipynb /home/jovyan/serverless/
=========================================================================================== test session starts ============================================================================================
platform linux -- Python 3.9.13, pytest-7.4.3, pluggy-1.3.0
rootdir: /home/jovyan
plugins: anyio-3.6.1, nbmake-1.4.6
collected 9 items                                                                                                                                                                                          

serverless/examples/01_vqe.ipynb .                                                                                                                                                                   [ 11%]
serverless/examples/02_qaoa.ipynb .                                                                                                                                                                  [ 22%]
serverless/getting_started/basic/01_running_program.ipynb .                                                                                                                                          [ 33%]
serverless/getting_started/basic/02_arguments_and_results.ipynb .                                                                                                                                    [ 44%]
serverless/getting_started/basic/03_dependencies.ipynb .                                                                                                                                             [ 55%]
serverless/getting_started/basic/04_distributed_workloads.ipynb .                                                                                                                                    [ 66%]
serverless/getting_started/basic/05_retrieving_past_results.ipynb .                                                                                                                                  [ 77%]
serverless/getting_started/experimental/file_download.ipynb .                                                                                                                                        [ 88%]
serverless/getting_started/experimental/running_programs_using_decorators.ipynb 
```

Logs of Scheculer
```
DEBUG 2023-11-15 14:03:32,212 ray.py:258 : Head node http://c-user-0d942d0d-head-svc:8265/ is not ready yet.
DEBUG 2023-11-15 14:03:33,217 ray.py:258 : Head node http://c-user-0d942d0d-head-svc:8265/ is not ready yet.
DEBUG 2023-11-15 14:03:34,226 ray.py:258 : Head node http://c-user-0d942d0d-head-svc:8265/ is not ready yet.
DEBUG 2023-11-15 14:03:35,231 ray.py:258 : Head node http://c-user-0d942d0d-head-svc:8265/ is not ready yet.
[...]

PR #1239, opened 2024-03-06
Use scripts instead of notebooks for tests

psschwei:
### Summary

Run tests against python scripts instead of jupyter notebooks

### Details and comments

There's two reasons for doing this:

* we're going to be moving the docs out of this repo in #1166 , so we'll need something to test against once that happens
* we get more granular logs on each job (rather than just a blanket pass/fail), which will partially help with things like #1094 

Note that the decorators test fails (xref #1238)

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -22,13 +22,19 @@ jobs:
           python-version: '3.9'
       - name: install dependencies
         shell: bash
-        run: pip install client/ && pip install nbmake pytest
-      - name: Run basic notebooks
+        run: pip install client/
+      - name: Run basic tests
         shell: bash
-        run: pytest --nbmake docs/getting_started/basic/
-      - name: Run experimental notebooks
+        run: |
+          cd tests/basic
+          for f in *.py; do echo "$f" && python "$f"; done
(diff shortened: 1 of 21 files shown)


psschwei:
We also now see all the warnings when the LocalProvider tries to use unimplemented file methods : https://github.com/Qiskit-Extensions/quantum-serverless/actions/runs/8177774823/job/22360183087?pr=1239

Tansito:
Oh wow, good refactorization @psschwei . Is it ready for review or are you testing something more?

psschwei:
> Is it ready for review or are you testing something more?

Ready for review :smile:

Tansito:
Do you update this pull request with the last change fixing the decorator, @psschwei ? I can take a look into this tomorrow 🙏

psschwei:
I'll rebase now

Tansito (commented):
I like the idea of the pull request. My only comment @psschwei is if you could check if we can remove `nbmake` dependency with this from the requirements:

https://github.com/Qiskit-Extensions/quantum-serverless/blob/main/client/requirements-dev.txt#L20

I think we don't need it with these last changes but I'm not sure 😅

Issue #1240, opened 2024-03-06
Deprecate services in favor of serializers


Tansito:
### What is the expected enhancement?

In #1147 I started a refactorization moving bussiness logic from `views` to `services`. Investing some more time learning about DRF I think Serializers can do the work that Services are doing right now simplifying the current logic. I will create a little PR testing this approach to see how it fits with `Programs` entity for example.

- [x] Upload end-point refactorization
  - #1243 
- [x] RunExisting end-point refactorization: 
  - #1245 
  - #1267
- [ ] Run end-point refactorization
  - #1286

PR #1245, opened 2024-03-12
Fix ExistingProgram serializer

### LINKED ISSUE #1238: Errors when executing running_programs_using_decorators notebook

Tansito:
### Summary

Fix #1238 

End-points `run` and `run_existing` where using different serializers. One was converting the `arguments` into a JSON and the another one to `str`. This pull request unify those serializers.

Later iterations of the work that we are doing in #1240 will improve this logic.

### DIFF
--- a/gateway/api/serializers.py
+++ b/gateway/api/serializers.py
@@ -71,7 +71,7 @@ class ExistingProgramSerializer(serializers.Serializer):
     """Serializer for launching existing program."""
 
     title = serializers.CharField(max_length=255)
-    arguments = serializers.JSONField()
+    arguments = serializers.CharField()
 
     def update(self, instance, validated_data):
         pass
--- a/gateway/api/services.py
+++ b/gateway/api/services.py
@@ -145,7 +145,7 @@ class JobService:
     @staticmethod
     def save(
         program: Program,
-        arguments,
+        arguments: str,
         author,
         jobconfig: JobConfig,
         token: str,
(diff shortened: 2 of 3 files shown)


psschwei:
Why'd you go for strings over json? (I don't have an opinion here, just curious)

Tansito:
My reasons for this approach were:
- Unify both serializers: `Program` & `ExistingProgram` (attention: we are using bad the serializers here because `arguments` go with `Job` model, not with `Program` but it could add more logic to fix and it's a refactorization that I will do in #1240 )
- For `Job` and `Program` arguments are just a text field and we are not using them in the code before storing them so instead of do Text -> JSON -> Text (without needing the json) I just preferred to let them as Text.

Tansito:
Tests fixed and ready for review!

psschwei (approved):

Issue #1251, opened 2024-03-14
Setup Python 3.10 as default and deprecate 3.8


Tansito:
### What is the expected enhancement?

Taking account last messages from #484 I think it would be nice to start to prepare the project to remove Python 3.8 support and run our images with Python 3.10 by default (right now we are building them with 3.9).

We would need to continue taking a look to Python 3.11 but not a priority I would say.

psschwei:
I think the two PRs above cover all we need here

psschwei:
This was done in https://github.com/Qiskit/qiskit-serverless/releases/tag/v0.10.0

PR #1258, opened 2024-03-26
Drop python 3.8 support

psschwei:
### Summary

Part of #1251 
Remove Python 3.8 support from builds / client / gateway / etc.

### Details and comments

Did not update https://github.com/Qiskit-Extensions/quantum-serverless/blob/841db743ff3334fa95e277c53093e30bd0a441d7/gateway/api/migrations/0012_jobconfig_python_version.py#L19 as it may still be a valid option there for historical data

### DIFF
--- a/gateway/api/models.py
+++ b/gateway/api/models.py
@@ -32,11 +32,9 @@ class JobConfig(models.Model):
         null=True,
     )
 
-    PYTHON_V3_8 = "py38"
     PYTHON_V3_9 = "py39"
     PYTHON_V3_10 = "py310"
     PYTHON_VERSIONS = [
-        (PYTHON_V3_8, "Version 3.8"),
         (PYTHON_V3_9, "Version 3.9"),
         (PYTHON_V3_10, "Version 3.10"),
     ]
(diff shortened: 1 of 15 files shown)


gateway/api/models.py
```diff
@@ -32,11 +32,9 @@ class JobConfig(models.Model):
         null=True,
     )
 
-    PYTHON_V3_8 = "py38"
```
Tansito, line 35: Yep, I totally agree with you, @psschwei. If we are not going to migrate it my proposal is more or less what you did here:
- Modify the serializer
- I would leave this model without changes

So the serializer is what we use to manage valid python versions. Other proposals @IceKhan13 , @akihikokuroda ?

akihikokuroda (approved):
LGTM. Thanks!

Tansito (commented):
My doubt continuous being the same. Should we remove the change in the model maintaining the serializer? Taking into account the we want to maintain the data in the DB and we are not applying a migration.

gateway/api/models.py
IceKhan13: I'm good with that :)

IceKhan13 (approved):

Tansito (approved):
Thank you, Paul!

Issue #1264, opened 2024-03-27
Client + gateway: improve error handling 


IceKhan13:
### What is the expected enhancement?

Errors are handled in unoptimal way. When request failed we throw generic expection. What we should do is provide explanation what actually happened.

Tansito:
I can take a look into this one as part of #1240

Tansito:
We are in a refactor plan so we will review this issue more in-depth. I will close this one in favor of new issues more near to the current status 🙏

PR #1267, opened 2024-03-28
Refactorization for run_existing end-point

Tansito:
### Summary

Continuing with #1240 and #1243 this is a proposal to move views and services logic to serializers.

The idea behind this refactorization is:
- Use more serializers in the different views
- Concentrate the access to the models in the serializers
- We move from a non-standard class: service, to a first-class member for DRF: serializers

This PR concentrates the work in the `run_existing` end-point.

### Details and comments
- Two new serializers are created:
  - **RunExistingProgramSerializer**: it will manage the input from the request for the rest of the logic
  - **RunExistingJobSerializer**: specific logic for `Job` in this end-point.
- Models optimization:
  - Field `created` is not editable
  - Fixed `__str__` method in `JobConfig` model
  - Fixed default value for `Job.status`. We were using `QUEUE` by default.
 
The rest changes are minor fixes:
- Removed Custom exceptions. They are not working properly. I will take a look later as part of  #1264 
- From #1245 fixed the type of `build_env_variables.arguments`


Tansito:
I didn't remove `services` or review if I can reuse a `serializer` in other end-point just to avoid to add more changes to the PR. I will do it later on.

akihikokuroda (approved):

PR #1271, opened 2024-03-28
Additional logging for auth errors on custom backends

### LINKED ISSUE #1263: Gateway: log authentication errors and reasons.

Currently if user does not pass custom authentication we just show auth error. 

We need to log out error in gateway with info what is missing for auth.

psschwei:
### Summary

Fixes #1263 

### Details and comments

Adds log messages for each condition that might fail in `CustomTokenBackend.authenticate()`. Also logs an error if a safe request comes back as not ok.

I opted for explicit `if` statements (rather than using `else` or `elif`) to try to make it clearer which error conditions were being checked (it was easy to miss which else lined up with which if when relying solely on whitespace...)

I don't have an easy way to test against a third-party auth service (like IQP) locally, so these may require some hand checking.

### DIFF
--- a/gateway/api/authentication.py
+++ b/gateway/api/authentication.py
@@ -36,7 +36,7 @@ def safe_request(request: Callable) -> Optional[Dict[str, Any]]:
         except Exception as json_exception:  # pylint: disable=broad-exception-caught
             logger.error(json_exception)
     if response is not None and not response.ok:
-        logger.error(f"{response.status_code}: {response.text}")
+        logger.error("%d : %s", response.status_code, response.text)
 
     return result
 
--- a/gateway/api/authentication.py
+++ b/gateway/api/authentication.py
@@ -91,22 +91,20 @@ class CustomTokenBackend(authentication.BaseAuthentication):
                             user = User(username=user_id)
                             user.save()
                     elif user_id is None:
-                        logger.warning("No user id.")
-
+                        logger.warning("Problems authenticating: No user id.")
                     else:  # not verified
-                        logger.warning("User is not verified.")
+                        logger.warning("Problems authenticating: User is not verified.")
 
                 else:  # verification_data is None
-                    logger.warning("No verification data returned from request.")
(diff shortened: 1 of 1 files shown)


gateway/api/authentication.py
```diff
@@ -88,6 +90,23 @@ def authenticate(self, request):
                         except User.DoesNotExist:
                             user = User(username=user_id)
                             user.save()
+                    if user_id is None:
+                        logger.warning("No user id.")
+
+                    if not verified:
+                        logger.warning("User is not verified.")
+
+                if verification_data is None:
```
akihikokuroda, line 99: Is there any reason using **if xxxx** instead of **else** here?

gateway/api/authentication.py
```diff
@@ -88,6 +90,23 @@ def authenticate(self, request):
...
+                    logger.warning("No verification data returned from request.")
```
akihikokuroda, line 100: Some of messages are not user friendly.  What causes no verification data or no authorization data?

akihikokuroda (commented):
Thanks for making changes.  I have minor comments.

gateway/api/authentication.py
psschwei: I found it hard to determine which else matched up with which if (one of the few times I wished python used brackets instead of whitespace), so I tried to make the blocks more explicit by using ifs

akihikokuroda: I probably like `else` with comments.  It makes clear that all cases (only 2 though) are covered.

gateway/api/authentication.py
psschwei: I added a prefix to the logs that there were problems authenticating to try to make things more clear. Not sure I can be more specific though, I think the goal is just to alert in all the different chances where something might go wrong, though I think we might have to wait until it actually goes wrong to know why it went wrong, if that makes sense

akihikokuroda (approved):
LGTM  Thanks!

PR #1211, opened 2024-02-05
Add catalog status

akihikokuroda:
### Summary
Fix:  #1210  

### Details and comments
Waiting for Catalog API design discussion

PR #1204, opened 2024-01-31
CatalogEntry API

akihikokuroda:
### Summary
Part of #1170 

### Details and comments
This PR adds [ModelView APIs](https://github.com/Qiskit-Extensions/quantum-serverless/issues/1170#issuecomment-1917931921) for CatalogEntry. 

The list method will be enhanced to handle the query string for `title`, `description` and `tags`.

PR #1276, opened 2024-04-05
Remove catalog implementation

Tansito:
### Summary

This PR reset changes from #1181, #1204, #1211


psschwei:
nit: since this is deleting a bunch of code, I'd retitle to something like "Remove catalog implementation"

Tansito:
How these migrations where not published I think I preferred to remove them but let me know if you prefer to create a new migration file. I would agree too.

akihikokuroda:
>How these migrations where not published I think I preferred to remove them but let me know if you prefer to create a new migration file. I would agree too.

What number is used for the next migration file?  Is it the next to the existing highest one or one of the removed ones? @Tansito

Tansito:
I think it got the last one but not 100% sure. Let me confirm it, @akihikokuroda , good question

gateway/api/migrations/0014_runtimejob.py
```diff
@@ -7,7 +7,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("api", "0013_program_groups_catalogentry"),
+        ("api", "0012_jobconfig_python_version"),
```
IceKhan13, line 10: Removing migrations or changing order of migration will break existing databases. If we want to remove code, related to models we need to remove models and generate new migration which will delete tables from datase.

Long story short: migrations are always append only :)

IceKhan13 (changes requested):

gateway/api/migrations/0014_runtimejob.py
IceKhan13: oops! Just saw your last comment about not published. 

I think it is still better to append migrations, if it is not hard to do

Tansito: Solved in [674a79c](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1276/commits/674a79c9350452c8bc97fef38866a5c8f933dc3b)

IceKhan13 (approved):
Thank you @Tansito !

akihikokuroda (approved):

PR #1279, opened 2024-04-09
add version api endpoint to gateway

### LINKED ISSUE #1277: Add version endpoint to gateway API

Provide a new API endpoint to return the version of the quantum-serverless.

akihikokuroda:
### Summary
fix #1277 

This PR adds `/version/` endpoint to the gateway.  It returns `{"version": "0.9.0"}` now.

### Details and comments

The version is defined as the global variable of the quantum-serverless helm chart.

### DIFF
--- a/gateway/main/urls.py
+++ b/gateway/main/urls.py
@@ -22,6 +22,7 @@ from rest_framework import routers, permissions
 from drf_yasg.views import get_schema_view
 from drf_yasg import openapi
 import probes.views
+import version.views as version
 
 schema = get_schema_view(  # pylint: disable=invalid-name
     openapi.Info(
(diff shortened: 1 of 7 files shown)


gateway/main/urls.py
```diff
@@ -22,6 +22,7 @@
 from drf_yasg.views import get_schema_view
 from drf_yasg import openapi
 import probes.views
+import version.views as version
```
Tansito, line 25: Maybe just:

```python
import version.views
```

So we can use later:

```python
path("version/", version.views.version, name="version"),
```

And maintain the same format than `probes`.

Tansito (approved):
LGTM, thank you, Aki. Just a minor comment from my side if you agree.

psschwei (commented):
LGTM aside from David's comment

gateway/main/urls.py
akihikokuroda: Thanks!  I update.

psschwei (approved):

PR #1280, opened 2024-04-10
Instances integration

### LINKED ISSUE #1278: Introduce instances to manage access to different resources

We want to integrate hubs & groups in Serverless to manage access to different resources. We are going to start with the entity `Programs`. The idea is:

- Add `instances` field to `Programs` so in that way we can filter the access to that resource.
- Add `instances` method to User to obtain the hubs & groups from that user.

Tansito:
### Summary

Close #1278 

This pull request introduces a proposal to manages instances in Gateway. There are two changes:

- `instances` in Program
- `QuantumUserProxy` to obtain `network` information from a specific user

So with `network` information you can generate an instance for the user: `["ibm-q/open/main"]` and we can filter by this value in `Program.instances = ["ibm-q", "ibm-q/open", "ibm-q/open/main"]`.

### Details and comments

Some things that I would like to take into account in this pull request to review:
- I decided to go through this path of `QuantumUserProxy` to continue without storing user information in our DB
- There were various options to configure `instances` field in programs:
  - **JSONField**
  - **ArrayField**
  - **M2M**
  - I decided to go with **JSONField** due to two reasons:
    - M2M requires to store `network` or `instance` information.
    - ArrayField could be a good option but it's only available for PostgresSQL, it's not support by SQLite what we use for tests. I would be open to change it but I was reviewing different performance charts and there is no so much difference between ArrayField and JSONField (from what I found).    
- **JSONField** offers us enough functionality with `contains` and `contained_by` to filter programs.


gateway/api/models_proxies.py
```diff
@@ -0,0 +1,76 @@
...
+
+import requests
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+
+from api.utils import safe_request
+
+
+class QuantumUserProxy(get_user_model()):
```
Tansito, line 11: I think in a near future we could even refactorize `authentication` to include login and validation logic in this proxy.

IceKhan13 (commented):
Maybe we can use Groups https://docs.djangoproject.com/en/5.0/topics/auth/default/#groups. 
It is built-in functionality of django that let's you control permissions of users. 
With that you will not need to create custom user model and have to deal with Json fields. 

All you will need to do is in `authentication.py` do something like:
```python
# authentication.py
...

def update_groups(user, user_instances):
  for instance in user_instances:
     group = Group.objects.get(name=instance)
     if group is None:
         # create group
         group = Group(name=instance)
         group.save()
     user.groups.add(group)
     return user

...

user_instances = ["bla/open/blah", ...]

if user_id is not None and verified:
    try:
        user = User.objects.get(username=user_id)
    except User.DoesNotExist:
        user = User(username=user_id)
        user = update_groups(user, user_instances)
        user.save()
...
```

In `models.py` then you will add many to many instances field

```python
# models.py

class Program(models.Model):
    ...
    instances = models.ManyToManyField(Group)
    ...
```

And then to check permission you might just check overlap of groups of program and user.

```python
# premissions.py

class UserCanViewProgram(permissions.BasePermission):
    """
    Custom permission to only allow users in groups see programs.
    """

    def has_object_permission(self, request, view, obj):
        if isinstance(obj, Program):
            return not set(obj.instances).isdisjoint(request.user.groups)
```


But I'm ok with any implementation you think works :)

Tansito:
I didn't know groups so let me check, @IceKhan13 thank you!

IceKhan13:
Ah, and I also see that you do requests for user instances all the time 🤔 
Good point

Tansito:
After some internal discussion we opted to go with M2M and Groups solution. I will close this pull request and come up with a new implementation.

PR #1281, opened 2024-04-11
Remove observability helm chart

psschwei:

The last meaningful contribution to these charts was made in Sept 2023. I'm not aware of anyone using them, so given that they are stale _and_ not in use, let's remove them.


Tansito (commented):
I think with these changes, Paul we need to update some workflows too:

- https://github.com/Qiskit-Extensions/quantum-serverless/blob/32802b8dafacc46783bf9655e1daa3b7a2423a67/.github/workflows/publish-helm.yaml#L14
- https://github.com/Qiskit-Extensions/quantum-serverless/blob/32802b8dafacc46783bf9655e1daa3b7a2423a67/.github/workflows/update-component-versions.yaml#L30

Tansito (approved):
I just got a fast review and I didn't find more references. LGTM, Paul. Thank you! 😄

PR #1285, opened 2024-04-12
Change the type of config data between client and gateway

### LINKED ISSUE #1268: Use JSON standar for all our calls to the gateway

Right now from client we are not specifying the format of the `Request` to the `Gateway` what it's generating a `QueryDict` and not a standard JSON in the `Request` type. This generates in the `serializers` not expected behaviors.

We should use in all our requests from `Client` the json format. Ref: https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests

This can generate extra modifications in the Gateway if any end-point fails due to the change in the format (special attention in the serializers).

akihikokuroda:
### Summary
Fix mismatch between the data type of the `config` in the post data and the data type in the data model
Fix #1268

### Details and comments
The mismatch made some special handling for the config data in the serializer.


Tansito (dismissed):
I think these changes are good but to avoid to use `QueryDict` in the request we need to specify that a request is a json and that's changing in `requests` the field `data` by `json` from what I could read here:

https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests

```
Please note that the above code will NOT add the Content-Type header (so in particular it will NOT set it to application/json).

If you need that header set and you don’t want to encode the dict yourself, you can also pass it directly using the json parameter (added in version 2.4.2) and it will be encoded automatically:

url = 'https://api.github.com/some/endpoint'
payload = {'some': 'data'}
r = requests.post(url, json=payload)
Note, the json parameter is ignored if either data or files is passed.
```

akihikokuroda:
@Tansito I misunderstood the issue.  I think this PR is still valid improvement.  I'll change the description of this PR and disassociates from the issue #1268.

gateway/api/serializers.py
```diff
@@ -116,7 +116,7 @@ class RunExistingProgramSerializer(serializers.Serializer):
 
     title = serializers.CharField(max_length=255)
     arguments = serializers.CharField()
-    config = serializers.CharField()
+    config = serializers.DictField()
```
Tansito, line 119: Curious to know, why did you decide to use `DictField` instead of `JSONField` here, Aki?

akihikokuroda: The `config` field of the `Job` is DictField so I changed the data format of the `config` sent by the client to Dict instead of str.

Tansito: Yes, the change I think is correct. But I was just curious about the `DictField` instead to use `JSONField`:

https://stackoverflow.com/questions/76178706/django-serializer-dictfield-vs-jsonfield

akihikokuroda: I haven't thought about it. I'll test it.

akihikokuroda: It worked by changing the post request with the Context-type to json.

Tansito (approved):
LGTM @akihikokuroda , should we open issues to do the same for `arguments` and `env_vars`?

akihikokuroda:
Thanks!  No, we don't unless we want to change the field type of in the models.  I don't know if it's a good thing to do changing the field type at this moment.  WDYT?

Tansito:
Mmm... Maybe we can open a discussion for this because I don't want to add too many comments in the PR. My proposal would be something like:
- Remove `arguments` from `Job` too and send them as standard JSON
- Send env_vars as standard JSON because at the end what we are storing is the encryption of arguments + env_vars
- If this option doesn't like send `arguments` as standard JSON and encrypt them as they were `env_vars`

akihikokuroda:
@Tansito Yes, new issue or discussion is good.  Would you create one?

PR #1286, opened 2024-04-12
Run refactorization

### LINKED ISSUE #1240: Deprecate services in favor of serializers

Tansito:
### Summary

Close #1240 

This PR continues the work of #1267. The logic is pretty similar the biggest difference is that here we remove `services` references because we are moving all the logic to the `serializers`, finally.

### Details and comments

- I removed `exceptions` too because they were not working as expected. I will continue this effort as part of #1264


gateway/api/views.py
```diff
@@ -250,59 +233,81 @@ def run(self, request):
...
-                    author=author,
-                    jobconfig=jobconfig,
-                    token=token,
-                    carrier=carrier,
+            arguments = serializer.data.get("arguments")
+            token = ""
+            if request.auth:
+                token = request.auth.token.decode()
+            job_serializer = self.get_serializer_run_and_run_existing_job(data={})
+            if not job_serializer.is_valid():
```
akihikokuroda, line 293: Why is this creating the empty serializer and validating it?

gateway/api/serializers.py
```diff
@@ -188,3 +188,67 @@ class RuntimeJobSerializer(serializers.ModelSerializer):
...
+        pass
+
+
+class RunProgramModelSerializer(serializers.ModelSerializer):
+    """
+    Program model serializer for the /run end-point
+    """
+
+    arguments = serializers.CharField(read_only=True)
+    config = serializers.CharField(read_only=True)
```
akihikokuroda, line 236: I wonder why these additional fields are necessary.  Are they used?

gateway/api/views.py
Tansito: That's a good question. I'm not convinced with what I did there. My assumption was like, I had some problems creating the serializer for the `Job` because the fields where relatively custom to validate:
```
job_serializer.save(
    arguments=arguments,
    author=author,
    carrier=carrier,
    token=token,
    program=program,
    config=jobconfig,
)
``` 
And the `is_valid` is mandatory if I want to call the `save` method. I accept new ideas to improve the approach 😅

gateway/api/serializers.py
Tansito: They are needed because if you pass to the `ModelSerializer` the data from the request that data contains `arguments` and `config` and it fails if you don't specify that those keys are not going to be used by the serializer (you do that with `read_only=True`).

Tansito: It fails because the model doesn't have those attribute.

gateway/api/serializers.py
```diff
@@ -188,3 +188,67 @@ class RuntimeJobSerializer(serializers.ModelSerializer):
...
+        )
+
+    def create(self, validated_data):
+        pass
+
+    def update(self, instance, validated_data):
+        pass
+
+
+class RunProgramModelSerializer(serializers.ModelSerializer):
```
akihikokuroda, line 230: These 2 serializers are confusing for me.  Is it possible merge there 2 into 1?

gateway/api/views.py
akihikokuroda: Can you put job object like
```
job_serializer = self.get_serializer_run_and_run_existing_job(
    data = job(
        arguments=arguments,
        author=author,
        carrier=carrier,
        token=token,
        program=program,
        config=jobconfig,
    )
)
if not job_serializer.is_valid():
...
```
then save it without additional fields?
```
 job = job_serializer.save()
```

gateway/api/serializers.py
Tansito: Hahaha I just had a meeting with @IceKhan13 due to this. I totally agree with you. They cannot be merged though, because it seems you cannot configure a field in a `ModelSerializer` (or a `Serializer` that creates an instance) that it's not in the model. That's why I needed to create two models:
- One serializer to check the request
- One model serializer to safe the `instance` without taking into account additional fields (read_only=True)

Something that I want to take a look more in-depth as a next step is to create a better structure like:
- serializers
  - run: for the end-point /run
    - BodySerializers: these are serializers that check the request.data
    - ModelSerializers: these are the serializers that receives the data from the BodySerializers and create the instances
- views
  - run: it uses its own BodySerializers and ModelSerializers

I think it integrates well with swagger documentation. But there were too many changes for this set of pull requests and I'm just trying to understand how Django works.

gateway/api/views.py
Tansito: I couldn't do that due to what I explained here: https://github.com/Qiskit-Extensions/quantum-serverless/pull/1286#discussion_r1567511446

`carrier` and `token` are fields that are not in the model so I would need to create two serializers if I want to use the serializer for those. What I could maybe do is the next one:
```python
job_serializer = self.get_serializer_run_and_run_existing_job(data={
        arguments=arguments,
        program=program,
        config=jobconfig,
})
if job_serializer.is_valid():
...
job = job_serializer.save(
  author=author,
  carrier=carrier,
  token=token,
)
```

And create a specific `JobSerializer` for this case. I remember that I tried it but I had some problems configuring it. I can do a branch on top of it just exploring how to improve. I will share it here.

Tansito: Let me know what you think about the changes for this review, Aki: [8184378](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1286/commits/81843781389b9940b9a4ef73838bbb83f61aaa63)

gateway/api/serializers.py
Tansito: Btw @akihikokuroda , what we can do it is rename them to something that it has more sense. Any proposal is welcome.

gateway/api/views.py
akihikokuroda: It looks much better.  Thanks!

gateway/api/serializers.py
akihikokuroda: I can not come up with any good names.

akihikokuroda (approved):
LGTM!  Thanks!

Tansito:
I'm going to wait for your merges @akihikokuroda and wait a bit for @psschwei or @IceKhan13 if they can take a look

psschwei:
Looks like docker/kubernetes tests are both failing with the same error in the running_programs_using_decorators notebook:
```
quantum_serverless.exception.QuantumServerlessException: 
| Message: Http bad request.
| Code: 400
| Details: {"config":["This field is required."]}
```

Does the decorators code need an update?

Tansito:
I found a problem with #1285 I'm fixing it now and put here the problem, @akihikokuroda

gateway/api/serializers.py
```diff
@@ -187,3 +187,71 @@ class RuntimeJobSerializer(serializers.ModelSerializer):
...
+    """
+    Program serializer for the /run end-point
+    """
+
+    title = serializers.CharField(max_length=255, allow_blank=False)
+    entrypoint = serializers.CharField(max_length=255)
+    artifact = serializers.FileField(allow_empty_file=False, use_url=False)
+    dependencies = serializers.CharField(allow_blank=False)
+    arguments = serializers.CharField(allow_blank=False)
+    config = serializers.CharField(allow_blank=False)
```
akihikokuroda, line 202: I think this was `JSONField`.  Then you can put back the changes in #1285.

Tansito:
Tests passed so here is what I found, @akihikokuroda . With `run` end-point if we set `config` the data that we generate in the client is the next one:

```python
Configuration(workers=None, min_workers=None, max_workers=None, auto_scaling=False, python_version='py39')
```

```json
{'title': 'my-first-pattern', 'entrypoint': 'basic-pattern.py', 'arguments': '{"test": "here it goes an argument"}', 'dependencies': '[]', 'config': {'workers': None, 'min_workers': None, 'max_workers': None, 'auto_scaling': False, 'python_version': 'py39'}}
```

Until here everything is fine with the changes from #1285 

The problem comes in the `gateway`:

```python
<QueryDict: {'title': ['my-first-pattern'], 'entrypoint': ['basic-pattern.py'], 'arguments': ['{"test": "here it goes an argument"}'], 'dependencies': ['[]'], 'config': ['workers', 'min_workers', 'max_workers', 'auto_scaling', 'python_version'], 'artifact': [<InMemoryUploadedFile: artifact.tar ()>]}>
```

I don't know why but Django transforms `config` into: 

```
'config': ['workers', 'min_workers', 'max_workers', 'auto_scaling', 'python_version']
```

What it generates the error. We didn't find this problem basically because we didn't have a test for `run` end-point with `config` and old logic was not taking into account `config` format. So I needed to restore the changes for `run` method from #1285 .

My question now is, with `Content-type: multipart/form-data` we cannot send JSON (I don't know if the problem is the content-type or the QueryDict). Should we send `config` as string in `run_existing` too to just not have different logic between these end-points, @akihikokuroda ?

gateway/api/serializers.py
Tansito: We almost write at the same time. Check my comment here, Aki: https://github.com/Qiskit-Extensions/quantum-serverless/pull/1286#issuecomment-2062135522

I changed to `JSONField` but if I change it the I receive an error: "JSON format invalid".

akihikokuroda:
I saw Django handled the Content-type and the config field without converting to the string.  You may merge this PR as-is.  I can try the #1285 later.

Tansito:
Ok, I'm going to prepare a test for this anyway before merge and put on-hold #1290 until we clarify this.

psschwei:
> With `run` end-point if we set `config` the data that we generate in the client is the next one:
> 
> ```python
> Configuration(workers=None, min_workers=None, max_workers=None, auto_scaling=False, python_version='py39')
> ```
> 
> ```
> {'title': 'my-first-pattern', 'entrypoint': 'basic-pattern.py', 'arguments': '{"test": "here it goes an argument"}', 'dependencies': '[]', 'config': {'workers': None, 'min_workers': None, 'max_workers': None, 'auto_scaling': False, 'python_version': 'py39'}}
> ```

Is that configuration part of our code somewhere? (wasn't clear if it was or just an example you made up) If so, do we need to update the python version to py310?

Tansito:
> Is that configuration part of our code somewhere? (wasn't clear if it was or just an example you made up) If so, do we need to update the python version to py310?

Oh no, no. It's just some examples that I'm doing to test things locally hehe.

gateway/api/serializers.py
```diff
@@ -138,9 +138,9 @@ def create(self, validated_data):
         pass
 
 
-class RunExistingJobSerializer(serializers.ModelSerializer):
+class RunAndRunExistingJobSerializer(serializers.ModelSerializer):
```
psschwei, line 141: nit: this name is a bit clunky, would it make sense to simplify to:

```suggestion
class RunJobSerializer(serializers.ModelSerializer):
```

gateway/api/serializers.py
```diff
@@ -138,9 +138,9 @@ def create(self, validated_data):
...
     """
-    Job serializer for the /run_existing end-point
+    Job serializer for the /run and /run_existing end-point
```
psschwei, line 143: long-term thought: would it make sense to refactor the API to just use a single `/run` endpoint and distinguish within the call whether the program/function/pattern/job/building_block/element/quantum_magic has already been uploaded or not?

gateway/api/v1/views.py
```diff
@@ -11,7 +11,6 @@
 from api.models import Program, Job, RuntimeJob
 from api.permissions import IsOwner
 from . import serializers as v1_serializers
-from . import services as v1_services
```
psschwei, line 14: this is maybe more for a later refactoring (change wasn't introduced in this PR, so let's not change now), but do we call non-v1 serializers here? if not, given that we're already in a `v1/` directory the prefix is a little redundant

psschwei (commented):
I have a question about names, but otherwise LGTM

gateway/api/serializers.py
Tansito: Definitely I think it's something that we can propose. I had in mind to continue the improvements in the gateway from an code organization point of view: https://github.com/Qiskit-Extensions/quantum-serverless/pull/1286#discussion_r1567511446

I considered too that we could refactor `/run_existing` end-point to `/upload` + `/run` in the client or something like that. Feel free to open an issue or a discussion because I think it deserves the times the refactorization.

gateway/api/v1/views.py
Tansito: Yep, I agree. It could be simplified.

gateway/api/serializers.py
Tansito: Fixed in [f824b4c](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1286/commits/f824b4ca34fa4d379f6220fc7e8922f6b4c40026)

PR #1287, opened 2024-04-14
Fix env_vars in QiskitPattern

### LINKED ISSUE #1283: Review environment variables functionality

When we `upload` a `QiskitPattern` we allow to pass environment variables: `env_vars`. But we are not sending them to the gateway.

In the same way these `env_vars` should be managed by the `Job` and being encrypted when we add them.

Store `env_vars` encrypted when we upload a `QiskitPattern`. Manage those `env_vars` in the Job when we create it.

akihikokuroda:
### Summary
fix #1283

### DIFF
--- a/client/quantum_serverless/core/job.py
+++ b/client/quantum_serverless/core/job.py
@@ -426,6 +426,7 @@ class GatewayJobClient(BaseJobClient):
                     "entrypoint": program.entrypoint,
                     "arguments": json.dumps(arguments or {}, cls=QiskitObjectsEncoder),
                     "dependencies": json.dumps(program.dependencies or []),
+                    "env_var": json.dumps(program.env_vars or []),
                 }
                 if config:
                     data["config"] = json.dumps(asdict(config))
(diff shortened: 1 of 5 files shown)


client/quantum_serverless/core/job.py
```diff
@@ -490,6 +491,7 @@ def upload(self, program: QiskitPattern):
                             "entrypoint": program.entrypoint,
                             "arguments": json.dumps({}),
                             "dependencies": json.dumps(program.dependencies or []),
+                            "env_vars": json.dumps(program.env_vars or {}),
```
Tansito, line 494: One question here: `env_vars` should be set with the `QiskitPattern` or in the `Job` like arguments? @psschwei @IceKhan13

Tansito:
Regardless my question, Aki. Can we update some of the tests to check that we are storing correctly the `env_vars`?

client/quantum_serverless/core/job.py
akihikokuroda: Here is my 2 cents.  For a configuration value, its default is in the environment variable with the pattern and the argument of the Job can overwrite it at the execution time.  So I think that the pattern is better have the env_vars.

akihikokuroda:
The test is added.

Tansito (approved):
LGTM, thanks Aki!

PR #1288, opened 2024-04-15
remove arguments from program

### LINKED ISSUE #1282: Remove arguments from QiskitPattern

As we discussed internally `QiskitPattern` should not send arguments when `upload`. Only the `run` method that generates the `Job` should do it.

akihikokuroda:
### Summary
Fix #1282


IceKhan13 (approved):
LGTM! Thank you @akihikokuroda !

Tansito:
Btw @akihikokuroda , @IceKhan13 sorry I didn't see this before. How we are ecrypting `arguments` in `Job` has it sense to remove `arguments` from `Job` too? They are in `env_vars`.

akihikokuroda:
@Tansito `arguments` should be given when the program is executed.  They are put into `evn_vars` of `Job` by serializer.  So it seems that the `arguments` field of `Job` is not filled.

Tansito:
I mean, they are being filled here:
https://github.com/akihikokuroda/quantum-serverless/blob/2cf2dfefc361367bcfa643bb93014237a6c87609/gateway/api/serializers.py#L167

But I agree with you, I don't think we are using them because if we access to them we use `get_arguments` that access to `env_vars` so probably we could remove that field from `Job` too (I didn't see it in-depth, just asking).

PR #1291, opened 2024-04-20
Groups and Permissions instances integration

### LINKED ISSUE #1278: Introduce instances to manage access to different resources

Tansito:
### Summary

Close #1278 

This pull request continues the work from #1280 implementing Groups and Permissions to guarantee access to protected Qiskit Functions through `Program.instances`.

### Details and comments

- With a quantum instance like `my-hub/my-group/my-project` three groups are generated: "my-hub", "my-hub/my-group" and "my-hub/my-group/my-project".
- `view_program` permission is assigned automatically to each new program when is created.
- `Program` model has a new permission `run_program` that must be assigned manually to a group to give run access.
- `Program.instances` is a M2M relationship with `Group`

I created a specific `fixture` where you can review the DB structure following this new data. I created three examples about how `View` and `Run` filter queries will work when we implement the end-points with this new architecture: `test_query_to_view_test_user_programs`, `test_query_to_view_test_user_2_programs` and `test_query_to_run_test_user_programs`.


akihikokuroda (approved):
LGTM!  I don't know the API interface of `iqpQcon`.  I assume that the test data are correct.

IceKhan13 (approved):
Let's give it a go :)

PR #1296, opened 2024-04-25
Add check for program execution failures

akihikokuroda:
### Summary
Docker notebook test doesn't catch failures of program executions.  This PR adds some scans in the execution logs to detect program execution failures.

### DIFF
--- a/.github/workflows/docker-build.yaml
+++ b/.github/workflows/docker-build.yaml
@@ -34,7 +34,7 @@ jobs:
             echo "OK"
             exit 0
           else
-            echo "not OK"
+            echo "FAILED: One or more programs failed during execution"
             cat basic.log
             exit 1
           fi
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-build.yaml
```diff
@@ -27,13 +27,43 @@ jobs:
...
           cd tests/basic
-          for f in *.py; do echo "$f" && python "$f"; done
+          for f in *.py; do echo "$f" && python "$f" &>> basic.log; done
+          done=$(cat basic.log | grep -c "DONE")
+          if [[ $done == 4 ]]
+          then
+            echo "OK"
+            exit 0
+          else
+            echo "not OK"
```
psschwei, line 37: nit: could we make the failure message a bit more explicit? 
maybe something like: `FAILED: One or more programs failed during execution`

psschwei (commented):
One nit, otherwise LGTM

psschwei (approved):

PR #1297, opened 2024-04-25
Custom base images for functions

IceKhan13:
### Summary

Allows usage of custom images for function execution

#### Details

User will need to do following steps:
- implement function template file
- build docker image with template file and push
- run function

Function template
```python
# runner.py
class Runner:
    def run(self, arguments: dict) -> dict:
        # this is just an example
        # your function can call for other modules, function, etc.
        return {
            **arguments,
            **{
                "answer": 42
            }
        }
```

Dockerfile
```Dockerfile
FROM icr.io/quantum-public/quantum-serverless-ray-node:0.9.0-py39

# install all necessary dependencies for your custom image

# copy our function implementation in `/runner.py` of the docker image
RUN mkdir /runner
COPY ./runner.py /runner/runner.py
```

Upload funciton with custom image

```python
serverless = ServerlessClient(...)

function_with_custom_image = QiskitFunction(
    title="custom-image-function",
    image="my-custom-function-image:latest"
)

serverless.upload(function_with_custom_image)
```


docs/examples/03_building_custom_image.ipynb
```diff
@@ -0,0 +1,253 @@
+{
```
psschwei, line 1: nit: the rest of the procedural notebooks are in `docs/getting_started/experimental`, whereas the other ones here (VQE / QAOA) are more application-like ... which begs the question of whether we need to think about how to organize our documentation (i.e. should we start making sections for end users, for function providers, etc.)

psschwei: And the pending move to the qiskit org probably provides a good cover for a docs reorg ...

psschwei, line 1: Ideally we should also add this to the e2e tests too

docs/examples/custom_function/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM icr.io/quantum-public/quantum-serverless-ray-node:0.9.0-py39
```
psschwei, line 1: It might be worth creating a template repo in the Qiskit org to store these files so folks can make use of it when writing their own functions. Also would give us a little insight into some of the third party functions out there in the wild

gateway/api/ray.py
```diff
@@ -165,10 +198,11 @@ def submit_job(job: Job) -> Job:
 
 
 def create_ray_cluster(
-    user: Any,
+    job: Job,
+    # user: Any,
     cluster_name: Optional[str] = None,
     cluster_data: Optional[str] = None,
-    job_config: Optional[JobConfig] = None,
+    # job_config: Optional[JobConfig] = None,
```
psschwei, line 205:
```suggestion
```
?

docs/examples/custom_function/Dockerfile
psschwei, line 1: Bump docker image tag to `0.10.1-py310`?

docs/examples/03_building_custom_image.ipynb
```diff
@@ -0,0 +1,253 @@
...
+   "metadata": {},
+   "source": [
+    "As a next step let's define and build our custom docker image.\n",
+    "\n",
+    "Dockerfile will be extending base serverless node image and adding required packages and structure to it. \n",
+    "\n",
+    "In our simple case it will look something like this\n",
+    "\n",
+    "```Dockerfile\n",
+    "FROM icr.io/quantum-public/quantum-serverless-ray-node:0.9.0-py39\n",
```
psschwei, line 51: Bump docker image tag to 0.10.1-py310?

docs/examples/03_building_custom_image.ipynb
IceKhan13: Agree with you :) 
I also was trying to figure out where to put those. I was trying to decide between `examples` or `deployment`

gateway/api/ray.py
```diff
@@ -76,17 +77,48 @@ def submit(self, job: Job) -> Optional[str]:
...
+                )
+                with open(
+                    os.path.join(
+                        working_directory_for_upload, DEFAULT_PROGRAM_ENTRYPOINT
+                    ),
+                    "w",
+                    encoding="utf-8",
+                ) as entrypoint_file:
+                    entrypoint_file.write(default_entrypoint_content)
+            elif bool(program.artifact):
```
Tansito, line 110: Curious about, can not we use `is not None` like in `image` here?

gateway/api/ray.py
```diff
@@ -76,17 +77,48 @@ def submit(self, job: Job) -> Optional[str]:
...
-                    "tmp",
-                    str(uuid.uuid4()),
+
+            # get artifact
+            working_directory_for_upload = os.path.join(
+                sanitize_file_path(str(settings.MEDIA_ROOT)),
+                "tmp",
+                str(uuid.uuid4()),
+            )
+            if program.image is not None:
```
Tansito, line 92: Can you confirm me if I understood this logic, @IceKhan13 ? 🙏 

- We have a template as the entrypoint
- We fill the template with default data
- We upload the template in user's OS as the program

gateway/api/ray.py
```diff
@@ -165,10 +198,10 @@ def submit_job(job: Job) -> Job:
 
 
 def create_ray_cluster(
-    user: Any,
+    job: Job,
+    # user: Any,
```
Tansito, line 202: Can we remove this line now?

gateway/api/serializers.py
```diff
@@ -254,5 +260,6 @@ def update(self, instance, validated_data):
         instance.env_vars = validated_data.get("env_vars", {})
         instance.artifact = validated_data.get("artifact")
         instance.author = validated_data.get("author")
+        instance.image = validated_data.get("image")
```
Tansito, line 263: I see we are only `uploading` by now. Is it because we are not adding the support to run it yet in this PR?

In case we need to add support for this in `RunProgramSerializer` and `RunProgramModelSerializer`.

Tansito (commented):
Thank you so much, @IceKhan13 I left a couple of comments but reviewing my comments + @psschwei comments is almost ready to merge 🙏

gateway/api/ray.py
IceKhan13: `program.artifact` always return `FileInstance` :) so it is never `None`

gateway/api/ray.py
IceKhan13: yes, user implements `Runner` template and we are calling `Runner.run` from this `main.py` entrypoint.

gateway/api/serializers.py
IceKhan13: That is a good point. Should we add run? Or we want to deprecate it overall? I think we are settling on `build` -> `upload` -> `run`. WDYT @Tansito ?

Tansito: Haha I didn't think about it so far. Mmm... I would leave this pull request as it is by now and let's discuss if we remove `run` and open another pull request for it. WDYT?

Tansito (approved):
Regardless @psschwei comments from my part LGTM! 🥳 

Thanks @IceKhan13 !!

docs/examples/03_building_custom_image.ipynb
akihikokuroda: and add to update-component-versions action?

docs/examples/custom_function/Dockerfile
akihikokuroda: and add to update-component-versions action for this, too ?

client/quantum_serverless/core/job.py
```diff
@@ -835,3 +803,101 @@ def _map_status_to_serverless(status: str) -> str:
...
+    Returns:
+        str: uploaded function name
+    """
+    response_data = safe_json_request(
+        request=lambda: requests.post(
+            url=url,
+            data={
+                "title": program.title,
+                "image": program.image,
+                "arguments": json.dumps({}),
```
akihikokuroda, line 828: unrelated question.  Is the `arguments` for the `program` used?

client/quantum_serverless/core/job.py
```diff
@@ -455,54 +455,22 @@ def upload(self, program: QiskitFunction):
...
-                    tar.add(fpath, arcname=filename)
-
-            # check file size
-            size_in_mb = Path(artifact_file_path).stat().st_size / 1024**2
-            if size_in_mb > MAX_ARTIFACT_FILE_SIZE_MB:
-                raise QuantumServerlessException(
-                    f"{artifact_file_path} is {int(size_in_mb)} Mb, "
-                    f"which is greater than {MAX_ARTIFACT_FILE_SIZE_MB} allowed. "
-                    f"Try to reduce size of `working_dir`."
+            elif program.entrypoint is not None:
```
akihikokuroda, line 465: Is it good to put a waring message when both image and entrypoint are set?

docs/examples/custom_function/Dockerfile
IceKhan13: done in https://github.com/Qiskit-Extensions/qiskit-serverless/pull/1297/commits/c3cd804b1dc4fee7d3050dd512683a6d64b15853

client/quantum_serverless/core/job.py
IceKhan13: probably you are right and it is not used anymore. I will look it up and try to fix it in following PRs

client/quantum_serverless/core/job.py
IceKhan13: I think it would be better to move all validation to QiskitFunction itself to reduce boilderplate code. Created an iissue for that. https://github.com/Qiskit-Extensions/qiskit-serverless/issues/1315

Thanks for suggestion!

akihikokuroda (approved):
Great work!  Thanks!

docs/examples/03_building_custom_image.ipynb
IceKhan13: Finally moved it to deployment as rst document.

IceKhan13: Created separate issue for that https://github.com/Qiskit-Extensions/qiskit-serverless/issues/1316

Not sure yet how to make e2e, since it will not work in local mode and in docker compose as well.

PR #1299, opened 2024-04-29
Enable or disable PVC functionality

Tansito:
### Summary

I would like to have the possibility to not create a PVC if there is one provided.

### DIFF
--- a/charts/quantum-serverless/templates/pvcs.yaml
+++ b/charts/quantum-serverless/templates/pvcs.yaml
@@ -1,3 +1,4 @@
+{{ if eq .Values.pvcEnable true }}
 {{ if eq .Values.platform "kind" }}
 apiVersion: v1
 kind: PersistentVolumeClaim
@@ -59,4 +60,5 @@ spec:
   resources:
     requests:
       storage: 1Gi
+{{ end }}
 {{ end }}


charts/quantum-serverless/templates/pvcs.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if eq .Values.pvcEnable true }}
```
IceKhan13, line 1: Should we add this flag to default `values.yaml` file? I'm not sure, but it might fail if none is defined.

Tansito: I totally forgot it, thanks @IceKhan13 🙏

Tansito: Solved it in [c1880a7](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1299/commits/c1880a769bcb69ca32f1f460a66d86b04b19dd70)

charts/quantum-serverless/values.yaml
```diff
@@ -131,9 +131,9 @@ postgresql:
 # PVC
 # ===================
 
+pvcEnable: true
```
psschwei, line 134:
```suggestion
createPvc: true
```

nit: this is more explicit about what this option does

Tansito: Solved in [bf42be0](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1299/commits/bf42be08a05504282e9a0c904d1c0cf698679290)

psschwei (approved):

PR #1304, opened 2024-05-01
Bump sqlparse verson

psschwei:
### Summary

Django uses an older version of sqlparse, and we need to bump to a newer one.


Tansito:
Thanks @psschwei 👍

Tansito (approved):

PR #1306, opened 2024-05-01
List end point implementation

### LINKED ISSUE #1305: Integrate list end-point with the new ACL

As the title says we want to implement the `list` introduced in #1293 with the ACL work introduced in #1291

Tansito:
### Summary

Close #1305 

This PR integrates ACL in the list end-point. So now we are going to be able not only our Qiskit Functions if not other Qiskit Functions that we can have access.

### Details and comments

- Use of `GenericViewSet` instead of `ModelViewSet` in `Programs`
- Added `list` end-point to the `ViewSet`
- Modified the `queryset` to include the new implementation for ACL

### DIFF
--- a/gateway/api/v1/views.py
+++ b/gateway/api/v1/views.py
@@ -8,7 +8,7 @@ from rest_framework.decorators import action
 
 
 from api import views
-from api.models import Program, Job, RuntimeJob
+from api.models import Job, RuntimeJob
 from api.permissions import IsOwner
 from . import serializers as v1_serializers
 
@@ -18,8 +18,8 @@ class ProgramViewSet(views.ProgramViewSet):  # pylint: disable=too-many-ancestor
     Quantum function view set first version. Use ProgramSerializer V1.
     """
 
-    queryset = Program.objects.all()
     serializer_class = v1_serializers.ProgramSerializer
+    pagination_class = None
     permission_classes = [permissions.IsAuthenticated]
 
     @staticmethod
@@ -46,8 +46,12 @@ class ProgramViewSet(views.ProgramViewSet):  # pylint: disable=too-many-ancestor
     def get_model_serializer_run_program(*args, **kwargs):
         return v1_serializers.RunProgramModelSerializer(*args, **kwargs)
 
-    def get_serializer_class(self):
-        return v1_serializers.ProgramSerializer
+    @swagger_auto_schema(
+        operation_description="List author Qiskit Patterns",
+        responses={status.HTTP_200_OK: v1_serializers.ProgramSerializer(many=True)},
+    )
+    def list(self, request):
+        return super().list(request)
 
     @swagger_auto_schema(
         operation_description="Upload a Qiskit Pattern",
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -122,6 +122,7 @@ class ProgramViewSet(viewsets.GenericViewSet):  # pylint: disable=too-many-ances
     def get_queryset(self):
         author = self.request.user
 
+        logger.info("ProgramViewSet get view_program permission")
         view_program_permission = Permission.objects.get(
             codename=VIEW_PROGRAM_PERMISSION
         )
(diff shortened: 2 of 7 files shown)


gateway/api/v1/views.py
```diff
@@ -18,8 +18,8 @@ class ProgramViewSet(views.ProgramViewSet):  # pylint: disable=too-many-ancestor
     Quantum function view set first version. Use ProgramSerializer V1.
     """
 
-    queryset = Program.objects.all()
```
psschwei, line 21: It doesn't appear this variable was used after it was defined.... was that some kind of Django config?

gateway/api/views.py
```diff
@@ -114,16 +116,63 @@ def get_model_serializer_run_program(*args, **kwargs):
     def get_serializer_class(self):
         return self.serializer_class
 
+    def get_object(self):
+        logger.warning("ProgramViewSet.get_object not implemented")
+
     def get_queryset(self):
-        # Allow unauthenticated users to read the swagger documentation
```
psschwei, line 118: will swagger access (whenever we get that actually working) now require authentication?

gateway/api/views.py
```diff
@@ -114,16 +116,63 @@ def get_model_serializer_run_program(*args, **kwargs):
...
+        logger.info("ProgramViewSet get view_program permission")
+        view_program_permission = Permission.objects.get(
+            codename=VIEW_PROGRAM_PERMISSION
         )
 
-    def perform_create(self, serializer):
-        serializer.save(author=self.request.user)
+        # Groups logic
+        user_criteria = Q(user=author)
+        author_criteria = Q(author=author)
```
psschwei, line 132: just doublechecking that user and author criteria are the same

psschwei (commented):
If I'm not mistaken, I don't think there's a way for us to test permissions e2e yet (i.e. there's nothing outside of user uploads for folks to list), so would like to get at least one more set of eyes on it before approving.

gateway/api/v1/views.py
IceKhan13: yes, it is a default django viewset attribute that specified what needs to be returned. It can be overwritten by `get_queryset` which is exactly doing later in this PR :)

Tansito: yep, exactly that!

gateway/api/views.py
```diff
@@ -114,16 +116,63 @@ def get_model_serializer_run_program(*args, **kwargs):
...
+        view_permission_criteria = Q(permissions=view_program_permission)
+        author_groups_with_view_permissions = Group.objects.filter(
+            user_criteria & view_permission_criteria
+        )
+        author_groups_with_view_permissions_count = Group.objects.filter(
```
IceKhan13, line 137: looks like we can save one query by using
```python
author_groups_with_view_permissions_count = author_groups_with_view_permissions.count()
```

gateway/api/views.py
Tansito: this is something I'm not sure, I was testing this locally and for me was working fine so I removed it but if you can double check perfect

(spoiler: you can enter in swagger but to run it correctly you need to add Authorization header authenticating in swager, what I think it has sense from my point of view)

gateway/api/views.py
Tansito: They are not exactly exactly the same in the sense that:

```
Program:
  author
  
Group:
  user
```

They are different fields in database so you need different criterias. If we had `user` instead of `author` we could reuse it.

gateway/api/views.py
Tansito: Good point, I think you are right! Let me double check and change it!

gateway/api/views.py
IceKhan13: it looks like @Tansito is:
- first, getting list of groups that user belongs to and program as permissions from
- then he selects programs that has this groups or user is the author of the program here https://github.com/Qiskit-Extensions/quantum-serverless/pull/1306/files#diff-2025150bf47484f2582f9816af400b1336b7ca6adcbb4faa61bb2d3330df856dR151

Am I right, @Tansito ?

IceKhan13 (approved):
one query save can be done if following PRs :) This one is looking good!

gateway/api/views.py
Tansito: Exactly, maybe reordering the blocks and put author criteria in programs logic is clearer than now.

gateway/api/views.py
Tansito: Done in [6ce8d69](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1306/commits/6ce8d69d39d6d8b95f714585253f123173bdcb65)

gateway/api/views.py
Tansito: Hope this change helps to clarify it:

[6ce8d69](https://github.com/Qiskit-Extensions/quantum-serverless/pull/1306/commits/6ce8d69d39d6d8b95f714585253f123173bdcb65)

psschwei (approved):

PR #1311, opened 2024-05-03
Fix proxy Dockerfile

Tansito:
### Summary

From #1232 it seems there were some errors with the Dockerfile not detected

### DIFF
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -4,7 +4,7 @@ FROM python:3.10.12-slim-buster
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
-RUN apt-get -y update && apt-get install  --no-install-recommends -y iptables=1.8.2-4 sudo=1.8.27-1+deb10u6 &&\
+RUN apt-get -y update && apt-get install  --no-install-recommends -y iptables=1.8.2-4 &&\
     apt-get clean &&\
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


akihikokuroda (commented):
`iptables=1.8.2-4` can be taken out, too.

akihikokuroda:
You can do like

`RUN pip install --upgrade pip==24.0 --no-cache-dir`

proxy/Dockerfile
```diff
@@ -4,24 +4,18 @@ FROM python:3.10.12-slim-buster
...
 WORKDIR /home/worker
 
-COPY --chown=worker:worker proxy/requirements.txt .
-RUN pip install -r requirements.txt --no-cache-dir
-
 ENV PATH="/home/worker/.local/bin:${PATH}"
 
+COPY --chown=worker:worker proxy/requirements.txt .
+# hadolint ignore=DL3013
+RUN pip install --upgrade pip --no-cache-dir &&\
```
Tansito, line 16: @akihikokuroda @psschwei I would like to know your opinion here. Do you think is a good idea ignore the version when upgrading pip? Instead of:

```py
RUN pip install --upgrade pip==24.0 --no-cache-dir
```

Thinking in that we are going to need to review this version from time to time.

Tansito: we just posted at the same time, @akihikokuroda 😂

psschwei: my two cents: if the old version of pip isn't causing problems, I wouldn't worry about upgrading it (there was a case a while back where version resolution through an old pip was very slow, but I don't think that's impacting the default pip in python 3.10...)

Tansito: Yep, and less in this case where dependencies are very small. I like your idea @psschwei , in case someone sees a problem let us know!

akihikokuroda (approved):
LGTM. Thanks!

PR #1313, opened 2024-05-06
Get end point

### LINKED ISSUE #1314: Creates get end-point with the new ACL

As the title says we want to implement the get introduced in https://github.com/Qiskit-Extensions/qiskit-serverless/pull/1293 with the ACL work introduced in https://github.com/Qiskit-Extensions/qiskit-serverless/pull/1291

Tansito:
### Summary

Close #1314 

This PR introduces `get` end-point to retrieve a Qiskit Function.

### Details and comments

- `retrieve` end-point implemented in Gateway
- remove `retrieve_by_title` method from serializers to use the new `get_object`
  - `upload` method continues only updating owners programs
- add support for the new end-point in client
- fixed a problem with the `list` end-point due to the new ACL logic

### DIFF
--- a/client/quantum_serverless/core/client.py
+++ b/client/quantum_serverless/core/client.py
@@ -357,6 +357,10 @@ class BaseClient(JsonSerializable):
     def list(self, **kwargs) -> List[QiskitFunction]:
         """Returns list of available programs."""
         raise NotImplementedError
+    
+    def get(self, title) -> QiskitFunction:
+        """Returns a specific program by title."""
+        raise NotImplementedError
 
 
 class BaseProvider(BaseClient):
@@ -480,9 +484,13 @@ class ServerlessClient(BaseClient):
     def file_upload(self, file: str):
         return self._files_client.upload(file)
 
-    def list(self, **kwargs) -> List[QiskitFunction]:
+    def list(self) -> List[QiskitFunction]:
         """Returns list of available programs."""
-        return self._job_client.get_programs(**kwargs)
+        return self._job_client.get_programs()
+    
+    def get(self, title) -> QiskitFunction:
+        """Returns a specific program by title."""
+        return self._job_client.get_program(title)
 
     def _verify_token(self, token: str):
         """Verify token."""
--- a/client/quantum_serverless/core/client.py
+++ b/client/quantum_serverless/core/client.py
@@ -354,10 +354,10 @@ class BaseClient(JsonSerializable):
         )
         return self.list(**kwargs)
 
-    def list(self, **kwargs) -> List[QiskitFunction]:
+    def list(self) -> List[QiskitFunction]:
         """Returns list of available programs."""
         raise NotImplementedError
-    
+
     def get(self, title) -> QiskitFunction:
         """Returns a specific program by title."""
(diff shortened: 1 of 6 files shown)


Tansito:
I discovered that how we are using `kwargs` now with the new `function.run` method we need to adapt the logic to support config and other params. Current logic like:

```py
job = function.run(arguments={ "test": "here it goes an argument" }, config=config)
```

It's not working. I'm going to create an issue about this (cc: @IceKhan13)

akihikokuroda:
The `title` in the db is not unique.  Should `retrieve` return more than one program?

Tansito:
> The title in the db is not unique. Should retrieve return more than one program?

It's not unique between all the programs but it's unique by author because if it exists it updates the program. The `get` it should always return one if not it can use `list`.

Tansito:
But you remembered me one thing @akihikokuroda , in the `upload` method I'm allowing that a user updates a Qiskit Function from another user if it has access. We should only allow `uploads` to Qiskit Functions created by the own owner.

akihikokuroda:
> But you remembered me one thing @akihikokuroda , in the upload method I'm allowing that a user updates a Qiskit Function from another user if it has access. We should only allow uploads to Qiskit Functions created by the own owner.

It may not make sense that the deployer and the consumer must be the same user to retrieve the function.  Am I misunderstanding someting?

Tansito:
> It may not make sense that the deployer and the consumer must be the same user to retrieve the function. Am I misunderstanding someting?

I don't know if I'm following you here, Aki. You can be consumer and deployer at the same time but in that case you can only have one Qiskit Function with the same title, due to the logic that we have:

https://github.com/Qiskit-Extensions/qiskit-serverless/blob/4c36828ebbbb578fe05a84bcc84961e305d201da/gateway/api/views.py#L246

It can happen that a Qiskit Function uploaded by other user and shared with you has the same name as one of yours. In that case, you can always change the title of your Qiskit Function.

Another situation is, you have a Qiskit Function as a deployer and you want to change it, run it as a consumer but not deploy it. We are not supporting that (at least right now). If you want to test a Qiskit Function you are going to need to create a different one from the Qiskit Function that you sharing basically.

akihikokuroda:
@Tansito I understand now. Would you add one or two more `program` instances in the fixtures so that it can test `list` and `retrieve` a little more?  Thanks!

Tansito:
I just added a test for it with my last change:

https://github.com/Qiskit-Extensions/qiskit-serverless/pull/1313/files#diff-4fc3a54d578fc5ec0af95b876b2c687cff200ff69fea692c1da2342877f480b0R101

Are you thinking in another use-case, @akihikokuroda ? Let me know if I'm forgetting something and I can add it without problem, thank you for the review! 🙏

akihikokuroda:
I'm thinking like
1. list returns more than 1 instance
2. deployer and author has instances with the same title and retrieve it.
Thanks!

Tansito:
From the tests that @akihikokuroda and I were preparing we saw that some logic seems a bit clunky for a good UX so we are reviewing how to solve it and we will prepare a solution for it.

PR #1318, opened 2024-05-07
Cancel

### LINKED ISSUE #1317: Cancel runtime jobs when our job is stopped

Middleware `Job.stop()` method stops pending backend runtime jobs in addition to the middleware job.

akihikokuroda:
### Summary
fix #1317 

### Details and comments

The optional `service` argument is added to the `job.stop()` method.  When the `service` is provided, the `stop` method stops the pending backend runtime jobs in addition to the middleware job.

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -30,10 +30,7 @@ from rest_framework.decorators import action
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
 
-from quantum_serverless.serializers.program_serializers import (
-    QiskitObjectsDecoder,
-)
-from qiskit_ibm_runtime import RuntimeInvalidStateError
+from qiskit_ibm_runtime import RuntimeInvalidStateError, QiskitRuntimeService 
 from utils import sanitize_file_path
 
 from .models import VIEW_PROGRAM_PERMISSION, Program, Job, RuntimeJob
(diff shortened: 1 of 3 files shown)


gateway/api/views.py
```diff
@@ -29,6 +29,11 @@
 from rest_framework.decorators import action
 from rest_framework.generics import get_object_or_404
 from rest_framework.response import Response
+
+from quantum_serverless.serializers.program_serializers import (
```
IceKhan13, line 33: I think we can remove this and just use `QiskitRuntimeService(**request.data.get("service"))`

gateway/requirements.txt
```diff
@@ -18,4 +18,5 @@ drf-yasg>=1.21.7
 cryptography>=41.0.1
 # Django dependency, but we need a newer version (IBMQ#246)
 sqlparse>=0.5.0
-
+qiskit_ibm_runtime
+quantum_serverless
```
IceKhan13, line 22: we do not need that with change above

gateway/requirements.txt
```diff
@@ -18,4 +18,5 @@ drf-yasg>=1.21.7
 cryptography>=41.0.1
 # Django dependency, but we need a newer version (IBMQ#246)
 sqlparse>=0.5.0
-
+qiskit_ibm_runtime
```
IceKhan13, line 21: let's put some version here :)

IceKhan13 (commented):
@akihikokuroda do you want to have it as intermediate step before figuring out solution without service passing?

Tansito:
> @akihikokuroda do you want to have it as intermediate step before figuring out solution without service passing?

I would agree as a solution by now at least.

gateway/requirements.txt
Tansito: Don't we continue needing `quantum_serverless` for `QiskitObjectsDecoder`? I think right now I would prefer to copy paste this decoder logic due to the problems that we can have between the rename from `quantum_serverless` and `qiskit_serverless`, WDTY?

akihikokuroda:
So far, I couldn't get the `super user` way.  The `token` or `service` is required to cancel the job.  The user has the `service` anyway so I added as the optional argument for `stop` method.

Tansito (approved):
From my part LGTM, @akihikokuroda . Thank you!

akihikokuroda:
> @akihikokuroda do you want to have it as intermediate step before figuring out solution without service passing?
@IceKhan13  I'll open an issue for the service less solution.

IceKhan13 (approved):
@akihikokuroda thank you!

PR #1320, opened 2024-05-08
Migrate run end-point and use function.run instead of client.run

### LINKED ISSUE #1321: /tests/basic and /tests/experimental are not updated
Execute CI workflow `Test QS on Kubernetes / tests`

It doesn't fail
Manual executions of the docs/getting_started/basic/*.ipynb fails

It should fail when the manual executions of the docs/getting_started/basic/*.ipynb fails

Tansito:
### Summary

Close #1321, #1323 

This PR covers two issues:
- Promote `run_existing` to `run` as standard way to run Qiskit Functions
- Migrated all the tests to the new interface taking advantage of the refactorization

### Details and comments

- Gateway:
  - Removed old `run` references and promoted `run_existing` to `run` end-point.
- Client:
  - Deprecate old `run_existing` and updated `run` methods
  - Fixed `list:get_programs` method
  - Updated `LocalClient` to support `list` method
- Tests: updated all the tests with the new interfaces


client/qiskit_serverless/core/job.py
```diff
@@ -306,66 +323,38 @@ def upload(self, program: QiskitFunction):
...
-        results = re.search("\nSaved Result:(.+?):End Saved Result\n", output)
-        result = ""
-        if results:
-            result = results.group(1)
-
-        job = Job(job_id=str(uuid4()), job_client=self)
-        entry = {"status": status, "logs": output, "result": result, "job": job}
-        self._jobs[job.job_id] = entry
-        return job
+        warnings.warn(
```
akihikokuroda, line 344: This method should not have been used by the user directly.  Can we just take out?

client/qiskit_serverless/core/job.py
```diff
@@ -480,39 +441,13 @@ def run_existing(
...
-                    json=data,
-                    headers={"Authorization": f"Bearer {self._token}"},
-                    timeout=REQUESTS_TIMEOUT,
-                )
-            )
-            job_id = response_data.get("id")
-            span.set_attribute("job.id", job_id)
-
-        return Job(job_id, job_client=self)
+        warnings.warn(
```
akihikokuroda, line 444: d should not have been used by the user directly. Can we just take out?

IceKhan13 (approved):
Ah, so much better! 
Now we have one UX flow to upload and run. Simplify handling of request and infrastructure a lot!

Thank you @Tansito !

client/qiskit_serverless/core/job.py
Tansito: Take out the warning do you mean, @akihikokuroda ?

akihikokuroda: I wonder if we can take out the method.

psschwei:
This is less about the changes and more just a nit on the PR text (which will become the commit message): can we also capture that we're moving from `client.run`  to `function.run` in there? Will be helpful later when searching changes locally.

Tansito:
How about this @psschwei ?

client/qiskit_serverless/core/job.py
Tansito: I'm not sure, let me check 👍

psschwei:
that works

psschwei:
LGTM pending resolution of Aki's comment

client/qiskit_serverless/core/job.py
Tansito: Is this what you were expenting, @akihikokuroda ? [2363e98](https://github.com/Qiskit/qiskit-serverless/pull/1320/commits/2363e98f01d00969e3bae867582c3ef9dda7373e)

akihikokuroda: Yes. Thanks!

akihikokuroda (approved):

PR #1324, opened 2024-05-09
Client get method

Tansito:
### Summary

As part of #1314 I implemented a `get` method only in the client as sugar syntax to get a Qiskit Function and avoid to iterate the list.

### DIFF
--- a/client/qiskit_serverless/core/client.py
+++ b/client/qiskit_serverless/core/client.py
@@ -358,6 +358,10 @@ class BaseClient(JsonSerializable):
         """Returns list of available programs."""
         raise NotImplementedError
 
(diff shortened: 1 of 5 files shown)


Tansito:
@akihikokuroda , @IceKhan13 , @psschwei , what do you think? Do you prefer to maintain the `list` until finish the `get` or this could work as workaround in the meantime?

akihikokuroda:
I don't have strong opinion but I prefer to maintain it for now.

psschwei:
If the old functionality still works, I'd just wait for the new function to be completely ready rather than adding a workaround

Tansito:
Ok, let's just leave as it is, then. Just in case we need to change the interface or something 👍

PR #1328, opened 2024-05-10
Fix some security concerns

akihikokuroda:
### Summary
Fix some security concerns.

### DIFF
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -572,7 +572,7 @@ class Job:
         """
         return self._job_client.filtered_logs(job_id=self.job_id, **kwargs)
 
-    def result(self, wait=True, cadence=5, verbose=False):
+    def result(self, wait=True, cadence=5, verbose=False, maxwait=0):
         """Return results of the job.
         Args:
             wait: flag denoting whether to wait for the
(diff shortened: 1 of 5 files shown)


client/qiskit_serverless/core/job.py
```diff
@@ -572,7 +572,7 @@ def filtered_logs(self, **kwargs) -> str:
         """
         return self._job_client.filtered_logs(job_id=self.job_id, **kwargs)
 
-    def result(self, wait=True, cadence=5, verbose=False):
+    def result(self, wait=True, cadence=5, verbose=False, maxwait=0):
```
Tansito, line 575: has it sense to just set `wait=False` by default and that's all? I think the final output is similar and it has more sense for me.

Tansito (approved):
Regardless my comment, I think it has sense a lot of sense these changes @akihikokuroda . Thank you!

client/qiskit_serverless/core/job.py
akihikokuroda: The scan suggested that loops needs static termination condition.  So I added not to wait forever.

Tansito:
I'm going to merge this to continue with the release @akihikokuroda , thank you!

PR #257, opened 2023-03-08
Ray head node proxy rework

akihikokuroda:
### Summary
Rework the proxy sidecar of the ray head node.

### Details and comments

This rework replaces the auth2-proxy with the gatekeeper for the proxy sidecar.
This change enables `Client Credentials Grant` to access to head node.
The auth2-proxy doesn't support `Client Credentials Grant`.

The request to the head node is redirected to the login screen when the request doesn't have the access token.  The request that has the access token is routed to the head node function.

This is an example script to get the access token from the Keycloak and send a request with the access token to the head node. 
```
#!/bin/bash
API=$1
export RESPONSE=$(curl --request POST \
  --url 'http:/KEYCLOAK-IP/realms/quantumserverless/protocol/openid-connect/token' \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data grant_type=password \
  --data client_id=rayclient \
  --data audience=rayclient \
  --data client_secret=CLIENTSECRET-CHANGEME \
  --data username=user \
  --data password=passw0rd)
						   
export A=$(echo $RESPONSE | jq .access_token)
echo $A
export TOKEN=${A//'"'/}

curl --request GET \
--header "authorization: Bearer $TOKEN" \
--header 'content-type: application/json' \
--url "http://HEADNODE-IP/$API"
```

PR #1331, opened 2024-05-12
Function uniqueness

Tansito:
### Summary

From #1313 it was a conversation around function uniqueness. This PR pretends to propose a solution for it with the next proposal:
- ACL _Program_ continues being the same: _program.instances_ are what provides external access to programs through VIEW and RUN permissions.
- _Namespace_ is the new entity that it manages proprietary Qiskit Functions:
  - **name**: string that will identify the namespace for the Qiskit Function: "ibm/transpiler"
  - **registry**: a namespace will have a registry associated so every image in that namespace will be uploaded there
  - **admin_group**: it's the group that will guarantee management over the namespace, if the user has this group it has the possibility to create and upload Qiskit Functions inside this namespace

### Details and comments

- New `Namespace` model
- `Program` now can has (or not) a `namespace`
- Updated the serializer to be able to receive the `namespace.name`
- Added the logic in the `update` end-point to manage access and update/create the Qiskit Function
- Tests updated


gateway/api/models.py
```diff
@@ -58,6 +58,27 @@ def __str__(self):
         return f"{self.id}"
 
 
+class Namespace(models.Model):
```
psschwei, line 61: Namespace has a [specific meaning](https://cloud.ibm.com/docs/Registry?topic=Registry-registry_overview#overview_elements_namespace) in the context of container registries, so we probably don't want to use that term here to mean something different.  

As I understand it, what we're trying to define here is more or less the author of a particular function, with links to the registry, namespace, and admin group. So something like `Provider` or `Author` might better represent the purpose of the object.

Tansito: Thanks for the feedback @psschwei ! `Author` is something that we are just using and I think it doesn't fit really well here. Probably `Provider` is a better option, I'm going to take a look and change it today 😄

Tansito: Solved in [0277a28](https://github.com/Qiskit/qiskit-serverless/pull/1331/commits/0277a280fa9a5e3c296a0c1bd37167aa7a8bc62a)

tests/basic/03_dependencies.py
```diff
@@ -6,7 +6,8 @@
     title="pattern-with-dependencies",
     entrypoint="pattern_with_dependencies.py",
     working_dir="./source_files/",
-    dependencies=["qiskit-experiments==0.6.0"],
```
Tansito, line 9: It seems `qiskit-experiments` is not working with the newer Qiskit version so I changed it by now for another thing to test dependencies installation. I accept other proposals.

gateway/api/serializers.py
```diff
@@ -24,23 +31,56 @@ class UploadProgramSerializer(serializers.ModelSerializer):
...
+
+    def check_provider_access(self, provider_name, author):
+        """
+        This method check if the author has access to the provider
+        """
+        provider = Provider.objects.filter(name=provider_name).first()
+        if provider is None:
+            logger.error("Provider [%s] does not exist.", provider_name)
+            return False
+        logger.error(
```
psschwei, line 56: is there a missing check here? I think this will log an error every time this function is run...

gateway/api/views.py
```diff
@@ -212,8 +212,33 @@ def upload(self, request):
...
+                if not user_has_access:
+                    # For security we just return a 404 not a 401
+                    return Response(
+                        {"message": f"Provider [{provider_name}] was not found."},
+                        status=status.HTTP_404_NOT_FOUND,
+                    )
+                program = serializer.retrieve_provider_function(
+                    title=title, provider_name=provider_name
+                )
+            else:
```
psschwei, line 237: is the logic for this else the same as for the `if provider_name is None` block on L217? if so, we should combine them

gateway/tests/api/test_v1_program.py
```diff
@@ -9,7 +9,7 @@
 from rest_framework import status
 from rest_framework.test import APITestCase
 
-from api.models import Job
+from api.models import Job, Program
```
psschwei, line 12: assuming since the gateway components aren't user-facing we don't need to rename this to Function :crossed_fingers:

gateway/tests/fixtures/fixtures.json
```diff
@@ -121,7 +125,7 @@
       "pk": 100,
       "fields": {
         "name": "runner",
-	"permissions": [61]
+	      "permissions": [61]
```
psschwei, line 128: is this indent needed?

gateway/api/serializers.py
Tansito: Good catch 🙏 , solved in: [edb541c](https://github.com/Qiskit/qiskit-serverless/pull/1331/commits/edb541c03fa80f2062da29da509590945c20f967)

gateway/tests/api/test_v1_program.py
Tansito: Correct, internally I'm not renaming models and so on. What I'm changing are errors messages or responses returned to the user. If you see any reference that I should change let me know 😄

gateway/tests/fixtures/fixtures.json
Tansito: Mmm... This is weird. In my IDE I see it like this:

Screenshot 2024-05-23 at 9 49 20 AM

I don't know why it appears like that in GitHub 🤔

gateway/api/views.py
Tansito: It's not exactly the same. Maybe it's a bit difficult to see it the idea is the next one:

- this `if` from L217 is to get the `provider_name`, it checks if the user sends a provider, if not we check if the provider is in the title
- `if` from L224 is for when you have definitely the `provider_name`

They are not the same but I think I can move first logic to make it less confuse.

Tansito: WDYT @psschwei ? [04a1c23](https://github.com/Qiskit/qiskit-serverless/pull/1331/commits/04a1c23d55ca6ad99b54e43c413c0b20c2ecb61b)

gateway/api/models.py
```diff
@@ -58,6 +58,27 @@ def __str__(self):
         return f"{self.id}"
 
 
+class Provider(models.Model):
```
akihikokuroda, line 61: How is the instance of the `Provider` created?  Is the new viewset necessary for the `Provider`?

Tansito: Good question. No, initially it's not going to be public. We are thinking to use the panel administrator or some kind of script in the first steps meanwhile we don't close the process to publish and manage a 3rd party function.

tests/basic/03_dependencies.py
akihikokuroda: I believe that the `docs/getting_started/basic/03_dependencies.ipynb` also has to be updated.

gateway/api/models.py
akihikokuroda: I see.  I believe that as an open source project, we need to provide some ways to fill the `provider` database.

Tansito: Mmm... A mock provider sounds really good. I agree. Can we add it with internal issue #257? I think it has more sense and we don't add more changes on this one. WDTY?

akihikokuroda: I'm  OK with a separate issue.  I would like an issue in this repo.  Thanks!

tests/basic/03_dependencies.py
akihikokuroda: It's better to revisit this test/doc.  Would you open an issue for that?

Tansito: Fixed here: [92cf627](https://github.com/Qiskit/qiskit-serverless/pull/1331/commits/92cf627a5a931cbe6b541f35354d1a8ec60ff617)

Yep, I will open an issue now to review it 👍

akihikokuroda (approved):
LGTM. Thanks!

tests/basic/03_dependencies.py
Tansito: Issue opened #1339

gateway/api/views.py
psschwei: yep, that works

Tansito:
I'm going to wait until tomorrow in case @psschwei and @IceKhan13 can double check. I started with #1340 in the meantime.

gateway/api/serializers.py
```diff
@@ -24,23 +32,67 @@ class UploadProgramSerializer(serializers.ModelSerializer):
...
+        This method returns a Program entry searching by the title and author, if not None
+        """
+        return Program.objects.filter(title=title, author=author).first()
+
+    def retrieve_provider_function(self, title, provider_name):
+        """
+        This method returns a Program entry searching by the title and provider, if not None
+        """
+        provider = Provider.objects.filter(name=provider_name).first()
+        return Program.objects.filter(title=title, provider=provider).first()
```
IceKhan13: I think you can do 
```python
Program.objects.filter(title=title, provider__name=provider_name).first()
```
and save one query to DB :)

PR #1338, opened 2024-05-17
Ubi gateway

akihikokuroda:
### Summary
Change the base image of the gateway pod to the UBI base image.

### Details and comments
I couldn't find the python 3.10 UBI base image.  There are only 3.9 and 3.11 UBI images.  This PR used the python 3.11 UBI image.  The Ray version is also changed to 2.22.0 for python 3.11 support.

There are issues in the decorator example with the 3.10 client.  The client has to be 3.11 to use the decorator.

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -14,7 +14,8 @@ ENV PYTHONUNBUFFERED 1
 USER 0
 COPY gateway/requirements.txt .
 RUN pip install -r requirements.txt --no-cache-dir
-RUN ln -s /usr/local/lib64/python3.11/site-packages/symengine /usr/local/lib/python3.11/site-packages/symengine
+RUN cp -r -n /usr/local/lib64/python3.11/site-packages/symengine /usr/local/lib/python3.11/site-packages
+RUN cp -r -n /usr/local/lib/python3.11/site-packages/symengine /usr/local/lib64/python3.11/site-packages
 COPY gateway .
 RUN chown -R 1000:100 /usr/src/app &&\
     mkdir /usr/src/app/media && chown 1000:100 /usr/src/app/media
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -1,4 +1,5 @@
-FROM registry.access.redhat.com/ubi9/python-311
+FROM registry.access.redhat.com/ubi9-minimal
+RUN microdnf install -y python3 python3-pip
 
 WORKDIR /usr/src/app
 
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -1,9 +1,9 @@
 FROM registry.access.redhat.com/ubi9-minimal:9.4-1134
 RUN microdnf install -y python3.11-3.11.7-1.el9_4.1 python3.11-pip-22.3.1-5.el9 python3.11-devel-3.11.7-1.el9_4.1
-RUN ln -s /usr/bin/python3.11 /usr/local/bin/python3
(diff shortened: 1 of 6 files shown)


psschwei (commented):
LGTM, though since this requires bumping the Ray version I think we should hold off on merging it until after the next release as per https://github.com/Qiskit/qiskit-serverless/pull/1272#issuecomment-2135297911

Tansito:
I'm going to restore the `on-hold`. How we are configuring ray here too let's merge first #1272 and #1343 and return to this 🫡

Tansito:
PR merged, @akihikokuroda . Feel free to update this whenever you can!

psschwei:
We may also want to re-re-add the changes from #554 so that we're testing against 3.11 too

akihikokuroda:
@psschwei @Tansito @IceKhan13 The **Test QS on Kubernetes / tests** fails because of run out of spaces.  The github action has limited resources and the test doesn't fit.  I wonder if we can disable the test.  WDYT?

Tansito:
> @psschwei @Tansito @IceKhan13 The Test QS on Kubernetes / tests fails because of run out of spaces. The github action has limited resources and the test doesn't fit. I wonder if we can disable the test. WDYT?

The test was useful in the past discovering issues in k8s. Maybe we can move this test to our internal pipeline, WDYT?

psschwei:
> Maybe we can move this test to our internal pipeline, WDYT?

That makes sense to me.

@akihikokuroda out of curiosity, how much bigger are the images using the UBI base than what we had before?

psschwei:
Another thought... could we use our staging cluster for the test?

akihikokuroda:
@psschwei The UBI image is big  

registry.access.redhat.com/ubi9/python-311 :latest - 1.1 GiB
python:3.10.12-slim-buster - 121.8 MiB
python:3.11-slim - 164.5 MiB 

I'll try to add python to a small UBI image instead of pre-made UBI image with python.

Tansito:
> registry.access.redhat.com/ubi9/python-311 :latest - 1.1 GiB

😮

akihikokuroda:
I changed the base image to registry.access.redhat.com/ubi9-minimal.  Its size is 25.2 MiB.  I added python3.11 and necessary python modules.

gateway/requirements.txt
```diff
@@ -20,3 +20,4 @@ cryptography>=41.0.7
 # Django dependency, but we need a newer version (IBMQ#246)
 sqlparse>=0.5.0
 qiskit_ibm_runtime>=0.22.0
+tzdata>=2024.1
```
Tansito, line 23: For what is this dependency, @akihikokuroda ?

Tansito (approved):

gateway/Dockerfile
```diff
@@ -1,20 +1,21 @@
...
 
 WORKDIR /usr/src/app
 
 # set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1
 ENV PYTHONUNBUFFERED 1
 
-# install psycopg2 dependencies
-RUN apt-get -y update &&\
-    apt-get install --no-install-recommends -y gcc=4:8.3.0-1 python3-dev=3.7.3-1 vim=2:8.1.0875-5+deb10u6 &&\
```
psschwei, line 11: Is it possible to keep vi/vim in this image? (we've used it for live debugging in the past)
or is there an editor in the UBI already?

psschwei (approved):

Issue #1336, opened 2024-05-16
docs/examples/02_qaoa doesn't work with "USE_RUNTIME_SERVICE = True"


akihikokuroda:
### Steps to reproduce the problem
Execute `docs/examples/02_qaoa` with `QiskitRuntimeService` for an account and `USE_RUNTIME_SERVICE = True`

### What is the current behavior?
> cloud simulators and Lab are now retired

So I tried to run the examples/02_qaoa with backend = "ibm_sherbrooke"  
I got
> IBMInputValueError(\nqiskit_ibm_runtime.exceptions.IBMInputValueError: \'The instruction u3 on qubits (0,) is not supported by the target system. Circuits that do not match the target hardware definition are no longer supported after March 4, 2024.

and
>See the transpilation documentation (https://docs.quantum.ibm.com/transpile) for instructions to transform circuits and the primitive examples (https://docs.quantum.ibm.com/run/primitives-examples) to see this coupled with operator transformations.

### What is the expected behavior?
Job is executed successfully on the quantum system.

akihikokuroda:
I added the transpiler step like this in qaoa.py: 
```
    if service is not None:
        # if we have service we need to open a session and create sampler
        service = arguments.get("service")
        backend = service.least_busy(operational=True, simulator=False, min_num_qubits=127)
        session = Session(service=service, backend=backend)
        options = Options()
        options.optimization_level = 3
        pm = generate_preset_pass_manager(optimization_level=1, backend=backend)
        ansatz = pm.run(ansatz)
```
then it's getting:

[...]

PR #1341, opened 2024-05-23
Fix QAOA example

akihikokuroda:
### Summary
fox #1336 

### Details and comments
**This example uses the real quantum system. It consumes a lot of usage time.**


Tansito (approved):
LGTM @akihikokuroda , thank you!

From what I saw the error in the pipeline is due to the fix that I added in #1331

IceKhan13 (approved):

Issue #1337, opened 2024-05-16
Migrate to RayJobs instead of managing RayClusters


psschwei:
### What is the expected enhancement?

From the Kuberay docs: "With RayJob, KubeRay automatically creates a RayCluster and submits a job when the cluster is ready. You can also configure RayJob to automatically delete the RayCluster once the Ray job finishes."

We currently spin up a RayCluster when a user submits a job and have a decent amount of code dedicated to managing said cluster. Now that RayJob has reached v1, we should consider migrating to use RayJobs, which will offload the cluster management piece, hopefully simplifying our code base.

More on RayJobs: https://docs.ray.io/en/master/cluster/kubernetes/getting-started/rayjob-quick-start.html

ElePT:
Moving away from ray.

PR #1343, opened 2024-05-28
Update kuberay operator, ray version

psschwei:


psschwei:
Holding until after next release and #1272 merges

Tansito:
Can we update Dockerfile too @psschwei ?

https://github.com/Qiskit/qiskit-serverless/blob/main/Dockerfile-ray-node

charts/qiskit-serverless/values.yaml
```diff
@@ -109,8 +109,8 @@ gateway:
 kuberayOperatorEnable: true
 kuberay-operator:
   image:
-    repository: kuberay/operator
-    tag: v1.0.0
+    repository: quay.io/kuberay/operator
```
Tansito, line 112: Curious about this change @psschwei , do we need to change the `Chart.yaml` too?

```
- name: kuberay-operator
```

What are we doing with this change exactly? (Sorry, I tried to find it in the documentation but I didn't find any reference).

psschwei: Ray now publishes their container images to quay, not docker hub, so we need to update the image name:
https://github.com/ray-project/kuberay/releases/tag/v1.1.0

Tansito: Awesome, thanks @psschwei . I also was taking a look into this link that they were mentioning in that release:

https://docs.ray.io/en/master/cluster/kubernetes/k8s-ecosystem/kueue.html#kuberay-kueue

Really interesting and probably related with your issue #1337

Tansito (approved):
Thank you so much, @psschwei LGTM

PR #1346, opened 2024-05-30
migrate vqe example to primitiveV2

akihikokuroda:
### Summary
V1 primitives have been deprecated and support will be dropped on July 15. Any code that uses Sampler and Estimator should switch to using SamplerV2 and EstimatorV2.

### Details and comments
This PR has fix for examples/01_vqe


docs/examples/01_vqe.ipynb
```diff
@@ -144,12 +144,12 @@
...
-    "service = None\n",
-    "if USE_RUNTIME_SERVICE:\n",
-    "    service = QiskitRuntimeService()\n",
-    "    backend = \"ibmq_qasm_simulator\"\n",
+    "service = QiskitRuntimeService(\n",
+    "    channel='ibm_quantum',\n",
+    "    instance='ibm-q/open/main',\n",
+    "    token='<Your Token>',\n",
+    "    verify=False\n",
+    ")\n",
```
psschwei, line 152: Is it not possible to run this example locally anymore?

akihikokuroda: It may not be simple differences to make it work locally.  I'll look into it in a separate issue.

psschwei: I don't think it's a big deal, but we probably should make a note about that early in the doc (since most of the rest of the examples we have can run locally)

psschwei (approved):

docs/examples/01_vqe.ipynb
akihikokuroda: The Aer simulator path is added for local execution.

Tansito (approved):
LGTM but I agree with @psschwei comment, it would be nice if we can add a note about run it in the cloud and how to obtain the token.

akihikokuroda:
> LGTM but I agree with @psschwei comment, it would be nice if we can add a note about run it in the cloud and how to obtain the token.

I added the Aer simulator path so now this works without the `service`.

PR #1348, opened 2024-05-31
Add helm option to disable ray tls

psschwei:
### Summary

Adds an option to disable TLS for Ray in the Helm chart, currently defaulting to enabled / true.

### Details and comments

When using Istio, we can make use of Istio's mTLS rather than needing to enable TLS explicitly in Ray. In that case, we don't need to do the various init config, either through cert manager or the creation of TLS certs.


akihikokuroda:
The proxy is using the certificate created in the init container in the Gateway pod.  The qiskit client sends the https requests to the backend.  The proxy intercepts the https requests by using the certificate.

psschwei:
> The proxy is using the certificate created in the init container in the Gateway pod. The qiskit client sends the https requests to the backend. The proxy intercepts the https requests by using the certificate.

Didn't know that... in that case, it's probably just simpler to allow setting the Ray TLS option (I think it's hardcoded right now)

Tansito:
from @akihikokuroda comment do you want to change something @psschwei or should I review this?

psschwei:
I made the change already, it's ready for review again

Tansito (approved):
LGTM thank you, Paul 👍

PR #1352, opened 2024-06-03
Issue 1314 | Client: get function by name

### LINKED ISSUE #1314: Creates get end-point with the new ACL

IceKhan13:
### Summary

Get function by name using client

```python

client = ServerlessClient(...)

my_function = client.get("my-function")
# <QiskitFunction "my-function">

job = my_function.run()

```

### Details and comments

List programs got feature to filter by `title` field. So, `get` client method is reusing this endpoint. It already had all access policies, so it make code changes pretty neat. And also it avoids problems with unique migrations, etc.

Closes #1314

### DIFF
--- a/client/qiskit_serverless/core/client.py
+++ b/client/qiskit_serverless/core/client.py
@@ -358,6 +358,10 @@ class BaseClient(JsonSerializable):
         """Returns list of available programs."""
         raise NotImplementedError
 
+    def get(self, title: str) -> Optional[QiskitFunction]:
+        """Returns qiskit function based on title provided."""
+        raise NotImplementedError
+
 
 class BaseProvider(BaseClient):
     """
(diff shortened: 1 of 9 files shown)


akihikokuroda (approved):
LGTM Thanks!

Tansito (approved):
LGTM, I didn't update the issue with the changes in #1331 so my fault here. An improvement that we can do in the next version (I'm fine with this approach by now) is to update the interface to guarantee the uniqueness of the function using the new provider interface:

```py
my_function = client.get("my-function")
# this get will look for this title in your private functions (author query)

my_function = client.get("provider/my-function")
# or
my_function = client.get(title="my-function", provider="provider")
# this get will look for this title in the provider
```

This way we can guarantee the uniqueness of the get method.

PR #1353, opened 2024-06-04
Client README.md file update

akihikokuroda:
### Summary
Update the out dated qiskit_serverless client README file

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -73,8 +73,8 @@ from qiskit_serverless import ServerlessClient, QiskitFunction
 from qiskit.circuit.random import random_circuit
 
 client = ServerlessClient(
-    token="awesome_token", 
-    host="http://localhost:59994",
+    token="<TOKEN>", 
+    host="<GATEWAY_ADDRESS>",
 )
 
 # create function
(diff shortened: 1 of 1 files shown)


client/README.md
```diff
@@ -21,105 +21,100 @@ Full docs can be found at https://qiskit-extensions.github.io/qiskit-serverless/
...
-       arguments={
-           "circuits": circuits
-       }
-   )
+from qiskit_serverless import ServerlessClient, QiskitFunction
+from qiskit.circuit.random import random_circuit
+
+client = ServerlessClient(
+    token="awesome_token", 
+    host="http://localhost:59994",
```
Tansito, line 77: Is it port `599994` the default one, @akihikokuroda ? Just to confirm 👍

Tansito (approved):
LGTM, thank you @akihikokuroda . I just looked for a double check about the port 👍

psschwei:
One thought - should we remove the examples from this doc and just link to the ones in the `docs/` folder? That way we only have one place we need to update when we do our quarterly refactoring...

client/README.md
akihikokuroda: oops.  I was going to change them before checkin.  I'll fix it.

akihikokuroda:
@psschwei That seems a good idea.   Or we may want to put a short migration guide here with the samples when we change the api.

PR #1357, opened 2024-06-06
add resources to gateway and scheduler containers

akihikokuroda:
### Summary
This PR adds `resources` in gateway and scheduler containers.

The `cpu` `limits` for gateway affects the startup time of the gateway pod.  The `readiness` probe configuration is changed to work with the  `cpu` `limits` value in this PR.

### DIFF
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -53,6 +53,7 @@ jobs:
             --set gateway.application.ray.cpu=1 \
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
+          sleep 120
       - name: setup python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -53,7 +53,8 @@ jobs:
             --set gateway.application.ray.cpu=1 \
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
-          sleep 120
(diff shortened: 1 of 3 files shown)


.github/workflows/kubernetes-deploy.yaml
```diff
@@ -53,6 +53,7 @@ jobs:
             --set gateway.application.ray.cpu=1 \
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
+          sleep 120
```
psschwei, line 56: instead of a sleep, could we do a readiness check instead? something like : 

`kubectl wait  --for=condition=Ready "$SOMEPOD" --timeout 5m`

akihikokuroda: It's better.

psschwei (approved):

charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -145,17 +145,21 @@ ingress:
...
+  limits:
+    cpu: "1500m"
+    memory: "1Gi"
+  requests:
+    cpu: "500m"
+    memory: "700Mi"
+scheduler:
+  resources:
+    limits:
+      cpu: "500m"
```
Tansito, line 158: From what I was seeing in the metrics it seems `scheduler` is using more `cpu` (like 3 cores), `memory` seems fine.

akihikokuroda: OK.  It is working in my system but I'll update to 1500m.

Tansito (approved):
LGTM, thank you Aki!

PR #1358, opened 2024-06-06
Client: Runtime service wrapper to track runtime jobs

IceKhan13:
### Summary

Runtime service wrapper `ServerlessRuntimeService` to track runtime jobs.
Add sessions closing to cancel job.

### Details and comments

Alternatively to proxy approach we can use runtime service client wrapper to get all requests for runtime serive and associate serverless jobs to runtime jobs and sessions.

```python
# function implementation

from qiskit_serverless import ServerlessRuntimeService

service = ServerlessRuntimeService(channel="...", token="...") # same as QiskitRuntimeService
...
backend = service.least_busy(operational=True, simulator=False)
...
with Session(service=service, backend=backend) as session:
    estimator = Estimator(session=session)
    ...
    job = estimator.run([(isa_circuit, isa_observables, theta_mid)])

```

```python
client = ServerlessClient(...)

function = client.get("my-function")

job = function.run()

job.stop(service = QiskitRuntimeService(...)) # this will cancel serverless job and all runtime jobs and sessions accociated
```


gateway/api/views.py
```diff
@@ -429,6 +429,11 @@ def stop(self, request, pk=None):  # pylint: disable=invalid-name,unused-argumen
                             except RuntimeInvalidStateError:
                                 logger.warning("cancel failed")
 
+                            if jobinstance.session_id:
```
akihikokuroda, line 432: Does it cancel the session for each job?

IceKhan13: yes, it is might be redundant, but it is ok. There might be a case when 2 or more sessions created per serverless job, so it is better to safely cancel them all.

akihikokuroda: OK. Does it return an error not an exception?

IceKhan13: it does not return anything. Down inside runtime client it returns `requests.Response`, but it is not propagated up

akihikokuroda: Does the cancel session cancel the jobs in the session?  If so, does it cause "cancel failed" for each job except one in the session?

client/qiskit_serverless/utils/runtime_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+        inputs: Dict,
+        options: Optional[Union[RuntimeOptions, Dict]] = None,
+        callback: Optional[Callable] = None,
+        result_decoder: Optional[
+            Union[Type[ResultDecoder], Sequence[Type[ResultDecoder]]]
+        ] = None,
+        session_id: Optional[str] = None,
+        start_session: Optional[bool] = False,
+    ) -> Union[RuntimeJob, RuntimeJobV2]:
+        runtime_job = super().run(
```
akihikokuroda, line 108: Is this the only method that requests runtine job?

gateway/api/views.py
IceKhan13: cancel session cancels all jobs in a session

IceKhan13: no, it does not cause cancel failed

client/qiskit_serverless/utils/runtime_service_client.py
IceKhan13: it looks like

akihikokuroda (approved):
Looks good.  It's much better solution than the proxy.  Thanks!

akihikokuroda:
We probably need 2 separate issues:

1. take out proxy
2. update documents for "ServerlessRuntimeService"

IceKhan13:
> It's much better solution than the proxy.

I'm not sure about that :) That is why I want to keep proxy as well. We will see what will work for us in the future. 

> update documents for "ServerlessRuntimeService"

Good point. We need to add it somewhere

psschwei (commented):
Just thinking out loud here (so haven't really worked through too many details), but would be possible to put this functionality into `ServerlessClient` itself? That way folks wouldn't have an opportunity to do things the "wrong" way when writing their functions...

IceKhan13:
I think what we should do for sure is put `QiskitRuntimeService` as constructor for `ServerlessClient`, so when we call `stop` we do not need to pass service separately. 
Something like:
```python
client = ServerlessClient(service=QiskitRuntimeService(...))

function = client.get("func")

job = function.run()

job.stop() # without service
```

and inside of function 
```python
from qiskit_serverless import get_service

service = get_service() # it is already wrapper
# ServerlessRuntimeService

backend = service.get_backend(...)
sampler = Sampler(backend=backend)

sampler.run()
```

Tansito:
> It's much better solution than the proxy.

Same opinion as @IceKhan13 . This idea came out from a discussion that we were having about the problematic of the current implementation of the proxy that it needs to setup the `verify=false`.

How user needs to setup something manually it doesn't matter if its a `verify=false` or `ServerlessRuntimeService`. The counterpart continues being the same: the user requires specific setup to user serverless. So, is it fine we for some jobs we don't related runtime with them?

For me `proxy` continues being interesting to explore but not ready to deploy yet. As we commented yesterday, @akihikokuroda .

Tansito:
Do you mind @IceKhan13 if we wait to merge it to confirm with Sanket this approach? I prefer to clarify all the constrains that this solution means with him.

psschwei:
> put `QiskitRuntimeService` as constructor for `ServerlessClient`

I think that makes sense.

IceKhan13:
@Tansito I do not mind :) btw, this solution does not break anything as it is decoupled from any internal logic. It's just optional client update.

Tansito:
Sanket agrees with the suggestion to include in the documentation a little suggestion to simplify providers implementations:

```py
from qiskit_serverless import ServerlessRuntimeService as QiskitRuntimeService
```

This way they should not notice too much difference.

gateway/api/models.py
```diff
@@ -218,3 +218,4 @@ class RuntimeJob(models.Model):
     runtime_job = models.CharField(
         primary_key=True, max_length=100, blank=False, null=False
     )
+    session_id = models.CharField(max_length=100, blank=True, null=True, default=None)
```
Tansito, line 221: I would like to maintain consistency in the names here `runtime_job`, `runtime_session` (I would agree that probably would better to have `job_id`, `session_id` but I don't think we would need to do a migration due to this).

client/qiskit_serverless/utils/runtime_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+    token = os.environ.get(ENV_JOB_GATEWAY_TOKEN)
+    if token is None:
+        logging.warning("Runtime job will not be associated with serverless job.")
+        return False
+
+    url = (
+        f"{os.environ.get(ENV_JOB_GATEWAY_HOST)}/"
+        f"api/{version}/jobs/{os.environ.get(ENV_JOB_ID_GATEWAY)}/add_runtimejob/"
+    )
+    response = requests.post(
```
Tansito, line 74: Are we not using `safe_json_request` here for any reason like [here](https://github.com/Qiskit/qiskit-serverless/blob/28b27e4c5f26544f5fc751fc6ff5f32f32432421/client/qiskit_serverless/core/job.py#L386)?

(Thinking in maybe not need to sanitize the output).

client/qiskit_serverless/utils/runtime_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+        url,
+        data={"runtime_job": runtime_job_id, "session_id": session_id},
```
Tansito, line 76: Is it possible to use `json` here to run a standard json call? (Maybe not)

client/qiskit_serverless/utils/runtime_service_client.py
```diff
@@ -0,0 +1,120 @@
...
+        headers={"Authorization": f"Bearer {token}"},
+        timeout=REQUESTS_TIMEOUT,
+    )
+    if not response.ok:
+        sanitized = response.text.replace("\n", "").replace("\r", "")
```
Tansito, line 81: Are we returning a weird response? I would like to annotate it to create an issue and fix it if so.

IceKhan13: that is totally weird, I do not know where did I get this :) fixed :)

client/qiskit_serverless/utils/runtime_service_client.py
IceKhan13: not needed, since requests doing it under the hood

client/qiskit_serverless/utils/runtime_service_client.py
IceKhan13: in this function we do not fail and do not need to raise any exceptions that is why vanilla requests is fine. If something goes wrong we need to warn in logs and that is it.

gateway/api/models.py
IceKhan13: ✅

client/qiskit_serverless/utils/runtime_service_client.py
Tansito: Oh, I see. Like a best effort. Got it!

client/qiskit_serverless/utils/runtime_service_client.py
Tansito: Correct me if I'm wrong here @IceKhan13 , in https://github.com/Qiskit/qiskit-serverless/pull/1285 we needed to change `data` by `json` because `requests` was not using `content-type` JSON by default:

https://requests.readthedocs.io/en/latest/user/quickstart/#more-complicated-post-requests

IceKhan13: OMG, that what you meant :) 

I though you wanted to wrap payload in `json.dumps` :) 

added in https://github.com/Qiskit/qiskit-serverless/pull/1358/commits/e96c3470bf1c431fc83b2bfbc736eebe522e09de

IceKhan13:
oh, wtf with kube tests now 🫤 https://github.com/Qiskit/qiskit-serverless/actions/runs/9485422545/job/26137478335?pr=1358

client/qiskit_serverless/utils/runtime_service_client.py
Tansito: well, you know, always the best one explaining itself, sorry 😅

Tansito:
> oh, wtf with kube tests now 🫤 https://github.com/Qiskit/qiskit-serverless/actions/runs/9485422545/job/26137478335?pr=1358

Maybe a random error? Let's wait after lunch and restart it to see if it continues 🤷‍♂️

Tansito (approved):
Thank you, Iskandar. Awesome work 🙏

Tansito:
@akihikokuroda one question, I assume #1360 should fix the problem from these actions, shouldn't it?

akihikokuroda:
@Tansito Yes, it should.

Tansito:
Let's update this with `main` again @IceKhan13 , I think you didn't get this last PR in your last update. Let's see if that fixes the problem 🤷‍♂️

IceKhan13:
let's see :)

psschwei:
If not #1373 should get you over the line

PR #1359, opened 2024-06-07
Client: get method with provider arguments

IceKhan13:
### Summary

Follow up to https://github.com/Qiskit/qiskit-serverless/pull/1352

Adding provider argument to `get` method.

```python
client = ...

function = client.get("my-func")
function = client.get("default/my-func")
function = client.get(title="my-func", provider="default")
```

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -112,10 +112,12 @@ class ProgramViewSet(viewsets.GenericViewSet):  # pylint: disable=too-many-ances
     def get_queryset(self):
         author = self.request.user
         title = self.request.query_params.get("title")
-        provider = self.request.query_params.get("provider")
+        provider_name = self.request.query_params.get("provider")
 
         serializer = self.get_serializer_upload_program(data=self.request.data)
-        provider, title = serializer.get_provider_name_and_title(provider, title)
+        provider_name, title = serializer.get_provider_name_and_title(
+            provider_name, title
+        )
 
         logger.info("ProgramViewSet get view_program permission")
         view_program_permission = Permission.objects.get(
@@ -144,8 +146,8 @@ class ProgramViewSet(viewsets.GenericViewSet):  # pylint: disable=too-many-ances
         )
         if title:
             title_criteria = Q(title=title)
-            if provider:
-                title_criteria = Q(title=title, provider__name=provider)
+            if provider_name:
+                title_criteria = Q(title=title, provider__name=provider_name)
             author_programs = Program.objects.filter(
                 (author_criteria | author_groups_with_view_permissions_criteria)
                 & title_criteria
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -333,6 +333,59 @@ class ProgramViewSet(viewsets.GenericViewSet):  # pylint: disable=too-many-ances
 
         return Response(job_serializer.data)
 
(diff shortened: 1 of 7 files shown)


akihikokuroda (approved):
LGTM. Thanks!

Tansito (commented):
How now the function is unique my approach would be to create a new GET method that will return one function instead to return the list with only one function and in the client iterate over the array. I think it has sense and is going to be useful for future developments.

IceKhan13:
But only title is never unique

IceKhan13:
so, if we only do `api/programs/get?title="my-function"` then we return `Program` named `my-function` where author is the user of request? 

and if we `api/programs/get?title="my-function"&provider="blah"` then we return by function name and provider?

Tansito:
> But only title is never unique. What do we do when user just requested my-function ?

If you request `my-function` from our current logic it will happen the next:
- There is no provider specified so we are going to look for it by the `author` that is doing the request
- `author` + `title` is unique, due to our logic in `upload`. If that title exists we update that function for that user, we don't create a new one

If you would like to get `my-function` from another user is something that we don't support by now and yes, if we would need to support that use-case definitely we would need to take a look to our entire logic.

Correct me if I'm wrong, Iskandar.

Tansito:
> so, if we only do api/programs/get?title="my-function" then we return Program named my-function where author is the user of request?

> and if we api/programs/get?title="my-function"&provider="blah" then we return by function name and provider?

Here I have more confidence about your Django knowledge than mine. I imagine that the end-point will need to look something like: `api/programs/get/my-function?provider="blah"` because Django is going to force us to do something like: `get/{title}`. I'm open to discuss if we should leave free that end-point for the id instead and we want to create another one like: `api/programs/get_by_title/my-function?provider="blah"`

IceKhan13:
CI is still failing due to https://github.com/Qiskit/qiskit-serverless/issues/1369

gateway/api/views.py
```diff
@@ -325,6 +297,68 @@ def run(self, request):
 
         return Response(job_serializer.data)
 
+    @action(methods=["GET"], detail=False, url_path="get_by_title/(?P<title>[^/.]+)")
+    def get_by_title(self, request, title):
```
Tansito, line 301: Sorry Iskandar, this is new for me. What are we doing with that `re`?

IceKhan13: in order to pass parameter `title` in a form of url `v1/api/programs/get_by_title/<title>` we need to add ths `url_path` parameter.

IceKhan13: it is basically defining path by some id

IceKhan13: if this id is not `primary key` and title is not a primary key

Tansito: ok, it's not like we are limitating the usage of spaces or something like that, just to verify that we are not changing this "Docker-Image-Program" due to that.

IceKhan13: yes, for spaces one just need to encode it to url str

On that note. We probably need to sanitize titles and do not allow spaces and special chars

Tansito: > On that note. We probably need to sanitize titles and do not allow spaces and special chars

Yes, totally agree on that.

Tansito (approved):
Thank you, Iskandar 🙏

PR #1360, opened 2024-06-07
Fix proxy container lint issue

akihikokuroda:
### Summary
Fix proxy container lint issue

### Details and comments
The ubi images path has 3 slashes.

### DIFF
--- a/.github/scripts/dockerfile-security.rego
+++ b/.github/scripts/dockerfile-security.rego
@@ -25,7 +25,7 @@ deny[msg] {
 deny[msg] {
     input[i].Cmd == "from"
     val := split(input[i].Value[0], "/")
-    count(val) > 2
+    count(val) > 3
     msg = sprintf("Line %d: use a trusted base image", [i])
 }
 
--- a/proxy/Dockerfile
+++ b/proxy/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/python-311
+FROM registry.access.redhat.com/ubi9/python-311@sha256:fae422b4e7dbc2e551919fb87b6004c06b731179be2ed2904427531a2d1ef3a2
 
 # set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1


.github/scripts/dockerfile-security.rego
```diff
@@ -25,7 +25,7 @@ deny[msg] {
 deny[msg] {
     input[i].Cmd == "from"
     val := split(input[i].Value[0], "/")
-    count(val) > 2
+    count(val) > 3
```
Tansito, line 28: Why are we changing this here @akihikokuroda ?

proxy/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9/python-311
+FROM registry.access.redhat.com/ubi9/python-311@sha256:fae422b4e7dbc2e551919fb87b6004c06b731179be2ed2904427531a2d1ef3a2
```
Tansito, line 1: do we want to start using `sha` from now on?

akihikokuroda: I got docker file lint error llike this: `-:1 DL3006 warning: Always tag the version of an image explicitly`.  So I put the `sha`.

.github/scripts/dockerfile-security.rego
akihikokuroda: It's a strange check.  The UBI image path is like this: `registry.access.redhat.com/ubi9/python-311` This path should be trusted but the lint fails.

psschwei: Yeah, I'm not sure how exactly the number of slashes corresponds to trustworthiness but que sera sera...

proxy/Dockerfile
psschwei: Do they also tag images or do we have to use the sha here? 
(I don't have a strong preference, just haven't worked with the UBI registry much before)

.github/scripts/dockerfile-security.rego
Tansito: Haha understood 👍

proxy/Dockerfile
Tansito: yeah, me neither, that's why I'm asking, I'm totally fine with this approach if it has sense with what we are using

Tansito (approved):
LGTM, Aki 👍

proxy/Dockerfile
akihikokuroda: @psschwei They have tag like this `1-62.1717085982`.  It doesn't look user friendly, either.

psschwei (approved):

PR #1365, opened 2024-06-12
Group qiskit serverless images in image registry

psschwei:
### Summary

This moves all the qiskit-serverless images under a top-level group, which will make image management easier in the future (as per infra team request). For example:

`qiskit-serverless-gateway:latest` --> `qiskit-serverless/gateway:latest`

### DIFF
--- a/.github/workflows/docker-verify.yaml
+++ b/.github/workflows/docker-verify.yaml
@@ -55,22 +55,22 @@ jobs:
       matrix:
         image:
           [
-            "qiskit-serverless-gateway:latest",
-            "qiskit-serverless-proxy:latest",
-            "qiskit-serverless-ray-node:latest-py39",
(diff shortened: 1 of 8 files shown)


Tansito (commented):
It's an interesting PR because it can be a good breaking change and we need to update the documentation too. How do you want manage this, @psschwei ? I see there are some files that are not being updated either so my proposal would be:

1.- Update hooks: `docker-verify`, `release-pull-request`, `update-component-versions`
2.- Open an issue related with this change to update documentation and last references from this `pull-request` before creating the release.

WDYT?

Tansito (approved):

psschwei:
Just to note - the new paths are already available in the registry

Tansito:
Thank you so much, Paul. Next week I will create the PR for the release and we can review too if there is any reference somewhere that we could forget.

PR #1366, opened 2024-06-12
Bump k8s version in ghactions

psschwei:
### Summary

We're running on v1.28 now, so should test on that version too


IceKhan13 (approved):
Let's merge as temp solution until resolution of the upsteam issue

akihikokuroda (approved):

PR #1370, opened 2024-06-12
Client: always cleanup artifact tar

IceKhan13:
### Summary

Fixes situation when something is happening during function upload and artifact tar remain in working directory


akihikokuroda (approved):
LGTM

Tansito (commented):
One doubt @IceKhan13, with this approach we could raise the exception without remove the tar file, couldn't we? Are you maybe missing a `finally` statement? Like:

```py
try:
  # Logic here
except Exception as error:
  raise QiskitServerlessException from error
finally:
    if os.path.exists(artifact_file_path):
        os.remove(artifact_file_path)
```

(Take it as pseudo code, I don't know 100% if this example is correct in python 😂)

IceKhan13:
you are absolutely right! haha! forgot to add `finally` What I was thinking?!

Tansito (approved):
Thank you, Iskandar!

PR #1371, opened 2024-06-12
Pin filelock in tox

psschwei:
### Summary

Version 3.15 seems to have broke our CI, so let's try sticking with an earlier version


IceKhan13 (approved):

psschwei:
The bad release of filelock has been pulled, so this fix is not needed

Issue #1270, opened 2024-03-28
Remove extra end-points for api/v1/


Tansito:
### Steps to reproduce the problem

In Swagger we can see that we have several number of end-points available to the user due to the use of `Viewset`. Taking into account that our API it's not a typical CRUD I propose to remove the use of `Viewset` for `GenericAPIView ` and specify manually those end-points that we are going to use.

Tansito:
Once I finish with #1240 I'm going to start taking a look into this

Tansito:
This issue can be closed from the last pull requests

PR #1377, opened 2024-06-17
Filter job end-points

Tansito:
### Summary

Part of #1270 , this PR pretends to clean-up some end-points from `Jobs` to only those ones that we are currently using.

### Details

End-Points removed:
- POST /jobs/: it's not possible to create jobs directly from the end-point
- PUT /jobs/{id}: I think we are not updating jobs from PUT call
- PATCH /jobs/{id}: I think we are not updating jobs from PATCH call
- DELETE /jobs/{id}: we are not removing jobs using DELETE call

### DIFF
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -54,7 +54,7 @@ jobs:
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
           GATEWAY=$(kubectl get pod -l app.kubernetes.io/name=gateway -o name)
-          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 7m
+          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 9m
       - name: setup python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -54,7 +54,7 @@ jobs:
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
           GATEWAY=$(kubectl get pod -l app.kubernetes.io/name=gateway -o name)
-          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 9m
+          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 5m
       - name: setup python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/kubernetes-deploy.yaml
+++ b/.github/workflows/kubernetes-deploy.yaml
@@ -54,7 +54,7 @@ jobs:
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
           GATEWAY=$(kubectl get pod -l app.kubernetes.io/name=gateway -o name)
-          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 5m
+          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 7m
       - name: setup python
         uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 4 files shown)


gateway/api/v1/views.py
```diff
@@ -69,13 +70,37 @@ class JobViewSet(views.JobViewSet):  # pylint: disable=too-many-ancestors
...
+
+    @swagger_auto_schema(
+        operation_description="Save the result of a job",
+        responses={status.HTTP_200_OK: v1_serializers.JobSerializer(many=False)},
+    )
+    @action(methods=["POST"], detail=True)
+    def result(self, request, pk=None):
+        return super().result(request, pk)
+
+    ### We are not returning serializers in the rest of the end-points
```
Tansito, line 102: We have a problem here because in the rest of the end-points we are not using serializers to format the response. I will open an issue for this, fix it would do the PR so big 😅

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
           GATEWAY=$(kubectl get pod -l app.kubernetes.io/name=gateway -o name)
-          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 5m
+          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 7m
```
Tansito, line 57: I increased a little bit the numbers because it was generating some false negatives the test.

Tansito:
@akihikokuroda @IceKhan13 let me know if we are using any of the end-points that I removed in this PR. All tests are passing, I was trying to look through the code and running some tests that I have locally just in case too but I could forget something perfectly.

gateway/api/v1/views.py
```diff
@@ -69,13 +70,37 @@ class JobViewSet(views.JobViewSet):  # pylint: disable=too-many-ancestors
     Job view set first version. Use JobSerializer V1.
     """
 
-    queryset = Job.objects.all()
     serializer_class = v1_serializers.JobSerializer
+    pagination_class = LimitOffsetPagination
```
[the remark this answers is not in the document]
IceKhan13: Just in case it is already a default for everything :) 

https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/main/settings.py#L244

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -54,7 +54,7 @@ jobs:
             --set gateway.application.limits.keepClusterOnComplete=false \
             .
           GATEWAY=$(kubectl get pod -l app.kubernetes.io/name=gateway -o name)
-          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 7m
+          kubectl wait --for=condition=Ready "$GATEWAY" --timeout 9m
```
psschwei, line 57: this shouldn't be taking that long... we should look into why the pods aren't coming ready

Tansito: Mmm... how can we test it? I mean, it's not something that I found in this PR but in others too

psschwei: actually, seems the gateway is crashing:
```
gateway-554bf468d7-8qnlb            0/1     CrashLoopBackOff   5 (57s ago)   5m4s
```

Tansito: Weird, locally I was not noticing that it was crashing. I will take a look into it 👍

Tansito: definitely in local doesn't crash to me 😅 

do you have `kind` locally @psschwei ?

psschwei: I just checked -- the gateway pod also restarted for me locally (same issue with probes failing), but after the first restart things righted themselves (eventually, it took about 90s for the pod to become ready)

psschwei (approved):

akihikokuroda (approved):
LGTM. Thanks!

PR #1386, opened 2024-06-24
Client: fix for Function post init

IceKhan13:
### Summary

Fix for Function post init.

### Details and comments

With current implementation `list` and `widget` method calls failed if user had function named `bla-bla/my-function` before introduction of `provider` field.


akihikokuroda (approved):
LGTM

Tansito (approved):
LGTM just a little change that I would do but I don't know how Python works in this case. Maybe it's better to run the second check first:
```py
self.provider is not None and self.provider != title_split[0]
```
It seems Python can manage it correctly anyway

PR #1388, opened 2024-06-24
Fail job on insufficient resources

### LINKED ISSUE #1379: Requesting more resources that you can have

When you request more resources that the cluster has available for you, you remain waiting for those resources. The problem is that if you ask for more resources that you can assign to your node you remain waiting for them indefinitely.

Create a parallel task with more cpus that you can assign to your node using the limites setup in the the `values.yaml`. The function remains in RUNNING state indefinitely.

The job should be marked as FAILED with the error.

psschwei:
### Summary

Fixes #1379 
Ray will wait indefinitely if a job has insufficient resources, so we need to explicitly fail jobs that don't have enough resources to run.

### Details and comments

When a user uses `distribute_task` within their functions, they may request more resources than they have available, which will result in something like this in the logs:

```
2024-06-24 14:23:25,045	INFO worker.py:1405 -- Using address 10.244.0.16:6379 set in the environment variable RAY_ADDRESS
2024-06-24 14:23:25,045	INFO worker.py:1540 -- Connecting to existing Ray cluster at address: 10.244.0.16:6379...
2024-06-24 14:23:25,056	INFO worker.py:1715 -- Connected to Ray cluster. View the dashboard at 10.244.0.16:8265 
(autoscaler +2s) Tip: use `ray status` to view detailed cluster status. To disable these messages, set RAY_SCHEDULER_EVENTS=0.
(autoscaler +2s) Error: No available node types can fulfill resource request {'CPU': 4.0, 'memory': 1.0}. Add suitable node types to this cluster to resolve this issue.
```

In this PR we look for the `No available node types can fulfill resource request` line in the logs and then fail the job if we find it.

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
@@ -74,6 +74,8 @@ spec:
             httpGet:
               path: /liveness
               port: http
+            initialDelaySeconds: 60
+            periodSeconds: 20
           readinessProbe:
             httpGet:
               path: /readiness
--- a/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
@@ -117,6 +117,10 @@ spec:
               value: {{ .Values.application.ray.minReplicas | quote }}
             - name: RAY_CLUSTER_WORKER_MAX_REPLICAS
               value: {{ .Values.application.ray.maxReplicas | quote }}
+            - name: LIMITS_CPU_PER_TASK
+              value: {{ .Values.application.ray.cpu | quote }}
+            - name: LIMITS_MEMORY_PER_TASK
+              value: {{ .Values.application.ray.memory | quote }}
 {{- if .Values.application.superuser.enable }}
             - name: DJANGO_SUPERUSER_USERNAME
               valueFrom:
--- a/gateway/api/management/commands/update_jobs_statuses.py
+++ b/gateway/api/management/commands/update_jobs_statuses.py
@@ -7,7 +7,7 @@ from django.core.management.base import BaseCommand
 
 from api.models import Job
 from api.ray import get_job_handler
-from api.schedule import check_job_timeout, handle_job_status_not_available
+from api.schedule import check_job_timeout, handle_job_status_not_available, fail_job_insufficient_resources
 from api.utils import ray_job_status_to_model_job_status, check_logs
 
 logger = logging.getLogger("commands")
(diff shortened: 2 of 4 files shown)


charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -74,6 +74,8 @@ spec:
             httpGet:
               path: /liveness
               port: http
+            initialDelaySeconds: 60
+            periodSeconds: 20
```
psschwei, line 78: Even if we don't go with this PR, we should add these configs to ensure we aren't failing the gateway prematurely...

Tansito:
Just to double check, @psschwei . Is this ready for review?

psschwei:
> Just to double check, @psschwei . Is this ready for review?

yes it is

Tansito:
I didn't forget about this @psschwei , I will try to take a look along this week, sorry 🙏

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
Tansito: I see, maybe this is the reason cause the gateway restart some times when we deploy it, could be?

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -57,6 +62,15 @@ def handle(self, *args, **options):
                 if job_handler:
                     logs = job_handler.logs(job.ray_job_id)
                     job.logs = check_logs(logs, job)
+                # check if job is resource constrained
```
Tansito, line 65: One question here @psschwei that it was not clear for me. I see that we are using `logs` , do we wanted to have this logic inside the `if`? Right now it's outside so probably this `logs` will be always None.

psschwei: I don't know if we need to move it inside the if, but the check should probably be against `job.logs` instead of the temp variable, i.e. `if no_resources_log in job.logs:` wdyt?

Tansito: Reviewing the code I think it has sense to move it under `job_handler`. Basically that's the point where we know that we have the cluster running. Before it I don't think it has sense to check it, probably. And yes, I agree. We should check `job.logs` 💯

gateway/main/settings.py
```diff
@@ -306,6 +306,8 @@
 # resources limitations
 LIMITS_JOBS_PER_USER = int(os.environ.get("LIMITS_JOBS_PER_USER", "2"))
 LIMITS_MAX_CLUSTERS = int(os.environ.get("LIMITS_MAX_CLUSTERS", "6"))
+LIMITS_CPU_PER_TASK = int(os.environ.get("LIMITS_CPU_PER_TASK", "4"))
+LIMITS_MEMORY_PER_TASK = int(os.environ.get("LIMITS_MEMORY_PER_TASK", "8"))
```
akihikokuroda, line 310: How do these variable work?

psschwei: cpus / ram per task are set in the helm config. I added both as envvars on the gateway pod, and this code here just reads those values in. that way we can tell the user what the limits are if they request too many

Tansito: just a friendly ping here @akihikokuroda if you can take a look 🙏

akihikokuroda (approved):

PR #1391, opened 2024-06-26
Remove old deprecated actions

Tansito:
### Summary

In this PR I remove some deprecated actions that are going to fail at the end of the month or are going to be deprecated.


psschwei (approved):
needs a rebase

PR #1394, opened 2024-06-27
Remove pre-delete hook

Tansito:
### Summary

I would like to propose to remove this hook. It fails and gets stuck the uninstall process of a release.


akihikokuroda (approved):
I'm OK removing this hook.  I'm thinking about adding the trouble shooting section in the doc.  WDYT?

Tansito:
> I'm thinking about adding the trouble shooting section in the doc. WDYT?

I think it has sense. I know @pandasa123 is working in the doc so we can talk with him about how he wants to manage it.

Tansito:
Btw, I was commenting with @psschwei that one of the things that I would like fix with the PR is that maybe we prefer not to remove clusters when we do an uninstall either.

psschwei (approved):

PR #1397, opened 2024-07-02
Remove workers from ray-cluster

Tansito:
### Summary

Due to some problems in the current configuration that we have in the cluster we cannot support workers right now. After some tests I could confirm that we can support the same features that we are supporting right now increasing the size of the head node.

So, this PR modifies the ray-cluster template removing the workers from the configuration.


Tansito:
@akihikokuroda reviewing the configuration of the cluster I had a question related with the values:

- useCertManager
- useTLS

Would has sense to only do `useCertManager` if we enable `useTLS`? Right now we are always adding the certs independently of that.

akihikokuroda:
> Would has sense to only do useCertManager if we enable useTLS? Right now we are always adding the certs independently of that.

It means that it doesn't create the self-signed certificates.  I think that's OK.  If the cluster doesn't have the cert manager, the TLS may not be needed.

Tansito:
And with Istio I suppose that they are not needed either, aren't they?

akihikokuroda:
@Tansito That's right.

psschwei (approved):
I would save updates to the TLS / certmanager options for another PR.

Tansito:
> I would save updates to the TLS / certmanager options for another PR.

Sure!

Issue #1400, opened 2024-07-08
Separate function logs from the other logs


akihikokuroda:
### What is the expected enhancement?
Functions may be provided by different parties from the party who executes the function.  The log from the function may be better to be separated from the part of log.

akihikokuroda:
Here is an idea to implement this.

The log entry from the function must be tagged (marked) like `[Function: <function name>] contents [End Function]` by putting the tag manually or by the function provided in the Serverless SDK.

Then

The scheduler filters the log entries and put into the new `function_log` item  of the `Job` and put the new gateway API to return from the item

OR
 
The gateway filters the log  and provide APIs that return the function log or the other part of log
 
WDYT @Tansito @psschwei @IceKhan13

akihikokuroda:
Do we need to emit the function log to somewhere?

psschwei:
One could make the argument that end users just need the function output/result and logs should only be available to the provider. That would probably require providers to generate something to return when jobs fail (so would shift some ownership to them), but it would also remove the need for us to filter logs (reducing our responsibility a bit, which I also like).

Tansito:
Yeah, we were commenting good approximations to this and due to the deadlines that we have our proposal would be:

- Logs from Provider Functions are going to be accesible only by user providers

So for this, the proposal in what I was thinking it's something similar to what Paul is saying:
- A user calls to `/job/:id/logs` end-point
  - if the user is a provider we return the logs (we know that a user is a provider because a provider has an `admin_group` assigned)
  - if not we return a message 
- A provider can obtain jobs id's from its functions through: `/provider/:id/function/:id/jobs`
- We need to close a structure for the result field to something like: { status: "ERROR_CODE_PRE_DEFINED", result: "" }
  - This way, as Paul said, providers can manage their errors and show to the user what they want

WDYT?

akihikokuroda:
> A user calls to /job/:id/logs end-point
if the user is a provider we return the logs (we know that a user is a provider because a provider has an admin_group assigned)
if not we return a message


Our **log** is a string returned by [ray.job_submission.JobSubmissionClient.get_job_logs](https://docs.ray.io/en/latest/cluster/running-applications/job-submission/doc/ray.job_submission.JobSubmissionClient.get_job_logs.html#ray.job_submission.JobSubmissionClient.get_job_logs). It is saved into `Job.logs` and retrieved via `api/v1/job/<job id>/logs`.
So this api is changed to return the log only for the provider.

[...]

PR #1406, opened 2024-07-11
add Function.get_jobs()

akihikokuroda:
### Summary
This the second part of #1400 

### Details and comments

This PR providers `Function.get_logs()` method that returns the job ids of jobs that executed the `function`.


psschwei:
The `running_programs_using_decorators.py` test is failing in kubernetes -- don't remember, is that one broken in general?

Tansito:
> The running_programs_using_decorators.py test is failing in kubernetes -- don't remember, is that one broken in general?

It should be working from what I remember, yep

akihikokuroda:
It's a cloudpickle one.  It must get broken when gateway moved to 3.11.

Tansito:
> It's a cloudpickle one. It must get broken when gateway moved to 3.11.

Oh, maybe. Maybe we can fix it if we move everything to python 3.11.

psschwei:
yeah, it's been failing for a couple of weeks now ([this](https://github.com/Qiskit/qiskit-serverless/actions/runs/9745549342) was as far back as I looked, but that was the end of the 1st page results)

gateway/tests/api/test_v1_program.py
```diff
@@ -414,3 +414,16 @@ def test_get_by_title(self):
...
+        """Tests run existing authorized."""
+
+        user = models.User.objects.get(username="test_user_2")
+        self.client.force_authenticate(user=user)
+
+        response = self.client.get(
+            "/api/v1/programs/1a7947f9-6ae8-4e3d-ac1e-e7d608deec82/get_jobs/",
+            format="json",
+        )
+        print(response.data)
```
Tansito, line 428: Probably here instead to print I would check if it returns the expected number of jobs

gateway/api/views.py
```diff
@@ -358,6 +358,19 @@ def _get_program_queryset_for_title_and_provider(
...
 
+    @action(methods=["GET"], detail=True)
+    def get_jobs(
+        self, request, pk=None
+    ):  # pylint: disable=invalid-name,unused-argument
+        """Returns jobs of the program."""
+        tracer = trace.get_tracer("gateway.tracer")
+        ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
+        with tracer.start_as_current_span("gateway.program.get_jobs", context=ctx):
+            program = Program.objects.filter(id=pk).first()
```
Tansito, line 369: We cannot return jobs only filtering by `id` because everyone could get `jobs` if it has that `id`. My recommendation for this would be:
- If the user is a provider returns all the jobs
- if the user is not a provider return **only** author jobs
- Let's use the serializer `JobSerializer` to return more info about the `Job`: 
  - https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/serializers.py#L102

Tansito (changes requested):

gateway/tests/api/test_v1_program.py
akihikokuroda: It was for debugging.  I'll take it out.  Thanks!

gateway/api/views.py
```diff
@@ -358,6 +358,30 @@ def _get_program_queryset_for_title_and_provider(
...
```
Tansito, line 369: We should check if `program` is **None** first.

Tansito (changes requested):

Tansito:
Can we add a test here too testing the client, Aki? https://github.com/Qiskit/qiskit-serverless/tree/main/tests/basic

function/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM icr.io/quantum-public/qiskit-serverless/ray-node:0.12.0-py310
```
Tansito, line 1: We have 0.14.0 version right now. We should update it here.

.github/workflows/docker-build.yaml
```diff
@@ -10,6 +10,8 @@ jobs:
     timeout-minutes: 45
     steps:
       - uses: actions/checkout@v4
+      - name: Build the function
+        run: docker build -t test_function:latest --build-arg TARGETARCH="amd64" -f ./function/Dockerfile .
```
Tansito, line 14: Is it possible to move this Dockerfile inside `tests/basic` and not use `root` folder?

Tansito: I would maybe change the name too as we did with `Sample-Docker` to avoid interactions with our CI from SPS.

function/runner.py
```diff
@@ -0,0 +1,25 @@
+from qiskit import QuantumCircuit
```
Tansito, line 1: `runner.py` files are duplicated, is it possible to remove one?

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -109,6 +109,10 @@ spec:
               value: {{ .Values.application.auth.token.verificationUrl | quote }}
             - name: SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD
               value: {{ .Values.application.auth.token.verificationField | quote }}
+{{- if .Values.application.authMockproviderRegistry }}
+            - name: SETTINGS_AUTH_MOCKPROVIDER_REGISTRY
+              value: {{ .Values.application.authMockproviderRegistry }}
```
Tansito, line 114: I would add a default value in the `values.yaml` like `icr.io` or `test` as you were using, wdyt?

client/qiskit_serverless/core/job.py
```diff
@@ -610,6 +614,32 @@ def get_program(
...
+        with tracer.start_as_current_span("program.get_by_title"):
+            response_data = safe_json_request(
+                request=lambda: requests.get(
+                    f"{self.host}/api/{self.version}/programs/get_by_title/{title}",
+                    headers={"Authorization": f"Bearer {self._token}"},
+                    params={"provider": provider},
+                    timeout=REQUESTS_TIMEOUT,
+                )
+            )
+            program_id = response_data.get("id", None)
```
Tansito, line 633: I would check if it's `None` here, Aki. And return and return an error if the function doesn't exist.

docker-compose-dev.yaml
```diff
@@ -45,6 +45,7 @@ services:
       - DJANGO_SUPERUSER_EMAIL=admin@noemail.com
       - SITE_HOST=http://gateway:8000
       - SETTINGS_AUTH_MECHANISM=mock_token
+      - SETTINGS_AUTH_MOCKPROVIDER_REGISTRY=test
```
Tansito, line 48: Can we add this to `docker-compose.yaml` too?

gateway/api/views.py
Tansito: We continue needing to check if `program` is `None` here.

tests/basic/06_function.py
```diff
@@ -0,0 +1,39 @@
+import os
```
Tansito, line 1: Thank you so much for the test, Aki. Really appreciate it 🙏

Tansito (changes requested):

akihikokuroda:
@Tansito Thanks for your review.  I updated according to your comments.  I couldn't make the function test in `Test QS on Kubernetes` because the resource available in the Github Action is not enough to build the all necessary container images.  The function test is added in the `Docker notebook tests`.

charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -27,7 +28,7 @@ application:
     maxReplicas: 4
     opensslImage: registry.access.redhat.com/ubi8/openssl:8.8-9
     kubectlImage: alpine/k8s:1.29.2@sha256:a51aa37f0a34ff827c7f2f9cb7f6fbb8f0e290fa625341be14c2fcc4b1880f60
-    proxyImage: "icr.io/quantum-public/qiskit-serverless/proxy:0.9.0"
+    proxyImage: "qiskit/qiskit-serverless-proxy:local"
```
Tansito, line 31: was this a desired change? I would maintain the old value, Aki. wdyt?

docker-compose.yaml
```diff
@@ -39,7 +39,8 @@ services:
       - DJANGO_SUPERUSER_EMAIL=admin@noemail.com
       - SITE_HOST=http://gateway:8000
       - SETTINGS_AUTH_MECHANISM=mock_token
-      - DATABASE_HOST=postgres
+      - SETTINGS_AUTH_MOCKPROVIDER_REGISTRY=test
+©      - DATABASE_HOST=postgres
```
Tansito, line 43:
```suggestion
      - DATABASE_HOST=postgres
```

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -31,11 +31,27 @@ jobs:
...
           docker image rm ray:test
-      - name: Build and load proxy
+      - name: Build and load function
         run: |
-          docker build -t proxy:test --build-arg TARGETARCH="amd64" -f ./proxy/Dockerfile .
-          kind load docker-image proxy:test
-          docker image rm proxy:test
+          docker build -t test_function:latest --build-arg TARGETARCH="amd64" -f ./tests/basic/function/Sample-Docker ./tests/basic
+          docker images
+          docker rmi moby/buildkit:latest
```
Tansito, line 38: Seeing that this configuration it's a bit weird right now, Aki, and just to not postpone too much this PR how would you see to remove by now the test from `kubernetes-deploy` and do wat @psschwei proposed one moment ago in another PR?

Tansito (changes requested):

charts/qiskit-serverless/charts/gateway/values.yaml
Tansito: An probably update the version.

akihikokuroda: It's my mistake :-(

Tansito:
@akihikokuroda I saw that you were making changes, let me know if you want that I review something 👍

akihikokuroda:
@Tansito Please review the changes.  Thanks!

.github/workflows/kubernetes-deploy.yaml
```diff
@@ -81,6 +78,7 @@ jobs:
           echo $GATEWAY_HOST
           # basic tests
           cd /home/runner/work/qiskit-serverless/qiskit-serverless/tests/basic
+          rm 06_function.py
```
psschwei, line 81: nit: should we put the function test in its own directory to prevent running it?

tests/basic/function/Sample-Docker
```diff
@@ -0,0 +1,15 @@
+FROM icr.io/quantum-public/qiskit-serverless/ray-node:0.14.0-py310
```
psschwei, line 1: depending on when this merges, we may need to drop the python version
```suggestion
FROM icr.io/quantum-public/qiskit-serverless/ray-node:0.14.0
```

psschwei (commented):
couple of very minor things, but otherwise LGTM

tests/basic/function/Sample-Docker
Tansito: From my last review here Aki, Paul I would like to keep it if both of you agree 👍 

https://github.com/Qiskit/qiskit-serverless/pull/1420#issuecomment-2250448999

akihikokuroda:
Are there any other changes required?  @Tansito @psschwei Thanks!

Tansito:
Sorry @akihikokuroda , I was turning off fires today and I couldn't have more time to see this. I will try to take a final look before finish the day 🙏

tests/basic/function/Sample-Docker
Tansito: I just saw your comment, Paul. I couldn't dedicate too much time to process everything, sorry. Tomorrow I'm going to read calmly everything and maybe we can connect to discuss it a bit. We can always do this change in that PR I assume 😄

Tansito (approved):
LGTM everything, Aki. Thank you!

PR #1407, opened 2024-07-12
Configure tasks to be able to run configurations

Tansito:
### Summary

This PR pretends to enable a way to pre-populate data outside the migrations. The main reason for this is in case we would have sensitive information and we can not publish we could use this approach to continue automatizing a bit the project.

The idea is that after `post_migrate` signal we will be able to run some tasks to configure some specific things.

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
@@ -176,9 +176,9 @@ spec:
                   name: {{ .Values.secrets.servicePsql.name }}
                   key: {{ .Values.secrets.servicePsql.key.databasePassword }}
             - name: PROVIDERS_CONFIGURATION
-              value: {{ .Values.tasks.providersConfiguration }}
+              value: "{{ .Values.tasks.providersConfiguration }}"
             - name: FUNCTIONS_PERMISSIONS
-              value: {{ .Values.tasks.functionsPermissions }}
+              value: "{{ .Values.tasks.functionsPermissions }}"
       {{- with .Values.nodeSelector }}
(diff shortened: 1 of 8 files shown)


akihikokuroda:
I'm not sure what is the use case and how to use this.  Would you give some examples?

Tansito:
This is temporary btw until we can enable the panel administrator and manage all these tasks there.

akihikokuroda:
I see.  It's picking up from the environment variables.  The values.yaml is updated, the gateway deployment instance is updated and the gateway pod is restarted.  Is it the way to use this?

Tansito:
> I see. It's picking up from the environment variables. The values.yaml is updated, the gateway deployment instance is updated and the gateway pod is restarted. Is it the way to use this?

That's exactly the idea. As an starting point we modify the `values.yaml` so we restart the application, after the migrations it loads the new data and everything continues working. Not ideal but I couldn't find a better way to manage it without introducing too many changes.

akihikokuroda:
I like the api idea better because it is not necessary to restart the gateway and it is probably easier to understand.  We can look into the api path later.

akihikokuroda (approved):

Tansito:
> I like the api idea better because it is not necessary to restart the gateway and it is probably easier to understand. We can look into the api path later.

It's an approach that I reviewed and the idea that I had was to have an environment variable like "ADMIN_USERS" where we could configure our id's. It would be another idea, yep.

psschwei (approved):
We're going to work on storing this info in the database in the future

Tansito:
> We're going to work on storing this info in the database in the future

I'm going to open an issue for this just after the release.

PR #1408, opened 2024-07-15
mount the dedicated storage for each function

akihikokuroda:
### Summary
Mount the dedicated storage for each function

### Details and comments

The dedicated storage area is mounted at `/function_data` for the function provider.  When the function doesn't have the associated provider, the storage mounted at `/function_date` is the same storage as `/data`.


Tansito:
My question to solve with this is... How are we going to make that users cannot access to this `sub_path` but providers can?

akihikokuroda:
Based on the current function design, when the Function is provided in the custom image, the code running in the ray nodes is the function code only.  The user code doesn't run in the ray nodes so it can not access to the directory.

charts/qiskit-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -119,6 +119,9 @@ data:
               - mountPath: /data
                 name: user-storage
                 subPath: {{`{{ user_id }}`}}
+              - mountPath: /function_data
+                name: user-storage
```
psschwei, line 123: It may be worth an internal discussion on whether we want / need to use a different volume for storing provider data

Tansito: Yeah, I agree I'm not sure about this solution either.

akihikokuroda: Yeah, I realized that it may not necessary when I implemented this.

charts/qiskit-serverless/charts/gateway/templates/rayclustertemplate.yaml
```diff
@@ -352,6 +355,9 @@ data:
...
+                subPath: {{`{{ function_data }}`}}
```
psschwei, line 360: Do we need to mount this with any kind of restricted permissions? In theory, and correct me if I'm wrong here, whoever is running the function (user or provider) would be able to access this location via the container...

akihikokuroda: My understanding is that with the current custom image function design, the only the code in the custom image is executed in the Ray node.  No user code is sent or executed in the Ray node so only the function code read/write to this directory.

psschwei: Users could in theory find a way around that restriction (if we misimplement something, CVS / exploits, etc.)  so it's a risk, we'd just need to determine if it's an acceptable one

Tansito: Could we just add the path when a function is from a provider? That could solve part of the problem. And correct me if I'm wrong here @akihikokuroda but all the providers will write here so they could potentially see files from other providers with this approach. What is the problem to use the name of provider as sub-path instead of `function_data`?

Tansito: My idea was that in that way we could attach the specific provider path only when it's a function from the specific provider (but I don't know if there are limitations in this approach).

akihikokuroda: >Could we just add the path when a function is from a provider? That could solve part of the problem. And correct me if I'm wrong here @akihikokuroda but all the providers will write here so they could potentially see files from other providers with this approach. What is the problem to use the name of provider as sub-path instead of function_data?

The `/<provider name> directory in the persistent volume is mounted to `/function_data` directory in the pod so each provider gets the separate space.

psschwei: assuming we aren't running as root, could we do something along the lines of `chmod 200 /function_data`  in the dockerfile, so the user could write to that location but not read files?

Tansito: > so the container could write to that location but not read files?

Well, this is something we can comment with Paco too, I'm not sure if the provider would need to read it at some moment 🤔

akihikokuroda: @psschwei @Tansito Is write only what they want?

Tansito: hahaha I just comment it the same at the same time, we can ask to Paco in the issue

psschwei: we could also use different uids for the provider and the user within the container (sort of like the old days when admins would install software on your laptop...)

akihikokuroda: >we could also use different uids for the provider and the user within the container (sort of like the old days when admins would install software on your laptop...)

It's not clear to me.  Function execution with 2 uids?

psschwei: I was thinking of something like the `/function_data` directory is "owned" by user 1002. Whereas the container gets run by user 1001, who has just enough permissions to execute the function.

In the end, whether we need to do something kind of restriction is ultimately going to depend on what the providers are comfortable with / require. So probably a good idea to check with Paco on what the requirements here are (and what level of risk is acceptable).

Tansito: After an internal discussion conversation is solved. We are going to assume that how this is a function for a provider we don't need to use permissions in this case.

I don't know what you think @akihikokuroda , @psschwei but maybe what we should do it's just add the volume `/function_data` when the function is from a providers? I don't see useful to it to the user having `/data` to be honest.

psschwei: If a function has some downloadable content, users would still need a place to get it from right? So I'd just leave it in. Since it sounds like we're going to end up with separate client for functions, may as well leave things as they are and review the whole design when its time to make the new client

Tansito: But for that we have `/data` folder, isn't it? But I agree, we can review it later once time we are sure about the entire workflow. Thanks Aki, Paul!

psschwei: I may have been confused. I thought you were asking if we should skip  mounting `/data` when it was a provider function.

akihikokuroda:
@Tansito @psschwei I would like to complete this.  What are the remaining concern?  Thanks!

psschwei:
Good from my side. Just need to resolve David's question about mounting `/data`

Tansito:
Sorry, I didn't read your answers yesterday.

charts/qiskit-serverless/charts/gateway/templates/rayclustertemplate.yaml
Tansito: Yeah, from a provider function I totally agree. My question is for those functions that are created by users (like the original ones).  It's like a path that it's not very useful for the user, isn't it? That's what I mean.

Tansito: Well, it's true that `/function_data` will be the same as `/data` for the user using `user.username` 👍

psschwei: I thought function-data was the provider space and data was for users ? since users would already go to data on their own functions

akihikokuroda:
The Function can write something that it wants to share with the user.  I'm planing providing the loggers to the function one for private logs and the other for the shared with the user.  The later log is written in the file in `/data` directory.

charts/qiskit-serverless/charts/gateway/templates/rayclustertemplate.yaml
akihikokuroda: For the provider, it may be useful to have both while they develop the function because it is the same environment when the function is executed by the user.

Tansito: > I thought function-data was the provider space and data was for users ?
yep

> since users would already go to data on their own functions
basically that's my comment, for functions that come from the user we are adding `/function_data` too. It's true that `/function_data` and `/data` are going to point to `user.username` path so not a bid deal but probably we can just avoid add `/function_data` for user functions.

BTW, another answer that I would understand is: "well, we don't overcomplicate more the template this way" and I would agree. Yep.

akihikokuroda: Django template in helm template is very tricky :-)

Tansito: Hahaha I agree, I agree. I'm just reviewing the rest of the code, Aki and I will update my review 👍

Tansito (approved):
Logic LGTM Aki, thank you! BTW something I would like to improve in the near future are logs. We need to introduce more logs in general if not debug this in case of an error would be a nightmare 😅

Issue #1411, opened 2024-07-16
Limitate that a provider can only update a Qiskit Function to a specific registry


akihikokuroda:
### What is the expected enhancement?

We can use the registry field from the Provider:
https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/models.py#L69

And validate that when a user uploads a qiskit function always starts with that path:
https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/serializers.py#L35

PR #1412, opened 2024-07-16
check regstry location of provider

akihikokuroda:
### Summary
fix: #1411

### DIFF
--- a/gateway/api/tasks/providers.py
+++ b/gateway/api/tasks/providers.py
@@ -43,7 +43,7 @@ def assign_admin_group():
 
             if created:
                 logger.info(  #  pylint: disable=logging-too-many-args
-                    "Provider [%s] created for admin [%s]",
+                    "Provider [%s] created for admin [%s] with registrt [%s]",
                     provider.name,
                     admin_group_name,
                     registry,
--- a/gateway/api/tasks/providers.py
+++ b/gateway/api/tasks/providers.py
@@ -20,25 +20,31 @@ def assign_admin_group():
     """
     try:
         providers_configuration = json.loads(settings.PROVIDERS_CONFIGURATION)
-    except json.JSONDecodeError as e:
-        logger.error("Assign admin group JSON malformed: %s", e)
+    except json.JSONDecodeError:
+        logger.error(
+            "Assign admin group JSON malformed in settings.PROVIDERS_CONFIGURATION"
+        )
         return
-    except Exception as e:  # pylint: disable=broad-exception-caught
-        logger.error("Assign admin group unexpected error: %s", e)
+    except Exception:  # pylint: disable=broad-exception-caught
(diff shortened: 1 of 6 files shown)


gateway/api/tasks/providers.py
```diff
@@ -20,25 +20,31 @@ def assign_admin_group():
...
-    except json.JSONDecodeError as e:
-        logger.error("Assign admin group JSON malformed: %s", e)
+    except json.JSONDecodeError:
+        logger.error(
+            "Assign admin group JSON malformed in settings.PROVIDERS_CONFIGURATION"
+        )
         return
-    except Exception as e:  # pylint: disable=broad-exception-caught
-        logger.error("Assign admin group unexpected error: %s", e)
+    except Exception:  # pylint: disable=broad-exception-caught
```
psschwei, line 28: did the log of the error get flagged on the security check? otherwise might be nice to keep for debugging

gateway/tests/api/test_v1_program.py
```diff
@@ -216,19 +216,19 @@ def test_upload_custom_image_without_access_to_the_provider(self):
                 "title": "Private function",
                 "dependencies": "[]",
                 "env_vars": env_vars,
-                "image": "icr.io/awesome-namespace/awesome-title",
-                "provider": "default",
+                "image": "docker.io/awesome-namespace/awesome-title",
+                "provider": "ibm",
```
psschwei, line 220: wouldn't ibm images always been in icr.io ? or is there an issue with using icr.io in the tests (looks like we're moving all the tests to docker.io, so maybe that's it?)

gateway/api/tasks/providers.py
akihikokuroda: Yes, the static scan complains about this.

psschwei (approved):

gateway/api/tasks/providers.py
```diff
@@ -20,25 +20,31 @@ def assign_admin_group():
...
             provider, created = Provider.objects.update_or_create(
-                name=provider_name, defaults={"admin_group": group}
+                name=provider_name,
+                defaults={"admin_group": group, "registry": registry},
             )
 
             if created:
-                logger.info(
+                logger.info(  #  pylint: disable=logging-too-many-args
                     "Provider [%s] created for admin [%s]",
```
Tansito, line 46: Correct me if I'm wrong Aki, don't we need here a `%s` to show the registry? This way we are only showing the two first.

Tansito (approved):
Just a comment left, thanks Aki 😄

gateway/api/tasks/providers.py
akihikokuroda: oh, right.  I'll add it.  Thanks!

PR #1410, opened 2024-07-16
Improved documentation for custom images

Tansito:
### Summary

Fix and improve a couple of references for the documentation about custom images

PR #1413, opened 2024-07-16
Fix tox configuration

Tansito:
### Summary

From #1412 and #1410 it seems the pipeline was failing due to a problem in `tox` configuration. Initially `install_command` should only accept one command so I moved the second one to `commands` parameter. Let's see if this fix the problem.

Spoiler: let me be honest I'm not sure if this change is correct or not 😂


akihikokuroda (approved):
LGTM. Thanks!

PR #1414, opened 2024-07-17
Fix tasks initialization

Tansito:
### Summary

We are having problems initializing the tasks so this new refactorization from the documentation hopefully fixes the problem:
https://docs.djangoproject.com/en/5.0/ref/signals/#post-migrate


psschwei (approved):
let's try it

akihikokuroda (approved):

PR #1416, opened 2024-07-18
Multiple groups in a provider

Tansito:
### Summary

This PR enables to have more than one group managing the resources of a provider. The idea is that more than one group could update a function if it would be needed.

This is needed due to that right now how we don't have a fully working workflow to enable that a provider publish a function so we need to give permissions to ourselves too to do it.

### Details and comments

To accomplish this use case we basically did the next steps:
- Removed `admin_group` field from Provider
- Added `admin_groups` field to the Provider as a M2M relationship
- Changed the current logic to support the change


akihikokuroda (approved):
LGTM.  Would you add example of the structure of user, groups and provider?  I assume there are user for consumer, provider and admin and groups for function admins and function user for each function.

Tansito:
What do you have in mind, @akihikokuroda ? Some kind of diagram? I can prepare it monday an update it to the PR

akihikokuroda:
Yeah, a diagram is great.thanks

Tansito:
Screenshot 2024-07-22 at 10 18 27 AM

Something like this would be what are you looking for, @akihikokuroda ?

akihikokuroda:
@Tansito It's a little different.  What I want is probably an instance diagram that has some example user, provider and groups.

Tansito:
I'm going to wait for #1406 before merge this. I think it's easier to migrate the logic from this PR to `admin_groups` than for the other one.

Tansito:
Initially is ready for review but Monday I will give it a last review just in case I missed any reference.

akihikokuroda (approved):

Issue #1034, opened 2023-10-18
Expose ability to set python version via client


IceKhan13:
### What is the expected behavior?

Expose ability to set python version via client.

Something like
```python
serverless.run(program, config=Configuration(python=Configuration.PY_39))
```

akihikokuroda:
closing as done by #1077

PR #1418, opened 2024-07-22
Build ray image with ubi base

psschwei:
### Summary

Build the ray image using a UBI base 

### Details and comments

This currently only tests x86, as I have no way to test an ARM image. Per Aki, the UBI base images work on Apple Silicon.

I've also left the old dockerfile (using the community Ray image as a base), in case that's helpful.

This also pins numpy to v1.x, as v2 causes a number of issues

PR #1420, opened 2024-07-23
Python 3.11 is only supported python version

psschwei:
### Summary

Only build and test against python 3.11

### Details and comments

Removes changes made in #1034


charts/qiskit-serverless/values.yaml
```diff
@@ -59,7 +59,7 @@ gateway:
       type: ClusterIP
       port: 8000
     ray:
-      nodeImage: "icr.io/quantum-public/qiskit-serverless/ray-node:0.14.0-py310"
+      nodeImage: "icr.io/quantum-public/qiskit-serverless/ray-node:0.14.0-py311"
```
Tansito, line 62: Should we just start naming it: `ray-node:0.14.0` and drop `py311`?

psschwei: works for me

psschwei:
We have a [whole mechanism](https://github.com/Qiskit/qiskit-serverless/blob/ff7bbfabfbb06f8e138ff58be505a5415e703efe/gateway/api/ray.py#L237-L250) to allow users to choose their Python version that will also need to be ripped out

Tansito:
> We have a [whole mechanism](https://github.com/Qiskit/qiskit-serverless/blob/ff7bbfabfbb06f8e138ff58be505a5415e703efe/gateway/api/ray.py#L237-L250) to allow users to choose their Python version that will also need to be ripped out

That's true, good catch. Probably better to address this in another PR.

Tansito:
> We have a [whole mechanism](https://github.com/Qiskit/qiskit-serverless/blob/ff7bbfabfbb06f8e138ff58be505a5415e703efe/gateway/api/ray.py#L237-L250) to allow users to choose their Python version that will also need to be ripped out

Good catch, thinking in this feature... Probably we could remove the config by now from the client and in the api use always the same image `node_image = settings.RAY_NODE_IMAGE`. With the providers I don't think it has sense to allow this kind of configuration from the user, wdyt? 🤔

Tansito:
Sorry Paul, missclick 😅

psschwei:
test failures should be resolved by https://github.com/Qiskit/qiskit-serverless/pull/1418/commits/1b3e8ee4e8b8910b1833cb4f78a27cce64e84c05 , so let's get the UBI PR merged first then return to this one

psschwei:
This also drops the number of gh actions we run to 10 (down from 18)

akihikokuroda:
> With the providers I don't think it has sense to allow this kind of configuration from the user, wdyt?

It maybe too late but it may be useful for the preparation for the next python version update.  The provider can test their function with the new python version in advance.

.github/workflows/update-component-versions.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           sed -i "s/ray-node:${OLDNUM}/ray-node:${NEWNUM}/" charts/qiskit-serverless/charts/gateway/values.yaml
           sed -i "s/qiskit-serverless\/proxy:${OLDNUM}/qiskit-serverless\/proxy:${NEWNUM}/" charts/qiskit-serverless/charts/gateway/values.yaml
           sed -i "s/tag: \"${OLDNUM}\"/tag: \"${NEWNUM}\"/" charts/qiskit-serverless/values.yaml
-          sed -i "s/tag: \"${OLDNUM}-py39\"/tag: \"${NEWNUM}-py39\"/" charts/qiskit-serverless/values.yaml
```
akihikokuroda, line 40: Is this still necessary?

psschwei: nope

psschwei:
> The provider can test their function with the new python version in advance.

Good point. Maybe when the project stabilizes we'll need to start cutting release candidates (similar to what qiskit does) for users to test new versions?

Tansito:
> The provider can test their function with the new python version in advance.

Yep, I didn't think about this but it's a good point for the future, Aki. I agree, let's do the next:
- Let's keep the configuration by now (I liked the idea of the releases candidates)
- Let's remove python 3.9 and 3.10 support only
- Let's maintain the `-py311`

Thank you guys! 🙏

gateway/api/models.py
```diff
@@ -42,10 +42,12 @@ class JobConfig(models.Model):
...
     PYTHON_V3_9 = "py39"
     PYTHON_V3_10 = "py310"
+    PYTHON_V3_11 = "py311"
     PYTHON_VERSIONS = [
         (PYTHON_V3_8, "Version 3.8"),
         (PYTHON_V3_9, "Version 3.9"),
         (PYTHON_V3_10, "Version 3.10"),
+        (PYTHON_V3_11, "Version 3.11"),
     ]
     python_version = models.CharField(
```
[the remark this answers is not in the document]
psschwei: I wasn't sure if we needed that for legacy data, but can try taking it out.

psschwei:
I'd still drop the python version. Testing against a release candidate shouldn't be exposed publicly, instead we'd set it up on a staging environment and run tests against that explicitly. Plus, if we're making this available for providers only then we'd need to wrap some kind of permissions around that field so that only providers could call it. Much simpler to just drop it completely.

gateway/api/models.py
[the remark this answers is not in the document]
Tansito: Not needed. In fact we didn't move data from our old database to the new one in the migration what it simplifies even more this.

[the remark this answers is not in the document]
Tansito: What we are going to need to test it's if somethings breaks if old data exists in the database. I can help with that test later.

Tansito:
@psschwei changes look good, if you remove `python_version` from the model I can help to test it 👍

Tansito:
And after reading you correctly, I agree. Let's see in future discussions how can we improve that way to facilitate the access to the providers to these nightlies versions.

psschwei:
@akihikokuroda @Tansito I think I've got the last bits of python_version out now, so this one is ready for another look

Tansito (commented):
@psschwei can we regenerate the migrations to have only one migration? For the rest of the things I see everything fine. I added @IceKhan13 too to the PR for his awareness 😄

I continue discussing with myself if this approach has sense in our context or not. But I don't see a way were we can support more than one python version without dying in the effort at this moment.

Tansito (approved):
LGTM Paul, thank you! I would like to know if @akihikokuroda and @IceKhan13 agrees with the change too

akihikokuroda (approved):

Tansito:
Btw @psschwei I think I added you some conflicts due to my merge 😅

Tansito:
@psschwei feel free to merge it so I can update #1416 and we merge both PRs today 💪

PR #1422, opened 2024-07-25
add versions for install modules in Dockerfile

akihikokuroda:
### Summary
add versions for install modules in Dockerfile


Dockerfile-ray-node
```diff
@@ -1,5 +1,5 @@
 FROM registry.access.redhat.com/ubi9-minimal:9.4-1134
-RUN microdnf install -y python3.11 python3.11-pip python3.11-devel vim-enhanced wget &&\
```
psschwei, line 2: If I'm remembering my errors correctly, I think we had to add wget for the ray node to work properly

psschwei:
Image builds without error on x86!

psschwei (approved):

PR #1424, opened 2024-07-26
Update ubi base image, drop legacy ray node

psschwei:
### Summary

Using latest version of UBI image
Also dropping legacy ray node build, as its working on both arm and x86


Tansito:
Whit what frequency this version changes, @psschwei . Do you know?

psschwei:
every six weeks or so

Tansito (approved):
Meh, not so bad then 🤷‍♂️ 

LGTM, Paul. Thank you!

PR #1426, opened 2024-07-30
Update the version for pyOpenSSl

Tansito:
### Summary

We are experiencing problems with some dependencies due to this error explained in this issue:

https://github.com/conda/conda/issues/13619#issuecomment-2075825121

The solution that we think it can work is to override the version that ray uses in its Dockerfile as we are doing in the PR.


akihikokuroda (approved):

Tansito:
It seems the partner is testing some configurations so I will put on-hold this PR until we can confirm that it's needed.

Tansito:
Problem solved by the other side so nothing to be added from our part 🎉

PR #1429, opened 2024-08-01
Improve registry validation

### LINKED ISSUE #1428: Allow upload functions without registry in development

Use `docker-compose up` and upload a function following the steps in: https://qiskit.github.io/qiskit-serverless/deployment/deploying_custom_image_function.html

If a user wants to upload a function with a docker image in development right now it returns an error because we are checking if the provider has a registry.

In development we should allow that a user uploads any local function without specify a registry.

Tansito:
### Summary

Fix #1428  

Let's simplify the logic of the registry validation to allow that if a registry is not specified the provider can use any registry that it wants. Specifically useful for development.


gateway/api/v1/serializers.py
```diff
@@ -60,10 +63,8 @@ def validate(self, attrs):
                 provider = title_split[0]
             provider_instance = Provider.objects.filter(name=provider).first()
             if provider_instance is None:
-                raise ValidationError(
-                    "Custom images are only available a valid provider."
-                )
-            if not provider_instance.registry or not image.startswith(
+                raise ValidationError(f"{provider} is not valid provider.")
+            if provider_instance.registry and not image.startswith(
```
IceKhan13, line 67: Are we sure here that `provider_instance` always has `registry` attribute?

Tansito: you mean if the `registry` from `provider_instance` can be `None`? The answer is yes. What I'm trying to do here is that if it's `None` it's ok, we accept any value in the image.

Tansito: Until now we were forcing it to have a value but there are situations where it can has no sense like in local development.

akihikokuroda (approved):
LGTM

IceKhan13 (approved):

PR #1436, opened 2024-08-02
set token permissions for workflows 

akihikokuroda:
### Summary
This PR fix "Token-Permissions" issues found by scorecard scan.

### Details and comments
Issues are listed here: https://github.com/Qiskit/qiskit-serverless/security/code-scanning

### DIFF
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -7,7 +7,7 @@ on:
       - main
 
 permissions:
-  contents: read
+  contents: write
 
 jobs:
   build_and_deploy_docs:
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -7,11 +7,13 @@ on:
       - main
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
   build_and_deploy_docs:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v4
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -1,7 +1,8 @@
 name: Build and deploy sphinx docs
 
 on:
-  permissions: contents: read
+  permissions:
+    contents: read
   workflow_dispatch:
   push:
     branches:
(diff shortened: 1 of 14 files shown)


.github/workflows/publish-docs.yaml
```diff
@@ -6,6 +6,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
```
psschwei, line 10: since this one is publishing docs, will it need write permissions?

.github/workflows/release-drafter.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
```
psschwei, line 9: I don't know what this action does exactly (I don't think I've ever run it), but it might need write permissions too

.github/workflows/publish-docs.yaml
akihikokuroda: Thanks for review.  This is the permission only for the github contents.  The list of operations are listed [here](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents).   It seems OK for me to put the `read` but I can change it to `write` for safe.

.github/workflows/release-drafter.yml
akihikokuroda: same as above.

akihikokuroda: It seems OK https://github.com/danyeaw/release-drafter/blob/ca78f853c4403f81294579c6fbff6acd2632795e/README.md

.github/workflows/publish-docs.yaml
akihikokuroda: This one needs `write`. https://github.com/peaceiris/actions-gh-pages.

.github/workflows/publish-docs.yaml
```diff
@@ -6,6 +6,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: write
```
psschwei, line 10: IIRC we should add the write permission just for the particular stage of the job that needs it (the deploy docs one) and leave the overall permissions as read

akihikokuroda: The permission can be set for each `Job` but not each `step`.  This workflow has only one job so it may not make difference but I changed the overall permission to `read` and the job permission to `write`.

psschwei (approved):

PR #1437, opened 2024-08-05
pin python module versions in workfows and add sha values for base docker images

akihikokuroda:
### Summary
pin python module versions in workfows and add sha values for base docker images

### DIFF
--- a/.github/workflows/client-verify.yaml
+++ b/.github/workflows/client-verify.yaml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install tox
         run: |
-          pip install tox
+          pip install tox==4.16.0
 
       - name: Run styles check
         run: tox -elint
--- a/.github/workflows/client-pypi-release.yaml
+++ b/.github/workflows/client-pypi-release.yaml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: '3.11'
       - name: Install Deps
-        run: pip install -U twine wheel
+        run: pip install -U twine==5.1.1 wheel==0.44.0
       - name: Build Artifacts
         run: |
           cd client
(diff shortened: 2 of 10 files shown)


psschwei:
Are these being recommended by something (scorecard maybe?)

akihikokuroda:
Yes, these are "Pinned Dependencies" in [here](https://github.com/Qiskit/qiskit-serverless/security/code-scanning)

Tansito (approved):
I see, LGTM Aki 👍

.github/workflows/client-verify.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install tox
         run: |
-          pip install tox
+          pip install tox==4.16.0
```
psschwei, line 42: Would it be worth installing the requirements-dev.txt file instead of a separate pip install here (and in the other actions)? Otherwise we'll have a lot of places that we have to update each time we want to upgrade a library...

akihikokuroda: The tox is installed by the actions for different components.  Where should the requirements.txt file put?  Repository root?  If so, should it be named like tox_version.txt instead of requirements.txt?  @psschwei WDYT?

psschwei: my initial thought was use the dev requirements file for client / docs / gateway / proxy ... that would cover everything except the kubernetes test...

maybe one requirements.txt in the main `test/` directory? 

not really a great answer here unfortunately... I'm fine with what ever you think is best @akihikokuroda  (including pinning the versions in the action)

akihikokuroda:
@psschwei I'll leave this for now.  I'm working for putting the sha value to the github actions.  It has similar situations.  We can address these issues at the same time later.  Thanks!

PR #1439, opened 2024-08-05
Fix docker compose failures

psschwei:
### Summary

Need to run docker compose as root so that our mounts work as expected.


Tansito (approved):
Thank you so much Paul ❤️

PR #1440, opened 2024-08-05
🏗️ Admin Panel application

Tansito:
### Summary

This PR pretends to integrate an administrator panel in `gateway` to manage in a better way the data from the database.

### Details and comments

I'm opening this PR to collect some feedback for the implementation. Basically the idea is to create a structure for future applications inside gateway project (like the scheduler). The division generated is the next one:

- I divided settings per application. **Important**: as you can see `scheduler` & `admin_panel` imports their settings from `api`. This is due to that `api` application contains the models and we cannot separate them right now. If we would have the models in other application we would not need to import the configuration from the `api`.
- I added a environment variable called `APPLICATION_MODE` that basically setups the configuration for the application depending if we are deploying the admin panel, the api or the scheduler.


Tansito:
@akihikokuroda , @IceKhan13 , @psschwei I would like to know what do you think about this approach to manage future applications inside the `gateway`. I would like collect some feedback before continue with the development. Any comment is welcome and other approaches too 🙏

gateway/main/settings_admin_panel.py
```diff
@@ -0,0 +1,41 @@
+"""
+Django settings for admin panel application.
+"""
+
+from .settings_api import *  # pylint: disable=wildcard-import, unused-wildcard-import
```
Tansito, line 5: Here you can take a look how we are managing settings from a base configuration.

gateway/main/urls.py
```diff
@@ -40,13 +41,23 @@
 
 router = routers.DefaultRouter()
 
-if settings.DEBUG:
-    urlpatterns = [
-        path("readiness/", probes.views.readiness, name="readiness"),
-        path("liveness/", probes.views.liveness, name="liveness"),
-        path("version/", version.views.version, name="version"),
+urlpatterns = [
```
Tansito, line 44: Here we can configure available end-points depending of the configuration.

gateway/main/wsgi.py
```diff
@@ -11,6 +11,16 @@
...
+if application_mode == "api":
+    DJANGO_SETTINGS_MODULE = "main.settings_api"
+elif application_mode == "scheduler":
+    DJANGO_SETTINGS_MODULE = "main.settings_scheduler"
+elif application_mode == "admin_panel":
+    DJANGO_SETTINGS_MODULE = "main.settings_admin_panel"
+else:
+    DJANGO_SETTINGS_MODULE = "main.settings_base"
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", DJANGO_SETTINGS_MODULE)
```
Tansito, line 24: The load of the settings is done here.

gateway/main/settings_admin_panel.py
Tansito: I left in the summary of the PR the reason behind we need to use `api` and not `base`, btw.

gateway/manage.py
```diff
@@ -14,7 +14,17 @@
...
+    if application_mode == "api":
+        DJANGO_SETTINGS_MODULE = "main.settings_api"
+    elif application_mode == "scheduler":
+        DJANGO_SETTINGS_MODULE = "main.settings_scheduler"
+    elif application_mode == "admin_panel":
+        DJANGO_SETTINGS_MODULE = "main.settings_admin_panel"
+    else:
+        DJANGO_SETTINGS_MODULE = "main.settings_base"
+
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", DJANGO_SETTINGS_MODULE)
```
Tansito, line 27: `manage.py` needs to load settings too because it's what we are using to run commands in the `scheduler`.

Tansito:
Added a diagram explaining the inheritance of the settings for each application to bring more clarity.

Tansito:
I will leave this refactorization proposal for an internal RFC once time we can have more time for this. Thanks team!

PR #1442, opened 2024-08-06
more pinning

akihikokuroda:
### Summary
Change the version numbers to sha values for the python modules

### DIFF
--- a/.github/workflows/client-verify.yaml
+++ b/.github/workflows/client-verify.yaml
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install tox
         run: |
-          pip install tox==4.16.0
+          pip install git+https://github.com/tox-dev/tox.git@2f7c60ec71d47f4099394af11999a840bc97c9a9 #tox==4.17.0
 
       - name: Run styles check
         run: tox -elint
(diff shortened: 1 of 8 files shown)


akihikokuroda:
There are still a few pinning issues remaining.  I'll work on them in the next PR.

.github/workflows/client-verify.yaml
```diff
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install tox
         run: |
-          pip install tox==4.16.0
+          pip install git+https://github.com/tox-dev/tox.git@2f7c60ec71d47f4099394af11999a840bc97c9a9 #tox==4.17.0
```
Tansito, line 42: Why are we installing it from git directly, Aki? I think I would prefer the other approach

Tansito: Basically it's my comment in the rest of the places where we are using git

akihikokuroda: That's the way to install using the hash value by pip.  The scan tool doesn't like the version nunber.

psschwei: I think it may be worth an internal discussion on whether we need to implement all these recommendations

Tansito: Agree, I don't know what best practice or security problem are we trying to cover with this to be honest 😅

akihikokuroda:
This seems too much.  I close this.

Issue #369, opened 2023-03-27
Limit dependencies to use in Nested Program


pacomf:
### What is the expected behavior?

We need a way to configure in the Middleware installation the dependencies to use (like a white list or black list, think about it), to restrict in someway external libraries to install and use it. Even it can be great if we can restrict access to external APIs, IPs or services (following the same approach with white list or black list).

It is a open question... think about it and design a proposal that can fit with that idea.

IceKhan13:
we can simply have list of backlisted packages and check them agains `dependencies` argument in `Program`

psschwei:
A framework for this was implemented in #1443 
Further work needs to be done to implement the full range of requirement specifiers in https://peps.python.org/pep-0508/
Then someone also has to come up with what we want to allowlist

PR #1443, opened 2024-08-06
Implement allowlist framework for dependencies

psschwei:

Allow operators to specify an allowlist of dependencies and allowed versions.

The allowlist is stored in a config file.

A sample allowlist might look like:

    allowlist = { "wheel": ["0.44.0", "0.43.2"] }

which would imply that the wheel package is allowed, but only versions 0.43.2 and 0.44.0 .

For this PR, the allowlist is empty, which means all dependencies are allowed, and it is stored locally. In a future PR, this should move into a Kubernetes configmap so that it can be updated without having to rebuild the container.

Each dependency can optionally specify a list of allowed versions. If the list is empty, then all versions of the dependency are allowed.

An example:

    allowlist = { "wheel": [] }

This could also be adapted to store a minimum allowed version instead of listing all allowed versions, depending on requirements.


psschwei:
still need to add tests for the new functionality...

psschwei:
ok, this should be ready for review now

gateway/api/v1/allowlist.json
```diff
@@ -0,0 +1 @@
+{}
```
Tansito, line 1: Curious about why do you opt by a JSON file to manage this, Paul? The problem that I see with this approach is that every time we want to modify this `white list` we are going to need to release a new version, aren't we?

What steps do we need to do in case we want to update this list in production, for example?

psschwei: Right now, just to make sure the whole process works (and also to be able to test).

Ultimately the allowlist should be mounted as a configmap on the gateway pod (meaning that file would live in the gateway helm templates directory), and since we deploy helm from the prod branch, all we'd need to do is update the prod branch to load the new configmap to the cluster.

Tansito: I see, so the idea is that we build a configmap and we get it using the environment variable, right?

psschwei: Pretty much. 
Though we wouldn't access as an envvar, but just [mount the configmap as a file on the pod](https://kubernetes.io/docs/concepts/configuration/configmap/#using-configmaps-as-files-from-a-pod). And we can just make sure the mount point is the same as the [default value](https://github.com/Qiskit/qiskit-serverless/blob/f490ec2abf94048876b984cf01d9aa3f3bd10ed9/gateway/main/settings.py#L357), and then we're good to go.

gateway/api/v1/serializers.py
```diff
@@ -43,6 +45,30 @@ def validate(self, attrs):
...
+        # allowlist stored in json config file (eventually via configmap)
+        # sample:
+        # allowlist = { "wheel": ["0.44.0", "0.43.2"] }
+        # where the values for each key are allowed versions of dependency
+        deps = json.loads(attrs.get("dependencies", None))
+        with open(settings.GATEWAY_ALLOWLIST_CONFIG, encoding="utf-8", mode="r") as f:
+            allowlist = json.load(f)
+
+        # If no allowlist specified, all dependencies allowed
+        if len(allowlist.keys()) > 0:
```
Tansito, line 58: should we check first is `allowlist` is `None`?

gateway/api/v1/serializers.py
```diff
@@ -43,6 +45,30 @@ def validate(self, attrs):
...
+            for d in deps:
+                dep, ver = d.split("==")
```
Tansito, line 60: How do you see here to force always to pin a version if you want to add a dependency?

I mean, check if the user sends: `wheel>=x.y.z`. So check if `dependencies` includes `>=` and if so raise a `ValidationError` to pin it. In that case we have one less corner to take care.

gateway/api/v1/serializers.py
```diff
@@ -2,9 +2,11 @@
 Serializers api for V1.
 """
 
+import json
 from rest_framework.serializers import ValidationError
 from api import serializers
 from api.models import Provider
+from main import settings
```
Tansito, line 9: I know that we are not using this in all the places (I'm changing it in the PR's that I have opened) but the standard way to access to the settings in django is using:

```py
from django.conf import settings
```

https://docs.djangoproject.com/en/5.0/topics/settings/#using-settings-in-python-code

gateway/api/v1/serializers.py
psschwei: I was not aware a user was allowed to specify anything other than `==` when passing dependencies

gateway/api/v1/serializers.py
psschwei: yeah, should catch possible errors here

gateway/api/v1/serializers.py
psschwei: well, it seems like the whole suite of [specifications](https://pip.pypa.io/en/stable/reference/requirement-specifiers/) is allowed....

Tansito: mmm... I didn't think about this until I saw your PR to be honest, sorry. Correct me if I'm wrong but basically if they use `SomeProject >= 1.2, < 2.0` we will return:

```
"Dependency SomeProject >= 1.2, < 2.0 is not allowed"
```

Maybe we can touch a little bit the message to say: `Dependency {dep} must be pinned or is not allowed`. Or something like that? Because I see a bit difficult to try to catch all the scenarios.

psschwei: yeah, we should probably force the user to either (a) specify a specific version, or (b) don't specify a version at all. the more esoteric cases (like version >3.0.1, < 4) are probably more work than it's worth to implement

gateway/api/v1/serializers.py
```diff
@@ -43,6 +45,30 @@ def validate(self, attrs):
...
             )
 
+        # validate dependencies
+        # allowlist stored in json config file (eventually via configmap)
+        # sample:
+        # allowlist = { "wheel": ["0.44.0", "0.43.2"] }
+        # where the values for each key are allowed versions of dependency
+        deps = json.loads(attrs.get("dependencies", None))
+        with open(settings.GATEWAY_ALLOWLIST_CONFIG, encoding="utf-8", mode="r") as f:
+            allowlist = json.load(f)
```
Tansito: I forgot this but should we add a try/catch here too? Just in case the file doesn't exist or the json is malformed?

psschwei: does this cover what you were thinking? https://github.com/Qiskit/qiskit-serverless/pull/1443/commits/dfc06056e27f30aa6f210608066d46803a5917d4

Tansito: Yep, I think so. Can we add a `log` warning us about this error? I assume we should always be able to load this file just in case.

gateway/api/v1/serializers.py
psschwei: Actually, we may be able to pull in the Ray logic to handle dependency parsing, need to do a little more investigation.

@Tansito since this PR is just getting the framework in place and thus not implementing any user-facing changes by populating the allowlist, let's merge this as is and then handle fine-tuning the dependency versioning in a follow up PR (that'll help keep the changes in each PR manageable)

Tansito: Sure, let me just give a last fast review taking that into account then.

Tansito (approved):
LGTM Paul, thank you!

psschwei:
xref #369

PR #1451, opened 2024-08-12
Remove retrieve Job end-point

Tansito:
### Summary

In this PR I remove two resources that we are not using right now:
- Retrieve Job end-point
- IsOwner permission: we are not using this permission configuration and the logic it's not updated with the current `provider` requirements.


Tansito:
Ok, nevermind, this end-point was being used in `result` method from `client`. Closing PR.

PR #1460, opened 2024-08-16
Catalog end-points

Tansito:
### Summary

This PR pretends to enable catalog end-points for our integration in IQP. For this integration we are creating two new end-points:
- `/api/v1/catalog`: that returns all the **public qiskit functions**
- `/api/v1/catalog/:id`: that returns a specific **public qiskit functions**

### Details about the implementation

-  What is a "Public Qiskit Function"?

Basically a _Public Qiskit Function_ it's a Qiskit Function that everyone should have access to view (but not run specifically), **even not authenticated users**.

For this we synced that all Qiskit Functions with the group "ibm-q/open/main" will be considered _Public Qiskit Functions_.

Due to this there some considerations to take into account:
- These end-points are public, so no-authenticated users are able to call to them too.
- There are 4 new fields:
  - _serializer only_ `available`: it will say to the catalog if the user (if exists in the request) has access or not
  - `type`: it will identify the type of the function: Generic, Application or Circuit. 
  - `url`, `icon_ulr`, `documentation_url` are basically fields that will contain URLs specific for the provider or the function
  - `additional_info`: contains a JSON structure with information for the catalog. Happy to discuss this approach, the decision behind it was that if the front-end requires a change we could easily change it without a new release or any change in the model.
- We didn't introduce yet how to populate these fields. The idea is try to have for the next week the `admin panel` and populate the info there.

For the rest of the things there are no more news: this PR introduces a new `view`, `serializers`, `tests` and swagger documentation for the new end-points.


psschwei:
> These end-points are public, so no-authenticated users are able to call to them too.

Do we anticipate endusers calling these APIs, or are they solely for the catalog? If it's the latter, we should probably restrict their usage somehow (either authentication or network policy restrictions or ...)

Tansito:
> Do we anticipate endusers calling these APIs, or are they solely for the catalog? If it's the latter, we should probably restrict their usage somehow (either authentication or network policy restrictions or ...)

Initially they are only designed for the catalog. There is a use case where a non-authenticated user in IQP could enter in the catalog page and it should be able to check the different public functions (in this case `availability` is set to False automatically).

I don't see probably the authentication but network policy restriction can have a lot of sense to be honest. We could take a look later, agree.

gateway/tests/api/test_v1_serializers.py
```diff
@@ -261,9 +261,6 @@ def test_upload_program_serializer_with_only_title(self):
 
     def test_upload_program_serializer_allowed_dependencies(self):
         """Tests dependency allowlist."""
-
-        print("TEST: Program succeeds if all dependencies are allowlisted")
-
```
psschwei, line 266: Did you want to remove these? If so, that's ok -- I just added those so I could see which output went with which test (which ideally we'd have for all tests).

gateway/api/serializers.py
```diff
@@ -242,3 +245,106 @@ class RuntimeJobSerializer(serializers.ModelSerializer):
...
+    class Meta:
+        model = Program
+
+    def get_available(self, obj):
+        """
+        This method populates available field.
+        If the user has RUN PERMISSION in any of its groups
+        available field will be True. If not, will be False.
+        """
+        author = self.context.get("author", None)
```
psschwei, line 277: nit

```suggestion
        user = self.context.get("author", None)
```

to avoid any confusion about intent (i.e. is the scope all functions the user has access to or only the functions the user is the author of)

gateway/api/serializers.py
```diff
@@ -242,3 +245,106 @@ class RuntimeJobSerializer(serializers.ModelSerializer):
...
+        representation["additional_info"] = json_additional_info
+        return representation
+
+    def get_available(self, obj):
+        """
+        This method populates available field.
+        If the user has RUN PERMISSION in any of its groups
+        available field will be True. If not, will be False.
+        """
+        author = self.context.get("author", None)
```
psschwei, line 330: s/author/user/

gateway/api/serializers.py
Tansito: I see, I like this distinction between user of the request and author of the function.

gateway/tests/api/test_v1_catalog.py
```diff
@@ -0,0 +1,131 @@
...
+
+        public_function = response.data[0]
+        self.assertEqual(public_function.get("available"), True)
+        self.assertEqual(public_function.get("title"), "Public-Function")
+        self.assertEqual(public_function.get("type"), Program.APPLICATION)
+
+    def test_catalog_retrieve_non_auth_user(self):
+        """Tests catalog retrieve non-authenticated."""
+        url = reverse(
+            "v1:catalog-detail", args=["1a7947f9-6ae8-4e3d-ac1e-e7d608deec82"]
```
akihikokuroda, line 64: I'm not sure how the `reverse` works.   Is this testing `retrieve` function?

gateway/api/serializers.py
Tansito: Applied change in [1c05730](https://github.com/Qiskit/qiskit-serverless/pull/1460/commits/1c057301ec0c23f735f9bf04ca9ef66ca2696851)

Let me know if this is what you were expecting, Paul.

gateway/api/serializers.py
Tansito: Applied change in [1c05730](https://github.com/Qiskit/qiskit-serverless/pull/1460/commits/1c057301ec0c23f735f9bf04ca9ef66ca2696851)

Let me know if this is what you were expecting, Paul.

gateway/tests/api/test_v1_catalog.py
Tansito: It's an utility from django, usually used in tests to not write every time the entire url:

https://docs.djangoproject.com/en/5.0/ref/urlresolvers/#reverse

In this case `v1:catalog-detail` points to `retrieve` but for example to access to the `programs/get_by_title` the reverse would be: `v1:programs-get-by-title`.

I usually get this info from this page:

Screenshot 2024-08-16 at 3 35 03 PM

I don't know, for me it's easier to manage it and we were using before in other tests too: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/tests/api/test_v1_program.py#L22

Happy to hear feedback if you prefer not use it. Unify approaches for me it's better 👌

akihikokuroda: I don't know how 

> v1:catalog-detail points to retrieve

Doesn't `v1:catalog-retrieve` work?

Tansito: > Doesn't v1:catalog-retrieve work?

Surprisingly not, I had the same doubt 😂 

`list`, `retrieve`, `update` are built in methods created by DRF ViewSets. I suppose that something below attaches the `retrieve` method from DRF to the `detail` name in Django. From what I saw was the only one method that has a _special_ name.

akihikokuroda: Have you tired just `v1:catalog`?   I have no idea so why the `detail` work.  Would you add comments on the `Reverse` with the `catalog-detail`?

psschwei (commented):
looks good from my side, will let Aki give final approval once his threads are resolved

gateway/tests/api/test_v1_catalog.py
Tansito: > Have you tired just v1:catalog?

Basically it follows the pattern: `<namespace>:<name>` where `<name>` is the name that django gives to the end-point (you can check an example in the image that I just shared a bit above).

> Would you add comments on the Reverse with the catalog-detail?

Sure, I will 👍

akihikokuroda (approved):
LGTM. Thanks!

Tansito:
Thanks @akihikokuroda , I'm just validating the format of the response before merge it and I'm almost have everything for the admin panel.

Tansito:
It seems everything looks so I'm going to merge to move forward with the admin-panel 👍

PR #1464, opened 2024-08-21
Admin panel

Tansito:
### Summary

This PR introduces a BackOffice for the project to help us with the management of the different resources running in the cluster.

- The back office is configured in: `/backoffice` end-point. 
- Some logic was added be able to search by different resources and some fields that can contain some IP are hidden.
- I refactorized a little bit the URLs file with the purpose to introduce `STATIC` folder in production and load the different css.
- To access to the admin panel we will need to create some admin users so we can just create them through our environment variables in our helm. I will take a look into this.


Tansito:
Btw, the good news is that as soon as this is merge it we will be able to remove the `tasks` that we created to fill data 🎉

akihikokuroda:
I'm not familiar this this admin module.  It looks useful.  Is it safe to enable this?  How is the user activated and the credentials are saved?

Tansito:
I need to admit that I'm not an expert either so I was basically reading a lot of documentation and asking:
https://docs.djangoproject.com/en/4.2/ref/contrib/admin/

> Is it safe to enable this?

I would say yes. It would be as safe as any UI that requires authentication that we could have.

> How is the user activated and the credentials are saved?

This is the process that I commented to you Aki, I don't know if you remember. Basically we can create an admin user with our current environment variables configuration: https://github.com/Qiskit/qiskit-serverless/blob/main/charts/qiskit-serverless/values.yaml#L93

This user is what we can use to access to the admin panel. I need to check how to create a secret in our Secrets-Manager for the initial password to use it in the environment variable. And with that we can create users for each of us.

Those admin users are saved the same as we are saving our current users. The only difference is that those users will have a `password` that django encrypts in database without problem.

akihikokuroda:
The secret manger is not installed in the cluster.  For now, we need to get console access (kubectl) and create the secret using it.

akihikokuroda:
As far as it safe.  I approve this PR.

akihikokuroda (approved):

Tansito:
> As far as it safe. I approve this PR.

Thanks @akihikokuroda , I will wait for @IceKhan13 in case he would have any recommendation too 👍 

> The secret manger is not installed in the cluster.

I confirm that we have connected the secrets-manager to our clusters, Aki. I learned today how it works in the new account, I can ping you tomorrow and show it to you.

Tansito:
Merging by now to move forward. Feel free to check it whenever you want @IceKhan13 👍

PR #1466, opened 2024-08-22
add filters to /job and /program

### LINKED ISSUE #1456: Gateway: implement sets of filter to return function by specific criteria

We want to have `jobs` and `programs` endpoints to have additional filtering parameter (https://<host>/v1/jobs?filter=<filter_name>).

Initially we need only 2 filters (names might change):
- `providers`
  - for `/programs?filter=providers` will return function uploaded by providers only
  - for `/jobs?filter=providers` will return only user jobs associated with providers only
- `user`
  - for `/programs?filter=user` will return function uploaded by this user only
  - for `/jobs?filter=user` will return only user jobs associated with only user programs

2 filters above are mutually exclusive.

akihikokuroda:
### Summary
fix #1456

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -115,9 +115,13 @@ class ProgramViewSet(viewsets.GenericViewSet):
         author = self.request.user
         title = self.request.query_params.get("title")
         provider_name = self.request.query_params.get("provider")
+        type_filter = self.request.query_params.get("filter")
 
         author_programs = self._get_program_queryset_for_title_and_provider(
-            author=author, title=title, provider_name=provider_name
+            author=author,
+            title=title,
+            provider_name=provider_name,
+            type_filter=type_filter,
         ).distinct()
 
         author_programs_count = author_programs.count()
@@ -305,7 +309,7 @@ class ProgramViewSet(viewsets.GenericViewSet):
         provider_name = self.request.query_params.get("provider")
 
         result_program = self._get_program_queryset_for_title_and_provider(
-            author=author, title=title, provider_name=provider_name
+            author=author, title=title, provider_name=provider_name, type_filter=None
         ).first()
 
         if result_program:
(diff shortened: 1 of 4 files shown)


IceKhan13:
Looks good! I'll try to run test it Monday for final approval :) Thanks Aki!

Tansito (commented):
I see the logic fine @akihikokuroda but I would like to propose a change in the logic as I was mentioning here: https://github.com/Qiskit/qiskit-serverless/issues/1456#issuecomment-2302236300

Instead to use `catalog` & `serverless` I would use `user` & `provider` as Iskandar proposed. The reason behind this is that If other client would like to retrieve `user` or `partner` would need to use `catalog` or `serverless` what probably it has not much sense.

Apart from that I see everything ok 👍

gateway/api/views.py
```diff
@@ -314,14 +318,38 @@ def get_by_title(self, request, title):
...
+        if type_filter:
+            if type_filter == "catalog":
+                view_permission_criteria = Q(permissions=view_program_permission)
+                groups_with_view_permissions = Group.objects.filter(
+                    view_permission_criteria
+                )
+                groups_with_view_permissions_criteria = Q(
+                    instances__in=groups_with_view_permissions
+                )
+                provider_exists_criteria = ~Q(provider=None)
```
Tansito, line 342: Just for curiosity what it does the `~` in front of the `Q`?

akihikokuroda: It's `NOT`.

akihikokuroda:
@Tansito The filter values are used (sent) by the current qiskit_ibm_catalog side.   So they must be these values until the catalog client is changed. They are not visible from the users so we can create a refactor PR later and fix both client and gateway at the same time.

Tansito:
Ooooh, ok, understood I didn't know that the catalog was already with those values.

IceKhan13 (approved):

Tansito:
Don't wait for my approval to merge it @akihikokuroda , what I saw I saw it ok and if one of us approves it it's more than enough 👍

PR #1468, opened 2024-08-23
Remove tasks

Tansito:
### Summary

Now that #1464 we are going to be able to remove tasks from our `gateway` thinking in that we are going to be able to use the admin panel to populate the needed data.


akihikokuroda (approved):

akihikokuroda:
One question 

How or where do we keep the provider and permission settings information?  It is in the values.yaml file but it will be removed, right?

Tansito:
> How or where do we keep the provider and permission settings information? It is in the values.yaml file but it will be removed, right?

It continues being in the database. These tasks were just to populate the database because we didn't any end-point or UI where we could setup providers and permissions.

akihikokuroda:
I'm thinking about "disaster recovery".  Can we assume that we never need to recover the data in the database?

Tansito:
> I'm thinking about "disaster recovery". Can we assume that we never need to recover the data in the database?

We should be prepared for any problem but we have configured in the databases security copies and so on every day or every X hours, I don't remember right now.

PR #1480, opened 2024-08-29
Rename function.get_jobs to function.jobs and make it return Job objects

### LINKED ISSUE #1463: Consistent get_jobs() return type

Currently `QiskitFunction.get_jobs()` returns a list of dictionaries, but `IBMServerlessClient.get_jobs()` return a list of `Job` objects. It'd be nice if they were consistent.

### LINKED ISSUE #1470: Provider can retrieve a job from its function

To retrieve a specific job we have `get_job_by_id `.  The problem is that we are limitating the retrieve of that job to the author and providers should be able to retrieve the job if it's a job from a owned function.

Providers should have access to all function jobs to check logs, errors and results.

### LINKED ISSUE #1479: Client: rename `get_jobs` to `jobs` in `QiskitFunction`

Client: rename `get_jobs` to `jobs` in `QiskitFunction`

Example:
```python
function: QiskitFunction = client.get("my-function")

function.jobs()
# [<Job | ....>, <Job | ....>]
```

akihikokuroda:
### Summary
fix #1479
fix #1463
fix #1470


akihikokuroda:
By importing the `Job` in the `Function`, it caused a circular dependency.  I had to workaround by putting the `import` inside of the method.  We probably need to refactor the client file structure.

client/qiskit_serverless/core/function.py
```diff
@@ -127,6 +127,10 @@ def get_jobs(self):
         Returns:
             Job ids : job executed this function
         """
+        from qiskit_serverless.core.job import (  # pylint: disable=import-outside-toplevel
```
Tansito, line 130: Let's put the import at the top, @akihikokuroda . As we are doing we similar approaches:

https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/job.py#L338

gateway/api/views.py
```diff
@@ -412,13 +413,8 @@ def get_jobs(
...
             else:
                 jobs = Job.objects.filter(program=program, author=request.user)
-            return Response(
-                list(
-                    jobs.values(
-                        "status", "result", "id", "created", "version", "arguments"
-                    )
-                )
-            )
+            serializer = JobSerializer(jobs, many=True)
```
Tansito, line 416: can we create a serializer that doesn't return logs and results? Thinking in return the minimum information as possible to reduce the size of the query

gateway/api/serializers.py
```diff
@@ -168,6 +168,7 @@ class JobSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Job
+        fields = "__all__"
```
Tansito, line 171: We should not return all `fields`, this value is overrided by `v1` serializers.

client/.pylintrc
```diff
@@ -81,7 +81,8 @@ disable=raw-checker-failed,
         suppressed-message,
         useless-suppression,
         deprecated-pragma,
-        use-symbolic-message-instead
+        use-symbolic-message-instead,
+        cyclic-import
```
Tansito, line 85: What did we need to add `cyclic-import` here?

Tansito (changes requested):

client/.pylintrc
akihikokuroda: Without it: lint error with:
```
lint: commands[1]> pylint -rn qiskit_serverless tests
************* Module tests.core.__init__
tests/core/__init__.py:1:0: R0401: Cyclic import (qiskit_serverless.core.function -> qiskit_serverless.core.job) (cyclic-import)
```

client/qiskit_serverless/core/function.py
akihikokuroda: I thought that putting it at the top causes the cyclic-import issue.  But I don't see it any longer.  I move it at the top.

akihikokuroda: Oh, it's failing after I move back at the top
```
________ ERROR collecting tests/serializers/test_program_serializers.py ________
ImportError while importing test module '/home/runner/work/qiskit-serverless/qiskit-serverless/client/tests/serializers/test_program_serializers.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.11.9/x64/lib/python3.11/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/serializers/test_program_serializers.py:23: in <module>
    from qiskit_serverless.serializers.program_serializers import (
qiskit_serverless/__init__.py:24: in <module>
    from .core import (
qiskit_serverless/core/__init__.py:54: in <module>
    from .client import (
qiskit_serverless/core/client.py:52: in <module>
    from qiskit_serverless.core.job import (
qiskit_serverless/core/job.py:66: in <module>
    from qiskit_serverless.core.function import QiskitFunction
qiskit_serverless/core/function.py:33: in <module>
    from qiskit_serverless.core.job import Job
E   ImportError: cannot import name 'Job' from partially initialized module 'qiskit_serverless.core.job' (most likely due to a circular import) (/home/runner/work/qiskit-serverless/qiskit-serverless/client/qiskit_serverless/core/job.py)
```

akihikokuroda: https://github.com/Qiskit/qiskit-serverless/pull/1480#issuecomment-2321486321

client/.pylintrc
akihikokuroda: https://github.com/Qiskit/qiskit-serverless/pull/1480#issuecomment-2321486321

gateway/api/serializers.py
```diff
@@ -168,6 +168,7 @@ class JobSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = Job
+        fields = ["id", "result", "status", "program", "created"]
```
Tansito, line 171: This is not needed because fields should be defined in its V1 Serializer:

https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/serializers.py#L156

To use that serializer you need to create an static method, Aki like here:
https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/views.py#L38

And you can override it here:
https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/views.py#L93

This is the way that we are managing the different serializers.

gateway/api/views.py
Tansito, line 416: This would need to be changed following my previous comment 👆

client/qiskit_serverless/core/function.py
Tansito: Thanks Aki, I will take a look into this 🤔 (cc/ to @IceKhan13)

Tansito (changes requested):

client/qiskit_serverless/core/function.py
Tansito: From what I was reading what you did here it's one of the solutions before refactoring the code so from my part this approach it's fine in case @IceKhan13 agrees too.

client/.pylintrc
Tansito: One question, Aki. If you do this local import: https://github.com/Qiskit/qiskit-serverless/pull/1480#discussion_r1742578677

does it continue being needed to add the `cyclic-import`?

akihikokuroda: Yes, it does.  Some tests fail like "tests/core/__init__.py:1:0: R0401: Cyclic import".  I could not make ignore it by the comment because it's an empty file.

IceKhan13 (approved):
It is ok for now. Reason why it is happening is `function` and `job` files are located in the same level in different files, therefore python is trying to initialize first file with reference to another file without being initialized. And you have circular references (one is referring to another in a cycle). 

Later on solution will be to move `Job` to higher level and refer job from `function` and `job_client` files.

client/qiskit_serverless/core/function.py
```diff
@@ -118,7 +118,7 @@ def run(self, **kwargs):
             config=config,
         )
 
-    def get_jobs(self):
```
Tansito, line 121: Something that we probably should do before merge is to add a deprecation warning here instead to remove it. Just to avoid a breaking change in the catalog until we release a new version.

Tansito: I can create an additional PR solving this.

Tansito (approved):

PR #1487, opened 2024-09-10
Additional fields for the UI

Tansito:
### Summary

This PR introduces `readable_name` and `readable_title` fields for the Provider and Program entities respectively to be used in the UI.


Tansito:
To be honest, I couldn't think in a better names for these fields. Any better proposal is welcome, team.

psschwei (approved):
assuming its easy enough to rename fields if we think of better names down the line

Issue #1236, opened 2024-02-27
Make `/static/` directory available on Kubernetes for Swagger docs


psschwei:
We added Swagger documentation in #1205 . 

The pages display fine locally (using docker compose or gunicorn directly), but when running in Kubernetes the page does not render properly as the swagger javascript files are not found (see https://github.com/Qiskit-Extensions/quantum-serverless/pull/1205#issuecomment-1965023194 and surround discussion for more details).

I think this is mainly an issue with how [Django serves static files](https://docs.djangoproject.com/en/5.0/howto/static-files/deployment/)... short version is in Kubernetes the `static/` directory isn't being served, which means none of the javascript loads, which means no swagger UI.

An appropriate solution would use one of the "production" methods in the above link to serve the `/static/` directory for our Django service.

psschwei:
https://github.com/Qiskit-Extensions/quantum-serverless/pull/1205#issuecomment-1965355457 shows a non-production way to make `/static/` available

david-alber:
@psschwei have you decided which of the production methods to serve the `/static/` directory is most suitable? 

If the issue is still up for grabs, I would like to give it a try :)

akihikokuroda:
@david-alber I've been working for another issue.  Please go a head.  Thanks!

psschwei:
Given the short-term plan is to move our docs over to IQP in  #1166 , I think we may want to get a little more clarity on that one (where docs will live, how they'll be served, etc.) before we spend too much time spinning up our own version of the docs

david-alber:
> Given the short-term plan is to move our docs over to IQP in #1166 , I think we may want to get a little more clarity on that one (where docs will live, how they'll be served, etc.) before we spend too much time spinning up our own version of the docs

Maybe the Qiskit Runtime API swagger docs can serve as inspiration: 
https://us-east.quantum-computing.cloud.ibm.com/openapi/

Tansito:
This was deployed in previous versions so closing the stale issue!

PR #1491, opened 2024-09-17
Change static files serving to whitenoise plugin

korgan00:
### Summary

Right now we have some limitations in the way we are serving static files because Django doesn't serve them by default. I use a plugin to implement better static files sharing.

### Details and comments

Using [whitenoise for django](https://whitenoise.readthedocs.io/en/stable/django.html) plugin to manage static files.

### DIFF
--- a/gateway/main/urls.py
+++ b/gateway/main/urls.py
@@ -16,7 +16,6 @@ Including another URLconf
 
 from django.conf import settings
 from django.conf.urls.static import static
-from django.views.static import serve
 from django.contrib import admin
 from django.urls import path, include, re_path
 from rest_framework import routers, permissions
--- a/gateway/main/urls.py
+++ b/gateway/main/urls.py
@@ -49,7 +49,6 @@ urlpatterns = [
     path("", include("django_prometheus.urls")),
     path("backoffice/", admin.site.urls),
     re_path(r"^api/v1/", include(("api.v1.urls", "api"), namespace="v1")),
-    re_path(r"^static/(?P<path>.*)$", serve, {"document_root": settings.STATIC_ROOT}),
 ]
 
 if settings.DEBUG:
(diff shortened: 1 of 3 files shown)


gateway/main/urls.py
```diff
@@ -49,7 +49,6 @@
     path("", include("django_prometheus.urls")),
     path("backoffice/", admin.site.urls),
     re_path(r"^api/v1/", include(("api.v1.urls", "api"), namespace="v1")),
-    re_path(r"^static/(?P<path>.*)$", serve, {"document_root": settings.STATIC_ROOT}),
```
Tansito, line 52: is this not need it anymore, Goyo?

psschwei:
Will this help with the issues in #1236 ?

korgan00:
I cannot be sure, but my intuition say that it should. Right now it works with docker.

psschwei:
Could we also check against Kubernetes? We should also test there since that's where we ultimately deploy. Docker can be useful for running code locally, but otherwise it can also get you into a bit of trouble down the line too...

You can use Kind to run a cluster locally.

psschwei:
ok, just checked and the docs DO show up on kubernetes :tada: 

image

psschwei (approved):

Tansito:
@psschwei could you double check too that the panel administrator loads correctly? 🙏

psschwei:
@Tansito that would be `/backoffice` right? (so locally, `http://localhost:8000/backoffice`)

Tansito:
Yes!

psschwei:
admin panel works

image

Tansito:
It worked with `DEBUG=0`, @psschwei ? Sorry for the double check, I remember having problems with that debug configuration.

psschwei:
No -- for my local setup I default to DEBUG=1, so still need to check with DEBUG=0...

psschwei (approved):
Confirmed that admin panel works with DEBUG=0

image

Tansito (approved):
Awesome! Thank you @psschwei for the double check, really appreciate it. And congrats @korgan00 for your first PR 🎉

PR #1493, opened 2024-09-18
allow provider to get the job by id

### LINKED ISSUE #1470: Provider can retrieve a job from its function

akihikokuroda:
### Summary
fix #1470 

### Details and comments
This PR changes allow the provider to retrieve the job in addition to the job author  in this method https://github.com/Qiskit/qiskit-ibm-catalog/blob/de2aaab8538e95942d1b7d98337c2c18093ec6f4/qiskit_ibm_catalog/catalog.py#L103

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -453,6 +453,18 @@ class JobViewSet(viewsets.GenericViewSet):
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.job.retrieve", context=ctx):
+            job = Job.objects.filter(pk=pk).first()
+            if job is None:
+                logger.warning("Job [%s] not found", pk)
+                return Response(status=404)
+            author = self.request.user
+            if job.program and job.program.provider:
+                provider_groups = job.program.provider.admin_groups.all()
+                author_groups = author.groups.all()
+                has_access = any(group in provider_groups for group in author_groups)
+                if has_access:
+                    serializer = self.get_serializer(job)
+                    return Response(serializer.data)
             instance = self.get_object()
             serializer = self.get_serializer(instance)
         return Response(serializer.data)
--- a/gateway/tests/api/test_job.py
+++ b/gateway/tests/api/test_job.py
@@ -83,6 +83,19 @@ class TestJobApi(APITestCase):
         self.assertEqual(jobs_response.data.get("status"), "SUCCEEDED")
         self.assertEqual(jobs_response.data.get("result"), '{"somekey":1}')
 
+    def test_job_provider_detail(self):
+        """Tests job detail authorized."""
+        user = models.User.objects.get(username="test_user_2")
+        self.client.force_authenticate(user=user)
+
+        jobs_response = self.client.get(
+            reverse("v1:jobs-detail", args=["1a7947f9-6ae8-4e3d-ac1e-e7d608deec86"]),
+            format="json",
+        )
+        self.assertEqual(jobs_response.status_code, status.HTTP_200_OK)
+        self.assertEqual(jobs_response.data.get("status"), "QUEUED")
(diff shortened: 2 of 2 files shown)


gateway/api/views.py
```diff
@@ -453,6 +453,18 @@ def retrieve(self, request, pk=None):  # pylint: disable=unused-argument
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.job.retrieve", context=ctx):
+            job = Job.objects.filter(pk=pk).first()
```
psschwei, line 456: If I'm understanding this right, if there was no job before we'd just return the serializer (since we didn't have a check), but now we'd be returning a 404... is that ok? (I don't know this area well enough to say, so figured I'd ask the question)

gateway/api/views.py
```diff
@@ -453,6 +453,18 @@ def retrieve(self, request, pk=None):  # pylint: disable=unused-argument
...
+            if job.program and job.program.provider:
+                provider_groups = job.program.provider.admin_groups.all()
+                author_groups = author.groups.all()
+                has_access = any(group in provider_groups for group in author_groups)
+                if has_access:
+                    serializer = self.get_serializer(job)
+                    return Response(serializer.data)
             instance = self.get_object()
             serializer = self.get_serializer(instance)
         return Response(serializer.data)
```
psschwei, line 470: In one case, we return a job and another we return an instance... in the same vein, is that ok?

akihikokuroda: The `instance` is the  job object.  So this returns a serialized job object.

gateway/api/views.py
akihikokuroda: It has been returning 404 when the job doesn't exist or the user is not authorized to access to the job.  Here is the test for it https://github.com/Qiskit/qiskit-serverless/blob/2dbdaf52389edb6b8d54b986f30be8fae844fb06/gateway/tests/api/test_job.py#L94

gateway/tests/api/test_job.py
```diff
@@ -83,6 +83,19 @@ def test_job_detail(self):
         self.assertEqual(jobs_response.data.get("status"), "SUCCEEDED")
         self.assertEqual(jobs_response.data.get("result"), '{"somekey":1}')
 
+    def test_job_provider_detail(self):
```
Tansito, line 86: @akihikokuroda , just to confirm, do we have a test that checks if the user is not a provider and either the author of the job returns a 404?

gateway/api/views.py
```diff
@@ -453,6 +453,18 @@ def retrieve(self, request, pk=None):  # pylint: disable=unused-argument
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.job.retrieve", context=ctx):
+            job = Job.objects.filter(pk=pk).first()
+            if job is None:
+                logger.warning("Job [%s] not found", pk)
+                return Response(status=404)
```
Tansito, line 459: I think I would add here a message as we have in other places like [here](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/views.py#L221). And I would try to use official status from django like [here](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/views.py#L222)

gateway/tests/api/test_job.py
akihikokuroda: Yes, it is `def test_not_authorized_job_detail(self)`

Tansito (approved):
LGTM @akihikokuroda , thank you!

PR #1496, opened 2024-09-18
Validate pk in job api url

akihikokuroda:
### Summary
Validate pk in the job api url to prevent SQL Injection

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -443,6 +443,10 @@ class JobViewSet(viewsets.GenericViewSet):
             pk,
             re.IGNORECASE,
         ):
+            logger.warning(
+                "Invalid job id format id[%s].",
+                pk,
+            )
             return None
         type_filter = self.request.query_params.get("filter")
         if type_filter:
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -438,7 +438,11 @@ class JobViewSet(viewsets.GenericViewSet):
 
     def get_queryset(self):
         pk = self.kwargs.get("pk")
-        if pk and not re.match("^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$", pk, re.IGNORECASE):
+        if pk and not re.match(
+            "^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$",
+            pk,
+            re.IGNORECASE,
+        ):
             return None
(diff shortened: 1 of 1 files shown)


gateway/api/views.py
```diff
@@ -436,6 +437,13 @@ def get_serializer_class(self):
         return self.serializer_class
 
     def get_queryset(self):
+        pk = self.kwargs.get("pk")
+        if pk and not re.match(
+            "^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$",
+            pk,
+            re.IGNORECASE,
+        ):
+            return None
```
psschwei, line 446: we should probably log a warning too if this happens

akihikokuroda: Added a log entry

Tansito: One question @akihikokuroda , is here where we need to check this? Because we are not using `pk` in the entire method 🤔

akihikokuroda: The issue is when the request is like this
```
https://qiskit-serverless-dev.quantum.ibm.com/api/v1/jobs/case+randomblob%281000000%29+when+not+null+then+1+else+1+end+/stop
```
It has a bad string in stead of the job id in the request url.  It causes a database request with the string.  The request that has the pk in the url always executes the `get_queryset` method before it calls the individual method.

Tansito: I need to say that I am not being able to replicate it

akihikokuroda: I haven't tried it by myself.  The issue says

> The query time is controllable using parameter value [case randomblob(1000000) when not null then 1 else 1 end ], which caused the request to take [334] milliseconds, parameter value [case randomblob(10000000) when not null then 1 else 1 end ], which caused the request to take [2,141] milliseconds, when the original unmodified query with value [id] took [824] milliseconds.

Tansito: My doubt here is that it surprises me that django orm doesn't manage this and we have similar end-points that passed the tests so I'm trying to figure out we are doing something that we should not do.

akihikokuroda: I've just tried to reproduce this issue but it didn't happen.  So I can not see what happened.  How about putting this fix in and see if the issue happens with it?

Tansito: Before merge it let me try to pass another scan Monday with security to confirm it and solve any doubt, Aki.

Tansito:
At the-end I think this PR doesn't apply, I think. Should we close it @akihikokuroda ?

akihikokuroda:
OK. I close

PR #1498, opened 2024-09-20
add catalog/by_title entry-point

akihikokuroda:
### Summary
add catalog/by_title entry-point

### DIFF
--- a/gateway/api/views.py
+++ b/gateway/api/views.py
@@ -858,6 +858,20 @@ class CatalogViewSet(viewsets.GenericViewSet):
 
         return Program.objects.filter(id=pk, instances=public_group).first()
 
+    def get_by_title_queryset(self, title, provider_name):
+        """
+        QuerySet to retrieve a specifc public programs in the catalog
+        """
+        public_group = Group.objects.filter(name=self.PUBLIC_GROUP_NAME).first()
+
+        if public_group is None:
+            logger.error("Public group [%s] does not exist.", self.PUBLIC_GROUP_NAME)
+            return []
+
+        return Program.objects.filter(
+            title=title, provider__name=provider_name, instances=public_group
+        ).first()
+
     def list(self, request):
         """List public programs in the catalog:"""
         tracer = trace.get_tracer("gateway.tracer")
@@ -877,6 +891,35 @@ class CatalogViewSet(viewsets.GenericViewSet):
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.catalog.retrieve", context=ctx):
             instance = self.get_retrieve_queryset(pk)
+            if instance is None:
+                return Response(
+                    {"message": "Qiskit Function not found."},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+
+            user = None
+            if request.user and request.user.is_authenticated:
+                user = request.user
+            serializer = self.get_serializer_retrieve_catalog(
+                instance, context={"user": user}
+            )
+        return Response(serializer.data)
+
+    @action(methods=["GET"], detail=False)
+    def by_title(self, request):
+        """Get a specific program in the catalog:"""
+        tracer = trace.get_tracer("gateway.tracer")
+        ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
+        with tracer.start_as_current_span("gateway.catalog.retrieve", context=ctx):
+            title = self.request.query_params.get("title")
+            provider_name = self.request.query_params.get("provider")
+            if not title or not provider:
+                return Response(
+                    {"message": "Qiskit Function not found."},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+
+            instance = self.get_by_title_queryset(title, provider_name)
             if instance is None:
                 return Response(
                     {"message": "Qiskit Function not found."},
--- a/gateway/tests/api/test_v1_catalog.py
+++ b/gateway/tests/api/test_v1_catalog.py
@@ -158,3 +158,83 @@ class TestCatalogApi(APITestCase):
         self.assertTrue(isinstance(public_function.get("additional_info"), dict))
         self.assertEqual(provider.get("name"), "default")
         self.assertEqual(provider.get("readable_name"), "Default")
+
+    def test_catalog_by_title_non_auth_user(self):
+        """Tests catalog retrieve non-authenticated."""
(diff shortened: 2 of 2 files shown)


gateway/api/views.py
```diff
@@ -890,3 +904,32 @@ def retrieve(self, request, pk=None):
                 instance, context={"user": user}
             )
         return Response(serializer.data)
+
+    @action(methods=["GET"], detail=False)
+    def by_title(self, request):
```
psschwei, line 909: Question: instead of having a separate endpoint to search by title, would it make sense to pass the search title as a parameter? something like `/api/v1/catalog?title=mytitle` ? Though it looks like we've already established a pattern with programs (`/programs` and `/programs/get_by_title`) so maybe its best to keep a consistent design...

Tansito: I think you have a point here @psschwei and I had the same doubt. The decision because I preferred this option was:
- `/api/v1/catalog?title=mytitle`: should return an `[]` of elements. So to maintain the coherence with the `retrieve` end-point and only return one function I preferred this approach.
- About `/programs/get_by_title` vs `by_title ` I think you are right and probably we should continue using `get_by_title` here too to maintain consistency at least in the `v1`.

Tansito: So basically this end-point should return only a `Function` not a `[Function]`. To have the same return as the `retrieve` by `id`.

Tansito: @akihikokuroda can we change `by_title ` by `get_by_title ` following @psschwei suggestion? 🙏

psschwei:
> Too many positional arguments (6/5)

Should we bump the default up a bit? seems like there's a few that are getting caught with 6/5, so maybe upping the limit is the right thing to do here?

akihikokuroda:
@psschwei The error didn't happen before.  I'm not sure if this check was not enabled or the default is lowered. @Tansito is trying to find what happened and making the right change.

Tansito:
@akihikokuroda it seems we found the culprit: 
- https://github.com/pylint-dev/pylint/releases/tag/v3.3.0
- https://github.com/pylint-dev/pylint/issues/9099

How the problem is not in our side feel free to restore [cf6e489](https://github.com/Qiskit/qiskit-serverless/pull/1498/commits/cf6e48946c77c9de43f03c6a2ec93552d05502ad) and I will open an issue Monday to monitor it and planify the refactorization if we consider it needed.

gateway/api/views.py
```diff
@@ -890,3 +904,32 @@ def retrieve(self, request, pk=None):
                 instance, context={"user": user}
             )
         return Response(serializer.data)
+
+    @action(methods=["GET"], detail=False)
```
Tansito, line 908: The output should be the same as the `retrieve` end-point so in this case detail must be True.

gateway/api/views.py
```diff
@@ -890,3 +904,32 @@ def retrieve(self, request, pk=None):
...
+        """Get a specific program in the catalog:"""
+        tracer = trace.get_tracer("gateway.tracer")
+        ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
+        with tracer.start_as_current_span("gateway.catalog.retrieve", context=ctx):
+            title = self.request.query_params.get("title")
+            provider_name = self.request.query_params.get("provider")
+            if not title or not provider:
+                return Response(
+                    {"message": "Qiskit Function not found."},
+                    status=status.HTTP_404_NOT_FOUND,
```
Tansito, line 919: Instead to return a 404 how we expected that they pass both params I would use a 400 BAD_REQUEST and clarify that both filters are needed in the message.

gateway/api/views.py
```diff
@@ -858,6 +858,20 @@ def get_retrieve_queryset(self, pk):
...
 
+    def get_by_title_queryset(self, title, provider_name):
+        """
+        QuerySet to retrieve a specifc public programs in the catalog
+        """
+        public_group = Group.objects.filter(name=self.PUBLIC_GROUP_NAME).first()
+
+        if public_group is None:
+            logger.error("Public group [%s] does not exist.", self.PUBLIC_GROUP_NAME)
+            return []
```
Tansito, line 869: Probably a copy paste error but I understand that in this case it should return a None instead of an array, shouldn't it?

Tansito (changes requested):

gateway/api/views.py
akihikokuroda: The doesn't work with a single object so I think the `detail` must be `False`. I verified that the exact responses are returned as the retrieve method.

gateway/api/views.py
akihikokuroda: These are exact same lines as https://github.com/Qiskit/qiskit-serverless/blob/e24c2a2543bf12b01588f1633b2f549a4285de50/gateway/api/views.py#L857
Either `[]` or `None` works in both methods.  I guess that they must be same.

gateway/api/views.py
Tansito: And if we don't have `detail` at all? Like, only the `GET` parameter.

gateway/api/views.py
Tansito: Yeah, I think this was a mistake from my side:
- `get_queryset` is fine
- `get_retrieve_queryset` I think it should return `None`

So `get_by_title_queryset` should return `None` too. Probably I copy-paste from `get_queryset` and I forgot to change it, yep!

gateway/api/views.py
akihikokuroda: The `detail` is a required parameter for the `action` decorator.  It must be there.

Tansito (changes requested):
As last thing Aki, remember to add the end-point to the V1 interfaces like in this example:

https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/views.py#L116

gateway/api/authentication.py
```diff
@@ -138,5 +138,23 @@ def authenticate(self, request):
                         )
                         provider.admin_groups.add(group)
                         logger.info("New provider created")
+                        return user, CustomToken(token.encode())
+
+                if token == settings.SETTINGS_AUTH_MOCK_TOKEN + "1":
```
Tansito, line 143: Why are we adding this, @akihikokuroda ?

akihikokuroda: Oh, it's my mistake.  I take out changes in this file.

Tansito (approved):
Thank you Aki! 🙏

PR #1500, opened 2024-09-24
Create new cluster for each job execution

psschwei:
### Summary

Create a new RayCluster each time we execute a job.

### Details and comments

This is a prerequisite for passing job arguments on disk as we'll need to mount them in a volume on the cluster, and the underlying location will change on each job.

It will also be better for resource management, as each job will get exactly the resources requested. Previously, if a user submitted two jobs they would run in the same cluster and sharing resources, so the resource requests wouldn't necessarily be honored exactly. 

Plus, this'll also be the default behavior after migrating to RayJobs (#1337), so its not something we weren't planning on doing eventually anyway.

### DIFF
--- a/gateway/api/schedule.py
+++ b/gateway/api/schedule.py
@@ -26,8 +26,7 @@ logger = logging.getLogger("commands")
 def execute_job(job: Job) -> Job:
     """Executes program.
 
-    1. check if cluster exists
-       1.1 if not: create cluster
+    1. create cluster
     2. connect to cluster
     3. run a job
     4. set status to pending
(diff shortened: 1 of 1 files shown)


psschwei:
We don't necessarily need this now, so will leave it up to reviewers on whether to merge this or close it.

psschwei:
Actually, the docker test failure here seems legit, and since I think this falls into the "nice to have" bucket right now, I don't think it makes sense to spend the time to iron that issue out right now.

Tansito:
The good thing at least is that right now in our cluster we are not allowing that:
https://github.com/Qiskit/qiskit-serverless/blob/main/charts/qiskit-serverless/charts/gateway/values.yaml#L45

We are using one here so good. Anyway I will open an issue around this, thanks Paul! 🙏

PR #1503, opened 2024-09-27
Remove simplejwt dependency

Tansito:
### Summary

This PR tries to remove SimpleJWT dependency that I think we are not using it right now.


Tansito:
All the pieces for the authentication are needed.

PR #1506, opened 2024-10-03
Use multistage build for gateway

psschwei:


gateway/Dockerfile
```diff
@@ -1,37 +1,61 @@
...
+COPY --from=BUILD ${MICRO_IMAGE_DIR}/ .
+
+# create symlinks for python
+RUN ln -s /usr/bin/python3.11 /usr/bin/python
+
+# Create project dir
 WORKDIR /usr/src/app
 
 # set environment variables
 ENV PYTHONDONTWRITEBYTECODE 1
```
Tansito, line 39: I saw that in your other PR you are modifying these lines:
https://github.com/Qiskit/qiskit-serverless/pull/1515/files#diff-577acb53f5b01ca2f90ce5ca71e940346262e7ad8d8ddbdbdb0ee728f090c372R12

How do you see to add that modification in this PR, @psschwei ?

Tansito (approved):
@akihikokuroda could you take a look to this PR? I would like a double check apart from my side around this change 🙏 

Regardless of my only comment I see everything fine but I would like another review from Aki if it's possible.

gateway/Dockerfile
psschwei: that's just to turn off a docker warning about how to assign envvar values. I can move it here if you want, but as long as it is in one PR it'll get into main

gateway/Dockerfile
```diff
@@ -1,37 +1,61 @@
...
+RUN dnf install --installroot ${MICRO_IMAGE_DIR} --nodocs -y \
+    python3.11-3.11.7 \
+    python3.11-devel-3.11.7 \
+    libstdc++ &&\
+    dnf upgrade --installroot ${MICRO_IMAGE_DIR} --nodocs -y && \
+    dnf clean all --installroot ${MICRO_IMAGE_DIR}
+
+# APP image from `scratch` which will be the final image
+# and remaining application requirements will be installed
+FROM scratch AS APP
```
akihikokuroda, line 27: I wonder why this APP image is necessary.  What is the difference between copying back the ${MICRO_IMAGE_DIR}/ into the BASE image and the APP image?

gateway/Dockerfile
```diff
@@ -1,37 +1,61 @@
...
 COPY gateway .
 RUN chown -R 1000:100 /usr/src/app &&\
     mkdir /usr/src/app/media && chown 1000:100 /usr/src/app/media
 
-# Need versions of pip/setuptools more recent than provided by UBI image
-RUN python3.11 -m ensurepip --upgrade
-
-# hadolint ignore=DL3013
-RUN pip install --upgrade --no-cache-dir pip>=24.2 &&\
-    pip install --upgrade --no-cache-dir setuptools>=72.1.0
```
akihikokuroda, line 29: Is the setuptools unnecessary after all or is the right version in ubi9-micro:9.4-15?

gateway/Dockerfile
psschwei: The BASE image still has things like a package manager and a shell that the gateway really doesn't need, whereas scratch / APP is empty except for what we explicitly add to it.  So that does two things: (a) it reduces the number of components that we have to manage for CVEs, and (b) it reduces the threat vectors for that particular container.

akihikokuroda: I may be wrong but I read 
```
COPY --from=BASE / ${MICRO_IMAGE_DIR}
```
copies everything in the BASE to ${MICRO_IMAGE_DIR} directory and
```
COPY --from=BUILD ${MICRO_IMAGE_DIR}/ .
```
copies everything in the BASE + additional pieces to the APP
So everything in the BASE is copied to APP.  Where am I misunderstanding?

psschwei: You could be right... I'm just copying from what [another team already did](https://github.ibm.com/IBM-Q-Software/qiskit-code-assistant-service/blob/7fd8d83b161c5c67d15c3159f41e2be51638794e/Dockerfile) and didn't look too closely at what they were doing, I just assumed it worked since it had already gone through review.

gateway/Dockerfile
psschwei: I was assuming that the scratch image doesn't have setuptools so it shouldn't need to be updated. but maybe it will be?

akihikokuroda: I'm not sure.  Scan will tell us :-)

akihikokuroda (approved):
I don't see any major issues.

gateway/Dockerfile
psschwei: all hail general scan :laughing:

PR #1508, opened 2024-10-07
None description shouldnt be updated

korgan00:


Tansito:
I think you need to run `tox -eblack` to fix the formatting, @korgan00

gateway/api/serializers.py
```diff
@@ -117,7 +117,11 @@ def update(self, instance, validated_data):
         instance.artifact = validated_data.get("artifact")
         instance.author = validated_data.get("author")
         instance.image = validated_data.get("image")
-        instance.description = validated_data.get("description")
+
+        description = validated_data.get("description")
+        if description is not None:
```
akihikokuroda, line 122: Nit: in Python
`if description:` is same as `if description is not None:` in most of the cases and it's simpler

akihikokuroda (approved):
There is one nit comment.  Other than that, LGTM!

Tansito (approved):
I agree with @akihikokuroda comment, apart from that LGTM too 👍

gateway/api/serializers.py
Tansito: thanks for the fast test, @korgan00 , understood!

PR #1513, opened 2024-10-10
Refactor client and fix tests

korgan00:
### Summary

Objectives:

- [x] Remove `BaseJobClient` and migrate it to `BaseClient`
- [x] Remove `RayJobClient` and migrate it to `RayClient`
- [x] Remove `LocalJobClient` and migrate it to `LocalClient`
- [x] Remove `GatewayJobClient` and migrate it to `ServerlessClient`
- [x] From `Job` rename the variable to use the `client` instead of `job_client`.
- [x] From `QiskitFunctions` rename the variable to use the `client` instead of `job_client`.

### Details and comments
Additionally:
 - Moved the `BaseClient` implementations (Ray, Local and Serverless) to their own files.
 - `BaseClient` now is officially an abstract class (`ABS` and `@abstractmethod`).
 - Fixed and barely improved tests.
 - Removed `ComputedResource` because it wasn't used correctly.


client/tests/utils.py
```diff
@@ -1,17 +1,19 @@
...
+def wait_for_ray_ready(connection_url: str, timeout: int = 60):
+    client = None
     must_finish = time.time() + timeout
-    while time.time() < must_finish:
-        if serverless.job_client() is not None:
-            break
-        time.sleep(1)
+    while time.time() < must_finish and not client:
+        try:
+            client = JobSubmissionClient(connection_url)
```
korgan00, line 14: Trying to know if Ray server started. I didn't found a better way but I'm sure it is. For the moment, this works.

client/tests/core/test_job.py
```diff
@@ -40,9 +46,10 @@ def test_save_result(self):
             )
             self.assertTrue(result)
 
+    @patch("requests.get", Mock(return_value=ResponseMock()))
```
korgan00, line 49: prevent token validation to fail.

korgan00:
I have to review the documentation.

Tansito:
I converted it to draft @korgan00 just to let the team knows that we want a review on this but we need to fix tests yet

client/tests/core/test_pattern.py
```diff
@@ -49,7 +48,7 @@ def test_program():
         assert job.in_terminal_state()
         assert job.status() == "DONE"
 
-        recovered_job = serverless.get_job_by_id(job.job_id)
+        recovered_job = serverless.get_job(job.job_id)
```
Tansito, line 51: I would agree to use the new method in the test but I would like to maintain backward compatibility for the old method too, @korgan00 . As mentioned above.

client/qiskit_serverless/core/client.py
```diff
@@ -182,70 +66,39 @@ def __init__(  # pylint:  disable=too-many-positional-arguments
...
         self.token = token
-        self.compute_resource = compute_resource
-        if available_compute_resources is None:
-            if compute_resource is not None:
-                available_compute_resources = [compute_resource]
-            else:
-                available_compute_resources = []
-        self.available_compute_resources = available_compute_resources
 
+    # ?¿¿?¿?
```
Tansito, line 74: > ?¿¿?¿?

Any question here? Hahaha

client/qiskit_serverless/core/client.py
```diff
@@ -182,70 +66,39 @@ def __init__(  # pylint:  disable=too-many-positional-arguments
...
 
         Returns:
             list of jobs.
         """
-        raise NotImplementedError
+        pass
 
-    def get_job_by_id(self, job_id: str) -> Optional[Job]:
+    @abstractmethod
+    def get_job(self, job_id: str) -> Optional[Job]:
```
Tansito, line 101: Here we are introducing a breaking change. What we usually do is maintain `get_job_by_id` with a `DeprecateWarning` and an internal call to `get_job`.

client/qiskit_serverless/core/client.py
```diff
@@ -27,131 +27,20 @@
...
-from qiskit_serverless.exception import QiskitServerlessException
 from qiskit_serverless.utils import JsonSerializable
-from qiskit_serverless.utils.json import safe_json_request
 from qiskit_serverless.visualizaiton import Widget
 
 TIMEOUT = 30
 
 
-@dataclass
-class ComputeResource:
```
Tansito, line 71: Here we are removing the entire `ComputeResource` reference. This change was not annotated in the PR, can you add this too, @korgan00 ?

Also, from my part it's fine because I think we are not using it but @IceKhan13 WDYT?

client/qiskit_serverless/core/clients/LocalClient.py
```diff
@@ -0,0 +1,230 @@
...
+    Job,
+    Configuration,
+)
+from qiskit_serverless.core.function import QiskitFunction
+from qiskit_serverless.exception import QiskitServerlessException
+from qiskit_serverless.serializers.program_serializers import (
+    QiskitObjectsEncoder,
+)
+
+TIMEOUT = 30
```
Tansito, line 58: Correct me if I'm wrong but this `TIMEOUT` is not being used (I didn't find any reference at least) and, in any case, I think we should use `REQUESTS_TIMEOUT` in `constants` probably.

client/qiskit_serverless/core/clients/LocalClient.py
```diff
@@ -0,0 +1,230 @@
...
+            QiskitFunction(
+                program.get("title"),
+                provider=program.get("provider", None),
+                raw_data=program,
+                client=self,
+            )
+            for program in self._patterns
+        ]
+
+    def get_program(
```
Tansito, line 191: I think this is a new API. My comment will be similar to my previous one. I'm fine with `get_program` but we need to maintain compatibility by now and support the old `get` method too, @korgan00 .

client/qiskit_serverless/core/clients/RayClient.py
```diff
@@ -0,0 +1,192 @@
...
+)
+from qiskit_serverless.core.job import (
+    Job,
+)
+from qiskit_serverless.core.function import QiskitFunction
+from qiskit_serverless.serializers.program_serializers import (
+    QiskitObjectsEncoder,
+)
+
+TIMEOUT = 30
```
Tansito, line 49: Similar feedback from my previous comment about `TIMEOUT`.

client/qiskit_serverless/core/clients/ServerlessClient.py
```diff
@@ -0,0 +1,583 @@
...
+from qiskit_serverless.core.function import QiskitFunction
+from qiskit_serverless.exception import QiskitServerlessException
+from qiskit_serverless.utils.json import safe_json_request
+from qiskit_serverless.utils.formatting import format_provider_name_and_title
+from qiskit_serverless.serializers.program_serializers import (
+    QiskitObjectsEncoder,
+    QiskitObjectsDecoder,
+)
+
+TIMEOUT = 30
```
Tansito, line 66: Similar feedback from my previous comment about `TIMEOUT`.

Tansito (changes requested):
I left several comments but really good job with this refactor @korgan00 , congrats. I'm very happy with the result.

Tansito:
Also @korgan00 , most of the problems in the tests are related with the changes in the API. I can help you tomorrow looking at them but with some of the changes that I'm requesting you we should be able to solve some of them.

client/qiskit_serverless/core/clients/LocalClient.py
korgan00: Sound like a bad copy and paste :)

korgan00:
Per offline conversations we also talked about changing `get_program` with `get_function` to make it more coherent with QiskitFunctions.

korgan00:
Since file management is a Serveless feature that is not available in Ray or Local, should we remove it from the Base? This way we will reduce the `NotImplementedError` exceptions.

IceKhan13:
> Since file management is a Serveless feature that is not available in Ray or Local, should we remove it from the Base? This way we will reduce the NotImplementedError exceptions.

Makes sense. Let's remove it from base.

---

Overall I like this PR. Much cleaner structure :) 

Request review once you are done with it. Thank you!

Tansito:
Moving PR to review 🎉

.github/workflows/docker-build.yaml
```diff
@@ -29,6 +27,7 @@ jobs:
         shell: bash
         run: |
           cd tests/basic
+          rm ./06_function.py
```
Tansito, line 30: These changes were necessary. Basically currently `06_function` doesn't work in this test with the current docker-compose configuration and it was generating problems in the execution of the test.

client/tests/core/test_pattern.py
Tansito: @korgan00 correct me if I'm wrong but this problem is not fixed, isn't it? I couldn't find where we are supporting the old reference to `get_job_by_id `

client/qiskit_serverless/core/clients/LocalClient.py
Tansito: Similar to my feedback about `get_job_by_id `, I don't think this problem is solved either. We can take a look tomorrow together if there is any doubt 👍

Tansito (changes requested):
I opened a couple of closed conversations about breaking changes in `get_job_by_id ` and `functions.get` methods. We can take a look tomorrow @korgan00  and clarify them. Maybe I missed them somewhere.

client/qiskit_serverless/core/client.py
```diff
@@ -254,16 +107,25 @@ def get_job_by_id(self, job_id: str) -> Optional[Job]:
...
+        Returns:
+            Job instance
+        """
+        warnings.warn(
+            "`get_job_by_id` method has been deprecated. "
+            "And will be removed in future releases. "
+            "Please, use `get_job` instead.",
+            DeprecationWarning,
+        )
+        return self.get_job(job_id)
```
korgan00, line 126: @Tansito this is the deprecated reference that you asked for.

client/qiskit_serverless/core/client.py
```diff
@@ -289,471 +151,73 @@ def run(
...
-
-    def delete_compute_resource(self, resource) -> int:
-        raise NotImplementedError(
-            "IBMServerlessClient does not support resources api yet."
+        """Returns program based on parameters."""
+        warnings.warn(
+            "`get` method has been deprecated. "
+            "And will be removed in future releases. "
+            "Please, use `get_function` instead.",
+            DeprecationWarning,
```
korgan00, line 203: @Tansito this is the other deprecated method

client/tests/core/test_pattern.py
korgan00: I sent a couple of messages showing the place where the deprecated functions are.
Additionally, the `test/basic/05_retrieving_past_resuilts.py` uses `get_job_by_id`, that is interesting because it is tested.

korgan00:
Since the client and the job files has huge changes, they are not shown by default in the changes preview.You have to load them intentionally. I put the deprecation redirections directly in the base file because is better to maintain than spread and repeat the code in every client implementation.

Tansito (approved):
Solved my questions and concerns with @korgan00 on a call. From my part everything looks good but @IceKhan13 if you can double check I would appreciate it 👍

client/qiskit_serverless/core/files.py
```diff
@@ -120,7 +120,7 @@ def delete(self, file: str, provider: Optional[str] = None) -> Optional[str]:
         """Deletes file uploaded or produced by the programs,"""
         tracer = trace.get_tracer("client.tracer")
         with tracer.start_as_current_span("files.delete"):
-            response_data = safe_json_request(
+            response_data = safe_json_request_as_dict(
```
IceKhan13, line 123: I think one day we will need to remove this and just use requests and do proper error handling with correct excetions, etc. 

Definitely not the highest priority. Maybe we can create good first issue or something to handle this in the future.

IceKhan13 (approved):
Waaaay cleaner now! 

Thank you @korgan00!

client/qiskit_serverless/core/files.py
korgan00: Yeah, i just tried to fix some types. There are too many `Any`s everywhere.

Tansito: Agree!

PR #1514, opened 2024-10-10
validate uuid id/pk

akihikokuroda:
### Summary
Validate UUID used as the id/pk for the model instance

### DIFF
--- a/gateway/api/utils.py
+++ b/gateway/api/utils.py
@@ -1,5 +1,6 @@
 """Utilities."""
 
+# pylint: disable=duplicate-code
 import base64
 from collections import OrderedDict
 import inspect
--- a/gateway/api/views/catalog.py
+++ b/gateway/api/views/catalog.py
@@ -5,6 +5,7 @@ Version views inherit from the different views.
 """
 import logging
 import os
+import re
 
 from django.conf import settings
 from django.contrib.auth.models import Group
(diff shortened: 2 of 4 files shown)


Tansito (changes requested):
Thanks for the PR, @akihikokuroda . You don't need to validate perse the UUID format. This is something that django is doing in its ORM. The problem is that we are not capturing the exception for that. The idea is to add a `try/except` in the queries where we are using the UUID.

akihikokuroda:
These changes check the uuid before django validates it.  So it stops raising exceptions and returns 404 for the requests.

Tansito:
> These changes check the uuid before django validates it. So it stops raising exceptions and returns 404 for the requests.

I know, I know. But we can rely on Django validation perfectly, we just need to capture the exception that it raises. We don't need to create our own validator for that.

akihikokuroda:
I believe that the django validation happens when it saves to or retrieves from the database.  It's much cleaner and safer if we validate it at the entry of the apis.

Tansito:
My proposal @akihikokuroda :
- We need to continue handling the exception in the queries anyway because we are not doing that right now. As the issue was proposing.
- I don't see bad to add additional checks but not through custom checks. We can use [UUID](https://docs.python.org/3/library/uuid.html) utility from python.
- Move the logic to utils and reuse it in those specific points.

akihikokuroda:
> handling the exception in the queries

Are the **queries** database query? Where do you propose to put the `try` clause?

The UUID utility seems for generating an UUID.  I don't see the method for validating it.

Tansito:
> Are the queries database query? Where do you propose to put the try clause?

It happens in every `find`, `filter`, etc that uses the `id` to search for something, example [here](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/views/catalog.py#L82).

You just need to encapsulate that call in a `try/except`, and in this case, return `None` is fine for example.

> The UUID utility seems for generating an UUID. I don't see the method for validating it.

If you try to create an UUID with a value that is not a valid UUID it will raise an exception that you can capture similar to what you were doing with the pattern.

Tansito:
Let's close this PR by now and we can reschedule it in the next weekly

PR #1515, opened 2024-10-10
Add support for running functions on GPUs

psschwei:
### Summary

Provides support for running a select group of functions on GPU nodes. Functions will need to use `distribute_task` to run workloads on the GPU (ex: `@distribute_task(target={"gpu": 1})`.

### Details and comments

To set the list of providers allowed to run functions on GPUs, edit the `gateway/api/v1/gpu-jobs.json` file. A sample config, allowing the `mockprovider` provider to run on GPUs would look like this:

```yaml
{ 
    "gpu-functions": { 
        "mockprovider": [ "my-first-pattern" ]
     }
}
```

The method for getting the list of GPU providers is the same as we previously used for allowlisting dependencies.

In a future iteration, this could also be limited to specific functions by checking the the program title matches an element in the list (in this example, that would be `my-first-pattern`).

Jobs and ComputeResources using GPUs are tracked via a new field in each model.

There is also a separate queue for provisioning GPU clusters. The current default is only 1 GPU job at a time, but that can be configured by setting a value for the `LIMITS_GPU_CLUSTERS` on the scheduler deployment.

CPU-only jobs are only allowed to run on CPU nodes, and GPU jobs can only run on GPU nodes. This is enforced by configuring a nodeSelector on the rayclustertemplate, the values of which can be configured in the Helm chart.


psschwei:
> Is it using the distribute_task annotation or the entire function will run in GPU?

There's two parts here:

As of now, to run on the GPU, functions will need to use `distribute_task` to let Ray know they'll need  GPU resources. If we didn't want to use distribute_task, we could alternatively  set `entrypoint_num_gpus` when the job is [submitted](https://github.com/Qiskit/qiskit-serverless/blob/cf655f26d93c7447c567d7b862872b553cdcc94d/gateway/api/ray.py#L149). 

But either way we go, each job will block a GPU until it completes. Nothing we can really do about that -- as serverless is currently written, we don't have the ability to release resources (CPU or GPU) in the middle of a job.

pacomf:
distribute_task annotation is ok.

Ohh so even if the method where the distribute_task annotation finished the resource is not released to be available again for other function?

pacomf:
Is it possible assign a classical CPU node for the main functions and just assign the GPU nodes per each distribute_task annotation call, so the GPU resource is unblocked when the annotated method finishes? (like Ray is doing :) )

psschwei:
> Is it possible assign a classical CPU node for the main functions and just assign the GPU nodes per each distribute_task annotation call, so the GPU resource is unblocked when the annotated method finishes? (like Ray allows :) )

No, not without doing a massive overhaul of serverless.

Tansito:
I'm going to try to give a review today @psschwei , thank you! 👍

gateway/api/v1/gpu-jobs.json
```diff
@@ -0,0 +1,3 @@
+{
```
Tansito, line 1: Just to double check: @psschwei . This approach would be very similar to this, right? 

https://github.com/Qiskit/qiskit-serverless/pull/1443#discussion_r1711688440

gateway/main/settings.py
```diff
@@ -367,12 +368,26 @@
...
+    "ibm-cloud.kubernetes.io/worker-pool-name: gpu-workers",
+)
+
 PROGRAM_TIMEOUT = int(os.environ.get("PROGRAM_TIMEOUT", "14"))
 
 GATEWAY_ALLOWLIST_CONFIG = str(
     os.environ.get("GATEWAY_ALLOWLIST_CONFIG", "api/v1/allowlist.json")
 )
 
+GATEWAY_GPU_JOBS_CONFIG = str(
```
Tansito, line 387: Similar to my previous comment, should we add this environment variable to the helm configuration?

Tansito (commented):
In general it looks really good. I just left a couple of comments related with the implementation, @psschwei 👍

gateway/main/settings.py
```diff
@@ -324,6 +324,7 @@
 # resources limitations
 LIMITS_JOBS_PER_USER = int(os.environ.get("LIMITS_JOBS_PER_USER", "2"))
 LIMITS_MAX_CLUSTERS = int(os.environ.get("LIMITS_MAX_CLUSTERS", "6"))
+LIMITS_GPU_CLUSTERS = int(os.environ.get("LIMITS_MAX_GPU_CLUSTERS", "1"))
```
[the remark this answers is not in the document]
psschwei: Yes, these could be added to Helm eventually. I didn't do it now since this is still a POC that we cause use the default value for.

gateway/api/ray.py
```diff
@@ -249,6 +249,14 @@ def create_ray_cluster(  # pylint: disable=too-many-branches
             job_config.auto_scaling = settings.RAY_CLUSTER_WORKER_AUTO_SCALING
         node_image = settings.RAY_NODE_IMAGE
 
+        # cpu job settings
+        node_selector_label = settings.RAY_CLUSTER_CPU_NODE_SELECTOR_LABEL
+        gpu_request = 0
+        # if gpu job, use gpu nodes and resources
+        if job.gpu:
+            node_selector_label = settings.RAY_CLUSTER_GPU_NODE_SELECTOR_LABEL
+            gpu_request = 1
```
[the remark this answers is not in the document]
psschwei: No, this is the number of GPUs to request for the job similar to CPU or RAM. We only have one per node currently, so this value should be constant. If that changes in the future, then we could make this configurable.

gateway/api/v1/gpu-jobs.json
psschwei: Yes

gateway/main/settings.py
psschwei: I'd wait until we're ready to make it a configmap... for now, easier to manage the file directly  in the internal repo

Tansito: Ok, you are planning to do the tests in the internal repo. Just to double check that I would not like to expose this info in the public repo.

Tansito (commented):
As a POC it looks good to me to start testing it in staging. It remains some things to close for a final version as I was pointing out but if they are closed before the deadline, perfect.

I can approve it @psschwei as soon as you can confirm me that the configuration for the `gpu-jobs.json` can be done in the private repository and it's not going to happen in the public one 👍 

@IceKhan13 if you can take a look in case I'm forgetting something I would appreciate the double check.

psschwei:
> confirm that the configuration for the gpu-jobs.json can be done in the private repository

Yes, that config will be done in the internal repo

Tansito (approved):
@IceKhan13 will be OOO some days, tests are passing and I would not like to delay more this feature to enable you to start testing it in staging, @psschwei so I'm going to approve it and we can add improvements later 👍

PR #1517, opened 2024-10-15
Function and jobs refactor - resolve circular dependencies

korgan00:


client/qiskit_serverless/core/client.py
```diff
@@ -140,76 +139,27 @@ def get_jobs(self, **kwargs) -> List[Job]:
...
-        """Return results."""
-
-    @abstractmethod
-    def logs(self, job_id: str) -> str:
-        """Return logs."""
-
-    @abstractmethod
-    def filtered_logs(self, job_id: str, **kwargs) -> str:
-        """Return filtered logs."""
-
```
korgan00, line 191: This methods are defined in new RunService and JobClient abstract classes.

client/qiskit_serverless/core/job.py
```diff
@@ -76,13 +77,44 @@ class Configuration:  # pylint: disable=too-many-instance-attributes
     auto_scaling: Optional[bool] = False
 
 
+class JobClient(ABC):
```
Tansito, line 80: As we commented through a call `JobClient` I would like to rename it to `JobService` to match it with the current `RunService`.

Apart from that I saw everything good, really good job here @korgan00

Tansito:
@IceKhan13 if you can take a look to this PR too and give some feedback to @korgan00 it would be nice!

client/qiskit_serverless/core/function.py
```diff
@@ -91,6 +96,70 @@ def __str__(self):
...
+        entrypoint: is a script that will be executed as a job
+            ex: job.py
+        env_vars: env vars
+        dependencies: list of python dependencies to execute a program
+        working_dir: directory where entrypoint file is located (max size 50MB)
+        description: description of a program
+        version: version of a program
+    """
+
+    _run_service: RunService = None
```
paaragon, line 146:
```suggestion
    _run_service: Optional[RunService] = None
```

paaragon (approved):
Very good job @korgan00 😄 

I have a question regarding `function.py` and `job.py`.

- `job.py` contains three classes: `Configuration`, `JobService` and `Job`
- `function.py` contains another three classes: `QiskitFunction`, `RunService` and `RunnableQiskitFunction`.

Would it make sense to have some of these classes in a separated file? I was thinking that maybe `RunService` and `JobService` could be in another file

client/qiskit_serverless/core/function.py
korgan00: In this case i think it shouldnt be optional because, to be runnable it needs an associated run service. If run service will be none, then it is just a `QiskitFunction`

korgan00:
> Very good job @korgan00 😄
> 
> I have a question regarding `function.py` and `job.py`.
> 
> * `job.py` contains three classes: `Configuration`, `JobService` and `Job`
> * `function.py` contains another three classes: `QiskitFunction`, `RunService` and `RunnableQiskitFunction`.
> 
> Would it make sense to have some of these classes in a separated file? I was thinking that maybe `RunService` and `JobService` could be in another file

Since they are very coupled to that files, I added it to the same file, but im ok with splitting it. What do you think @IceKhan13 , @Tansito ?

IceKhan13 (approved):
woohoo, circular deps gone! 
LGTM
THank you!

Tansito:
I think by now having them in the same file is good enough, as we commented we are not expecting to re use them in other places in the near future, @korgan00 👍

Tansito (approved):

PR #1519, opened 2024-10-16
Feat/increate request timeout for upload

paaragon:
### Summary

Increase request timeout for streaming requests when uploading file


Tansito (changes requested):
We should add this change into the `download` method too:

https://github.com/Qiskit/qiskit-serverless/blob/f5dc4752911a77a9d203b74abebce643c7856de3/client/qiskit_serverless/core/files.py#L70

Tansito:
To solve the issue in the PR btw @paaragon you just need to pass the black command

PR #1521, opened 2024-10-16
feat: improve instances many to many in backoffice

paaragon:
### Summary

This PR improves the "Instances" field in the form located at `/backoffice/api/program/add/`.

It uses the `filter_horizontal` described [here](https://stackoverflow.com/questions/5385933/a-better-django-admin-manytomany-field-widget)

### Screenshot

image


Tansito (approved):
It works like a charm @paaragon , thank you!

PR #1524, opened 2024-10-22
chore(deps): bump pin on numpy requirement

robertodr:
### Summary
NumPy is only needed to run tests and examples. I've moved it to `requirements-dev.txt` as otherwise downstream packages cannot update to NumPy >=2.0.

### DIFF
--- a/client/requirements-dev.txt
+++ b/client/requirements-dev.txt
@@ -18,4 +18,4 @@ requests-mock>=1.11.0
 testcontainers>=4.7.2
 tox>=4.0.0
 setuptools>=70.0.0
-numpy>=1.26.4, <2
+numpy>=1.17,<3
--- a/client/requirements-dev.txt
+++ b/client/requirements-dev.txt
@@ -18,4 +18,3 @@ requests-mock>=1.11.0
 testcontainers>=4.7.2
 tox>=4.0.0
 setuptools>=70.0.0
-numpy>=1.17,<3
--- a/client/requirements-dev.txt
+++ b/client/requirements-dev.txt
@@ -18,3 +18,4 @@ requests-mock>=1.11.0
 testcontainers>=4.7.2
 tox>=4.0.0
 setuptools>=70.0.0
+numpy>=1.26.4, <2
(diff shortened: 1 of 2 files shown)


paaragon (approved):

Tansito (commented):
First of all thank you so much for your contribution @robertodr 

`client` is our library used in our `ray-node` image. Taking this into account `numpy` is a library very used in our entire stack so I don't think we can remove it so easily because it could be needed to run other functions that count with it.

My proposal would be to follow a similar requirement as [Qiskit](https://github.com/Qiskit/qiskit/blob/main/requirements.txt#L2):
```
numpy>=1.17,<3
```

So v2 would be supported.

robertodr:
@Tansito I would be happy with lifting the version pinning as you suggest. However, keeping a package you don't directly depend on in your `install_requires`, just because other dependents _might_ need it, is, as far as I know, bad practice. The dependents should explicitly list that package in their `install_requires`.

Tansito:
Well, our use case it's a bit special due to the usage that we are doing with our docker image. But thinking a bit more about it I think you are right here and it's better to move it to our -dev dependencies. Could you update your change with the same restriction as qiskit as we were discussing above?

Apart from that I think everything is fine 👍

Tansito:
It seems there is an error in one of the tests @robertodr , I will take a look into it and let you know

Tansito:
@robertodr from the tests it seems `numpy` is a required dependency, without it serverles is not able to run the functions. Could you apply the restriction in the `requirements.txt`?

We will take a look in the meantime what dependency requires to have `numpy` declared there, it's not clear. Thank you!

Tansito:
sorry @robertodr it seems we continue with some incompatibilities, tomorrow I will take a look and let you know 🙏

Tansito:
@robertodr I could finally identify the problem and we will take a look as soon as we can. I created this issue to keep you updated about the advances. I will close this PR in the meantime and open a new one as soon as we have it fix it!

robertodr:
Ok thanks. Will subscribe to that issue

PR #1526, opened 2024-10-23
Filter by access in the program list for catalog client

Tansito:
### Summary

Until now we were returning in the catalog all those functions that had view access. With this PR the filter will only return those functions with view access by the user.

### DIFF
--- a/gateway/tests/fixtures/fixtures.json
+++ b/gateway/tests/fixtures/fixtures.json
@@ -79,7 +79,10 @@
       "image": "icr.io/awesome-namespace/awesome-title",
       "author": 2,
       "env_vars": "{\"PROGRAM_ENV1\": \"VALUE1\", \"PROGRAM_ENV2\": \"VALUE2\"}",
-      "provider": "bfe8aa6a-2127-4123-bf57-5b547293cbea"
+      "provider": "bfe8aa6a-2127-4123-bf57-5b547293cbea",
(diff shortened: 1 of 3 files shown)


gateway/tests/fixtures/fixtures.json
```diff
@@ -79,7 +79,10 @@
       "image": "icr.io/awesome-namespace/awesome-title",
       "author": 2,
       "env_vars": "{\"PROGRAM_ENV1\": \"VALUE1\", \"PROGRAM_ENV2\": \"VALUE2\"}",
-      "provider": "bfe8aa6a-2127-4123-bf57-5b547293cbea"
+      "provider": "bfe8aa6a-2127-4123-bf57-5b547293cbea",
```
akihikokuroda, line 82: Would you add another api.program that has only 100 instead of adding 100 here so that it can make sure this instance is not returned?

akihikokuroda (changes requested):
The code changes look good.  Would you add some notes that describe the expected return instances?  Probably as comments?

Tansito:
Redhat is returning a 504 downloading the image, I will wait an retry the action later @akihikokuroda

Tansito:
@akihikokuroda it should be everything fixed now 👍

akihikokuroda (approved):
LGTM!

PR #1528, opened 2024-10-25
implement error message in job

korgan00:
### Summary

> If job.status() is ERROR, return raw result in error_message() and raise an exception in result()
> else return an empty string in error_message() and return raw result in result()

### DIFF
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -150,13 +150,12 @@ class Job:
                 if verbose:
                     logging.info(count)
 
-        if self.status() == "ERROR":
-            raise QiskitServerlessException(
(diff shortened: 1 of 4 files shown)


client/qiskit_serverless/core/job.py
```diff
@@ -145,6 +150,11 @@ def result(self, wait=True, cadence=5, verbose=False, maxwait=0):
                 if verbose:
                     logging.info(count)
 
+        if self.status() == "ERROR":
+            raise QiskitServerlessException(
```
pandasa123, line 154: This should return the message that's collected in `error_message()`

Tansito:
@pandasa123 we should update documentation with this right, where should we take a look?

PR #1530, opened 2024-10-29
Change tox testenv:black envdir

korgan00:
### Summary

I realized that if i run `tox -eblack && tox -elint` dependencies are installed twice everytime i execute it.
If we change the `envdir` of black, we have the dependecies duplicated in both directories but they are installed only the first time so it is much faster.

Is there any better option @IceKhan13 ?

### DIFF
--- a/client/tox.ini
+++ b/client/tox.ini
@@ -24,7 +24,6 @@ commands =
   python -m pytest -v
 
 [testenv:lint]
-envdir = .tox/lint
 skip_install = true
 commands =
   black --diff --check .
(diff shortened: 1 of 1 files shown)


Tansito (approved):
From my part this looks like a typo from our side but @IceKhan13 has more experience than me with `tox.ini` so I would wait to merge it until his approval.

Tansito (changes requested):
@IceKhan13 , @korgan00 just a little comment from @garrison in this PR: https://github.com/Qiskit/qiskit-ibm-catalog/pull/33

I think it's exactly what we are trying to solve in this PR.

Tansito (approved):
🚀

PR #1535, opened 2024-11-04
Fix build stage name

Tansito:
### Summary

This PR should remove the warnings generated at build time in the gateway image:

```
#1 WARN: StageNameCasing: Stage name 'BASE' should be lowercase (line 5)
#1 WARN: StageNameCasing: Stage name 'BUILD' should be lowercase (line 9)
#1 WARN: StageNameCasing: Stage name 'APP' should be lowercase (line 27)
```

### DIFF
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -2,11 +2,11 @@ ARG MICRO_IMAGE_DIR=/ubi-micro-img
 
 # BASE image using UBI 9 micro where the
 # application and requirements will be installed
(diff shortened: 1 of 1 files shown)


akihikokuroda (changes requested):
I believe that `COPY --from=xxxx` also need to be changed.

Tansito:
Good catch @akihikokuroda , thanks!

akihikokuroda (approved):
LGTM.  Thanks!

PR #1539, opened 2024-11-19
Logs format improvements

Tansito:
### Summary

After reviewing our security scans I tried to unify a little bit our logs. We need more work on this field but step by step.

This PR just reviews the format of some of our logs there is no new feature or bing change introduced.


korgan00 (approved):
LGTM
Why not using string interpolation?

Tansito:
> Why not using string interpolation?

At some point we decided to start using this approach due to a problem that I don't remember right now. But I think definitely we should try to use them again. I can open an issue to migrate the current loggers to use them because they are quite faster.

PR #1540, opened 2024-11-22
Catalog returns only the functions that you can run

Tansito:
### Summary

This is PR continues the work done in #1526 

The end-point that returns the list of the Functions for the Catalog is returning all the those Functions with `view` permission. This was generating some frustrations because `view` permission doesn't give you access to run them. So with this PR now we will return only those functions for what the user has `run` permission.

This PR also introduces a refactor around how `GET` end-points from `Program` view were managing the access to the model. With the purpose to continue adding good practices to the base code we are starting to introduce some new concepts:
- Now the access to the model is being done through a `Repository` class. In this case we started with the `get` methods for the `Program` view. We will continue the refactor in future PR.
- Swagger was updated
- Now `list` end-point doesn't filter by title. We are doing that using the end-point `get_by_title`.
- Updated test due to the change for the catalog filter.

### DIFF
--- a/gateway/api/views/programs.py
+++ b/gateway/api/views/programs.py
@@ -104,28 +104,6 @@ class ProgramViewSet(viewsets.GenericViewSet):
     def get_object(self):
         logger.warning("ProgramViewSet.get_object not implemented")
 
-    def get_queryset(self):
-        author = self.request.user
-        title = sanitize_name(self.request.query_params.get("title"))
-        provider_name = sanitize_name(self.request.query_params.get("provider"))
-        type_filter = self.request.query_params.get("filter")
-
-        author_programs = self._get_program_queryset_for_title_and_provider(
-            author=author,
-            title=title,
-            provider_name=provider_name,
-            type_filter=type_filter,
-        ).distinct()
-
-        author_programs_count = author_programs.count()
-        logger.info(
-            "ProgramViewSet get author [%s] programs [%s]",
-            author.id,
-            author_programs_count,
-        )
-
-        return author_programs
-
     def get_run_queryset(self):
         """get run queryset"""
         author = self.request.user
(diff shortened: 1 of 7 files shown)


gateway/tests/fixtures/fixtures.json
```diff
@@ -49,7 +49,7 @@
       "password": "pbkdf2_sha256$390000$kcex1rxhZg6VVJYkx71cBX$e4ns0xDykbO6Dz6j4nZ4uNusqkB9GVpojyegPv5/9KM=",
       "is_active": true,
       "groups": [
-          101
+          100
```
Tansito, line 52: This is to change the user group to the runner one that has the `run` permission.

gateway/api/views/programs.py
```diff
@@ -167,7 +147,28 @@ def list(self, request):
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.program.list", context=ctx):
-            serializer = self.get_serializer(self.get_queryset(), many=True)
+
+            author = self.request.user
+            type_filter = self.request.query_params.get("filter")
+
+            if type_filter == "serverless":
```
Tansito, line 154: @korgan00 let's comment about these values, Monday. I think we can use the `Enum` approach that we used in the last PR but how they don't accept strings if we don't use `StrEnum` we will need to compare them in a different way probably.

Tansito: @korgan00 I just solved it in [86d1c3a](https://github.com/Qiskit/qiskit-serverless/pull/1540/commits/86d1c3aed10666695f87c9cb92e091c4a4452ce0)

and I could find a way to setup a Enum with strings for Python 3.9 👍

korgan00 (approved):
LGTM

PR #1543, opened 2024-11-25
Refactor: integration tests

korgan00:


Tansito:
The good thing @korgan00 is that for me, locally, all the tests are working. So as we were assuming probably something related with your computer.

There are some problems with a couple of tests I think due to permissions. I will try to take a look these days and let you know but awesome work here ❤️  

Screenshot 2024-12-03 at 12 41 48 PM

Tansito (changes requested):
Four things for the PR, @korgan00:

1. Thanks to the refactor now I think we can remove the action that it was running manually the tests: [docker-build.yaml](https://github.com/Qiskit/qiskit-serverless/blob/main/.github/workflows/docker-build.yaml)
2. If I pass the tests locally an archive is created under `tests/` with the format `downloaded_*.tar`. I propose to add that pattern to the `.gitignore` to avoid push those files.

I think these two can be managed in another PR:

3. It would be nice if we can refactor too the action for [LocalClient](https://github.com/Qiskit/qiskit-serverless/blob/main/.github/workflows/notebook-local-verify.yaml) as we commented in the internal issue.
4. `06_function.py` is not being tested. This is normal because it was not being tested before either but the preparation can be a bit different. Let's connect Monday and I can give you some guidance around it.

Overall, awesome job here again, @korgan00. It looks awesome!

Tansito (approved):
LGTM @korgan00 

We can continue working in the last two points in other PRs 😄

PR #1546, opened 2024-12-03
Add support new files

korgan00:
### Summary

Add support for the new files list method in the client

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -400,8 +400,10 @@ class ServerlessClient(BaseClient):
     def files(self, function: QiskitFunction) -> List[str]:
         """Returns list of available files produced by programs to download."""
         return self._files_client.list(function)
-    
-    def provider_files(self, function: QiskitFunction, provider: Optional[str] = None) -> List[str]:
+
+    def provider_files(
+        self, function: QiskitFunction, provider: Optional[str] = None
+    ) -> List[str]:
         """Returns list of available files produced by programs to download."""
         return self._files_client.provider_list(function, provider)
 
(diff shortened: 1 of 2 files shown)


Tansito:
@korgan00 can you put `data-folder` branch as base? In `main` you are not going to have the required end-points to pass the tests

Tansito:
@korgan00 I just saw too that your branch is based on `main`. You will need to apply a rebase if you can over `data-folder`. I can take a look to it Monday if you want.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -397,9 +397,15 @@ def function(
...
     #####################
 
-    def files(self, provider: Optional[str] = None) -> List[str]:
+    def files(self, function: QiskitFunction) -> List[str]:
         """Returns list of available files produced by programs to download."""
-        return self._files_client.list(provider)
+        return self._files_client.list(function)
+
+    def provider_files(
+        self, function: QiskitFunction, provider: Optional[str] = None
```
Tansito, line 405: Technically you can obtain the title and the provider from the `QiskitFunction` instance:
- https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/function.py#L71

So you would not need to pass the provider to the function. My proposal would be to pass only the Qiskit Function in obtain the info from the instance, @korgan00 . WDYT?

And maybe we could add a check in `provider_files` to return an error if the function doesn't have a provider.

client/qiskit_serverless/core/files.py
```diff
@@ -54,6 +55,7 @@ def __init__(self, host: str, token: str, version: str):
         self.host = host
         self.version = version
         self._token = token
+        self._files_url = os.path.join(self.host, "api", self.version, "files")
```
Tansito, line 58: This change LGTM 👍

Tansito (changes requested):

Tansito (approved):
LGTM @korgan00 🎉

PR #1529, opened 2024-10-29
Files list end point refactor

Tansito:
### Summary

This PR is part of an effort to stabilize `files` API end-points in #1509 

This PR focuses on `files/` and `files/provider/` end-points to be able to list the files in the working directories for the user and for the provider.

### Details and comments
- `files/` end-point was separated in two:
  -  `files/` now contains the paths: `username/` and `username/provider/function` with the purpose to separated files from functions from the user's root folder
  - `files/provider/` to manage the path `provider/function` so now each function will have its own folder in the provider directory.
- A new service `FileStorage` was created to manage the logic of access to this storage for providers and users with the purpose to simplify the view.

PR #1547, opened 2024-12-03
Download end point refactor

Tansito:
### Summary

Following the work done in #1529 this PR refactors the `download` end-point. The process followed was very similar:

This PR enables `files/download` and `files/provider/download` end-points to be able to download the files from the user and provider directory.

### Details and comments

- `files/download` end-point was separated in two:
  -  `files/download` now contains the paths: `username/` and `username/provider/function`. Following the previous development, with the purpose to separate files to be downloaded from functions from the user's root folder.
  - `files/provider/download` to manage the path `provider/function` so now each function will have its own folder in the provider directory to download the files.


gateway/api/views/files.py
```diff
@@ -276,48 +288,152 @@ def provider_list(self, request):
...
+                    {"message": "Requested file was not found."},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+
+            file_wrapper, file_type, file_size = result
+            response = StreamingHttpResponse(file_wrapper, content_type=file_type)
+            response["Content-Length"] = file_size
+            response[
+                "Content-Disposition"
+            ] = f"attachment; filename={requested_file_name}"
```
korgan00, line 436:
```suggestion
            if not self.user_has_provider_access(request.user, provider_name):
                return Response(
                    {"message": f"Provider {provider_name} doesn't exist."},
                    status=status.HTTP_404_NOT_FOUND,
                )

            return _download_file_response(
                request.user, 
                function_title, 
                provider_name, 
                working_dir, 
                requested_file_name
            )
```

I think all the code is duplicated in here. Maybe, we can extract it to a helper function.

gateway/api/services/file_storage.py
```diff
@@ -41,6 +44,21 @@ class FileStorage:  # pylint: disable=too-few-public-methods
         provider_name (str | None): name of the provider in caseis needed to build the path
     """
 
+    @staticmethod
+    def file_extension_is_valid(file_name: str) -> bool:
```
korgan00, line 48:
```suggestion
    def is_valid_extension(file_name: str) -> bool:
```
Since it is part of the `FileStorage` you can omit `file` and since it is a boolean function, I always try to start with the verb.

gateway/api/views/files.py
```diff
@@ -276,48 +288,152 @@ def provider_list(self, request):
...
-                # look for file in user's folder
-                filename = os.path.basename(requested_file_name)
-                user_dir = os.path.join(
-                    sanitize_file_path(settings.MEDIA_ROOT),
-                    sanitize_file_path(user_dir),
+            username = request.user.username
+            requested_file_name = sanitize_file_name(
+                request.query_params.get("file", None)
+            )
+            provider_name = sanitize_name(request.query_params.get("provider", None))
```
korgan00, line 304: Should we check if the user has provider access if provider is provided?

korgan00 (approved):
Looks good to me. Just minor code improvements and a question.

gateway/api/services/file_storage.py
```diff
@@ -122,3 +140,42 @@ def get_files(self) -> list[str]:
...
+                file_size = os.path.getsize(path_to_file)
+
+                return file_wrapper, file_type, file_size
+        except FileNotFoundError:
+            logger.warning(
+                "Directory %s does not exist for file %s.",
+                path_to_file,
+                file_name_path,
+            )
+            return None
```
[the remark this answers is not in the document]
Tansito: +1 you are totally right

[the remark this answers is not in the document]
Tansito: Removed the exception here: [d55c449](https://github.com/Qiskit/qiskit-serverless/pull/1547/commits/d55c4498cc6257405d7ee999b5bd98f70a1eef2f)

gateway/api/utils.py
```diff
@@ -448,3 +448,14 @@ def create_gpujob_allowlist():
...
+
+def sanitize_file_name(name: str | None):
+    """Sanitize name of a file"""
+    if name:
+        sanitized_name = ""
+        for c in name:
+            if c.isalnum() or c in ["_", "-", "."]:
+                sanitized_name += c
+        return sanitized_name
+    return name
```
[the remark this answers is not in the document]
Tansito: I usually don't like to use regex because they are not so easy to read than the original approach. Let's do one thing, I can agree with the change if every time we add a regex we add a comment explaining what the regex is doing.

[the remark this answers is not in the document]
Tansito: Done here: [e774e07](https://github.com/Qiskit/qiskit-serverless/pull/1547/commits/e774e07807ec446f8c4dd4e6b386a8afeb319d12)

gateway/api/views/files.py
```diff
@@ -276,48 +288,152 @@ def provider_list(self, request):
...
+            function_title = sanitize_name(request.query_params.get("function", None))
+            working_dir = WorkingDir.PROVIDER_STORAGE
+
+            if (
+                requested_file_name is None
+                or function_title is None
+                or provider_name is None
+            ):
```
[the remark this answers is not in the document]
Tansito: Done here: [b2de4b1](https://github.com/Qiskit/qiskit-serverless/pull/1547/commits/b2de4b173af582f98201dea8ff1a82adc673325f)

gateway/api/views/files.py
Tansito: I agree but as commented before I was planning to finish with the four end-points to migrate all of them to make use of [program repository](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/repositories/programs.py).

gateway/api/services/file_storage.py
Tansito: Done here: [cee8159](https://github.com/Qiskit/qiskit-serverless/pull/1547/commits/cee8159201661eb441795b6f3f37382899deee1c)

gateway/api/views/files.py
Tansito: Yes, it's being checked inside `self.get_function`. If the user doesn't have access `get_function` returns a `None` and the view returns a `404`. 

We log in case the user doesn't have access in the traces internally.

Tansito:
Changes applied @korgan00 , thank you so much!

Feel free to take a last view tomorrow morning and if you see everything fine feel free to merge it yourself 😄

gateway/api/utils.py
```diff
@@ -421,13 +421,10 @@ def create_dependency_allowlist():
...
-        sanitized_name = ""
-        for c in name:
-            if c.isalnum() or c in ["_", "-", "/"]:
-                sanitized_name += c
-        return sanitized_name
-    return name
+    if not name:
+        return name
+    # Remove all characters except alphanumeric, _, -, /
+    return re.sub("[^a-zA-Z0-9_\\-/]", "", name)
```
korgan00, line 427:
```suggestion
    return re.sub("[^a-zA-Z0-9_-/]", "", name)
```
I think you don't need the backslash.

gateway/api/views/files.py
```diff
@@ -276,48 +288,148 @@ def provider_list(self, request):
...
+                return Response(
+                    {"message": f"Provider {provider_name} doesn't exist."},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+
+            function = self.get_function(
+                user=request.user,
+                function_title=function_title,
+                provider_name=provider_name,
+            )
```
korgan00, line 405: Since `get_function` check the user has provider access. Do we want to recheck it again?
[We are not doing it](https://github.com/Qiskit/qiskit-serverless/pull/1547#discussion_r1882822838) in the `download` function

korgan00 (approved):
Looks good, two minor comments.

gateway/api/utils.py
Tansito: It's not mandatory but it's recommended. In practice, for this specific use case, both will likely work the same way. However, the escaped version \- is considered a best practice. It explicitly tells the regex parser to treat the hyphen as a literal character and avoids potential ambiguity with the character interpretation.

gateway/api/views/files.py
Tansito: This is a different use case because this is the end-point for the providers. Here the check is to clarify if the user has access to the provider not to the function:
- user_has_provider_access
- user_has_access_to_provider_function

Tansito: So, returning to your previous comment that I think I misunderstood it. No, in the `download` end-point is not needed to check if the user has access to the provider. It only needs to have access to the function.

gateway/api/utils.py
korgan00: Ok, I didn't know about it, then you can replicate in the other regex. Thank you for the information

gateway/api/utils.py
```diff
@@ -448,3 +445,11 @@ def create_gpujob_allowlist():
...
 
     return gpujobs
+
+
+def sanitize_file_name(name: str | None):
+    """Sanitize the name of a file"""
+    if not name:
+        return name
+    # Remove all characters except alphanumeric, _, ., -
+    return re.sub("[^a-zA-Z0-9_\\.-]", "", name)
```
korgan00, line 455:
```suggestion
    return re.sub("[^a-zA-Z0-9_\\.\\-]", "", name)
```

korgan00 (approved):
LGTM

PR #1549, opened 2024-12-10
Add localclient to integration tests

korgan00:


client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -41,7 +41,6 @@
 
 from qiskit_serverless.core.constants import (
     OT_PROGRAM_NAME,
-    ENV_JOB_ARGUMENTS,
```
Tansito, line 44: Probably this is for another PR because we will need to take a look but I think we can remove all the references for `ENV_JOB_ARGUMENTS` in the `client`. In the `gateway` it will require a bit more analysis but I think so too.

client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -112,11 +111,16 @@ def run(
             **(saved_program.env_vars or {}),
             **{OT_PROGRAM_NAME: saved_program.title},
             **{"PATH": os.environ["PATH"]},
-            **{ENV_JOB_ARGUMENTS: json.dumps(arguments, cls=QiskitObjectsEncoder)},
         }
 
+        with open("arguments.serverless", "w", encoding="utf-8") as f:
```
Tansito, line 116: How would you see to move the name of the file to a constant? Probably with the migration to object storage we are going to need to modify it and having only one reference would be cool.

tests/docker/conftest.py
```diff
@@ -5,15 +5,32 @@
...
 from qiskit_serverless import ServerlessClient, QiskitFunction
+from qiskit_serverless.core.clients.local_client import LocalClient
 
 resources_path = os.path.join(
     os.path.dirname(os.path.abspath(__file__)), "source_files"
 )
 
 
+@fixture(scope="module", params=["serverless", "local"])
+def any_client(request):
```
Tansito, line 16: Instead to use `any_client` what do you think about using `base_client`? Not a strong opinion here but in reality all our clients inheritance from `BaseClient` if I remember correctly so I think it has sense.

Tansito (approved):
In general I see everything fine, @korgan00 . Just a couple of comments that are not required to be honest. Thank you so much for this PR 🙏

client/qiskit_serverless/core/clients/local_client.py
korgan00: I didnt removed it because of `ray_client` that it is using it and I don't know if it will raise any problem.

Tansito: Absolutely, for another PR. I will create an issue annotating this improvement.

Tansito (approved):

Tansito:
I'm not sure if I could merge this, @korgan00 so I leave my approval and feel free to merge it if you finished all the last changes 🙏

PR #1550, opened 2024-12-13
Add support new download files endpoint

korgan00:


Tansito (approved):
Very straightforward, thanks @korgan00 !

PR #1553, opened 2024-12-16
Trace decorator

korgan00:
### Summary

Remove most of the `trace with ...` lines and replace them with a decorator.


Tansito (approved):
LGTM

PR #1509, opened 2024-10-07
Improve data directory functionality

Tansito:
### Summary

This PR pretends to standardize the way that users will access to the file system to upload and interact with files in the Qiskit Functions. The new interface will generate the next UX:

- **User persona**

```python
from qiskit_ibm_catalog import QiskitFunctionsCatalog

catalog = QiskitFunctionsCatalog(token=...)

function = catalog.load(".../...")

# It will list files in the user's function folder
catalog.files(function)

# It will upload a file in the user's function folder
catalog.file_upload(filename, function)

# It will download a file in the user's function folder
catalog.file_download(filename, function)

# It will delete a file in the user's function folder
serverless.file_delete(filename, function)
```

- **Provider persona**

```python
from qiskit_ibm_catalog import QiskitFunctionsCatalog

catalog = QiskitFunctionsCatalog(token=...)

function = catalog.load(".../...")

# It will list files in the user's function folder
catalog.provider_files(function)

# It will upload a file in the user's function folder
catalog.provider_file_upload(filename, function)

# It will download a file in the user's function folder
catalog.provider_file_download(filename, function)

# It will delete a file in the user's function folder
catalog.provider_file_delete(filename, function)
```

The functionality inside the Qiskit Function will remain the same. The same two folders to manage the different kind of content:
- `/data`: to store files for the user
- `/function_data`: to store files for the partners 

### Details and comments

- [x] End-Points migrated
- [x] Qiskit Serverless client migrated
- [x] Scheduler migrated
- [x] Documentation updated
- [x] Qiskit IBM Catalog migrated

PR #1559, opened 2024-12-17
Remove extension limitation in files

Tansito:
### Summary

This PR will continue the work from #1509 . In this case we are removing the limitation to only list and download tar and h5 files.


korgan00 (approved):
LGTM

PR #1556, opened 2024-12-17
Client - files delete refactor

korgan00:
### Summary

depends of https://github.com/Qiskit/qiskit-serverless/pull/1554

PR #1561, opened 2024-12-19
Integration tests fix

korgan00:
### Summary

Depends on https://github.com/Qiskit/qiskit-serverless/pull/1556

### DIFF
--- a/client/qiskit_serverless/core/files.py
+++ b/client/qiskit_serverless/core/files.py
@@ -130,7 +130,9 @@ class GatewayFilesClient:
                 os.path.join(self._files_url, "provider", "download"),
             )
 
-    def upload(self, file: str, provider: Optional[str] = None) -> Optional[str]:
+    def upload(
+        self, file: str, function: QiskitFunction, provider: Optional[str] = None
+    ) -> Optional[str]:
         """Uploads file."""
         tracer = trace.get_tracer("client.tracer")
         with tracer.start_as_current_span("files.upload"):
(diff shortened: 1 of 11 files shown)


Tansito (approved):
LGTM, I left you a couple of comments @korgan00 that we can discuss tomorrow. Also the error in the test will be fixed with #1560 .

client/qiskit_serverless/core/files.py
```diff
@@ -138,7 +138,7 @@ def upload(
         with tracer.start_as_current_span("files.upload"):
             with open(file, "rb") as f:
                 with requests.post(
-                    os.path.join(self._files_url, "upload"),
+                    os.path.join(self._files_url, "upload/"),
```
Tansito, line 141: This change was due to use a `POST` without a finished slash: https://docs.djangoproject.com/en/dev/ref/settings/#std-setting-APPEND_SLASH

Honestly, I couldn't look at that configuration yet so when I review it we can decide if we should start adding the slash at the end in all our end-points or what.

gateway/api/services/file_storage.py
```diff
@@ -151,13 +151,19 @@ def get_file(self, file_name: str) -> Optional[Tuple[FileWrapper, str, int]]:
...
-            file_wrapper = FileWrapper(file_object)
+        file_type = mimetypes.guess_type(path_to_file)[0]
+        file_size = os.path.getsize(path_to_file)
 
-            file_type = mimetypes.guess_type(path_to_file)[0]
-            file_size = os.path.getsize(path_to_file)
+        # We can not use context manager here. Django close the file automatically:
+        # https://docs.djangoproject.com/en/5.1/ref/request-response/#fileresponse-objects
+        file_wrapper = FileWrapper(
+            open(path_to_file, "rb")  # pylint: disable=consider-using-with
```
Tansito, line 160: I don't know why the `gateway` tests were working before but it seems that to transfer a file we don't need to setup a context manager. It's being specified in the documentation that I just pasted there.

tests/docker/test_docker_experimental.py
```diff
@@ -92,19 +95,61 @@ def test_upload_download_delete(self, serverless_client: ServerlessClient):
...
+        print(serverless_client.file_delete(filename, function))
 
         print("::: files after delete:::")
-        files = serverless_client.files()
+        files = serverless_client.files(function)
+        print(files)
+
+        assert (file_count - len(files)) == 1
+
+    def test_provider_upload_download_delete(self, serverless_client: ServerlessClient):
```
Tansito, line 115: I just added an e2e test for the provider end-points too.

gateway/api/services/file_storage.py
```diff
@@ -151,13 +151,19 @@ def get_file(self, file_name: str) -> Optional[Tuple[FileWrapper, str, int]]:
             )
             return None
 
-        with open(path_to_file, "rb") as file_object:
-            file_wrapper = FileWrapper(file_object)
+        file_type = mimetypes.guess_type(path_to_file)[0]
+        file_size = os.path.getsize(path_to_file)
```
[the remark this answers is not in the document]
Tansito: Ups, probably I just left those doing tests let me check and I remove them if that's the case

PR #1567, opened 2025-01-14
Feat/create result storage class

paaragon:
### Summary

This PR implements functionality as part of the user story to allow users to save larger results.

To achieve this, it introduces the ResultStorage function, designed to serve as a service for storing job results efficiently.

### Details and comments

- Results are saved to disk


Tansito:
Let's use `cos-result` as the feature branch

PR #1569, opened 2025-01-16
Store the result in the object storage

Tansito:
### Summary

This PR will save the result of a Job in the object storage instead of the database.


gateway/api/services/result_storage.py
```diff
@@ -0,0 +1,78 @@
...
+    ENCODING = "utf-8"
+
+    def __init__(self, username: str):
+        """Initialize the storage path for a given user."""
+        self.user_results_directory = os.path.join(
+            settings.MEDIA_ROOT, username, "results"
+        )
+        os.makedirs(self.user_results_directory, exist_ok=True)
+
+    def __build_result_path(self, job_id: str) -> str:
```
Tansito, line 27: @paaragon to maintain the same semantic as we had in `FileStorage` let's use `__get_result_path` instead of build, WDYT?

gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,67 @@
+"""
+Access policies implementation for Job access
+"""
+import logging
+from django.contrib.auth import get_user_model
+from api.models import Job
+
+User = get_user_model()
```
Tansito, line 8: I see that we are only using `User` as a type. The problem is that I'm receiving a warning in VSCode when we use it:

```
Variable not allowed in type expressionPylance[reportInvalidTypeForm]
(variable) User: Any
```

How we cannot import directly the user model due to [E5142](https://github.com/pylint-dev/pylint-django/issues/278) error until we find a workaround my recommendation would be just remove it by now.

gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+logger = logging.getLogger("gateway")
+
+
+class JobAccessPolocies:  # pylint: disable=too-few-public-methods
```
Tansito, line 14: I think know you can remove this `pylint` warning now

gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    The main objective of this class is to manage the access for the user
+    to the Job entities.
+    """
+
+    @staticmethod
+    def can_access(user: User, job: Job) -> bool:
+        """
+        Checks if the user has access to save the result of a Job:
```
Tansito, line 23: Correct me if I'm wrong here, @paaragon . This method is to check if the user can retrieve the job so we should update the text to something like:

```
Checks if the user has access to the Job
```

Or something like that. It's not specific to `save` the result.

gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,67 @@
...
+
+        Returns:
+            bool: True or False in case the user has access
+        """
+
+        is_provider_job = job.program and job.program.provider
+        if is_provider_job:
+            provider_groups = job.program.provider.admin_groups.all()
+            author_groups = user.groups.all()
+            has_access = any(group in provider_groups for group in author_groups)
```
Tansito, line 37: Here @paaragon you can use the `Provider.can_access`: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/access_policies/providers.py#L19

I think it does exactly the same thing that you have in those 3 lines of code. It checks if the user has access to that provider.

gateway/api/repositories/jobs.py
```diff
@@ -0,0 +1,31 @@
...
+            id (str): id of the job
+
+        Returns:
+            Job | None: job with the requested id
+        """
+
+        result_queryset = Job.objects.filter(id=job_id).first()
+
+        if result_queryset is None:
+            logger.warning("Job [%s] was not found", id)
```
Tansito, line 29: I think we are using `logger.info` in these cases. `warning` maybe is too much, WDYT?

gateway/api/services/result_storage.py
```diff
@@ -0,0 +1,78 @@
...
+
+        Returns:
+            Tuple containing:
+            - FileWrapper for the file
+            - File MIME type
+            - File size in bytes
+        """
+        result_path = self.__get_result_path(job_id)
+        if not os.path.exists(result_path):
+            logger.warning(
```
Tansito, line 43: How we are assuming that some jobs are not going to have this file let's use `logger.info` here.

gateway/api/services/result_storage.py
```diff
@@ -0,0 +1,78 @@
...
+        os.makedirs(self.user_results_directory, exist_ok=True)
+
+    def __get_result_path(self, job_id: str) -> str:
+        """Construct the full path for a result file."""
+        return os.path.join(
+            self.user_results_directory, f"{job_id}{self.RESULT_FILE_EXTENSION}"
+        )
+
+    def get(self, job_id: str) -> Optional[str]:
+        """
```
Tansito, line 32: I think you need to update here the `Args` and `Returns` params.

gateway/api/services/result_storage.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, username: str):
+        """Initialize the storage path for a given user."""
+        self.user_results_directory = os.path.join(
+            settings.MEDIA_ROOT, username, "results"
+        )
+        os.makedirs(self.user_results_directory, exist_ok=True)
+
+    def __get_result_path(self, job_id: str) -> str:
+        """Construct the full path for a result file."""
```
Tansito, line 26: I think you need to update here the `Args` and `Returns` params.

gateway/api/views/jobs.py
```diff
@@ -51,10 +53,39 @@ class JobViewSet(viewsets.GenericViewSet):
...
+        Returns the default serializer class for the view.
+        """
         return self.serializer_class
 
+    @staticmethod
+    def get_serializer_job(*args, **kwargs):
+        """
+        Returns a `JobSerializer` instance
+        """
+        return v1_serializers.JobSerializer(*args, **kwargs)
```
Tansito, line 69: Take a look how we are using the Serializers in the Program, @paaragon :
- In V1 we specify the V1 serializers: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/views/programs.py#L25
- And in views we are using the default one: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/views/programs.py#L64

So here you should use the `JobSerializer` and this one in the v1 view.

gateway/api/views/jobs.py
```diff
@@ -51,10 +53,39 @@ class JobViewSet(viewsets.GenericViewSet):
...
+
+    @staticmethod
+    def get_serializer_job_without_result(*args, **kwargs):
+        """
+        Returns a `JobSerializerWithoutResult` instance
+        """
+        return v1_serializers.JobSerializerWithoutResult(*args, **kwargs)
```
Tansito, line 76: Similar to my previous comment about `JobSerializer`

gateway/api/views/jobs.py
```diff
@@ -51,10 +53,39 @@ class JobViewSet(viewsets.GenericViewSet):
...
+
     def get_queryset(self):
+        """
```
Tansito, line 79: Thanks @paaragon 🙏

gateway/api/views/jobs.py
```diff
@@ -76,24 +107,36 @@ def retrieve(self, request, pk=None):  # pylint: disable=unused-argument
...
+            job = self.jobs_repository.get_job_by_id(pk)
             if job is None:
+                logger.info("Job [%s] nor found", pk)
+                return Response(
+                    {"message": f"Job [{pk}] nor found"},
+                    status=status.HTTP_404_NOT_FOUND,
+                )
+
+            if not JobAccessPolocies.can_access(author, job):
                 logger.warning("Job [%s] not found", pk)
```
Tansito, line 121: We can remove this logger because I think we are logging the error in the `can_access`., WDYT?

gateway/api/views/jobs.py
```diff
@@ -76,24 +107,36 @@ def retrieve(self, request, pk=None):  # pylint: disable=unused-argument
         tracer = trace.get_tracer("gateway.tracer")
         ctx = TraceContextTextMapPropagator().extract(carrier=request.headers)
         with tracer.start_as_current_span("gateway.job.retrieve", context=ctx):
-            job = Job.objects.filter(pk=pk).first()
+
+            author = self.request.user
+            job = self.jobs_repository.get_job_by_id(pk)
             if job is None:
+                logger.info("Job [%s] nor found", pk)
```
Tansito, line 114: I think we can remove this logger message, we are just logging it in the method if I remember correctly 👍

gateway/api/views/jobs.py
```diff
@@ -116,29 +159,26 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
-            attempts_left = 10
-            while not saved:
-                if attempts_left <= 0:
-                    return Response(
-                        {"error": "All attempts to save results failed."}, status=500
-                    )
+            author = self.request.user
+            job = self.jobs_repository.get_job_by_id(pk)
+            if job is None:
+                logger.info("Job [%s] nor found", pk)
```
Tansito, line 165: Similar to my previous comments, how we are logging it in the repository there is no need to log it the error here too.

gateway/api/views/jobs.py
```diff
@@ -116,29 +159,26 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
-                    logger.warning(
-                        "Job [%s] record has not been updated due to lock. Retrying. Attempts left %s",  # pylint: disable=line-too-long
-                        job.id,
-                        attempts_left,
-                    )
-                    continue
-                time.sleep(1)
+            can_access = JobAccessPolocies.can_save_result(author, job)
+            if not can_access:
+                logger.info("Job [%s] nor found for author %s", pk, author.username)
```
Tansito, line 173: Similar to my previous comments, how we are logging it in the can_save_result there is no need to log it the error here too.

gateway/api/views/jobs.py
```diff
@@ -116,29 +159,26 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
-                    saved = True
-                except RecordModifiedError:
-                    logger.warning(
-                        "Job [%s] record has not been updated due to lock. Retrying. Attempts left %s",  # pylint: disable=line-too-long
-                        job.id,
-                        attempts_left,
-                    )
-                    continue
-                time.sleep(1)
+            can_access = JobAccessPolocies.can_save_result(author, job)
```
Tansito, line 171: I would change probably the name of the variable to something like: `can_save_result`. Thinking in can be confusing with the `can_access` method, WDYT @paaragon ?

Tansito:
@paaragon they look like a lot of comments but they are minor details, the development is pretty good. Thank you so much! ❤️

Tansito:
@paaragon I was running some tests and I see everything fine. Approve this PR and we merge it to test it in staging.

pink-zebra234 (approved):

paaragon (approved):

PR #1570, opened 2025-01-16
Feat/save result into result store

paaragon:
### Summary

This PR continues the use case about saving job results in files instead of database.

This modifies the POST /result endpoint to save the result using the ResultStorare


gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,36 @@
...
+class JobAccessPolocies:  # pylint: disable=too-few-public-methods
+    """
+    The main objective of this class is to manage the access for the user
+    to the Job entities.
+    """
+
+    @staticmethod
+    def can_save_result(user, job: Job) -> bool:
+        """
+        Checks if the user has access to a Provider:
```
Tansito, line 21: Probably @paaragon here you want to modify the sentence to something like:

```
Checks if the user has access to save the result of a Job:
```

gateway/api/access_policies/jobs.py
```diff
@@ -0,0 +1,36 @@
...
+            job: Job instance against to check the access
+
+        Returns:
+            bool: True or False in case the user has access
+        """
+
+        has_access = user.id == job.author.id
+        if not has_access:
+            logger.warning(
+                "User [%s] has no access to job [%s].", user.username, job.author
```
Tansito, line 34: What do you think about having a more specific sentence?

```
"User [%s] has no access to save the result of the job [%s].", user.username, job.author
```

gateway/api/repositories/jobs.py
```diff
@@ -0,0 +1,34 @@
...
+        Args:
+            id (str): id of the job
+
+        Returns:
+            Job | None: job with the requested id
+        """
+
+        id_criteria = Q(id=job_id)
+
+        result_queryset = Job.objects.filter(id_criteria).first()
```
Tansito, line 29: From what I know `Q` purpose in Django is to build complex queries. How in this case the function `get_job_by_id` leaves pretty clear the what we are doing I think with just:

```
result_queryset = Job.objects.filter(id=job_id).first()
```

Would be more than enough. Let me know @paaragon if you prefer that approach.

gateway/tests/resources/fake_media/test_user/results/1a7947f9-6ae8-4e3d-ac1e-e7d608deec82.json
```diff
@@ -0,0 +1 @@
+{"ultimate": 42}
```
Tansito, line 1: @paaragon this file should be removed from git, I think, shouldn't it?

Tansito (changes requested):

gateway/tests/resources/fake_media/test_user/results/8317718f-5c0d-4fb6-9947-72e480b8a348.json
```diff
@@ -0,0 +1 @@
+{"ultimate": 42}
```
Tansito, line 1: @paaragon similar to the previous one, can this file be deleted? I don't see is needed in any test but just to confirm.

paaragon: Oh yes... is not needed in this branch but in [feat/retrieve-job-results-from-result-store](https://github.com/Qiskit/qiskit-serverless/pull/1575) I'm removing it from here

Tansito (approved):

PR #1571, opened 2025-01-17
Trial mode added for Qiskit Functions

Tansito:
### Summary

This PR enables to configure Qiskit Functions with a trial mode. The functionality is simple:

- Now the Program model has a `trial_instances` new field that will store the groups from those users that are going to run the specific function in a trial mode.
- The new trial mode consists in add a new environment variables "ENV_JOB_TRIAL" with the **string values** "True" or "False".

### Details and comments

Main modifications applied in this PR:

- Program model:
  - new field `trial_instances`
  -  `instances` field now has a relational name in the ORM `program_instances` instead of just `program`. This was needed to avoid conflicts in the namespace of Django's ORM.
- Job model:
  - new field `trial` to identify easily if the Job is a trial mode or not
- I did a refactor in the tests because some UUIDs were being reused and causing problems (I will leave some comments about this)
- Due to a bug that I found obtaining the Qiskit Functions from a user I modified a bit the `run` end-point:
  - The biggest change was that I removed the old `run_query` to use the new one method from the Repository class
- If a job is in a trial mode is checked in the serializer when creating the job


Tansito:
@pandasa123 I think the most interest point in this PR for you is this one:

> The new trial mode consists in add a new environment variables "ENV_JOB_TRIAL" with the string values "True" or "False".

Let me know if you agree with the name or you prefer to change it

gateway/tests/fixtures/fixtures.json
```diff
@@ -125,12 +128,15 @@
...
+      ],
+      "trial_instances": [
+        100
       ]
     }
   },
   {
     "model": "api.job",
-    "pk": "1a7947f9-6ae8-4e3d-ac1e-e7d608deec82",
+    "pk": "57fc2e4d-267f-40c6-91a3-38153272e764",
```
Tansito, line 139: These changes @paaragon about the `pk` in the fixtures was because the UUID between Functions and Jobs were being reused and it was created me some problems so I just updated them.

gateway/api/migrations/0033_job_trial_program_trial_instances_and_more.py
```diff
@@ -0,0 +1,33 @@
...
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AddField(
+            model_name="program",
+            name="trial_instances",
+            field=models.ManyToManyField(
+                blank=True, related_name="program_trial_instances", to="auth.group"
+            ),
+        ),
+        migrations.AlterField(
```
Tansito, line 26: For who is interested on, this `AlterField` doesn't apply to the Database so we continue providing backward compatibility in the database in case we need to do rollback. It only affects to the ORM.

gateway/api/serializers.py
```diff
@@ -208,6 +211,22 @@ class RunJobSerializer(serializers.ModelSerializer):
     class Meta:
         model = Job
 
+    def is_trial(self, function: Program, author) -> bool:
```
Tansito, line 214: Serializers logic will need a big refactor but it's too much for the purpose of this PR so I just left it by now here.

gateway/api/views/programs.py
```diff
@@ -104,45 +101,6 @@ def get_serializer_class(self):
     def get_object(self):
         logger.warning("ProgramViewSet.get_object not implemented")
 
-    def get_run_queryset(self):
```
Tansito, line 107: I needed to apply this change in this PR due to a bug in this method. The change here was basically:
- remove `get_run_queryset`
- use the function repository class [here](https://github.com/Qiskit/qiskit-serverless/pull/1571/files#diff-60276657248dc9f5babc8ac88d48a0709ec594e741be0bc0b12103419e37a5d6R212)

Tansito, line 107: I needed to apply this change in this PR due to a bug in this method. The change here was basically:
- remove `get_run_queryset`
- use the function repository class [here](https://github.com/Qiskit/qiskit-serverless/pull/1571/files#diff-60276657248dc9f5babc8ac88d48a0709ec594e741be0bc0b12103419e37a5d6R212)

Tansito:
I will work to solve this e2e tests, probably is related with the change in the `run` method.

paaragon (approved):

pandasa123:
For `ENV_JOB_TRIAL` , let's update this to `ENV_ACCESS_TRIAL`. Otherwise, this looks great!

PR #1577, opened 2025-01-30
Feat/remove result from list endpoint

paaragon:
### Summary

This PR removes the result from the response in the list endpoint

### DIFF
--- a/gateway/api/views/jobs.py
+++ b/gateway/api/views/jobs.py
@@ -7,8 +7,6 @@ import json
 import logging
 import os
 
-from django.db.models import Q
-
 # pylint: disable=duplicate-code
 from opentelemetry import trace
 from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import OTLPSpanExporter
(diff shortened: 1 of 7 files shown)


gateway/api/views/jobs.py
```diff
@@ -145,10 +145,10 @@ def list(self, request):
 
             page = self.paginate_queryset(queryset)
```
Tansito, line 146: @paaragon let's use this opportunity to migrate the `queryset` too. The idea as we commented is to put all the logic related with the ORM in the Repositories. In this case the philosophy would be similar to what we did in the `list` for functions: https://github.ibm.com/IBM-Q-Software/qiskit-serverless/blob/main/gateway/api/views/programs.py#L104

Tansito (commented):
Thanks for the refactor @paaragon ! We couldn't remove the `get_queryset` but for another chance.

As my last comment before merge I think we can remove the method `filter_queryset`, we are not using it from what I know. With that I think we are going to be able to merge this.

Tansito (approved):
LGTM!

PR #1578, opened 2025-01-31
Create is_trial method for new trial mode

Tansito:
### Summary

Continuing with the work done in #1571 this PR introduces a new method in the Qiskit Serverless client to interact with the new trial functionality.

The new method `is_trial` will check the value of its environment variable and return True or False depending of the value.


Tansito:
@pandasa123 I created this PR as an addition that I think it could improve the DX from this functionality. We didn't have it in mind but let me know what you think. And if you agree with the naming `is_trial` also.

pandasa123 (approved):

PR #1579, opened 2025-02-04
Remove scorecard action

Tansito:
### Summary

We are not using this action and currently is failing. How we don't have time to maintain it and it's not being priority for us we will remove it until we have time to make it work again.


academix-yeah (approved):

paaragon (approved):

Tansito:
Thanks @academix-yeah 😉

PR #1583, opened 2025-02-11
feat: job provider list endpoint

paaragon:
### Summary

This PR creates a new endpoint to allow providers to retrieve all jobs from a function they own.

### Details and comments

The endpoint support pagination and jobs are returned without results

### DIFF
--- a/gateway/api/repositories/jobs.py
+++ b/gateway/api/repositories/jobs.py
@@ -5,7 +5,7 @@ import logging
 from typing import List
 from django.db.models import Q
 from api.models import Job
-from api.models import Program
+from api.models import Program as Function
 
 logger = logging.getLogger("gateway")
 
@@ -33,7 +33,7 @@ class JobsRepository:  # pylint: disable=too-few-public-methods
 
         return result_queryset
 
-    def get_program_jobs(self, program: Program, ordering="-created") -> List[Job]:
+    def get_program_jobs(self, function: Function, ordering="-created") -> List[Job]:
         """
         Retrieves all program's jobs.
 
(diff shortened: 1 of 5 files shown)


gateway/api/views/jobs.py
```diff
@@ -145,6 +154,65 @@ def list(self, request):
             serializer = self.get_serializer_job_without_result(queryset, many=True)
         return Response(serializer.data)
 
+    @action(methods=["GET"], detail=False, url_path="provider")
+    def provider_list(self, request):
```
Tansito, line 158: Remember @paaragon that you need to include this new end-point in the v1 too:
https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/v1/views/jobs.py

Add to it the swagger configuration also.

gateway/api/v1/views/jobs.py
```diff
@@ -52,6 +52,14 @@ def retrieve(self, request, pk=None):
     def list(self, request):
         return super().list(request)
 
+    @swagger_auto_schema(
+        operation_description="List provider Jobs",
+        responses={status.HTTP_200_OK: v1_serializers.JobSerializer(many=True)},
```
Tansito, line 57: In reality here the correct serializer is the `JobSerializerWithoutResult`, isn't it?

Tansito (approved):
@paaragon good work! I only found a comment to fix but feel free to merge after fix it 😄

gateway/api/v1/views/jobs.py
academix-yeah: done!

gateway/api/repositories/jobs.py
```diff
@@ -32,6 +33,20 @@ def get_job_by_id(self, job_id: str) -> Job:
 
         return result_queryset
 
+    def get_program_jobs(self, program: Program, ordering="-created") -> List[Job]:
```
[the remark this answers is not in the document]
paaragon: Done!

PR #1586, opened 2025-02-19
Authentication logic refactor

Tansito:
### Summary

This PR will refactor the entire authentication flow to prepare the project to introduce additional services like IBM Cloud. This is a 100% refactor, there is no change in the logic or the output. So the main purpose is the simplification and improvement of the codebase.

### Details and comments
- The current logic was divided in two additional classes: 
  - AuthenticationUseCase: it will manage the main flow for the authentication process. User authentication, validation and repository access.
  - QuantumPlatformService: it will manage all the access to these 3rd party end-points.  
-  QuantumUserProxy was removed in favor of refactoring some logic between QuantumPlatformService and UserRepository classes what it has more sense from a responsibility point of view.
- Environment variables were simplified also to reduce the number of required configurations along the authentication flow.


gateway/api/services/authentication/quantum_platform.py
```diff
@@ -0,0 +1,175 @@
...
+    def __init__(self, authorization_token):
+        self.auth_url = (
+            f"{settings.QUANTUM_PLATFORM_API_BASE_URL}/api/users/loginWithToken"
+        )
+        self.verification_url = f"{settings.QUANTUM_PLATFORM_API_BASE_URL}/api/users/me"
+        self.instances_url = f"{settings.IQP_QCON_API_BASE_URL}/network"
+        self.authorization_token = authorization_token
+        self.access_token = None
+
+    def _get_network(self, access_token: str):
```
Tansito, line 29: This is something very interesting @paaragon . If I use `__get_network` here I cannot mock the method because as a private method the test cannot access to the method.

I didn't remember this so probably we should stop using that nomenclature, what do you think?

gateway/api/repositories/users.py
```diff
@@ -2,16 +2,42 @@
...
+User = get_user_model()
+logger = logging.getLogger("gateway.repositories.user")
+
+
+class UserRepository:
     """
     The main objective of this class is to manage the access to the model
     """
 
+    def get_or_create_by_id(self, user_id: str) -> type[AbstractUser]:
```
Tansito, line 23: I also found a way to type that we are returning a user with `type[AbstractUser]`. This way the linter doesn't complain 😂

gateway/api/authentication.py
```diff
@@ -28,81 +26,24 @@ class CustomToken:
...
-
-        elif auth_header is None:
+        authorization_token = None
+        auth_header = request.META.get("HTTP_AUTHORIZATION")
+        if auth_header is None:
             logger.warning(
-                "Problems authenticating: User did not provide authorization token."
+                "Problems authenticating: user did not provide authorization token."
             )
+            return quantum_user, CustomToken(authorization_token.encode())
```
paaragon, line 37: If `authorisation_token` is `None` at this point, wouldn't `authorisation_token.encode()` raise an error?

paaragon (changes requested):
Hi @Tansito . Very good job! this organisation looks so clean and performant.

Just one little change request 😉

gateway/api/authentication.py
Tansito: Good catch @paaragon , thank you!

Tansito: Solved in [d779f7e](https://github.com/Qiskit/qiskit-serverless/pull/1586/commits/d779f7e6d7ddba235211e9602060d0bf2666acc0)

paaragon (approved):

PR #1590, opened 2025-02-26
fix: remove job result from backoffice

paaragon:
### Summary

This PR removes the job results from the admin backoffice


Tansito (approved):
Thanks @paaragon 🙏

PR #1591, opened 2025-02-27
Feat: migrate log format to fstring

paaragon:
### Summary

This PR changes all the logging messages from lazy logging with `%` to strings.


Tansito:
After discussing with @paaragon we decided to continue using `%` in the backend instead of f-string due to the performance parsing the strings in the logs. Thank you so much @paaragon for the review on this! 😄

PR #1593, opened 2025-02-27
IBM Cloud Integration

Tansito:
### Summary

This pull request introduces a series of improvements and optimizations across various components of the authentication and service integration modules. 

The most important new feature is the possibility to authenticate with IBM Cloud. This way we have three different channels:
- Local
- IBM Quantum
- IBM Cloud

The first are the channels we were using until now and this PR doesn't add any breaking change on those features. IBM Cloud authentication will need to use the new interface:

```python
serverless = IBMServerlessClient(channel="ibm_cloud", token="IBM_CLOUD_API_KEY", instance="IBM_CLOUD_CRN")
```

The rest of the channels, IBM Quantum and Local, remain the same with the exception that now you would have two ways to authenticate:

```python
serverless = IBMServerlessClient(token="IQP_TOKEN")
serverless = IBMServerlessClient(channel="ibm_quantum", token="IQP_TOKEN")
```

**Important**

`instance` parameter will be used only for IBM Cloud crn.


Tansito:
Last thing remains, we just need to save the id from the access group instead the name.

paaragon (approved):

PR #1600, opened 2025-03-06
Fix/save account

paaragon:
### Summary

This PR fixes the `IBMServerlessClient.save_account` method


.gitignore
```diff
@@ -101,3 +101,5 @@ docs/_build/
 .jupyter/
 .ipython/
 .virtual_documents/
+
```
Tansito, line 104: I don't think this is needed for the PR, @paaragon

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -552,6 +552,7 @@ def __init__(
             channel: identifies the method to use to authenticate the user
         """
         token = token or QiskitRuntimeService(name=name).active_account().get("token")
+        print(token)
```
Tansito, line 555: I think we should remove this print

Tansito (changes requested):

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -564,6 +565,8 @@ def save_account(
         token: Optional[str] = None,
         name: Optional[str] = None,
         overwrite: Optional[bool] = False,
+        instance: Optional[str] = None,
+        channel: Optional[Channel] = Channel.IBM_QUANTUM,
```
Tansito, line 569: Also @paaragon , we are going to need to change this logic a little bit to send a string instead of the enum so we are more similar with runtime. I just adapted the logic here so it should almost copy/paste: https://github.com/Qiskit/qiskit-serverless/pull/1593/commits/41081818be788edd338ba9c2fa37e63e5fcff487#diff-fdc3aef5e1a595990db1b71f1ca195cbdfc589291d1f23c213c5fe735c8b0611R130

Let me know if you would have any doubt 👍

Tansito (changes requested):

Tansito (approved):
LGTM @paaragon , thank you!

PR #1605, opened 2025-03-13
feat: handle 500 errors / improve dev UX

paaragon:
### Summary

This PR:
- handles 500 errors to return them as JSON instead of HTML
- Modify docker-compose-dev.yaml to hot reload gateway changes

### Details and comments

To test the change:
1. set `DEBUG=0` in `services.dateway.environment` inside `docker-compose-dev.yaml`.
2. run `docker-compose -f docker-compose-dev.yaml watch`.
3. raise an Error somewhere in the code.
4. Make an http request to the code with the previous error (you should receive the error in json format).
5. Remove the Error raise and make the http request again. You should see the change sync and the response without errors.

### DIFF
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -36,6 +36,7 @@ services:
     command: gunicorn main.wsgi:application --bind 0.0.0.0:8000 --workers=4
     ports:
       - 8000:8000
+    user: "root"
     environment:
       - DEBUG=1
       - RAY_HOST=http://ray-head:8265
@@ -55,6 +56,11 @@ services:
       - program-artifacts:/usr/src/app/media/
     depends_on:
       - postgres
+    develop:
+      watch:
+        - path: ./gateway
+          action: sync
+          target: /usr/src/app
   scheduler:
     container_name: scheduler
     build:
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -36,7 +36,7 @@ services:
     command: gunicorn main.wsgi:application --bind 0.0.0.0:8000 --workers=4
     ports:
       - 8000:8000
-    user: "root"
+    user: "root" # we use the root user here so the docker-compose watch can sync files into the container
     environment:
       - DEBUG=1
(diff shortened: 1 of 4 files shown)


gateway/main/urls.py
```diff
@@ -23,6 +23,9 @@
 from drf_yasg import openapi
 import probes.views
 import version.views
+from api.utils import custom_server_error
+
+handler500 = custom_server_error
```
Tansito, line 28: @paaragon I think we can use this approach from DRF I don't know if you had the opportunity to take a look:

https://www.django-rest-framework.org/api-guide/exceptions/#rest_frameworkexceptionsserver_error

It would be just to use the next line:
```
handler500 = 'rest_framework.exceptions.server_error'
```

I didn't know about it, I just saw it after reading the documentation about the `handler500` variable. And we can remove the method `custom_server_error`.

gateway/api/utils.py
```diff
@@ -200,7 +201,7 @@ def generate_cluster_name(username: str) -> str:
         generated cluster name
     """
     pattern = re.compile("[^a-zA-Z0-9-.]")
-    cluster_name = f"c-{re.sub(pattern,'-',username)}-{str(uuid.uuid4())[:8]}"
+    cluster_name = f"c-{re.sub(pattern, '-', username)}-{str(uuid.uuid4())[:8]}"
```
Tansito, line 204: I don't think this change is needed in this PR.

docker-compose-dev.yaml
```diff
@@ -36,6 +36,7 @@ services:
     command: gunicorn main.wsgi:application --bind 0.0.0.0:8000 --workers=4
     ports:
       - 8000:8000
+    user: "root"
```
Tansito, line 39: @paaragon can we add a comment about why are we using the root here? I think it's interesting to annotate that this root is for the `watch`.

Tansito (changes requested):

Tansito (approved):
Thanks @paaragon this looks really nice 😄

PR #1611, opened 2025-04-04
feat: set function as disabled

paaragon:
### Summary

This PR introduces two new attributes to the Program model:

disabled (BooleanField): Flags whether the program is currently disabled.
disabled_message (TextField, optional): A custom message returned when a disabled program is attempted to be run.

If a program is marked as disabled, any attempt to run it will now return an HTTP 423 Locked response, along with the disabled_message if provided. This prevents new jobs from being launched for disabled programs.

### Details and comments

- Added disabled and disabled_message fields to the Program model.
- Updated the /run endpoint to check for disabled before enqueuing a job.
- Return HTTP 423 with a message when a disabled program is requested.


korgan00 (approved):
LGTM

gateway/api/models.py
```diff
@@ -75,13 +75,20 @@ class Program(ExportModelOperationsMixin("program"), models.Model):
...
+
     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
     created = models.DateTimeField(auto_now_add=True, editable=False)
 
     title = models.CharField(max_length=255, db_index=True)
     readable_title = models.CharField(
         max_length=255, null=True, blank=True, default=None
     )
+
+    disabled = models.BooleanField(default=False, null=True)
```
Tansito, line 88: The only thing that I would change from this PR is the usage of `null` here. I don't think it has sense to configure it as `True` taking into account that we are setting up a `default` value. So I think `null` should be `False` here.

Tansito: Done in [75ad3a3](https://github.com/Qiskit/qiskit-serverless/pull/1611/commits/75ad3a329950c9c683e49d7d00f86575cd1ac072)

PR #1613, opened 2025-04-07
feat: maintenance mode

paaragon:
### Summary

This PR adds the MAINTENANCE environment variable.

If MAINTENANCE is true, queued jobs are not scheduled.

Also, if the user tries to create a new job during the MAINTENANCE, it will return an error

### DIFF
--- a/Dockerfile-ray-node
+++ b/Dockerfile-ray-node
@@ -1,7 +1,7 @@
-FROM registry.access.redhat.com/ubi9-minimal:9.5-1738816775
+FROM registry.access.redhat.com/ubi9-minimal:9.6-1747218906
 RUN microdnf install -y \
-    python3.11-3.11.9 \
-    python3.11-devel-3.11.9 \
+    python3.11-3.11.11 \
+    python3.11-devel-3.11.11 \
     wget-1.21.1
 RUN microdnf clean all
 RUN ln -s /usr/bin/python3.11 /usr/local/bin/python3 && \
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -2,11 +2,11 @@ ARG MICRO_IMAGE_DIR=/ubi-micro-img
 
 # BASE image using UBI 9 micro where the
 # application and requirements will be installed
-FROM registry.access.redhat.com/ubi9-micro:9.5-1738816380 AS base
+FROM registry.access.redhat.com/ubi9-micro:9.6-1745521186 AS base
 
 # BUILD image using UBI 9 where the dependencies that
 # require installing with a package manager will be installed
-FROM registry.access.redhat.com/ubi9:9.5-1738814488 AS build
+FROM registry.access.redhat.com/ubi9:9.6-1747219013 AS build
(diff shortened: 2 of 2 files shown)


korgan00 (approved):
LGTM

gateway/api/views/programs.py
```diff
@@ -205,6 +206,13 @@ def run(self, request):
                 )
                 return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
+            if settings.MAINTENANCE:
```
Tansito, line 209: Mmm... I don't know if I agree with this. I think even if the scheduler is in maintenance mode we should continue letting create jobs. Those jobs are going to be QUEUED but that's not problematic meanwhile we don't process them, WDYT @korgan00 ?

gateway/main/settings.py
```diff
@@ -319,6 +319,7 @@
 LIMITS_CPU_PER_TASK = int(os.environ.get("LIMITS_CPU_PER_TASK", "4"))
 LIMITS_GPU_PER_TASK = int(os.environ.get("LIMITS_GPU_PER_TASK", "1"))
 LIMITS_MEMORY_PER_TASK = int(os.environ.get("LIMITS_MEMORY_PER_TASK", "8"))
+MAINTENANCE = os.environ.get("LIMITS_MEMORY_PER_TASK", "false") == "true"
```
Tansito, line 322: I think the environment variable that we want to use here is "MAINTENANCE", probably a typo.

gateway/tox.ini
```diff
@@ -21,7 +21,7 @@ deps = -rrequirements.txt
        -rrequirements-dev.txt
 commands =
   pip check
-  python manage.py test
+  python manage.py test {posargs}
```
Tansito, line 24: For what is this change? Do you know, @korgan00 ?

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -133,6 +133,8 @@ spec:
               value: {{ .Values.application.ray.gpu | quote }}
             - name: LIMITS_MEMORY_PER_TASK
               value: {{ .Values.application.ray.memory | quote }}
+            - name: MAINTENANCE
+              value: {{ .Values.application.maintenance | quote }}
```
Tansito, line 137: If we don't block the gateway this "MAINTENANCE" could be removed I think.

Tansito (changes requested):

gateway/tox.ini
korgan00: It is to be able to run certain tests using the tox command args

gateway/tests/api/test_v1_program.py
```diff
@@ -8,6 +8,7 @@
 from django.urls import reverse
 from rest_framework import status
 from rest_framework.test import APITestCase
+from django.test import override_settings
```
Tansito, line 11: Same as my previous comment here, @korgan00

Tansito (approved):

PR #1615, opened 2025-04-07
Add running substatuses

korgan00:
### Summary

Add substatus for running status in order to allow functions creators to explicitly change it during execution.


gateway/api/models.py
```diff
@@ -190,6 +213,9 @@ class Job(models.Model):
         choices=JOB_STATUSES,
         default=QUEUED,
     )
+    sub_status = models.CharField(
+        max_length=30, choices=SUB_STATUSES, default=None, null=True, blank=True
```
Tansito, line 217: @korgan00 how do you see to configure `max_length` as 255. I don't think there is a real reason for that 30 and we can avoid future problems in case we want a greater value.

gateway/api/views/jobs.py
```diff
@@ -249,6 +249,73 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
+                # If we import this with the regular imports,
+                # update jobs statuses test command fail.
+                # it should be something related with python import order.
+                from api.management.commands.update_jobs_statuses import (  # pylint: disable=import-outside-toplevel
+                    update_job_status,
+                )
+
+                update_job_status(job)
+
+                if job is None:
```
Tansito, line 278: I would do this check even before `update_job_status` to avoid problem with the update process.

gateway/api/views/jobs.py
```diff
@@ -249,6 +249,73 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
+                    return Response(
+                        {
+                            "message": "Cannot update 'sub_status' when is not"
+                            f" in RUNNING status. (Currently {job.status})"
+                        },
+                        status=status.HTTP_403_FORBIDDEN,
+                    )
+
+                job.sub_status = sub_status
+                job.save()
```
Tansito, line 307: I would like to propose another approach for this `save`. It could happen that as we are doing `update_job_status` to ensure that the Job is updated, since we retrieve the information from the DB the `status` could change also. So I would like to propose move these lines:
```
job.sub_status = sub_status
job.save()
```

To a new method in `JobRepository` called `jobs_repository.update_job_statuses` that will do something like:
```
updated = Job.objects.filter(id=id, status=Job.RUNNING).update(sub_status=sub_status)
```

This way we can ensure that the update happens at the moment the Job was in a Running state and we could check with `updated != 1` if we had success or not updating the object.

WDYT @korgan00 ?

gateway/api/v1/views/jobs.py
```diff
@@ -70,6 +70,13 @@ def provider_list(self, request):
     def result(self, request, pk=None):
         return super().result(request, pk)
 
+    @swagger_auto_schema(
+        operation_description="Update the sub status of a job",
```
Tansito, line 74: I'm not 100% sure but I think we need to add `sub_status` here, `swagger_generator` won't fill that info directly by itself.

Tansito (commented):
Happy to discuss the changes proposed with you, @korgan00

gateway/api/models.py
korgan00: I just took the longest one and rounded up :)
I'm ok with 255.

gateway/api/views/jobs.py
korgan00: I'm ok with everything but, what happens if we are not succeed? In my opinion, we checked everything before and if we fail is because of a race condition so, we can safely say that is not an error.

korgan00: I'm ok with everything but, what happens if we are not succeed? In my opinion, we checked everything before and if we fail is because of a race condition so, we can safely say that is not an error.

Tansito:
Just to double check with you @pandasa123 . PR is almost ready to merge so just to validate that this is what we are expecting from the product side:
- New status definition: https://github.com/Qiskit/qiskit-serverless/blob/ae0c667b84d81d4402bd9aa8493e44556c63eafd/client/qiskit_serverless/core/job.py#L328
- Way to update partners can update the status: https://github.com/Qiskit/qiskit-serverless/blob/ae0c667b84d81d4402bd9aa8493e44556c63eafd/tests/docker/source_files/pattern_with_sub_status_and_wait.py

Let us know if you would like to do any last minute modification 😄

Tansito:
From @pandasa123 he wants to add a little modification to the new statuses names:

```
"RUNNING: MAPPING",
"RUNNING: OPTIMIZING_FOR_HARDWARE",
"RUNNING: WAITING_FOR_QPU",
"RUNNING: EXECUTING_QPU",
"RUNNING: POST_PROCESSING",
```

Just add the "FOR" in some of the statuses.

gateway/api/views/jobs.py
```diff
@@ -249,6 +249,73 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
             serializer = self.get_serializer(job)
         return Response(serializer.data)
 
+    @action(methods=["POST"], detail=True)
```
Tansito, line 252: Let's start using "PATCH" for those methods that will update a existed entry in the DB (we will see how to manage the migration for other end-points).

gateway/api/views/jobs.py
```diff
@@ -249,6 +249,73 @@ def result(self, request, pk=None):  # pylint: disable=invalid-name,unused-argum
...
+                        {
+                            "message": "Cannot update 'sub_status' when is not"
+                            f" in RUNNING status. (Currently {job.status})"
+                        },
+                        status=status.HTTP_403_FORBIDDEN,
+                    )
+
+                self.jobs_repository.update_job_sub_status(job, sub_status)
+
+                return Response({"message": "Sub status updated correctly"})
```
Tansito, line 309: Also, in a PATCH end-point, let's return the entity updated if everything was ok. In this case, let's return the Job with the `JobSerializerWithoutResult`, result would not be needed at this point.

Tansito (approved):
I think now is perfect, @korgan00 . We are fine to merge it and testing it in staging 😄 

Good work, @korgan00 , @paaragon !

PR #1623, opened 2025-04-25
Update Django

Tansito:
### Summary

Update Django to `5.2`


Tansito:
Ok it looks like we are not compatible with v5. We will update our v4 and take a look later to upgrade to v5.

PR #1625, opened 2025-04-29
Improve HTTP server configuration

Tansito:
### Summary

This PR introduces several improvements in our current HTTP Server configuration after reading carefully the documentation: https://docs.gunicorn.org/en/stable/settings.html

The main improvement that I was trying to accomplish was to be able to apply different configurations depending of the environment through `values.yaml`. I added also the typical parameters that users usually modify to increase stability: workers, threads, max_requests, max_requests_jitter and timeout.

Being able to configure `workers` is very important because different environments have different resources so we could increase or decrease this number depending of the environment where we are deploying in order: `2-4 x $(NUM_CORES)`

Also the combination of `max_requests` and `max_requests_jitter` is important to configure because they configure the thresholds used to restart the different workers to prevent memory leaks (this could be easily one of the problems we were having in the system).


korgan00 (approved):
LGTM

PR #1626, opened 2025-04-30
Return function type

korgan00:


Tansito (approved):
Integration tests on k8s fixed and the changes LGTM, @korgan00 . Thank you so much 😄

PR #1627, opened 2025-04-30
Fix custom image integration test

korgan00:


Tansito:
for when you return, @korgan00 , remember to update this branch with the changes applied in #1626 to avoid problems with the tests in k8s

avilches:
Are we going to continue this PR? @korgan00 @Tansito

PR #1633, opened 2025-05-05
Enable access to swagger

Tansito:
### Summary

Due to the refactor in the authentication we disabled access to swagger. This PR enables this access again.


korgan00 (changes requested):
I think we can improve this like:

```python
def isPublicEndpoint(path):
  return any(path in request.path for path in PUBLIC_ENDPOINTS)
  
[...]

public_access = isPublicEndpoint(request.path)
```

korgan00 (approved):

PR #1638, opened 2025-05-13
Extend Python version support in serverless client

ElePT:
### Summary

This PR extends the Python version support of qiskit-serverless to match qiskit's (3.9-3.13).  I found that `ray[data]` was the limiting factor for supporting `python != 3.11`, so I made it an optional requirement that can be installed through `pip install qiskit-serverless[ray-data]`. Users interested in ray-data will still be limited to python 3.11 but all other users can now use any of the supported versions.

Other changes in this PR include:
- Updated client tests to check 3.9-3.11
- Updated `pyproject.toml` and `setup.py` files
- Added a `version.py` file for compatibility with `pyproject.toml`
- Some minor docs improvements


client/requirements-optional.txt
```diff
@@ -0,0 +1 @@
+ray[default,data]>=2.30, <3
```
Tansito, line 1: As we commented internally Elena I was looking for references to `ray[data]` and I don't think we are using it anymore. I think it's safe to remove this optional package and we simplify the README also.

korgan00 (approved):
LGTM

client/requirements-optional.txt
Tansito, line 1: Taking into account that we are removing `data` we are able to remove this file also, don't we @ElePT ?

Tansito (approved):
LGTM @ElePT , good job here 🎉

PR #1642, opened 2025-05-21
Update to Django 5

Tansito:
### Summary

This PR pretends to update Django project to the last major version.


Tansito:
It looks like with the middleware approach the problem was solved. I'm not going to open the PR yet. I will double check:
- Dependencies to update to the official versions where they support Django v5
- Django panel administrator. We don't have tests for the Django panel and we need to be sure that it's working for our common processes.

gateway/main/urls.py
```diff
@@ -66,4 +63,5 @@
     re_path(r"^redoc/$", schema.with_ui("redoc", cache_timeout=0), name="schema-redoc"),
 ]
 if settings.DEBUG:
+    urlpatterns += [path("", include("django_prometheus.urls"))]
```
Tansito, line 66: I moved our prometheus url to DEBUG until we have time to do the integration with Grafana

gateway/main/middleware.py
```diff
@@ -0,0 +1,18 @@
+"""
+Custom middleware used by Qiskit Serverless API
+"""
+
+
+def add_slash(get_response):
+    """
+    This middleware allows to do calls with and without slashes:
```
korgan00, line 8: :clap:

korgan00 (approved):
Everything looks good

gateway/main/middleware.py
Tansito: I'm going to have nightmares with this 😂

PR #1644, opened 2025-05-23
update qaoa example

psschwei:
### Summary

Was trying out the QAOA example and noticed that it wasn't working. With the changes made here, it does.


ElePT:
I started a follow-up PR here: https://github.com/Qiskit/qiskit-serverless/pull/1645/files

PR #1655, opened 2025-06-09
Improve Windows support for the client

paaragon:
### Summary

This PR fixes the GatewayFilesClient url composition to add support for windows

### Details and comments

Instead of using os.path.join for create the url, this change uses a utility to join paths for urls

### DIFF
--- a/.github/workflows/client-pypi-release.yaml
+++ b/.github/workflows/client-pypi-release.yaml
@@ -26,7 +26,7 @@ jobs:
           python setup.py sdist
           python setup.py bdist_wheel
         shell: bash
-      - uses: actions/upload-artifact@89ef406dd8d7e03cfd12d9e0a4a378f454709029 #4.3.5
+      - uses: actions/upload-artifact@4.4.3
         with:
           path: ./client/dist/qiskit_serverless*
       - name: Publish to PyPi
--- a/.github/workflows/docs-verify.yaml
+++ b/.github/workflows/docs-verify.yaml
@@ -25,7 +25,7 @@ jobs:
           tox -edocs
       - name: Upload docs artifact
         if: always()
-        uses: actions/upload-artifact@89ef406dd8d7e03cfd12d9e0a4a378f454709029 #4.3.5
+        uses: actions/upload-artifact@4.4.3
         with:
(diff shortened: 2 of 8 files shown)


Tansito (commented):
I don't disregard with the solution @paaragon but remember that the issue was not only for this line of code. You need to apply this patch in every place where we are manipulating url paths with `os.path.join`.

client/arguments.serverless
```diff
@@ -0,0 +1 @@
+{"circuit": {"__type__": "QuantumCircuit", "__value__": "eJwL9Az29gzhY2JgYoAAxoJCBu40Bg4gk4UBAUDSjFA2WDw5syi5NLNE19DMCKamuraQESzNyFjIgApgWmG2MMNlOJANxQKCnI0j3BNLUkEmgo1lAhHMEAtY0YxHB1EwncwM7ASUhqSkwxQzwBUz4VDs6hwEUwx2DsjQ/0gApg4AA6wiEA=="}, "array": {"__type__": "ndarray", "__value__": "eJyb7BfqGxDJyFDGUK2eklqcXKRupaBuk2ahrqOgnpZfVFKUmBefX5SSChJ3S8wpTgWKF2ckFqQC+RpGOgqGmjoKtQpkAy4GMHB1YIACABOFG/M="}}
```
korgan00: I think we don't need this, I think it is a test subproduct.

client/qiskit_serverless/utils/urls.py
```diff
@@ -0,0 +1,10 @@
+"""
+Utility functions for URL path manipulation.
+"""
+
+
+def url_path_join(base, *parts):
```
korgan00: can you set types here?

korgan00 (changes requested):
LGTM, some small changes

client/arguments.serverless
paaragon: you are right. Removed

client/qiskit_serverless/utils/urls.py
paaragon: done!

Tansito (approved):

PR #1656, opened 2025-06-11
format error messages correctly

korgan00:

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -99,7 +99,6 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
         token: Optional[str] = None,
         instance: Optional[str] = None,
         channel: str = Channel.IBM_QUANTUM.value,
-        verbose: bool = False,
     ):
         """
         Initializes the ServerlessClient instance.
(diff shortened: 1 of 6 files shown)


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -720,6 +717,8 @@ def _upload_with_artifact(  # pylint:  disable=too-many-positional-arguments
             span.set_attribute("function.provider", response_data.get("provider", "na"))
             response_data["client"] = client
             response_function = RunnableQiskitFunction.from_json(response_data)
+    except QiskitServerlessException as error:  # pylint: disable=broad-exception-caught
```
Tansito: I think the pylint here is not needed.

tests/tox.ini
```diff
@@ -20,7 +20,7 @@ deps = -rrequirements-dev.txt
 commands =
   pip install ../client
   pip check
-  python -m pytest -v --order-dependencies
+  python -m pytest -v --order-dependencies --capture=no
```
Tansito: I think this capture can be removed also.

Tansito (commented):
LGTM @korgan00 just a couple of questions.

ElePT (approved):
LGTM, as discussed offline, if we ever find that we need to disable specific messages it's better to target them with another mechanism (instead of the verbose flag).

Tansito (approved):
Perfect @korgan00 thank you!

PR #1657, opened 2025-06-11
Improved support for Runtime 0.40.0

Tansito:
### Summary

This PR updates the project to be able to use the new value `ibm_quantum_platform` from Qiskit Runtime 0.40.0 version.

### Details and comments

- Support in the client:

The functionality is basically the same as `ibm_cloud`. The new value, `ibm_quantum_platform`, was added as a new channel. The output is the next one:

```python
# Current usage with ServerlessClient
serverless = ServerlessClient(
    host="http://localhost:8000",
    channel="ibm_quantum",
    token="YOUR_IQP_TOKEN"
)

# Future usage of ServerlessClient
serverless = ServerlessClient(
    host="http://localhost:8000",
    channel="ibm_quantum_platform", # "ibm_cloud" value will work the same until its deprecation
    instance="YOUR_CRN_INSTANCE",
    token="YOUR_API_TOKEN"
)

# Current usage with IBMServerlessClient
serverless = IBMServerlessClient(
    channel="ibm_quantum",
    token="YOUR_IQP_TOKEN"
)

# Future usage of IBMServerlessClient
serverless = IBMServerlessClient(
    channel="ibm_quantum_platform", # "ibm_cloud" value will work the same until its deprecation
    instance="YOUR_CRN_INSTANCE",
    token="YOUR_API_TOKEN"
)
```

- Support in the API:
  - IBM Cloud value was change by IBM Quantum Platform
  - Quantum Platform was changed to IBM Quantum
  - IBM Cloud was changed to IBM Quantum Platform

Naming was adapted to reflect the changes in the API. Partners will start receiving `ibm_quantum_platform` as environment variable in their Functions instead of `ibm_cloud`.

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -130,7 +130,7 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
         try:
             channel_enum = Channel(channel)
         except ValueError as error:
-            raise QiskitServerlessException(
+            raise ValueError(
                 "Your channel value is not correct. Use one of the available channels: "
                 f"{Channel.LOCAL.value}, {Channel.IBM_QUANTUM.value}, "
                 f"{Channel.IBM_CLOUD.value}, {Channel.IBM_QUANTUM_PLATFORM.value}"
@@ -587,7 +587,7 @@ class IBMServerlessClient(ServerlessClient):
         try:
             channel_enum = Channel(channel)
         except ValueError as error:
-            raise QiskitServerlessException(
+            raise ValueError(
                 "Your channel value is not correct. Use one of the available channels: "
                 f"{Channel.LOCAL.value}, {Channel.IBM_QUANTUM.value}, "
                 f"{Channel.IBM_CLOUD.value}, {Channel.IBM_QUANTUM_PLATFORM.value}"


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -570,6 +580,22 @@ def __init__(
             channel: identifies the method to use to authenticate the user
         """
         token = token or QiskitRuntimeService(name=name).active_account().get("token")
+        channel = channel or QiskitRuntimeService(name=name).active_account().get(
```
Tansito, line 584: @ElePT we are using `save_account` from Runtime to manage the account object. They way that we retrieve the info is with this logic. I don't know if there is a better way but happy to hear improvements on this.

ElePT: This is how I would do it too. If I come up with any other way I'll let you know.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -570,6 +580,22 @@ def __init__(
...
+            "channel"
+        )
+        try:
+            channel_enum = Channel(channel)
+        except ValueError as error:
+            raise QiskitServerlessException(
+                "Your channel value is not correct. Use one of the available channels: "
+                f"{Channel.LOCAL.value}, {Channel.IBM_QUANTUM.value}, {Channel.IBM_CLOUD.value}"
```
ElePT: The error message should show the new iqp channel too (the line is probably going to be too long after the suggestion, sorry)

```suggestion
                "Your channel value is not correct. Use one of the available channels: "
                f"{Channel.LOCAL.value}, {Channel.IBM_QUANTUM_PLATFORM.value}, {Channel.IBM_QUANTUM.value},  {Channel.IBM_CLOUD.value}"
```

client/qiskit_serverless/core/enums.py
```diff
@@ -12,3 +12,4 @@ class Channel(str, Enum):
     LOCAL = "local"
     IBM_QUANTUM = "ibm_quantum"
     IBM_CLOUD = "ibm_cloud"
+    QUANTUM_PLATFORM = "ibm_quantum_platform"
```
ElePT:
```suggestion
    IBM_QUANTUM_PLATFORM = "ibm_quantum_platform"
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -570,6 +581,23 @@ def __init__(
...
+            raise QiskitServerlessException(
+                "Your channel value is not correct. Use one of the available channels: "
+                f"{Channel.LOCAL.value}, {Channel.IBM_QUANTUM.value}, "
+                f"{Channel.IBM_CLOUD.value}, {Channel.IBM_QUANTUM_PLATFORM.value}"
+            ) from error
+
+        if channel_enum is not Channel.IBM_QUANTUM:
+            instance = instance or QiskitRuntimeService(name=name).active_account().get(
+                "instance"
+            )
```
korgan00, line 599: `ServerlessClient` raise an exception if `instance is not None` when `Channel` is `IBM_QUANTUM`, should we do it here too?

gateway/api/use_cases/authentication.py
```diff
@@ -40,13 +40,13 @@ def __init__(
         self.public_access = public_access
 
     def _get_authentication_service_instance(self) -> AuthenticationBase:
-        if self.channel == Channel.IBM_CLOUD:
+        if self.channel == Channel.IBM_QUANTUM_PLATFORM:
```
ElePT, line 43: With this change, would authentication with "ibm_cloud" still work? I haven't had the chance to test it but it looks like it might fail (unless I am missing something being set to iqp under the hood)

ElePT (commented):
Thanks David! I looked through the PR and agree with the overall changes, but had a few comments about the implementation.

korgan00 (commented):
Minor comment

client/qiskit_serverless/core/clients/serverless_client.py
Tansito: done! [9d385a0](https://github.com/Qiskit/qiskit-serverless/pull/1657/commits/9d385a0eadadfb86bebd81b793dcbc229fcf2cdf)[9d385a0](https://github.com/Qiskit/qiskit-serverless/pull/1657/commits/9d385a0eadadfb86bebd81b793dcbc229fcf2cdf)

client/qiskit_serverless/core/enums.py
Tansito: Done! [9d385a0](https://github.com/Qiskit/qiskit-serverless/pull/1657/commits/9d385a0eadadfb86bebd81b793dcbc229fcf2cdf)[9d385a0](https://github.com/Qiskit/qiskit-serverless/pull/1657/commits/9d385a0eadadfb86bebd81b793dcbc229fcf2cdf)

gateway/api/use_cases/authentication.py
Tansito: Yes, this is what I was commenting you the other day. We basically don't send the channel if not that if you send us an instance we assume you will use the `ibm_cloud` / `ibm_quantum_platform`: [ref](https://github.com/Qiskit/qiskit-serverless/pull/1657/files#diff-fd3c3b60712a1b62ca11717fd45330b50cf56d4ef179eef2dc2015db0ba1d9e9R45). We can do this because we don't support the `instance` for `ibm_quantum` channel.

The idea is, internally, we just don't make use of channels. Thinking that, in the near future, how we are going to have only one channel we are going to be able to remove this logic (let me know if I would be wrong).

client/qiskit_serverless/core/clients/serverless_client.py
Tansito: In this case it's not needed because how this class extends `ServerlessClient` that logic it will run with the init [here](https://github.com/Qiskit/qiskit-serverless/pull/1657/files#diff-fdc3aef5e1a595990db1b71f1ca195cbdfc589291d1f23c213c5fe735c8b0611R601)

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -130,16 +130,27 @@ def __init__(  # pylint:  disable=too-many-positional-arguments
         try:
             channel_enum = Channel(channel)
         except ValueError as error:
-            raise QiskitServerlessException(
+            raise ValueError(
```
Tansito, line 133: Following @korgan00 feedback I just changed `QiskitServerlessException` by `ValueError` as a better way to manage this exception specifically.

gateway/api/use_cases/authentication.py
ElePT: ok, got it, so we manually "set" the channel by checking if an instance has been provided or not

Tansito: exactly!

korgan00 (approved):

PR #1662, opened 2025-06-18
Prettify error message on jobs

korgan00:

### DIFF
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -189,7 +189,10 @@ class Job:
 
     def error_message(self):
         """Returns the execution error message."""
-        return self._job_service.result(self.job_id) if self.status() == "ERROR" else ""
+        error = (
+            self._job_service.result(self.job_id) if self.status() == "ERROR" else ""
+        )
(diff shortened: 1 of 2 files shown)


client/qiskit_serverless/core/job.py
```diff
@@ -189,7 +189,10 @@ def filtered_logs(self, **kwargs) -> str:
 
     def error_message(self):
         """Returns the execution error message."""
-        return self._job_service.result(self.job_id) if self.status() == "ERROR" else ""
+        error = (
+            self._job_service.result(self.job_id) if self.status() == "ERROR" else ""
+        )
+        return error.strip('"').encode().decode("unicode_escape")
```
ElePT: Thanks @korgan00, I know it's a small change but a unit test would be good to see the effect of it.

Tansito: I miss a `try/except` in case we could have problems transforming the string for whatever reason. I think we should always return the error as it is in case we cannot transforming it so the user can work with it.

Tansito (approved):
LGTM, @korgan00. Feel free to update the branch to fix the errors and merge it if everything goes fine. Thanks!

PR #1663, opened 2025-06-23
Include qpy-compat option from Qsikit 2.1.0

Tansito:
### Summary

[Qiskit 2.1.0](https://github.com/Qiskit/qiskit/releases/tag/2.1.0) changed the way they were managing `symengine` package. From previous versions they made that `symengine` was not needed anymore but that only works for versions greater than 2.1.0.

For versions lower than 2.1.0 we need to change the way we are installing qiskit adding the option `qpy-compat`.


Tansito:
This PR should fix the problems in #1662

korgan00 (approved):

PR #1666, opened 2025-06-26
Enable instance initialization

Tansito:
### Summary

This PR configures the instance in the initialization of the client to enable to use the instance when you are using the `ibm_quantum` channel.

### Details and comments

To be able to send the instance in the initialization some changes in the API needed to be done. Now we send also the `channel` so we can use internally the instance independently from the `channel` that the user selects.


ElePT (commented):
Just looking at the changes, they look good, and I think having everything explicitly defined (channel, instance, etc) in the intialization makes the pipeline cleaner. However, the main point we need to be careful about is the user interaction with these changes. How do you suggest to go about the testing process, does it make to run some local tests or wait for the deployment in dev?

korgan00 (approved):

Tansito:
> How do you suggest to go about the testing process, does it make to run some local tests or wait for the deployment in dev?

Yes, I basically setup the configuration locally and I tested it. I can show you how to do it next week if you would have any doubt 👍

PR #1669, opened 2025-06-30
Dep version end point

korgan00:


gateway/api/use_cases/dependencies_versions.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import Optional
+from django.contrib.auth.models import AbstractUser
+
+from api.utils import create_dynamic_dependencies_whitelist
+
+
+logger = logging.getLogger("gateway.use_cases.authentication")
+
+
+class AvailableDynamicDependenciesUseCase:  # pylint: disable=too-few-public-methods
```
Tansito: We should probably modify the configuration for `too-few-public-methods` if not we are going to need to add it in every use case class.

gateway/api/use_cases/dependencies_versions.py
```diff
@@ -0,0 +1,22 @@
+"""Authentication use case to manage the authentication process in the api."""
+
+import logging
+from typing import Optional
+from django.contrib.auth.models import AbstractUser
+
+from api.utils import create_dynamic_dependencies_whitelist
+
+
+logger = logging.getLogger("gateway.use_cases.authentication")
```
Tansito:
```suggestion
logger = logging.getLogger("gateway.use_cases.dependencies_versions")
```

Tansito: I'm not sure we are using _ or - here

gateway/api/use_cases/dependencies_versions.py
Tansito: I would match more or less the name of the class with the file, WDYT @korgan00 ? My proposal: `AvailableDependenciesVersions`.

gateway/api/use_cases/dependencies_versions.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class AvailableDynamicDependenciesUseCase:  # pylint: disable=too-few-public-methods
+    """
+    This class will available dynamic dependencies.
+    """
+
+    def execute(self) -> Optional[type[AbstractUser]]:
```
Tansito: We are not returning an `AbstractUser` here, isn't it?

gateway/api/v1/views/dependencies_versions.py
```diff
@@ -0,0 +1,44 @@
...
+        status.HTTP_200_OK: openapi.Response(
+            description="List of strings",
+            schema=openapi.Schema(
+                type=openapi.TYPE_ARRAY, items=openapi.Items(type=openapi.TYPE_STRING)
+            ),
+        )
+    },
+)
+@api_view(["GET"])
+@permission_classes([permissions.AllowAny])
```
Tansito: Inside API end-points we are only allowing IsAuthenticated permission. We need to modify some logic to allow not authenticated end-points.

gateway/api/v1/views/dependencies_versions.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Prepare de output for the end-point
+    """
+    return [str(dep) for dep in data.values()]
+
+
+@swagger_auto_schema(
+    method="get",
+    operation_description="Get the list of available "
+    "additional dependencies for creating functions",
```
Tansito: We are using addition, dynamic or not using at all any word to describe this functionality. I would like to fusion a bit the idea around: `AvailableDependenciesVersions`, for example.

I accept other ideas.

gateway/api/v1/views/dependencies_versions.py
```diff
@@ -0,0 +1,44 @@
...
+from rest_framework.response import Response
+from rest_framework.decorators import api_view, permission_classes
+from api.use_cases.dependencies_versions import (
+    AvailableDynamicDependenciesUseCase,
+)
+
+
+def serialize_output(data: Dict[str, Requirement]):
+    """
+    Prepare de output for the end-point
```
Tansito:
```suggestion
    Prepare the output for the end-point
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -177,6 +177,16 @@ def _verify_credentials(self):
                 "Credentials couldn't be verified."
             ) from reason
 
+    def dependencies_versions(self):
+        """Get the allowed additional dependencies and versions."""
```
Tansito: Similar to other comments, we should unify how we want to describe the functionality, `additional`, `dynamic`, `available`...

gateway/api/use_cases/dependencies_versions.py
korgan00: I think this is better in another PR:
https://github.com/Qiskit/qiskit-serverless/pull/1673

korgan00:
Yes, review it again

gateway/api/v1/route_registry.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    _routes: List[URLPattern] = []
+
+    @classmethod
+    def register(cls, url_path: str, view_func: Callable, name: Optional[str] = None):
+        """
+        register a route to serve as endpoint.
+        """
+        route_name = name or view_func.__name__.replace("_", "-")
```
Tansito, line 22: Is this replacement really needed @korgan00 ? Just curious in case we need to add a regex as a parameter in the URL that requires to use `_`.

gateway/api/v1/endpoint_decorator.py
```diff
@@ -0,0 +1,36 @@
...
+            parts = module_path.split(".views.", 2)
+
+            if len(parts) == 2:
+                relative_path = parts[1]
+            else:
+                relative_path = module_path  # fallback
+
+            generated_name = relative_path.replace(".", "-").replace("_", "-")
+        else:
+            generated_name = name.replace("_", "-")
```
Tansito, line 27: Is this not something similar to what we are doing in `RouteRegistry` in reality?

gateway/api/v1/endpoint_decorator.py
```diff
@@ -0,0 +1,36 @@
+"""
+Endpoint decorator Module
+"""
+
+from functools import wraps
+from typing import Callable
+from api.v1.route_registry import RouteRegistry
+
+
+def endpoint(url_path: str, name: str = None):
```
Tansito, line 10: Should be this decorator outside `v1`?

gateway/api/v1/route_registry.py
```diff
@@ -0,0 +1,34 @@
+"""
+RouteRegistry Module
+"""
+
+from typing import Callable, List, Optional
+from django.urls import path
+from django.urls.resolvers import URLPattern
+
+
+class RouteRegistry:
```
Tansito, line 10: Should be this class outside `v1`?

Tansito (approved):
LGTM, we can review the comments next week with @korgan00

gateway/api/v1/route_registry.py
korgan00: This is the name to do the reverse in tests. It only replace the underscores from the function name in the case where `name` is not defined.
So, I think this shouldn't be a problem.

gateway/api/v1/endpoint_decorator.py
korgan00: Yes, it is similar, but this is more complete, maybe we should take this one.

gateway/api/v1/endpoint_decorator.py
korgan00: I was thinking on it, but, at the moment I dont know how we will manage different versions. So I think we should have a call to think on that.

gateway/api/v1/endpoint_decorator.py
Tansito: I agree let's unify them, then

PR #1670, opened 2025-06-30
Improve probe configuration

Tansito:
### Summary

This PR improves the configuration for the probe end-points to have more time to manage the probes and reduce the possibility that the cluster could restart the API for false positives.

### DIFF
--- a/gateway/probes/views.py
+++ b/gateway/probes/views.py
@@ -1,11 +1,9 @@
-from django.http import HttpResponse
-import os
-import requests
+from django.http import JsonResponse
 
 
 def readiness(request):
-    return HttpResponse("Hi! I'm ready.")
+    return JsonResponse({"status": "alive"})
 
 
 def liveness(request):
-    return HttpResponse("Hi! I'm alive.")
+    return JsonResponse({"status": "ready"})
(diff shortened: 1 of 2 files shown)


gateway/probes/views.py
```diff
@@ -1,11 +1,9 @@
...
 
 
 def readiness(request):
-    return HttpResponse("Hi! I'm ready.")
+    return JsonResponse({"status": "alive"})
 
 
 def liveness(request):
-    return HttpResponse("Hi! I'm alive.")
+    return JsonResponse({"status": "ready"})
```
korgan00: I think `alive` and `ready` are switch

korgan00 (changes requested):

gateway/probes/views.py
Tansito: hehe you are right 😂

Tansito: Solved in [bac2e0d](https://github.com/Qiskit/qiskit-serverless/pull/1670/commits/bac2e0d4732d1297d990ea2a5a6e6b007ab1f32f)

korgan00 (approved):

PR #1671, opened 2025-07-01
Feat/arguments migration to cos

paaragon:
### Summary

This PR changes the way we handle job arguments to store them in COS instead of database.

### Details and comments

We still maintain the old "arguments.serverless" file in case some non-upgraded function is using it but it will be removed in the future


client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -108,13 +109,23 @@ def run(
...
+
         env_vars = {
             **(saved_program.env_vars or {}),
             **{OT_PROGRAM_NAME: saved_program.title},
             **{"PATH": os.environ["PATH"]},
+            **{"DATA_PATH": data_path},
         }
 
-        with open(JOB_ARGUMENTS_FILE, "w", encoding="utf-8") as f:
+        job_id_gateway = os.environ.get(ENV_JOB_ID_GATEWAY)
```
korgan00: Should we check if this is not `None` or maybe have a default value?

client/qiskit_serverless/serializers/program_serializers.py
```diff
@@ -77,12 +77,19 @@ def object_hook(self, obj: Any) -> Any:
...
     """Parses arguments for program and returns them as dict.
-
     Returns:
         Dictionary of arguments.
     """
     arguments = "{}"
-    if os.path.isfile(JOB_ARGUMENTS_FILE):
-        with open(JOB_ARGUMENTS_FILE, "r", encoding="utf-8") as f:
+    job_id_gateway = os.environ.get(ENV_JOB_ID_GATEWAY)
+    data_path = os.environ.get("DATA_PATH", "/data")
```
korgan00: Should we have a constant for `DATA_PATH`?

docker-compose-dev.yaml
```diff
@@ -1,7 +1,7 @@
 # compose config for running images based on local files
 services:
   ray-head:
-    user: "0"
+    user: "root"
```
korgan00, line 4: Is this change finally needed?

gateway/api/ray.py
```diff
@@ -112,24 +113,24 @@ def submit(self, job: Job) -> Optional[str]:
...
-            ) as f:
-                if job.arguments:
+            storage = ArgumentsStorage(job.author.username)
+            arguments = storage.get(job.id) or job.arguments
+            arguments_file = os.path.join(
+                working_directory_for_upload, "arguments.serverless"
+            )
+
+            # DEPRECATED: arguments is now saved in /:username/:jobid,
+            # arguments.serverless is going to be deprecated
```
korgan00, line 124: Is this some that could affect users?

gateway/api/ray.py
```diff
@@ -144,7 +145,7 @@ def submit(self, job: Job) -> Optional[str]:
             )
             ray_job_id = retry_function(
                 callback=lambda: self.client.submit_job(
-                    entrypoint=entrypoint,
+                    entrypoint=f"python {program.entrypoint}",
```
korgan00, line 148: why this change?

gateway/api/services/arguments_storage.py
```diff
@@ -0,0 +1,79 @@
...
+    ARGUMENTS_FILE_EXTENSION = ".json"
+    ENCODING = "utf-8"
+
+    def __init__(self, username: str):
+        self.user_arguments_directory = os.path.join(
+            settings.MEDIA_ROOT, username, "arguments"
+        )
+        os.makedirs(self.user_arguments_directory, exist_ok=True)
+
+    def __get_arguments_path(self, job_id: str) -> str:
```
korgan00: The general private method convention is single underscore. Doble underscore results in [name mangling](https://realpython.com/ref/glossary/name-mangling/) that is not something you need here.

```suggestion
    def _get_arguments_path(self, job_id: str) -> str:
```

korgan00 (changes requested):
LGTM, only minimum comments and questions

gateway/api/services/arguments_storage.py
```diff
@@ -0,0 +1,79 @@
...
+        if not os.path.exists(arguments_path):
+            logger.info(
+                "Arguments file for job ID '%s' not found in directory '%s'.",
+                job_id,
+                self.user_arguments_directory,
+            )
+            return None
+
+        try:
+            with open(arguments_path, "r", encoding="utf-8") as arguments_file:
```
korgan00:
```suggestion
            with open(arguments_path, "r", encoding=self.ENCODING) as arguments_file:
```

gateway/api/ray.py
paaragon: The original value was the variable entrypoint, which was declared as entrypoint = f"python {program.entrypoint}".
Since this was the only place the variable was used, I chose to remove it and write the value directly here. If you prefer the previous version I can revert the change (may be declaring the variable closer to this point)

gateway/api/ray.py
paaragon: Right now, `arguments.serverless` is still created so it won't affect. In the future, this file won't be created anymore so those functions that are still taking arguments from `arguments.serverless` will fail

docker-compose-dev.yaml
paaragon: I checked, and user: "0" and user: root are equivalent, but I prefer seeing root rather than "0"

client/qiskit_serverless/serializers/program_serializers.py
paaragon: you are right

client/qiskit_serverless/core/clients/local_client.py
paaragon: I think if we don't have this value... we shouldn't even execute the job. I would prefer to raise an exception if this doesn't exists. @Tansito what do you think?

gateway/api/ray.py
korgan00: No, its ok. Thank you!

gateway/api/ray.py
korgan00: Ok, lets keep an eye on that.

client/tests/serializers/test_program_serializers.py
```diff
@@ -55,12 +58,31 @@ def test_runtime_service_serialization(self):
...
+        os.makedirs(self.arguments_dir, exist_ok=True)
+
+        self.original_data_path = os.environ.get("DATA_PATH")
+        os.environ["DATA_PATH"] = self.test_data_dir
+
+    def tearDown(self):
+        if self.original_data_path is not None:
+            os.environ["DATA_PATH"] = self.original_data_path
+        elif "DATA_PATH" in os.environ:
+            del os.environ["DATA_PATH"]
```
korgan00: We can use the `DATA_PATH` constant

client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -108,13 +109,23 @@ def run(
                     [sys.executable, "-m", "pip", "install", dependency]
                 )
         arguments = arguments or {}
+        data_path = "./"
+
         env_vars = {
             **(saved_program.env_vars or {}),
             **{OT_PROGRAM_NAME: saved_program.title},
             **{"PATH": os.environ["PATH"]},
+            **{"DATA_PATH": data_path},
```
korgan00: We can use the `DATA_PATH` constant

korgan00 (dismissed):

client/qiskit_serverless/core/clients/local_client.py
Tansito: I agree, this is one of the main variables from the Job and it should be always there. I see fine to raise an exception in that case.

client/qiskit_serverless/serializers/program_serializers.py
Tansito: I suppose this was resolved 👍

docker-compose-dev.yaml
Tansito: I suppose this conversation was also solved 👍

gateway/api/ray.py
Tansito: Yep, we need to test this well before merge it. Let's see if we can stabilize the migration to cloud to merge it and test it in staging.

client/qiskit_serverless/serializers/program_serializers.py
```diff
@@ -77,12 +77,19 @@ def object_hook(self, obj: Any) -> Any:
...
 def get_arguments() -> Dict[str, Any]:
     """Parses arguments for program and returns them as dict.
-
     Returns:
         Dictionary of arguments.
     """
     arguments = "{}"
-    if os.path.isfile(JOB_ARGUMENTS_FILE):
-        with open(JOB_ARGUMENTS_FILE, "r", encoding="utf-8") as f:
+    job_id_gateway = os.environ.get(ENV_JOB_ID_GATEWAY)
```
Tansito, line 85: In fact here is probably where we really want to check if the env_var exists. I would double check just in case and raise an exception.

client/qiskit_serverless/serializers/program_serializers.py
```diff
@@ -77,12 +77,19 @@ def object_hook(self, obj: Any) -> Any:
...
+    data_path = os.environ.get(DATA_PATH, "/data")
```
Tansito, line 92: I would add a comment here that `DATA_PATH` is specifically for tests and local development. In k8s we use the default `/data`, WDYT?

client/qiskit_serverless/serializers/program_serializers.py
```diff
@@ -77,12 +77,19 @@ def object_hook(self, obj: Any) -> Any:
...
+    arguments_dir = f"{data_path}/arguments"
+    arguments_file_path = f"{arguments_dir}/{job_id_gateway}.json"
+
+    os.makedirs(arguments_dir, exist_ok=True)
+
+    if os.path.isfile(arguments_file_path):
```
Tansito: Should we raise an exception if the file didn't exist or is not a file? I think so.

gateway/tests/api/test_ray.py
```diff
@@ -6,7 +6,7 @@
 from unittest.mock import MagicMock
 
 import requests_mock
-from django.conf import settings
+from main import settings
```
Tansito: I think these settings was correctly imported before: https://docs.djangoproject.com/en/5.2/topics/settings/#using-settings-in-python-code

client/qiskit_serverless/core/clients/local_client.py
paaragon: Done

client/qiskit_serverless/core/clients/local_client.py
paaragon: done

Tansito:
Pending from the fix in https://github.com/Qiskit/qiskit-serverless/pull/1699 and the new patch with that fix

Tansito (approved):
Really good job here, @paaragon 😄

PR #1672, opened 2025-07-01
Use `AccountManager` in `IBMServerlessClient`

ElePT:
…rom qiskit-ibm-runtime

### Summary
This PR refactors the IBM serverless client authentication to rely on the `AccountManager` class of `qiskit-ibm-runtime`. This allows to offer the same account resolution logic as the Runtime Client, including loading an account by name, which fixes https://github.com/Qiskit/qiskit-ibm-catalog/issues/56.

### Details and comments
I realized that this class does not have unit tests in the repo. Should we add a test file for the authentication at least?

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -38,6 +38,7 @@ from typing import Optional, List, Dict, Any, Union
 import requests
 from opentelemetry import trace
 from qiskit_ibm_runtime import QiskitRuntimeService
+from qiskit_ibm_runtime.accounts import AccountManager, Account
 
 from qiskit_serverless.core.constants import (
     REQUESTS_TIMEOUT,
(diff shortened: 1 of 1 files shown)


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -577,31 +578,75 @@ def __init__(
...
+            account = AccountManager.get(name=name)
+        elif channel:
+            if channel and channel not in [
+                "ibm_cloud",
+                "ibm_quantum",
+                "ibm_quantum_platform",
+            ]:
+                raise ValueError(
+                    "'channel' can only be 'ibm_cloud', 'ibm_quantum', or 'ibm_quantum_platform"
+                )
```
Tansito: I would try to use the same logic that we are using in other parts: https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/clients/serverless_client.py#L130

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -577,31 +578,75 @@ def __init__(
...
+            if token:
+                account = Account.create_account(
+                    channel=channel,
+                    token=token,
+                    instance=instance,
+                )
+            else:
+                account = AccountManager.get(name=name, channel=channel)
```
Tansito:
```suggestion
                account = AccountManager.get(channel=channel)
```

Naive question here, if `name` exists we are not going reach never to this point, I think. I understand that what we want to do here is get the default account from the channel (?)

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -577,31 +578,75 @@ def __init__(
...
+        elif any([token]):
+            # Let's not infer based on these attributes as they may change in the future.
+            raise ValueError(
+                "'channel' is required if 'token' is specified but 'name' is not."
+            )
+
+        # channel is not defined yet, get it from the AccountManager
+        if account is None:
+            account = AccountManager.get()
```
Tansito, line 661: So, at this point no `token`, no `channel`, no `name` is defined. How the `AccountManager` knows the account to get? Just as curiosity.

Tansito (commented):
In general everything looks good to me. I just added a couple of comments to solve before merge.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: Correct, the name is not used, this is a small oversight from the code in `qiskit-ibm-runtime`.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: The account manager has defaults for the channel and config file name, which are then used to fetch the previously saved information. It's a bit of an obscure way to represent that we are filling up the missing information with defaults. I don't love it: https://github.com/Qiskit/qiskit-ibm-runtime/blob/f2b8e2095667cd4cc50614a2df9d3d0782bf08a5/qiskit_ibm_runtime/accounts/management.py#L60-L63

Tansito (approved):
I could confirm that everything is working from my side. Feel free to merge it when you have that test ready. Good work here! 🎉

ElePT:
Thanks @Tansito. I ended up adding 3 new unit tests, one for `save_account` + instantiation by name (the use case we modified in this PR), and two more to check that wrong instance formats raise a `QiskitServerlessError` instead of a qiskit runtime error. Additionally, I made the `account` a public attribute. Lint complained about using a private attribute in the unit tests and thinking about it, I see value in making it accessible and allowing users to check what account has been loaded in the client (allowing to check, for example, `serverless.account.instance`).

Tansito (approved):
🎉

PR #1675, opened 2025-07-03
Fix incompatibility with old authentication processes

Tansito:
### Summary

Due to the changes in the https://github.com/Qiskit/qiskit-serverless/pull/1666 I detected an incompatibility problem in the old authentication process.

This last PR was basically sending the channel as a way to be able to send the `instance` with `ibm_quantum`. All the tests passed because we were testing that the new functionality was working with the same interface as we had until now but we didn't do any regression test. That's when I detected that in the next use case:
```python
QiskitFunctionsCatalog(channel="ibm_quantum_platform", token="YOUR_TOKEN", instance="YOUR_CRN")
```

We were not sending the channel in previous versions and due to the new logic we were assigning in the backend `ibm_quantum` to this authentication process. This PR should fix that problem as we are validating that if the user is not sending a channel and it's sending the CRN we assume it's an old version and we maintain that logic to support compatibility with previous versions.

### DIFF
--- a/gateway/api/authentication.py
+++ b/gateway/api/authentication.py
@@ -36,10 +36,14 @@ class CustomTokenBackend(authentication.BaseAuthentication):
 
         crn = request.META.get("HTTP_SERVICE_CRN", None)
         channel_header = request.META.get("HTTP_SERVICE_CHANNEL", None)
-        if channel_header is None:
-            # this is to maintain compatibility for older versions
-            # that are not being managing channel
+        # This is to maintain compatibility for older versions:
+        #   - IQP use case: Catalog(token=""")
+        #   - IBM cloud use case: Catalog(channel=", token="", crn="")
+        # Originally we were not sending the channel in <=0.24.0 versions
+        if channel_header is None and crn is None:
             channel_header = Channel.IBM_QUANTUM.value
+        if channel_header is None and crn is not None:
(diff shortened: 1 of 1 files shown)


gateway/api/authentication.py
```diff
@@ -36,10 +36,14 @@ def authenticate(self, request):
 
         crn = request.META.get("HTTP_SERVICE_CRN", None)
         channel_header = request.META.get("HTTP_SERVICE_CHANNEL", None)
-        if channel_header is None:
-            # this is to maintain compatibility for older versions
-            # that are not being managing channel
+        # This is to maintain compatibility for older versions:
+        #   - IQP use case: Catalog(token=""")
+        #   - IBM cloud use case: Catalog(channel=", token="", crn="")
+        # Originally we were not sending the channel in <=0.24.0 versions
```
ElePT:
```suggestion
        # This is to maintain backwards compatibility with previous user patterns.
        # In qiskit-serverless <=0.24.0 , the channel input was not provided to the 
        # authenticator, but inferred from other inputs:
        #   - IQP use case: Catalog(token=""")
        #   - IBM cloud use case: Catalog(channel=", token="", crn="")
        
```

ElePT (approved):
I just added a suggestion to clarify the comment. LGTM too.

PR #1681, opened 2025-07-15
Improve error reporting after failed `run`

ElePT:
### Summary
The current implementation of the `Job` class relies on the function entrypoint implementation to properly handle error reporting and return the error trace as part of the result if something goes wrong. In practice, this is rarely happening, resulting in most function failures to show up as empty `QiskitServerlesException`s. The function logs however often store useful information about the error trace.

This PR adds a mechanism that scans the logs for an exception and returns that line in the case that `result` is `None`. So instead of 

```
QiskitServerlessException                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 result = job.result()
      2 print(result)

File ~/qiskit_workspace/.venv_benchmark/lib/python3.11/site-packages/qiskit_serverless/core/job.py:229, in Job.result(self, wait, cadence, verbose, maxwait)
    226 results = self._job_service.result(self.job_id)
    228 if self.status() == "ERROR":
--> 229     raise QiskitServerlessException(results)
    231 if isinstance(results, str):
    232     try:

QiskitServerlessException: {}
```

Users would see a more actionable message, such as:

```
---------------------------------------------------------------------------
QiskitServerlessException                 Traceback (most recent call last)
Cell In[8], line 1
----> 1 result = job.result()
      2 print(result)

File ~/qiskit_workspace/qiskit-serverless/client/qiskit_serverless/core/job.py:238, in Job.result(self, wait, cadence, verbose, maxwait)
    234         raise QiskitServerlessException(results)
    235     else:
    236         # If no result returned (common with import errors), 
    237         # try to match on error trace in logs to point to source of error
--> 238         raise QiskitServerlessException(self.filtered_logs(include=r'(?i)error|exception'))
    240 if isinstance(results, str):
    241     try:

QiskitServerlessException: ModuleNotFoundError: No module named 'circuit_function.utils'; 'circuit_function' is not a package
```

### Details and comments

1. If the function developers don't add any logging capabilities, this improvement will not change that. For example, some functions simply return `"No available logs. Contact the provider"`. But this feature is still helpful for custom function development, where we don't have any issues reporting logs.

2. I am considering adding the full error trace instead of just the exception, although this is anyways available in the logs, it might help to see something like:

```
INFO job_manager.py:531 -- Runtime env is setting up.\nTraceback (most recent call last):\n  File "/tmp/ray/session_2025-07-15_11-14-11_975056_1/runtime_resources/working_dir_files/_ray_pkg_545018ec1c39fc2c/circuit_function/circuit_function.py", line 21, in <module>\n    from circuit_function.utils import (\n  File "/tmp/ray/session_2025-07-15_11-14-11_975056_1/runtime_resources/working_dir_files/_ray_pkg_545018ec1c39fc2c/circuit_function/circuit_function.py", line 21, in <module>\n    from circuit_function.utils import (\nModuleNotFoundError: No module named \'circuit_function.utils\'; \'circuit_function\' is not a package\n'
```

Again, this would only be visible if the developers choose to log these error traces.

### DIFF
--- a/client/qiskit_serverless/core/clients/local_client.py
+++ b/client/qiskit_serverless/core/clients/local_client.py
@@ -163,7 +163,25 @@ class LocalClient(BaseClient):
 
     def filtered_logs(self, job_id: str, **kwargs):
         """Return filtered logs."""
-        raise NotImplementedError
+        all_logs = self.logs(job_id=job_id)
+        included = ""
+        include = kwargs.get("include")
+        if include is not None:
+            for line in all_logs.split("\n"):
+                if re.search(include, line) is not None:
+                    included = included + line + "\n"
+        else:
+            included = all_logs
+
+        excluded = ""
+        exclude = kwargs.get("exclude")
+        if exclude is not None:
+            for line in included.split("\n"):
+                if line != "" and re.search(exclude, line) is None:
+                    excluded = excluded + line + "\n"
+        else:
+            excluded = included
+        return excluded
 
     #########################
     ####### Functions #######
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -228,12 +228,12 @@ class Job:
         if self.status() == "ERROR":
             if results:
                 raise QiskitServerlessException(results)
-            else:
-                # If no result returned (common with import errors),
-                # try to match on error trace in logs to point to source of error
-                raise QiskitServerlessException(
-                    self.filtered_logs(include=r"(?i)error|exception")
-                )
+
+            # If no result returned (common with import errors),
+            # try to match on error trace in logs to point to source of error
+            raise QiskitServerlessException(
+                self.filtered_logs(include=r"(?i)error|exception")
+            )
 
         if isinstance(results, str):
             try:
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -183,8 +183,8 @@ class Job:
     def filtered_logs(self, **kwargs) -> str:
         """Returns logs of the job.
         Args:
-            include: rex expression finds match in the log line to be included
-            exclude: rex expression finds match in the log line to be excluded
+            include: regex expression finds matching line in the log to be included
+            exclude: regex expression finds matching line in the log to be excluded
         """
         return self._job_service.filtered_logs(job_id=self.job_id, **kwargs)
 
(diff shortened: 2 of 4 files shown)


ElePT:
> Now the tests are returning the current problems. Two things to fix:
> 
> * `job.run` doesn't exist (I'm not sure what are you trying to check here)
> * you are not waiting to finish the job to check the status. I would recommend you to use `job.result` because it's going to wait until the job finishes (correctly or incorrectly)

Thank you, this was me multitasking and not having the setup to test locally, so I didn't pay too much attention to the test. Indeed, `job.run` makes 0 sense, I meant `function.run`.

client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -163,7 +163,25 @@ def logs(self, job_id: str):
...
+
+        excluded = ""
+        exclude = kwargs.get("exclude")
+        if exclude is not None:
+            for line in included.split("\n"):
+                if line != "" and re.search(exclude, line) is None:
+                    excluded = excluded + line + "\n"
+        else:
+            excluded = included
+        return excluded
```
Tansito, line 184: It took me a bit to follow the logic so I tried to clean it a bit. Let me know what do you think about this proposal, Elena:

```py
logs = self.logs(job_id=job_id).splitlines()

include_pattern = kwargs.get("include")
exclude_pattern = kwargs.get("exclude")

if include_pattern:
    include_re = re.compile(include_pattern)
    logs = [line for line in logs if include_re.search(line)]
    
if exclude_pattern:
    exclude_re = re.compile(exclude_pattern)
    logs = [line for line in logs if line and not exclude_re.search(line)]

return "\n".join(logs) + "\n"
```

We first split `logs` so we don't need to do `all_logs.split("\n")` in every include or exclude.
I also added `re.compile` instead to use `re.search` directly so we compile the regular expression once instead to  do it in every step of the for (sorry, I have a personal fight with @korgan00 about regular expressions and their optimizations, too many discussions about them haha).

The part that I doubt the most is the return. I was not sure if we should do: `return "\n".join(logs) + "\n"` or just `return "\n".join(logs)` but basically I tried to reduce the number of ifs and make the code a bit easy to follow.

It's probably that I also missed a use case or something. Happy to discuss the solution.

ElePT: To be honest, all I did here is copy-paste the `filtered_logs` implementation from the `ServerlessClient`. But I like your suggestions. Maybe in the interest of time we could merge the current implementation and I could open a follow-up to make both methods more efficient?

Tansito: Thanks Elena! Yep, sounds good to me.

Tansito (approved):
LGTM, thanks for the improvement Elena  🚀

PR #1695, opened 2025-08-01
Improve get jobs and get job logic with new search parameters

paaragon:
### Summary

This PR adds the new jobs and provider_jobs parameters in the client:
- status
- created_after
- function_name (only for jobs)

And in the backend the changes applied are:
- Migrated `/jobs` and `/jobs/provider` endpoints to new architecture
- Added new filters: `status`, `function` and `created_after` date
- Added tests to cover new filters
- Added a new `ProgramSummarySerializer` so we return less info about the program in the job lists

### Details and comments

After trying multiple solutions, I think the best pattern is to return a QuerySet in the repository but not propagate it to the endpoint. The use-case should return a business object, not the QuerySet.

I created the `create_paginated_response` utility method that should be used in future endpoints to maintain consistency between responses.

Also, the previous response was returning different response types. I removed the non-paginated response because it wasn't being used in the client.

### DIFF
--- a/gateway/tests/api/test_job.py
+++ b/gateway/tests/api/test_job.py
@@ -214,7 +214,7 @@ class TestJobApi(APITestCase):
         self.client.force_authenticate(user=user)
         provider = "default"
         function = "Docker-Image-Program"
-        created_after = "2023-02-02T00:00:00.000000Z"
+        created_after = "2023-02-02"
 
         jobs_response = self.client.get(
             reverse("v1:jobs-provider-list"),


gateway/api/repositories/jobs.py
korgan00, line 1: clean and neat

gateway/api/use_cases/get_jobs.py
```diff
@@ -0,0 +1,54 @@
...
+        created_after: Optional[datetime] = None,
+        function_name: Optional[str] = None,
+    ):
+        self.user = user
+        self.limit = limit
+        self.offset = offset
+        self.type_filter = type_filter
+        self.status = status
+        self.created_after = created_after
+        self.function_name = function_name
```
korgan00: I have two comments here:

 - To remove the "too many positional arguments" maybe you can use the `JobRepository` >`JobFilter` `dataclass`. I think it will also be more readable.
 - I was migrating the `files` endpoint and I put the args in the execute instead of the init. I think it is good to have consensus on this. In my opinion, most of this parameters are not a "permanent" state, are more related with the action. Also it simplifies the code, we don't have to set object variables, just use them.

What do you think?

gateway/api/use_cases/get_provider_jobs.py
```diff
@@ -0,0 +1,79 @@
...
+logger = logging.getLogger("gateway.use_cases.get_jobs")
+
+
+class ProviderNotFoundException(Exception):
+    """Provider not found or access denied."""
+
+
+class FunctionNotFoundException(Exception):
+    """Function not found or access denied."""
+
```
korgan00: In the files migration i created a generic `NotFoundException` in the domain folder: https://github.com/Qiskit/qiskit-serverless/pull/1696/files#diff-e6da6866795b4c157d9b9e887eb27622f036a9dc512d30e5097a301df1d5001c

Also I created an additional decorator to quickly manage the possible `NotFoundException`s (and other that can be created later) and create the corresponding `Response`. The only cons is that you have to set the message when you create the exception to inform about the issue.

gateway/api/v1/views/get_provider_jobs.py
```diff
@@ -0,0 +1,167 @@
...
+            {"message": f"Provider {input_data['provider']} doesn't exist."},
+            status=status.HTTP_404_NOT_FOUND,
+        )
+    except FunctionNotFoundException:
+        return Response(
+            {
+                "message": f"Qiskit Function {input_data['provider']}/{input_data['function_name']} doesn't exist."  # pylint: disable=line-too-long
+            },
+            status=status.HTTP_404_NOT_FOUND,
+        )
```
korgan00: In the files migration i created a generic `NotFoundException` in the domain folder: https://github.com/Qiskit/qiskit-serverless/pull/1696/files#diff-e6da6866795b4c157d9b9e887eb27622f036a9dc512d30e5097a301df1d5001c

Also I created an additional decorator to quickly manage the possible `NotFoundException`s (and other that can be created later) and create the corresponding `Response`. The only cons is that you have to set the message when you create the exception to inform about the issue.

gateway/api/v1/views/get_jobs.py
```diff
@@ -0,0 +1,166 @@
...
+
+    return {
+        "user": user,
+        "limit": limit,
+        "offset": offset,
+        "type_filter": type_filter,
+        "status": status_filter,
+        "created_after": created_after,
+        "function_name": function_name,
+    }
```
korgan00: I think this is a bit complex to read. Maybe a `Serializer` class fits better in here.

gateway/api/v1/views/get_provider_jobs.py
```diff
@@ -0,0 +1,167 @@
...
+
+    return {
+        "user": user,
+        "limit": limit,
+        "offset": offset,
+        "status": status_filter,
+        "created_after": created_after,
+        "provider": provider,
+        "function_name": function_name,
+    }
```
korgan00: Same here, maybe we can try with a `Serlializer` class

gateway/api/v1/views/utils.py
korgan00, line 1: cool

gateway/api/v1/urls.py
```diff
@@ -46,4 +46,4 @@
     basename=v1_views.CatalogViewSet.BASE_NAME,
 )
 
-urlpatterns = router.urls + RouteRegistry.get()
+urlpatterns = RouteRegistry.get() + router.urls
```
korgan00: nice catch

korgan00 (approved):
Good job! this is nice.
Since I was working in the [files migration](https://github.com/Qiskit/qiskit-serverless/pull/1696) in parallel, naturally we didn't developed same solution. I have some lines to discuss.

If you have some time, take a look to the PR and review it because you probably have some comments for me :)

gateway/api/use_cases/get_jobs.py
paaragon: you are right, I though about passing JobFilters to the use case and discard it but it will be a better solution than passing all the individual filters.

I also like the idea of moving the arguments to the execute method.

korgan00:
the `jobs_new` folder should be renamed to jobs when we delete the `jobs.py` file. Then we should be able to delete also all the endpoints `name` properties because it will be auto-calculated

gateway/api/repositories/functions.py
```diff
@@ -222,6 +222,44 @@ def get_function_by_permission(
...
         return self.get_user_function(author=user, title=function_title)
 
+    def get_function(
+        self,
+        function_title: str,
+        provider_name: Optional[str],
+    ) -> Optional[Function]:
+        """
+        This method returns the specified function if the user is
+        the author of the function or it has a permission.
```
Tansito: I think we need to update the description as this method doesn't check the permission.

Tansito: Also the arguments need to be updated.

gateway/tests/api/test_job.py
```diff
@@ -137,9 +187,71 @@ def test_job_provider_list_ok(self):
...
+        self.assertEqual(jobs_response.data.get("count"), 1)
+        self.assertEqual(jobs_response.data.get("results")[0]["status"], "SUCCEEDED")
+
+    def test_job_provider_list_created_after(self):
+        """Tests job provider list filtered by created_after."""
+        user = models.User.objects.get(username="test_user_2")
+        self.client.force_authenticate(user=user)
+        provider = "default"
+        function = "Docker-Image-Program"
+        created_after = "2023-02-02T00:00:00.000000Z"
```
Tansito: If I can be honest I don't think the user will introduce us this format in the date. Can we change the tests with `created_after` with a more realistic date? Like: `2023-02-02`, for example.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -192,10 +192,29 @@ def dependencies_versions(self):
...
+        Returns:
+            List[Job]: List of Job objects matching the criteria.
+        """
         limit = kwargs.get("limit", 10)
         kwargs["limit"] = limit
         offset = kwargs.get("offset", 0)
         kwargs["offset"] = offset
+        status = kwargs.get("status", None)
+        kwargs["status"] = status
+        created_after = kwargs.get("created_after", None)
```
Tansito, line 214: I would probably validate here also that the `created_after` is a valid date with [fromisoformat](https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat).

Tansito (changes requested):
I just left a couple of comments plus we need to add also in the `client` the same filters as we added for the `jobs` method.

client/qiskit_serverless/core/clients/serverless_client.py
korgan00: Per offline conversations, we will delay this for another PR.

Tansito (approved):
Now LGTM! Thanks @paaragon , @korgan00 !

PR #1696, opened 2025-08-01
Migrate files endpoints

korgan00:
### Summary

Remove the ViewSet and create a file for each endpoint.

### Details and comments

In general, there is a lot of duplicated code. I commented some ideas about reducing redundancy but we want to finish all the migrations to look at the duplicated code and think in the solution that better fits.

The InputSerializer is a Django Serializer. It has a [`.create()`](https://www.django-rest-framework.org/api-guide/serializers/#saving-instances) method that I thought it was related with DB but it is not (not mandatory). In case that there are too many args in the use case, we can pack them into a dataclass.


gateway/api/v1/views/files/delete.py
```diff
@@ -0,0 +1,109 @@
+"""
+API V1: Delete file end-point.
+"""
+# pylint: disable=duplicate-code
+from typing import cast
+from drf_yasg import openapi
+from drf_yasg.utils import swagger_auto_schema
+from django.contrib.auth.models import AbstractBaseUser
```
Tansito: As discussed internally it's preferable to use `AbstractUser` as it contains most of the data from the user. Not only the authentication data.

Let's change it in all the views.

gateway/api/v1/views/files/download.py
```diff
@@ -0,0 +1,119 @@
...
+            type=openapi.TYPE_STRING,
+            required=False,
+        ),
+    ],
+    responses={
+        status.HTTP_200_OK: openapi.Response(
+            description="Requested file",
+            content_type="application/octet-stream",
+            schema=openapi.Schema(type=openapi.TYPE_FILE),
+            examples={
```
Tansito: We can remove this example as it's not accurate.

gateway/api/v1/views/files/provider_delete.py
```diff
@@ -0,0 +1,117 @@
...
+            type=openapi.TYPE_STRING,
+            required=True,
+        ),
+    ],
+    responses={
+        status.HTTP_200_OK: openapi.Response(
+            description="Requested file",
+            content_type="application/octet-stream",
+            schema=openapi.Schema(type=openapi.TYPE_FILE),
+            examples={
```
Tansito: Let's use the `delete` definition.

gateway/api/v1/views/files/provider_download.py
```diff
@@ -0,0 +1,119 @@
...
+        openapi.Parameter(
+            "provider",
+            openapi.IN_QUERY,
+            description="Provider name",
+            type=openapi.TYPE_STRING,
+            required=True,
+        ),
+    ],
+    responses={
+        status.HTTP_200_OK: openapi.Response(
```
Tansito, line 77: Let's use the `download` definition.

PR #1704, opened 2025-08-15
🏗️ Update action versions in client-verify

Tansito:
### Summary

This PR fixes the problem found in the actions in the PRs: #1696 and #1703

### DIFF
--- a/.github/workflows/client-verify.yaml
+++ b/.github/workflows/client-verify.yaml
@@ -26,9 +26,9 @@ jobs:
         working-directory: ./client
 
     steps:
-      - uses: actions/checkout@4.2.2
+      - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@5.3.0
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
@@ -40,7 +40,7 @@ jobs:
           echo TOX_ENV=$tox_env >> $GITHUB_ENV
       - name: Install tox
         run: |
-          pip install tox==4.16.0
+          pip install tox>=4.0.0
       - name: Run styles check
         run: tox -elint
       - name: Test using tox environment


.github/actions/helm-lint/action.yaml
```diff
@@ -5,19 +5,19 @@ runs:
...
       with:
-        version: v3.12.1
+        version: v3.18.5
 
     # Python is required because `ct lint` runs Yamale (https://github.com/23andMe/Yamale) and
     # yamllint (https://github.com/adrienverge/yamllint) which require Python
-    - uses: actions/setup-python@5.3.0
+    - uses: actions/setup-python@v5
       with:
-        python-version: '3.10'
```
ElePT, line 16: Interesting that this was using python 3.10

.github/actions/icr-build-and-push-images/action.yaml
```diff
@@ -1,35 +0,0 @@
-name: 'Build and push images'
```
ElePT, line 1: Out of curiosity, do you know when this action was used? Happy to clean it up if it's no longer used.

.github/workflows/client-verify.yaml
```diff
@@ -38,9 +38,9 @@ jobs:
           tox_env="-epy${pver/./}"
           echo tox_env
           echo TOX_ENV=$tox_env >> $GITHUB_ENV
-      - name: Install tox
+      - name: Install requirements
         run: |
-          pip install tox==4.16.0
+          pip install -r requirements.txt -r requirements-dev.txt
```
ElePT: Naive question. If we only need tox here, does it make to install the full requirements? Won't this make CI much slower?

.github/actions/helm-lint/action.yaml
Tansito: Yep, I didn't remember it to be honest 😂

.github/actions/icr-build-and-push-images/action.yaml
Tansito: Some months ago we were doing the entire release from the public repository and we were publishing the images from here. Now, due to our internal requirements we needed to move this process to our internal repository (1PL, you know).

.github/workflows/client-verify.yaml
Tansito: Good question, I think we could. My idea here was that, how we are using later on `tox -e lint` and `tox -e py311`, If I'm not wrong, those commands install the requirements anyway. So:
- With this change we prevent to have packages around the code that we can forget to update in the future.
- Those dependencies are going to be installed by `tox` anyway so I don't expect an increase on time

But we can test it. These tests are currently taking 5/6 minutes. Let's do the change and review the new times.

Tansito: It looks like we reduced the times to 4 minutes so your assumption was correct, @ElePT. Thank you! Let's apply your suggestion and let me some minutes to review the rest of the actions in case I applied a similar change somewhere else.

Tansito: Done! Thanks for the review, Elena!

PR #1705, opened 2025-08-19
Remove `ibm_quantum` channel to support runtime 0.41+

ElePT:


Tansito (commented):
I see everything fine, I only one comment related with the [Channel enum](https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/enums.py#L13). Should we just remove from the client also? It doesn't affect to the `gateway` I think and this logic it's only used in the client so I think it's safe to remove it also.

ElePT:
I have applied your suggestion and also modified the pin on the `qiskit-ibm-runtime` package to `>=0.40.1, <1`. 0.40.1 is the first working version post-IQP-migration that I think makes sense to support for consistency with the channels supported in this serverless version.

PR #1707, opened 2025-08-29
DNM: WIP runtime wrapper

ElePT:


tests/staging/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+        token=os.environ["QISKIT_IBM_TOKEN"],
+        instance=os.environ["QISKIT_IBM_INSTANCE"],
+    )
+    # # Initialize serverless folder for current user
+    # function = QiskitFunction(
+    #     title="hello-world",
+    #     entrypoint="hello_world.py",
+    #     working_dir=resources_path,
+    # )
+    # serverless.upload(function)
```
korgan00: I think this can be safely deleted

tests/staging/test_staging.py
```diff
@@ -0,0 +1,45 @@
...
+
+        function = QiskitFunction(
+            title="test-runtime-wrapper",
+            entrypoint="pattern_with_runtime_wrapper.py",
+            working_dir=resources_path,
+        )
+        staging_client.upload(function)
+
+        functions = {f.title: f for f in staging_client.list()}
+        my_pattern_function = functions.get("test-runtime-wrapper")
```
korgan00: `staging_client.upload` returns the uploaded function as `RunnableFunction` and `staging_client.function` allows you to retrieve it directly. Why list and map them?

tests/staging/test_staging.py
```diff
@@ -0,0 +1,45 @@
...
+
+resources_path = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "source_files"
+)
+
+
+class TestFunctionsStaging:
+    """Test class for integration testing in staging environment."""
+
+    @mark.order(1)
```
korgan00: you are not ordering dependencies in the command, so you can delete this line.

korgan00 (changes requested):

tests/staging/test_staging.py
ElePT: Yep, this is code I didn't clean up from my tests.

gateway/api/views/jobs.py
```diff
@@ -124,13 +123,11 @@ def add_runtimejob(
...
+        """List RuntimeJob objects associated to Job"""
         job = self.get_object()
-        runtimejobs = RuntimeJob.objects.filter(job=job)
-        ids = []
-        for runtimejob in runtimejobs:
-            ids.append(runtimejob.runtime_job)
-        return Response(json.dumps(ids))
+        runtimejobs = job.runtime_jobs.all()
+        serializer = RuntimeJobSerializer(runtimejobs, many=True)
+        return Response({"runtime_jobs": serializer.data})
```
ElePT: I went multiple times back and forth with these endpoints. In the end decided to keep both the add and the list one for various reasons. I ran into limitations in result serialization, where the automatic pipeline would not return a dict for "runtime_jobs", instead it would return a list. I tried to figure this out but I think I need to improve my django knowledge. In any case:

- Making formatting changes is easier if we have a dedicated endpoint
- Keeping the add/list also adds transparency of what the endpoint is for in my opinion
- It creates separation of concerns. I don't see the need to have everything together in `retrieve`

releasenotes/notes/add-runtime-integration-6dcfa0fe966daea1.yaml
```diff
@@ -0,0 +1,82 @@
+---
+features:
```
ElePT, line 2: I tried adding a release note, and realized that I am only used to communicating interface changes. I think we should also have API release notes but I am not sure how these should be structured.

releasenotes/notes/add-runtime-integration-6dcfa0fe966daea1.yaml
```diff
@@ -0,0 +1,82 @@
...
+      session_id = job.runtime_sessions()[0]
+      # out = "session_id_1"
+      session_job_ids =  job.runtime_jobs(session_id)
+      # in this example, only job ids 1 and 3 correspond to session_id_1:
+      # out = ["job_id_1", "job_id_3"]
+
+
+upgrade:
+  - |
+    Enhanced ``job.stop()`` behavior to use the newly introduced features. ``job.stop()`` now attempts
```
Tansito: If I remember correctly the interface `job.stop` was deprecated in favor of `job.cancel` to follow a similar UI as Runtime 👀

releasenotes/notes/add-runtime-integration-6dcfa0fe966daea1.yaml
Tansito: Yes, let's discuss the better way to handle this!

tests/docker/conftest.py
```diff
@@ -46,15 +46,6 @@ def set_up_serverless_client():
         host=os.environ.get("GATEWAY_HOST", connection_url),
         instance=os.environ.get("GATEWAY_INSTANCE", "an_awesome_crn"),
     )
-
-    # Initialize serverless folder for current user
-    function = QiskitFunction(
```
Tansito, line 51: I think the error from the integration test comes from this change. Why are we removing this `hello-world`? (a naive question, I don't remember in detail this configuration for the test, maybe @korgan00 has more context).

ElePT: yeah, that makes sense, I originally had 2 of these conftests but then decided to move the new tests to the docker folder and use the "smallest" conftest. Up until this point I hadn't understood what this part of the code was for. I will undo this change.

releasenotes/notes/add-runtime-integration-6dcfa0fe966daea1.yaml
ElePT: Hmmm I did miss that, I am surprised I never saw the warnings running my tests...

ElePT: Aha, so only `job.stop` was deprecated, not `service.stop(job_id)`. Should we add a new `service.cancel()` for consistency? Keeping track of these naming changes can be challenging.

ElePT: Anyways, addressed in [afe300b](https://github.com/Qiskit/qiskit-serverless/pull/1707/commits/afe300b8be00219af7854eb36f9427e86717f7ca).

tests/docker/conftest.py
korgan00: The hello world was about initializing something. I think it ensures that some registers are properly initialized before doing anything else, I thought that was fixed at some point but maybe not...

client/qiskit_serverless/core/clients/ray_client.py
```diff
@@ -159,6 +159,10 @@ def filtered_logs(self, job_id: str, **kwargs) -> str:
         """Return filtered logs."""
         raise NotImplementedError
 
+    def runtime_jobs(self, job_id: str, runtime_session: Optional[str] = None):
```
korgan00: Why do we have this here? We are not having it in the `BaseClient`. Maybe this is something that can live in the `ServerlessClient` only and we don't have the `NotImplementedError`.

What do you think? @ElePT, @Tansito

client/qiskit_serverless/core/clients/local_client.py
```diff
@@ -88,6 +88,10 @@ def job(self, job_id: str) -> Optional[Job]:
     def jobs(self, **kwargs) -> List[Job]:
         return [job["job"] for job in list(self._jobs.values())]
 
+    def runtime_jobs(self, job_id: str, runtime_session: Optional[str] = None):
```
korgan00: same here

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -382,9 +383,20 @@ def stop(self, job_id: str, service: Optional[QiskitRuntimeService] = None):
...
+                )
+                data = {
+                    "service": json.dumps(service, cls=QiskitObjectsEncoder),
+                }
+            except InvalidAccountError:
+                warnings.warn(
+                    "No QiskitRuntimeService can be associated to the given token and instance. "
+                    "Continuing without a QiskitRuntimeService."
+                )
+                data = {"service": json.dumps(None, cls=QiskitObjectsEncoder)}
```
korgan00:
```suggestion
        if not service:
            try:
                service = QiskitRuntimeService(
                    channel=self.channel, instance=self.instance, token=self.token
                )
            except InvalidAccountError:
                warnings.warn(
                    "No QiskitRuntimeService can be associated to the given token and instance. "
                    "Continuing without a QiskitRuntimeService."
                )
                service = None

        data = {
            "service": json.dumps(service, cls=QiskitObjectsEncoder),
        }
```

I think this is easier to read.

I dont know why, but it seems that it is not taking lines properly. It takes from beginning of the function `stop` until `response_data = ...`

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -426,6 +439,48 @@ def logs(self, job_id: str):
...
+            )
+        )
+
+        if runtime_session:
+            return [
+                job.get("runtime_job")
+                for job in response_data.get("runtime_jobs", [])
+                if job.get("runtime_session") == runtime_session
+            ]
+        return [job.get("runtime_job") for job in response_data.get("runtime_jobs", [])]
```
korgan00, line 461: What is the case where we don't have a `runtime_session` and we have `runtime_jobs`?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -426,6 +439,48 @@ def logs(self, job_id: str):
...
+                ),
+                timeout=REQUESTS_TIMEOUT,
+            )
+        )
+        out_sessions = []
+        for job in response_data.get("runtime_jobs", []):
+            session = job.get("runtime_session")
+            if session and not session in out_sessions:
+                out_sessions.append(session)
+        return out_sessions
```
korgan00: I think this can be simplified like this:

```suggestion
        runtime_jobs = response_data.get("runtime_jobs", [])
        out_sessions = {job.get("runtime_session") for job in runtime_jobs}
        return list(dict.fromkeys(out_sessions))
```

client/qiskit_serverless/core/job.py
```diff
@@ -102,6 +103,10 @@ def result(self, job_id: str) -> Any:
     def logs(self, job_id: str) -> str:
         """Return logs."""
 
+    @abstractmethod
+    def runtime_jobs(self, job_id: str, runtime_session: Optional[str] = None):
```
korgan00: Can you add the return type?

client/qiskit_serverless/core/job.py
```diff
@@ -180,6 +185,16 @@ def logs(self) -> str:
         """Returns logs of the job."""
         return self._job_service.logs(self.job_id)
 
+    def runtime_jobs(self, runtime_session: Optional[str] = None):
+        """Returns associated runtime jobs if any."""
+        return self._job_service.runtime_jobs(
+            self.job_id, runtime_session=runtime_session
+        )
+
+    def runtime_sessions(self):
```
korgan00: We should add this method to the `JobService` and add the return type annotation

gateway/api/use_cases/jobs/stop.py
```diff
@@ -29,45 +29,88 @@ def execute(self, job_id: UUID, service: str) -> str:
...
             job.save(update_fields=["status"])
+            status_messages.append("Job has been stopped.")
+        else:
+            status_messages.append("Job already in terminal state.")
 
+        # Unit tests send a None directly, but the client sends a serialized None
+        if service_str:
+            service = json.loads(service_str, cls=json.JSONDecoder)
+        else:
+            service = None
```
korgan00:
```suggestion
        service = None
        if service_str:
            service = json.loads(service_str, cls=json.JSONDecoder)
```

I use to use this form. Not a correction, just suggestion. I try to avoid `else` when possible, it use to improve readability.

gateway/api/use_cases/jobs/stop.py
```diff
@@ -29,45 +29,88 @@ def execute(self, job_id: UUID, service: str) -> str:
...
+                            )
+                        except RuntimeInvalidStateError:
+                            status_messages.append(
+                                f"Runtime job {job_id_str} could not be canceled (invalid state)."
+                            )
+                else:
+                    status_messages.append(
+                        f"Runtime job {job_id_str} not found in runtime service. "
+                        "Check that credentials used to authenticate match."
+                    )
```
korgan00:
```suggestion
        if runtime_jobs and service:
            service_config = service["__value__"]
            qiskit_service = QiskitRuntimeService(**service_config)
            qiskit_api_client = qiskit_service._get_api_client()

            stopped_sessions = []
            for runtime_job_entry in runtime_jobs:
                job_id_str = runtime_job_entry.runtime_job
                session_id_str = runtime_job_entry.runtime_session
                job_instance = qiskit_service.job(job_id_str)

                if not job_instance:
                    status_messages.append(
                        f"Runtime job {job_id_str} not found in runtime service. "
                        "Check that credentials used to authenticate match."
                    )
                    continue
                if session_id_str:
                    if session_id_str in stopped_sessions:
                        continue
                    try:
                        qiskit_api_client.cancel_session(
                            session_id_str
                        )
                        status_messages.append(
                            f"Canceled runtime session: {session_id_str} and associated runtime jobs."
                        )
                        stopped_sessions.append(session_id_str)
                    except Exception as e:
                        status_messages.append(
                            f"Runtime session {session_id_str} could not be canceled. Exception: {e}"
                        )
                else:
                    try:
                        job_instance.cancel()
                        status_messages.append(
                            f"Canceled runtime job [{job_id_str}]."
                        )
                    except RuntimeInvalidStateError:
                        status_messages.append(
                            f"Runtime job {job_id_str} could not be canceled (invalid state)."
                        )
```
Some code readability improvements.
I think this function is being too large, lets move this into an auxiliary function.

gateway/api/use_cases/jobs/stop.py
```diff
@@ -29,45 +29,88 @@ def execute(self, job_id: UUID, service: str) -> str:
...
-                if not was_running:
-                    message = "Job was already not running."
+                if was_running:
+                    status_messages.append("Ray job was running and has been stopped.")
+                else:
+                    status_messages.append("Ray job was already not running.")
             else:
                 logger.warning(
-                    "Compute resource is not accessible %s", job.compute_resource
+                    "Ray compute resource is not accessible %s", job.compute_resource
```
korgan00: Should we give the user information about what utility are we using to run jobs? Should we mention `Ray`?

gateway/api/schedule.py
```diff
@@ -25,12 +25,13 @@
...
-    ):
-        logger.debug("Job [%s] will be run on GPU nodes", job.id)
-        job.gpu = True
+    if job.program:
+        if (
+            job.program.provider
+            and job.program.provider.name in gpujobs["gpu-functions"].keys()
+        ):
+            logger.debug("Job [%s] will be run on GPU nodes", job.id)
+            job.gpu = True
```
korgan00:
```suggestion
    if (
        job.program
        and job.program.provider
        and job.program.provider.name in gpujobs["gpu-functions"].keys()
    ):
        logger.debug("Job [%s] will be run on GPU nodes", job.id)
        job.gpu = True
```

client/qiskit_serverless/core/clients/ray_client.py
ElePT: I tried a couple of time removing these, but I always ended up trapped in lint complaints that made me add the methods to the base class and having to add the not implemented error. The source issue was the job method, that called the service method implicitly, and hence required the method to be implemented in the base class. I can try again though, see if I missed something.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: There are 3 execution modes in runtime:
- session mode -> where users have a session id
- batch mode -> where I believe the batch is considered a "session" and users also have a session id
- job mode -> jobs are sent individually and have no session id

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: I am not sure this refactor covers the cases where `job.get("runtime_session")` returns `None`. I can test it though.

gateway/api/use_cases/jobs/stop.py
ElePT: Maybe not, but as a user "compute resource is not accessible" was not very informative. Is it talking about the quantum computer? the serverless resource? We can maybe use "cluster" instead of "ray"? Something to indicate that we are not talking about qiskit runtime.

client/qiskit_serverless/core/clients/serverless_client.py
korgan00: then you can do:
```
out_sessions = {job.get("runtime_session") for job in runtime_jobs if job}
```

Isn't it?

gateway/api/use_cases/jobs/stop.py
```diff
@@ -20,54 +22,120 @@ class StopJobUseCase:
...
+                    qiskit_api_client,
+                    status_messages,
+                    stopped_sessions,
+                )
+
+        self._stop_ray_job_if_active(job, status_messages)
+
+        return " ".join(status_messages)
 
+    @staticmethod
```
korgan00: You can make them non-static since the UseCase should be instanced. Also you can make `status_messages` and `stopped_sessions` a class attribute to prevent spread them.

gateway/api/use_cases/jobs/stop.py
Tansito: I would say "cluster" is also probably not accurate. How about this:

- "Qiskit Serverless Job was running and has been stopped."
- "Qiskit Serverless Job is not accessible %s"

My reasons behind it. It doesn't matter what we use behind the scenes, everything is wrapped as a Qiskit Serverless Job. So if we use Ray or CE it doesn't matter, it's just that we cannot access to our Job (managed by Ray or other technology in the future).

gateway/api/use_cases/jobs/stop.py
```diff
@@ -20,54 +22,120 @@ class StopJobUseCase:
...
+                else:
+                    status_messages.append("Ray job was already not running.")
             else:
                 logger.warning(
-                    "Compute resource is not accessible %s", job.compute_resource
+                    "Ray compute resource is not accessible %s", job.compute_resource
                 )
-
-        return message
+                status_messages.append("Ray compute resource not accessible.")
```
korgan00:
```suggestion
        if not (job.compute_resource and job.compute_resource.active):
            return
            
        job_handler = get_job_handler(job.compute_resource.host)
        if job_handler is None:
            logger.warning(
                "Ray compute resource is not accessible %s", job.compute_resource
            )
            status_messages.append("Ray compute resource not accessible.")
            return
            
        was_running = job_handler.stop(job.ray_job_id)
        if was_running:
            status_messages.append("Ray job was running and has been stopped.")
        else:
            status_messages.append("Ray job was already not running.")
```

To prevent hadoken code :)

gateway/api/use_cases/jobs/stop.py
ElePT: now that we are at it, why class attributes and not instance attributes?

korgan00: yeah, instance attributes, sorry my bad on naming :)

gateway/api/views/jobs.py
```diff
@@ -124,13 +123,11 @@ def add_runtimejob(
 
     @_trace
     @action(methods=["GET"], detail=True)
-    def list_runtimejob(
+    def list_runtime_jobs(
```
Tansito: Now that we are refactoring our API I would like to propose a change in these names to follow a bit more the REST principals:

- POST: runtime-jobs  
- GET: runtime-jobs

So the client would be:
```
requests.get(
                f"{self.host}/api/{self.version}/jobs/{job_id}/runtime-jobs/",
```

and

```
requests.post(
                f"{self.host}/api/{self.version}/jobs/{job_id}/runtime-jobs/",
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -406,6 +415,7 @@ def result(self, job_id: str):
                 headers=get_headers(
                     token=self.token, instance=self.instance, channel=self.channel
                 ),
+                params={"with_result": "true"},
```
Tansito, line 418: Thanks @ElePT ❤️

client/qiskit_serverless/core/job.py
```diff
@@ -397,6 +418,37 @@ def _map_status_to_serverless(status: str) -> Tuple[str, Union[str, None]]:
...
+    Returns:
+        QiskitRuntimeService wrapped into a ServerlessRuntimeService instance that allows
+        to associate serverless job ids to runtime job and sesison ids.
+    """
+
+    return ServerlessRuntimeService(
+        channel=channel or os.environ["QISKIT_IBM_CHANNEL"],
+        instance=instance or os.environ["QISKIT_IBM_INSTANCE"],
+        token=token or os.environ["QISKIT_IBM_TOKEN"],
+        url=url or None,
```
Tansito: How would you see to use here:

```
url=url or os.environ["QISKIT_IBM_URL"],
```

I'm proposing this change because it's possible that we are going to want to test a Function from a partner in staging and we will be able to use our staging environment from Runtime without any change in the Functions (QISKIT_IBM_URL should contain the url for staging).

gateway/api/use_cases/jobs/stop.py
```diff
@@ -19,55 +21,110 @@ class StopJobUseCase:
...
+            self.status_messages.append(
+                "QiskitRuntimeService not found, cannot stop runtime jobs."
+            )
+        elif not runtime_jobs:
+            self.status_messages.append(
+                "No active runtime job ID associated with this serverless job ID."
+            )
+        else:
+            service_config = service["__value__"]
             qiskit_service = QiskitRuntimeService(**service_config)
```
Tansito, line 61: I have a question here @ElePT , do you know if we initialize like this runtime do we store something in disk? Or is it totally stateless?

gateway/api/v1/serializers.py
```diff
@@ -202,15 +202,33 @@ class Meta(serializers.RunJobSerializer.Meta):
...
+        fields = ["runtime_job", "runtime_session"]
+
+
 class JobSerializer(serializers.JobSerializer):
     """
     Job serializer first version. Include basic fields from the initial model.
     """
 
     program = ProgramSerializer(many=False)
+    runtime_jobs = RuntimeJobSerializer(many=True, read_only=True)
```
Tansito: I understood from you comment [here](https://github.com/Qiskit/qiskit-serverless/pull/1707/files#r2416929286) that you were not going to return the runtime jobs in the job serializer. Is this a logic that you forgot to remove from your tests or is it something that you are definitely using?

gateway/api/views/jobs.py
Tansito: Sounds good to me 😄

gateway/tests/api/test_v1_program.py
```diff
@@ -444,34 +444,76 @@ def test_upload_provider_function_with_description(self):
...
 
         user = models.User.objects.get(username="test_user")
         self.client.force_authenticate(user=user)
-        programs_response = self.client.post(
-            "/api/v1/jobs/8317718f-5c0d-4fb6-9947-72e480b8a348/add_runtimejob/",
+
+        job_id = "8317718f-5c0d-4fb6-9947-72e480b8a348"
+
+        response = self.client.post(
+            f"/api/v1/jobs/{job_id}/add_runtime_jobs/",
```
Tansito: Until now we were using the `reverse` method to obtain the different urls for the tests like [here](https://github.ibm.com/IBM-Q-Software/qiskit-serverless/blob/main/gateway/tests/api/test_job.py#L34). Said that, I would like to know your opinion, @ElePT , @korgan00 . I have different feelings about the use of `reverse` as it hides too much for me the URL.

It was the recommended way in Django for the testing approach but I'm open to start using URLs if we decide so. I would just like to standardize the way we manage it.

gateway/api/use_cases/jobs/stop.py
ElePT: It should be stateless, the moment `execute` finishes there is no longer a handle to the runtime service, and nothing should be stored in disk (credentials are stored in a json file only if you call `QiskitRuntimeService.save_account()`, which we are not calling here).

client/qiskit_serverless/core/clients/serverless_client.py
korgan00: Maybe i'm a bit fussy but "with_result" sounds a bit colloquial. Maybe "include_result" sound more formal?

client/qiskit_serverless/core/job.py
ElePT: Making this change essentially makes the URL a mandatory parameter, which I am not super convinced about. When `url` is None, the qiskit runtime service automatically defaults to the production URL, which is I believe the expected default. If we change this to `QISKIT_IBM_URL`, we are asking users (function developers) to explicitly provide the URL for production (either through an env variable or through the `url` parameter). I don't expect developers to even know the production URL though, and asking them to set the variable to None doesn't seem like a very transparent user pattern to me...

gateway/api/views/jobs.py
ElePT: This sounds good, I will review though if we should use dash or underscore, I don't remember how other endpoints are set up.

gateway/tests/api/test_v1_program.py
korgan00: Yeah, I agree with stantarization here. I see `reverse` like a good method to manage small url changes but also sounds like typical thing that is not so useful as it seems. Anyway, I prefer to keep it at this moment.

gateway/api/views/jobs.py
ElePT: Done in [c097ff7](https://github.com/Qiskit/qiskit-serverless/pull/1707/commits/c097ff72c6c29cf07d9ed842b8b3c91560bb5e88)

gateway/tests/api/test_v1_program.py
ElePT: I personally prefer the full url, but mostly because I am not familiar with `reverse` and I find it very hard to infer the url at first sight. I can change to reverse here for consistency.

client/qiskit_serverless/core/job.py
Tansito: I see your point, other proposal then:

```
url=url or os.environ.get("QISKIT_IBM_URL", None),
```

In this case if the environment variable doesn't exist it uses `None` following the same use case that you are proposing but also it allows us to modify the URL if we need it like for staging, for example, or in a pentest environment.

gateway/tests/api/test_v1_program.py
Tansito: For me is also easier to read with the full url to be honest and I agree it's hard to infer it. We can keep it and make the change in the rest of tests in a future PR.

gateway/api/views/jobs.py
Tansito: There is no specific standard for - or _ I think. How we are using _ in other end-points I think it's fine if we maintain it, looks good to me, thanks!

gateway/api/views/jobs.py
```diff
@@ -99,35 +99,32 @@ def get_queryset(self):
...
-    def list_runtime_jobs(
-        self, request, pk=None
-    ):  # pylint: disable=invalid-name,unused-argument
-        """List RuntimeJob objects associated to Job"""
-        job = self.get_object()
-        runtimejobs = job.runtime_jobs.all()
-        serializer = RuntimeJobSerializer(runtimejobs, many=True)
-        return Response({"runtime_jobs": serializer.data})
+        # POST: associate runtime jobs to serverless job
+        if request.method == "POST":
```
Tansito, line 111: The good thing is that, in a next PR, when we move this end-point under v1 following the new architecture, we are going to be able to separate this logic in two files. For me it's fine as it is by now. Thanks!

client/qiskit_serverless/core/clients/serverless_client.py
Tansito: Maybe, but to be honest, it's not very transparent to the user and change it it's not trivial because we will need to support backward compatibility so... Not a big deal I would say here.

gateway/api/v1/serializers.py
ElePT: Yes, I am not sure I am using this, I will do another pass at the code to clean it up.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: applied a version of it in [e987056](https://github.com/Qiskit/qiskit-serverless/pull/1707/commits/e9870567128d5fd9918a7f79d6bf04b60403e603)

PR #1708, opened 2025-09-04
Feat migrate all jobs endpoints

paaragon:
### Summary

This PR migrates all the jobs endpoints to the new architecture with use cases.

Endpoints:

- GET jobs
- GET provider-jobs
- GET jobs/{id}
- POST jobs/{id}/result
- GET jobs/{id}/logs
- PATCH jobs/{id}/sub_status
- POST jobs/{id}/stop

### DIFF
--- a/Dockerfile-ray-node
+++ b/Dockerfile-ray-node
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9-minimal:9.6-1755695350
+FROM registry.access.redhat.com/ubi9-minimal:9.6-1758184547
 
 ARG DYNAMIC_DEPENDENCIES_FILE
 
--- a/client/requirements.txt
+++ b/client/requirements.txt
@@ -2,7 +2,7 @@ ray[default]>=2.30, <3
 requests>=2.32.2, <3
 importlib-metadata>=5.2.0, <9
 qiskit[qpy-compat]>=1.4, <3
-qiskit-ibm-runtime>=0.40.1, <1
+qiskit-ibm-runtime~=0.41.0
 # Make sure ray node and notebook node have the same version of cloudpickle
 cloudpickle==2.2.1
 tqdm>=4.66.3, <5
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -6,7 +6,7 @@ FROM registry.access.redhat.com/ubi9-micro:9.6-1754467928 AS base
 
 # BUILD image using UBI 9 where the dependencies that
 # require installing with a package manager will be installed
-FROM registry.access.redhat.com/ubi9:9.6-1756915113 AS build
+FROM registry.access.redhat.com/ubi9:9.6-1758184894 AS build
 ARG MICRO_IMAGE_DIR
 
 # Copy the BASE image into the BUILD image
--- a/gateway/requirements.txt
+++ b/gateway/requirements.txt
@@ -19,7 +19,7 @@ drf-yasg>=1.21.10, <2
 # Django dependency, but we need a newer version (IBMQ#246)
 sqlparse>=0.5.0, <1
 qiskit[qpy-compat]>=1.4, <3
-qiskit-ibm-runtime~=0.40.1
+qiskit-ibm-runtime~=0.41.0
 tzdata>=2024.1
 django-cors-headers>=4.4.0, <5
 whitenoise>=6.7.0, <7


gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -0,0 +1,47 @@
...
+            NotFoundError: If the job does not exist.
+
+        Returns:
+            str: Job logs if accessible, otherwise a message indicating no logs are available.
+        """
+        job = self.jobs_repository.get_job_by_id(job_id)
+        if job is None:
+            raise NotFoundError(f"Job [{job_id}] not found")
+
+        if job.program and job.program.provider:
```
Tansito, line 41: I would use here the `can_access` job policy. I think they are the same but correct me if I'm wrong.

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -0,0 +1,47 @@
...
+            provider_groups = set(job.program.provider.admin_groups.all())
+            user_groups = set(user.groups.all())
+            if provider_groups & user_groups:
+                return job.logs
+            return NO_LOGS_MSG
```
Tansito: I think we should return a `Forbidden` here if the user doesn't have access.

gateway/api/use_cases/jobs/stop.py
```diff
@@ -0,0 +1,73 @@
...
+                if not was_running:
+                    message = "Job was already not running."
+            else:
+                logger.warning(
+                    "Compute resource is not accessible %s", job.compute_resource
+                )
+
+        return message
+
+    def _get_runtime_job(self, job):
```
Tansito: We should refactor this in its own repository called `RuntimeJob`. In general any use of the ORM from Django must be in a repository.

gateway/api/v1/views/utils.py
```diff
@@ -87,3 +91,62 @@ def parse_positive_int(param: str, default: int) -> int:
     if value < 0:
         raise ValueError(f"{param} must be non-negative")
     return value
+
+
+def error_schema(example_msg: str, description: str = "Error response"):
```
Tansito: I think we can create a `swagger_utils` or something like that managing these methods for our swagger definition.

gateway/api/use_cases/jobs/get_logs.py
paaragon: Should we return a NotFound error instead? we are returning it in similar cases

Tansito: My two cents here, in this specific case I think it has sense to return a `Forbidden`. We are using `NotFound` to obfuscate the fact that if exist or not the resource. But in this case you know that it exists (due to the job) but it just you don't have access to it.

gateway/api/v1/views/jobs_new/stop.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    message = serializers.CharField()
+
+    def create(self, validated_data):
+        raise NotImplementedError
+
+    def update(self, instance, validated_data):
+        raise NotImplementedError
+
```
korgan00: I think you can just delete this.

gateway/api/v1/views/jobs_new/stop.py
```diff
@@ -0,0 +1,78 @@
...
+    service = serializers.CharField(required=False, default=None)
+
+    def create(self, validated_data):
+        """Not implemented - this serializer is for validation only."""
+        raise NotImplementedError
+
+    def update(self, instance, validated_data):
+        """Not implemented - this serializer is for validation only."""
+        raise NotImplementedError
```
korgan00: I think you can just delete this.

paaragon: Done

gateway/api/v1/views/jobs_new/stop.py
paaragon: Done

gateway/api/use_cases/jobs/get_logs.py
paaragon: Done

gateway/api/use_cases/jobs/stop.py
paaragon: Done!

gateway/api/v1/views/utils.py
paaragon: Done! I created the swagger_utils.py file

Tansito: Correct me If I'm wrong here @paaragon, you created the new file but you didn't remove the old one, did you?

gateway/api/v1/views/utils.py
```diff
@@ -87,3 +91,28 @@ def parse_positive_int(param: str, default: int) -> int:
...
+            type=openapi.TYPE_OBJECT,
+            properties={
+                "message": openapi.Schema(type=openapi.TYPE_STRING, example=example_msg)
+            },
+            required=["message"],
+        ),
+    )
+
+
+class SanitizedCharField(serializers.CharField):
```
Tansito: Also @paaragon , @korgan00 how do you see to move this utility specific for serializers under `serializer_utils` or something like that? To start cleaning a bit this utils file.

paaragon: Thanks @Tansito I have apply your suggestions.

I think a serialises utils is a good idea

gateway/api/access_policies/providers.py
```diff
@@ -29,10 +29,10 @@ def can_access(user, provider: Provider) -> bool:
         """
 
         user_groups = user.groups.all()
-        admin_groups = provider.admin_groups.all()
-        has_access = any(group in admin_groups for group in user_groups)
-        if not has_access:
+        admin_groups = provider.admin_groups.all() if provider is not None else []
```
korgan00:
```suggestion
        admin_groups = provider.admin_groups.all() if provider else []
```

I think falsy is always a good condition in this case. The other classic falsys are `false` and 0 that I think is a good solution for them too.

gateway/api/access_policies/providers.py
```diff
@@ -29,10 +29,10 @@ def can_access(user, provider: Provider) -> bool:
...
+        user_is_admin = any(group in admin_groups for group in user_groups)
```
korgan00: This is not a recommendation or correction, you dont have to do anything here.

I just found another way to do this with `set intersection`. I think it is interesting to know about this methods and I wanted to highlight it here.

```python
a = [1, 2, 3, 4, 5]
b = [1, 3, 5, 6]
set(a).intersection(b) # [1, 3, 5]
```

gateway/api/utils.py
```diff
@@ -66,33 +65,32 @@ def retry_function(  # pylint:  disable=too-many-positional-arguments
...
-    result = None
-    name = function_name or inspect.stack()[1].function
-
-    while run < num_retries and not success:
-        run += 1
-
-        logger.debug("[%s] attempt %d", name, run)
+    name = function_name or getattr(callback, "__name__", "<callback>")
 
+    for attempt in range(1, num_retries + 1):
```
korgan00, line 70: Good refactor

gateway/api/access_policies/providers.py
paaragon: Thanks @korgan00

gateway/api/utils.py
paaragon: Thanks! 😄

gateway/api/access_policies/providers.py
paaragon: I think your suggestion is way more readable so I'm applying it

Tansito:
Really cool @paaragon ! Thank you so much 😄

PR #1712, opened 2025-09-11
Remove Catalog end-points

Tansito:
### Summary

This PR removes the different end-points that were managing the catalog information in the old IQP platform. Due to the migration to the new IQP these end-points are not used anymore and we should be able to remove them now.

### DIFF
--- a/client/requirements-dev.txt
+++ b/client/requirements-dev.txt
@@ -1,5 +1,5 @@
 coverage>=5.5
-pylint>=2.17.7
+pylint>=2.17.7, <4
 nbqa>=1.7.1
 treon>=0.1.4
 pytest>=6.2.5
--- a/docs/requirements-doc.txt
+++ b/docs/requirements-doc.txt
@@ -1,5 +1,5 @@
 coverage>=5.5
-pylint>=2.17.7
+pylint>=2.17.7, <4
 nbqa>=1.7.1
 treon>=0.1.4
 pytest>=6.2.5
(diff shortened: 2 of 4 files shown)


gateway/api/authentication.py
korgan00, line 1: I `CustomTokenBackend` still used?

gateway/api/v1/views/__init__.py
```diff
@@ -4,6 +4,5 @@
 Specification for V1 version of the API.
 """
 
-from .catalog import CatalogViewSet
```
korgan00, line 7: If we are not using this, we can remove all the specific serializers from `api/serializers.py` and `api/v1/serializers.py`

gateway/api/authentication.py
Tansito: ~~Yes, it's for when we want to run something locally so we can use `awesome_token`. I will double check it tomorrow just in case I'm confusing with another thing, @korgan00~~

gateway/api/v1/views/__init__.py
Tansito: Good catch, agree. Curious why the linter didn't say anything 🤔

gateway/api/authentication.py
Tansito: Nvm, I totally missed the reference. The `CustomTokenBackend` is the logic that we follow to authenticate the user so yes, it's yet being used.

gateway/api/v1/views/__init__.py
Tansito: Done in [1f8ae37](https://github.com/Qiskit/qiskit-serverless/pull/1712/commits/1f8ae374a8999af19fcf195c95f15fd385bded81)!

PR #1720, opened 2025-09-22
Add provider to the api filters for jobs

korgan00:

### DIFF
--- a/tests/docker/test_docker.py
+++ b/tests/docker/test_docker.py
@@ -381,14 +381,12 @@ class TestFunctionsDocker:
         assert len(offset_jobs) == 1
         assert offset_jobs[0].job_id == job_1_1.job_id
 
-
         while job_1_1.status() == "RUNNING":
             sleep(1)
         while job_1_2.status() == "RUNNING":
             sleep(1)
         while job_2.status() == "RUNNING":
             sleep(1)
-            
+
         succeeded_jobs = serverless_client.jobs(status="SUCCEEDED")
         assert len(succeeded_jobs) >= 3
-


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -213,8 +213,16 @@ def jobs(self, **kwargs) -> List[Job]:
         kwargs["status"] = status
         created_after = kwargs.get("created_after", None)
         kwargs["created_after"] = created_after
+
         function_name = kwargs.get("function_name", None)
```
Tansito: I know that this is not the purpose of the PR but I forgot it from the previous one. I would want to propose change `function_name` to just `function` and send the Qiskit Function in that parameter. I think that that's what we are doing in the rest of the methods. WDYT @korgan00 ?

Tansito: Sorry, maybe I was not clear enough, @korgan00 . I wanted to say like this: https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/clients/serverless_client.py#L534

Sending the QiskitFunction instance and obtaining the info from it.

client/qiskit_serverless/core/function.py
```diff
@@ -235,10 +235,7 @@ def jobs(self):
                     f"Function validation failed. Validation errors:\n {error_string}",
                 )
 
-        jobs = self._run_service.jobs(
-            title=self.title,
-            provider=self.provider,
-        )
+        jobs = self._run_service.jobs(function=self)
```
Tansito: Question @korgan00 , should not we have here also **kargs? (I'm not sure)

korgan00: Lets create a couple of tests

tests/docker/test_docker.py
```diff
@@ -313,11 +312,53 @@ def test_execute_functions_in_parallel(self, serverless_client: ServerlessClient
...
+        while job_1.status() == "QUEUED" or job_1.status() == "INITIALIZING":
             sleep(1)
 
         while job_2.status() == "QUEUED" or job_2.status() == "INITIALIZING":
             sleep(1)
 
         assert job_1.status() == "RUNNING"
         assert job_2.status() == "RUNNING"
+
+    def test_get_filtered_jobs(self, serverless_client: ServerlessClient):
```
Tansito, line 327: This looks good @korgan00 , can we also add tests to filter by the function, the status and the date? 🙏

korgan00: When you get jobs through the function you already filter by function.
I will check the status and date.

Tansito (approved):
I love you @korgan00 , thank you so much for this ❤️

PR #1721, opened 2025-09-22
Filter by status correctly

korgan00:


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -210,6 +211,7 @@ def jobs(self, **kwargs) -> List[Job]:
         offset = kwargs.get("offset", 0)
         kwargs["offset"] = offset
         status = kwargs.get("status", None)
+        status, _ = _map_status_from_serverless(status)
```
Tansito: Correct me if I'm wrong here @korgan00 . We are not mapping status **FROM** serverless if not **TO** serverless, aren't we?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -260,6 +262,7 @@ def provider_jobs(self, function: QiskitFunction, **kwargs) -> List[Job]:
         kwargs["function"] = function.title
         kwargs["provider"] = function.provider
         status = kwargs.get("status", None)
+        status, _ = _map_status_to_serverless(status)
```
Tansito: One question here, @korgan00 . What happens exactly if the status is `None`? Sorry it's not clear for me looking at the code.

korgan00:
```python
    try:
        status_translation = INVERSE_STATUS_MAP[status]
    except KeyError:
        return status, None
```
It should rise a `KeyError` and we return the provided status parameter. I don't know if it is the desired behaviour, it is mimicking the other function (`...from_serverless`).

Tansito: In this case we may want to return Jobs with any status. So in that case `status` would be `None`. If it's `None` we probably should not the map and that's all.

I the original use case I think it has sense because we are going to receive always the `status` from the API but in this case the client will not always provide the `status`.

korgan00: But, if the status is none, the exception will return None, None. That is the desired behaviour right?

PR #1724, opened 2025-09-29
Logs Storage service

Tansito:
### Summary

This PR implementes a new Storage for logs and applies a little refactor to improve the way we are managing the paths so we can avoid errors like #1715 

### Details and comments

- Created `LogsStorage` class to manage the access to the new logs
  - Paths are going to be the same as `FileStorage` but with and additional subpath:
      - For user logs in Custom Functions: `/{username}/logs/{job_id}.log`
      - For user logs in Partner Functions: `/{username}/{provider_name}/{function_title}/logs/{job_id}.log`
      - For partner logs in Partner Functions: `/{provider_name}/{function_title}/logs/{job_id}.log`
- Created `PathBuilder` to manage in the different storages the generated paths
- Moved storages under `storage` folder

### DIFF
--- a/gateway/api/services/storage/logs_storage.py
+++ b/gateway/api/services/storage/logs_storage.py
@@ -34,14 +34,14 @@ class LogsStorage:
         function_title: str,
         provider_name: Optional[str],
     ) -> None:
-        self.sub_path = PathBuilder.get_sub_path(
+        self.sub_path = PathBuilder.sub_path(
             working_dir=working_dir,
             username=username,
             function_title=function_title,
             provider_name=provider_name,
             extra_sub_path=self.PATH,
         )
-        self.absolute_path = PathBuilder.get_absolute_path(
+        self.absolute_path = PathBuilder.absolute_path(
             working_dir=working_dir,
             username=username,
             function_title=function_title,
--- a/gateway/api/services/storage/path_builder.py
+++ b/gateway/api/services/storage/path_builder.py
@@ -83,7 +83,7 @@ class PathBuilder:
         return sanitize_file_path(path)
 
     @staticmethod
-    def get_sub_path(
+    def sub_path(
         working_dir: WorkingDir,
         username: str,
         function_title: str,
@@ -126,7 +126,7 @@ class PathBuilder:
         return sub_path
 
     @staticmethod
-    def get_absolute_path(
+    def absolute_path(
         working_dir: WorkingDir,
         username: str,
         function_title: str,
@@ -134,7 +134,8 @@ class PathBuilder:
         extra_sub_path: Optional[str],
     ) -> str:
         """
-        This method returns the aboslute path for the required interaction.
+        This method returns the aboslute path for the required interaction
+        and it creates it if it doesn't exist.
 
         Args:
             working_dir (WorkingDir): configuration for the generation of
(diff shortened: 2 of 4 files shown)


gateway/api/services/storage/path_builder.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        path = os.path.join(provider_name, function_title)
+
+        if extra_sub_path is not None:
+            path = os.path.join(path, extra_sub_path)
+
+        return sanitize_file_path(path)
+
+    @staticmethod
+    def get_sub_path(
```
korgan00: I think `get` here is not very meaningful, it use to be used in a normal class to retrieve an attribute and this is an static class.
In a builder the common words are: `build`, `resolve`, `make`, `compose`...
But in this case the class itself has the `build` word and it could be redundant, so maybe we can just remove the `get`: `PathBuilder.sub_path` or `PathBuilder.absolute_path`.
This also applies for `absolute_path`.

gateway/api/services/storage/path_builder.py
```diff
@@ -0,0 +1,169 @@
...
+class PathBuilder:
+    """
+    This class manages the logic for the user and provider paths generated in the Object Storage
+    """
+
+    @staticmethod
+    def __get_user_sub_path(
+        username: str,
+        function_title: str,
+        provider_name: Optional[str],
```
korgan00, line 26: Maybe we can just send the whole function right here as well as we are doing in the client. This is not a blocker, just an idea. Same in the `FileStorage`, I reviewed and everywhere where we have `FileStorage`, we have the `function`.

gateway/api/services/storage/logs_storage.py
```diff
@@ -0,0 +1,92 @@
...
+        )
+        self.absolute_path = PathBuilder.get_absolute_path(
+            working_dir=working_dir,
+            username=username,
+            function_title=function_title,
+            provider_name=provider_name,
+            extra_sub_path=self.PATH,
+        )
+
+    def _get_log_path(self, job_id: str) -> str:
```
korgan00:
```suggestion
    def _get_logs_path(self, job_id: str) -> str:
```
I think logs it is usually plural always because you use to have multiple log lines.

gateway/api/services/storage/__init__.py
korgan00, line 1: how about exporting all the storages here so we can import it as a package?
`from api.services.storage import FileStorage, WorkingDir, LogsStorage`

korgan00 (changes requested):
Nice! I have some small comments

gateway/api/services/storage/__init__.py
Tansito: Done [3218530](https://github.com/Qiskit/qiskit-serverless/pull/1724/commits/32185301c6c32c014530ed083c69104dd1202f19)!

gateway/api/services/storage/logs_storage.py
Tansito: Done [8d14e9b](https://github.com/Qiskit/qiskit-serverless/pull/1724/commits/8d14e9b6734167fb324a19c0d394cc8a3cc90623)!

gateway/api/services/storage/path_builder.py
Tansito: After an internal talk we are going to delay this refactor for another PR to reduce the noise and also leave some work for the new members that will come as a `good-first-issue`.

gateway/api/services/storage/path_builder.py
Tansito: Done [40932ec](https://github.com/Qiskit/qiskit-serverless/pull/1724/commits/40932eceab572b16bad877435fdd2834f642cd4b)!

PR #1726, opened 2025-10-01
🏗️ Fix conflict naming between serializers

Tansito:
### Summary

This PR pretends to fix a problem from #1708 due to that Serializers with the same name between different views were generating errors when we try to access to our swagger documentation.


Tansito:
Closing this proposal in favor of a more consensual solution with the team in #1727

PR #1727, opened 2025-10-06
fix: swagger serializer name collision

paaragon:
### Summary

This PR fixes the `/swagger` endpoint that was broken because of serialises name collision.

### Details and comments

In the future, every time we write an input serializer, we should add:

```
 class Meta:
        """Meta class to define input serializer name"""

        ref_name = "{Your Endpoint}InputSerializer"
```


korgan00:
Maybe we can do it only when we have collisions instead of spread all the code with this.

PR #1733, opened 2025-10-14
Refactor endpoints post Runtime integration

ElePT:
### Details and comments
Follow-up of #1707

### DIFF
--- a/gateway/api/v1/views/jobs/runtime_jobs.py
+++ b/gateway/api/v1/views/jobs/runtime_jobs.py
@@ -10,6 +10,7 @@ from uuid import UUID
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework import permissions, serializers, status
 from rest_framework.decorators import api_view, permission_classes
+from rest_framework.exceptions import MethodNotAllowed
 from rest_framework.request import Request
 from rest_framework.response import Response
 


gateway/api/use_cases/jobs/associate_runtime_jobs.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+            message = (
+                f"RuntimeJob object [{runtime_job}] created "
+                f"for serverless job id [{job_id}]."
+            )
+        except Exception as e:
+            message = (
+                f"Failed to create RuntimeJob object "
+                f"[{runtime_job}] for job id [{job_id}]. Error: {e}"
+            )
```
korgan00, line 52: Maybe we can log it too to have information about this exception

```suggestion
            message = (
                f"Failed to create RuntimeJob object "
                f"[{runtime_job}] for job id [{job_id}]. Error: {e}"
            )
            logger.error(message)
```

korgan00 (changes requested):
Code wise it looks good to me.
Should we control something about permissions in the `Post`?

Tansito:
@ElePT I think you said this was ready for review, right? Can I click on "Ready of review"?

gateway/api/v1/views/jobs/runtime_jobs.py
```diff
@@ -0,0 +1,111 @@
...
+    responses={
+        status.HTTP_200_OK: serializers.Serializer(),
+        **standard_error_responses(
+            bad_request_example="Got empty `runtime_job` field. Please, specify `runtime_job`.",
+            not_found_example="Job [XXXX] not found",
+            unauthorized_example="Authentication credentials were not provided or are invalid.",
+        ),
+    },
+)
+@endpoint("jobs/<uuid:job_id>/runtime_jobs", name="jobs-runtime-jobs")
```
Tansito, line 89: @korgan00 , @ElePT how the end-point is identified with this I would like to divide this end-point in two, one for GET and another one for POST (two separated files).

Tansito (changes requested):

gateway/api/v1/views/jobs/runtime_jobs.py
Tansito: NVM, it looks like it's not possible in Django Rest Framework 🤦

gateway/api/use_cases/jobs/associate_runtime_jobs.py
```diff
@@ -0,0 +1,56 @@
...
+        job = self.jobs_repository.get_job_by_id(job_id)
+        if job is None:
+            raise NotFoundError(f"Job [{job_id}] not found")
+
+        try:
+            RuntimeJob.objects.create(
+                job=job,
+                runtime_job=runtime_job,
+                runtime_session=runtime_session,
+            )
```
avilches: I know accessing to the Django orm in use cases is tempting, but if we are following the Repository approach, all the database accesses (that means, any call to `Model.objects.whatever()`) should be done through a repository.

avilches: In this specific case, this use case is very simple. I guess you can get rid of the use case and call to the repository directly from the endpoint. Having one single call to one repository in the controller is acceptable.

gateway/api/use_cases/jobs/get_runtime_jobs.py
```diff
@@ -0,0 +1,36 @@
...
+
+        Raises:
+            NotFoundError: If the job does not exist or access is denied.
+        """
+        job = self.jobs_repository.get_job_by_id(job_id)
+
+        if job is None:
+            raise NotFoundError(f"Job [{job_id}] not found")
+
+        return job.runtime_jobs.all()
```
avilches: Same, just create (or reuse) a repository `RuntimeRepository` with this logic and call it from the endpoint. In the repository, you can safely call to any `Model.objects.whatever()` method you want :)

gateway/api/use_cases/jobs/associate_runtime_jobs.py
ElePT: This is a good point, this case is so simple that I struggled to find the balance between following the approach of other endpoints and adding too much boilerplate. I am fine updating it.

gateway/api/views/jobs.py
```diff
@@ -3,6 +3,7 @@
```
Tansito: Initially now that we removed the end-points from this file we should be able to remove it entirely 👍

ElePT: ok, I will give it a try

korgan00 (approved):
If the rest of comments are adressed I approve :)

gateway/api/v1/views/jobs/runtime_jobs.py
```diff
@@ -0,0 +1,111 @@
...
+        runtime_job = serializer.validated_data.get("runtime_job")
+        runtime_session = serializer.validated_data.get("runtime_session")
+        message = AssociateRuntimeJobsUseCase().execute(
+            job_id, runtime_job, runtime_session
+        )
+        return Response({"message": message})
+
+    # GET
+    out_runtime_jobs = GetRuntimeJobsUseCase().execute(job_id)
+    return Response(serialize_output(out_runtime_jobs))
```
avilches:
```suggestion
    if request.method == "GET":
        out_runtime_jobs = GetRuntimeJobsUseCase().execute(job_id)
        return Response(serialize_output(out_runtime_jobs))
        
    raise MethodNotAllowed(request.method)
```

Tansito: Good suggestion, apart from this I saw everything ok

PR #1734, opened 2025-10-15
Improve the management for deactivated users

Tansito:
### Summary

This PR pretends to improve the management that the authentication flow is doing to guarantee the access to the service. It introduces a new access policy for the Users that double check the `is_active` parameter under the `User` object.

### DIFF
--- a/releasenotes/notes/accepted-users-0cd36a473605a3ff.yaml
+++ b/releasenotes/notes/accepted-users-0cd36a473605a3ff.yaml
@@ -1,0 +1,5 @@
+---
+upgrade:
+  - |
+    Added a new AccessPolicy for User model to be able to allow or
+    not the access to the service by the `is_active` parameter.


gateway/api/access_policies/users.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    The main objective of this class is to manage the access for the user
+    to the application.
+    """
+
+    @staticmethod
+    def can_access(user: type[AbstractUser]) -> bool:
+        """
+        Checks if the user has access to the application.
+        For that it will need to be active.
```
ElePT: Small clarification (I wasn't sure if it was referring to the user or the application until I saw the code):

```suggestion
        For that, the user will need to be active.
```

gateway/api/use_cases/authentication.py
```diff
@@ -71,6 +72,10 @@ def execute(self) -> Optional[type[AbstractUser]]:
 
         access_groups = authentication_service.get_groups()
         quantum_user = self.user_repository.get_or_create_by_id(user_id=user_id)
+        if not UserAccessPolicies.can_access(quantum_user):
+            raise exceptions.AuthenticationFailed(
+                "Your user was deactivated. Please contact to support for reactivaton."
+            )
```
ElePT: Which support is this referring to? the function provider support? IBM support?

```suggestion
                "Your user was deactivated. Please contact support for reactivaton."
            )
```

gateway/tests/api/authentication/custom_token/test_ibm_quantum_platform.py
```diff
@@ -104,3 +107,78 @@ def test_default_authentication_workflow(
...
+    @responses.activate
+    def test_default_authentication_workflow_with_inactive_account(
+        self,
+        mock_get_api_keys_details: MagicMock,
+        mock_get_resource_instance: MagicMock,
+        mock_list_access_groups: MagicMock,
+    ):
+        """
+        This test verifies the entire flow of the custom token authentication
+        for a not verified user.
```
ElePT:
```suggestion
        for a deactivated user.
```

gateway/tests/api/authentication/custom_token/test_ibm_quantum_platform.py
```diff
@@ -104,3 +107,78 @@ def test_default_authentication_workflow(
...
+                "account_id": "abc18abcd41546508b35dfe0627109c4",
+            },
+            status=200,
+        )
+
+        custom_auth = CustomTokenBackend()
+        request = MagicMock()
+        request.META = {
+            "HTTP_SERVICE_CHANNEL": "ibm_quantum_platform",
+            "HTTP_AUTHORIZATION": "Bearer AWESOME_TOKEN",
```
ElePT, line 170: I am always unsure of when the bearer token is used vs the API key (which we call token in the runtime context).

korgan00 (approved):
LGTM

gateway/api/use_cases/authentication.py
Tansito: Good catch, IBM support I would say.

gateway/tests/api/authentication/custom_token/test_ibm_quantum_platform.py
Tansito: We can discuss it in detail tomorrow but basically the Bearer is an HTTP standard for authentication headers. You usually build your header with the format: `Bearer <token>` from a HTTP perspective. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Authentication).

The Bearer schema it's not the only one we can follow for the authentication, another common ones are the Basic and Digest, for example. But I would say Bearer is the most common.

gateway/api/use_cases/authentication.py
Tansito: Done [91f0ad6](https://github.com/Qiskit/qiskit-serverless/pull/1734/commits/91f0ad6e12d26d737c60dda2c2de6821e2e591cd)!

PR #1735, opened 2025-10-15
Update the limits for logs storage

Tansito:
### Summary

This PR propose a new limitation in the way we are managing logs. With the purpose to reduce the size in our database and until we can finish the migration I would like to reduce logs storage to only those logs with less than 50MB.

### DIFF
--- a/gateway/api/ray.py
+++ b/gateway/api/ray.py
@@ -120,7 +120,7 @@ class JobHandler:
             storage = ArgumentsStorage(
                 job.author.username, program.title, provider_name
             )
-            arguments = storage.get(job.id) or job.arguments
+            arguments = storage.get(job.id)
             arguments_file = os.path.join(
                 working_directory_for_upload, "arguments.serverless"
             )
--- a/gateway/api/serializers.py
+++ b/gateway/api/serializers.py
@@ -287,7 +287,6 @@ class RunJobSerializer(serializers.ModelSerializer):
             trial=trial,
             status=status,
             program=program,
-            arguments=arguments,
             author=author,
             config=config,
         )


korgan00 (approved):

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -56,15 +57,18 @@ def update_job_status(job: Job):
             job.env_vars = "{}"
 
     if job_handler:
+        # let's store logs only for Functions from providers
```
ElePT: We replaced this by the size limitation, right?
```suggestion
```

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -56,15 +57,18 @@ def update_job_status(job: Job):
...
-        job.logs = check_logs(logs, job)
-        # check if job is resource constrained
-        no_resources_log = "No available node types can fulfill resource request"
-        if no_resources_log in job.logs:
-            job_new_status = fail_job_insufficient_resources(job)
-            job.status = job_new_status
-            # cleanup env vars
-            job.env_vars = "{}"
+        logs_size_mgb = sys.getsizeof(logs) / (1024**2)
+        if logs_size_mgb < 50:
```
ElePT: I think I would rather truncate the logs than not save them at all, just an opinion. In any case, if we followed this approach, I would add a log saying something like "Logs exceeded maximum allowed size (50 MB) and could not be stored."

korgan00: I agree, I prefer to truncate them

gateway/api/management/commands/update_jobs_statuses.py
Tansito: Done! (I did a little refactor for the truncate so the commit with the change is around 😂 )

gateway/api/management/commands/update_jobs_statuses.py
Tansito: Done! (I did some commits to implement it 😂 )

korgan00 (approved):
In general, it looks good to me. I think that maybe we can rethink the name of the `check_logs` method because it is doing more things like trimming the logs and that is not something you expect of a "check" function. Maybe `validate_logs`, or `prepare_logs_for_store`? I don't know.

PR #1736, opened 2025-10-20
fix: arguments not saved in db

paaragon:
### Summary

This PR removes the arguments from DB save so they are only saved in the COS storage


Tansito (changes requested):
As we commented internally we think is safe to remove the `job.arguments` from this line also: https://github.com/Qiskit/qiskit-serverless/blob/d1e0d7f863d9b857a1baacefde82dddd2816e6ea/gateway/api/ray.py#L123

Issue #1676, opened 2025-07-04
Docs: Update getting started content to qiskit 2.x


ElePT:
The basic examples used in the documentation are severely outdated and cannot run with qiskit 2.x. The code for the following tutorials should be reviewed and updated:

- [ ] https://github.com/Qiskit/qiskit-serverless/blob/main/docs/getting_started/basic/01_running_program.ipynb
- [ ] https://github.com/Qiskit/qiskit-serverless/blob/main/docs/getting_started/basic/02_arguments_and_results.ipynb
- [ ] https://github.com/Qiskit/qiskit-serverless/blob/main/docs/getting_started/basic/03_dependencies.ipynb
- [ ] https://github.com/Qiskit/qiskit-serverless/blob/main/docs/getting_started/basic/04_distributed_workloads.ipynb
- [ ] https://github.com/Qiskit/qiskit-serverless/blob/main/docs/getting_started/basic/05_retrieving_past_results.ipynb

## Points to look out for:

1. Replace any outdated class with the 2.0 equivalent, as pointed out in the 1.0 and 2.0 migration guides:
    - https://quantum.cloud.ibm.com/docs/en/migration-guides/qiskit-1.0
    - https://quantum.cloud.ibm.com/docs/en/migration-guides/qiskit-2.0
2. Confirm that notebook can run, fix what doesn't work
3. Update outdated text descriptions to match current state of the code
4. Include versions table at the end of each tutorial to document the package versions used to run

ItamarGoldman:
I can't assign myself to this issue. I will start working on it.

PR #1737, opened 2025-10-20
Updating Qiskit Serverless docs to support Qiskit 2.x

ItamarGoldman:
### Summary
Change the Jupyter Notebook and the python code to support Qiskit 2.x.
Issue #1676

### Details and comments
This PR only changed the Qiskit related issues in the document. Now the circuit is going through a pass manager and the tutorial use `Sampler` and `AerSimulator` to run the circuit.

- [x] Fix Tutorial 1
- [x] Fix Tutorial 2
- [x] Fix Tutorial 4

### DIFF
--- a/docs/getting_started/basic/source_files/function.py
+++ b/docs/getting_started/basic/source_files/function.py
@@ -8,22 +8,22 @@ from qiskit_serverless import save_result
 # all print statement will be available in job logs
 print("Running function...")
 
-# creating circuit
+# Step 1: Create a Bell-state circuit
 circuit = QuantumCircuit(2)
 circuit.h(0)
 circuit.cx(0, 1)
 circuit.measure_all()
 
-# Using a fake backend for the example
+# Step 2: Transpile for a fake backend (for demonstration)
 backend = FakeVigoV2()
 pm = generate_preset_pass_manager(backend=backend, optimization_level=1)
 isa_circuit = pm.run(circuit)
 
-# running Sampler primitive
+# Step 3: Run the Sampler primitive
 sampler = Sampler(backend)
 quasi_dists = sampler.run([isa_circuit]).result()[0].data.meas.get_counts()
 
-# save results of function execution,
-# which will be accessible by calling `.result()`
+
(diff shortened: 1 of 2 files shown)


Tansito:
I don't know why in this PR the integration tests don't advance 🤔

docs/getting_started/basic/source_files/function.py
```diff
@@ -1,6 +1,8 @@
+"""function for jupyter notebook."""
 from qiskit import QuantumCircuit
-from qiskit.primitives import StatevectorSampler as Sampler
-
+from qiskit.transpiler import generate_preset_pass_manager
+from qiskit_ibm_runtime import SamplerV2 as Sampler
+from qiskit_aer import AerSimulator
```
ElePT: I recommend importing a fake backend from `qiskit_ibm_runtime` instead of instantiating the aer simulator directly. Fake backends contain useful information for the transpiler and the workflow is more similar to what a workflow with a real backend would look like. For example, you can pick `FakeVigoV2` as a 5 qubit fake backend:

```suggestion
from qiskit_ibm_runtime.fake_provider import FakeVigoV2
```

docs/getting_started/basic/source_files/function.py
```diff
@@ -12,11 +14,16 @@
 circuit.cx(0, 1)
 circuit.measure_all()
 
+# Using AerSimulator for the example
+backend = AerSimulator()
```
ElePT:
```suggestion
# Using a fake backend for the example
backend = FakeVigoV2()
```

docs/getting_started/basic/source_files/function.py
ItamarGoldman: Thanks! I will change as you suggested and will update the other tutorials to use `FakeVigoV2` as the backend.

docs/getting_started/basic/source_files/function.py
ItamarGoldman: changed as suggested

Tansito:
Another run that got stuck in the integration tests, it's only happening in this PR for whatever reason.

PR #1739, opened 2025-10-21
Enable/Disable deployments

Tansito:
### Summary

This PR allows us to enable/disable custom deployments for the API and the Scheduler.


charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -354,7 +357,7 @@ spec:
               value: {{ .Values.application.limits.maxJobsPerUser | quote }}
             - name: LIMITS_MAX_CLUSTERS
               value: {{ .Values.application.limits.maxComputeResources | quote }}
-            - name: LIMITS_GPU_CLUSTERS
+            - name: LIMITS_MAX_GPU_CLUSTERS
```
Tansito, line 360: I also edited the name for this environment variable that it was wrong in our helm (I discovered it after some internal tests).

PR #1741, opened 2025-10-22
Remove public group env var

Tansito:
### Summary

From #1712 we removed the end-points related with the catalog. This PR removes an environment variable used in that development that now we can delete.


ElePT (approved):
LGTM

PR #1744, opened 2025-10-28
Add release notes infra

ElePT:
### Summary
Add necessary structure to build release notes as part of `tox -edocs` and deploy them in our documentation page (GitHub pages). There should be a follow-up to increase automatization and automatically feed the release notes into the changelog when tagging a new release.

### DIFF
--- a/releasenotes/notes/new-job-filters-f1e38cee26594fbe.yaml
+++ b/releasenotes/notes/new-job-filters-f1e38cee26594fbe.yaml
@@ -1,0 +1,18 @@
+---
+upgrade:
+  - |
+    Enhanced ``client.jobs()`` and ``function.jobs()`` behavior with
+    new filters to improve the management of the different jobs. Now you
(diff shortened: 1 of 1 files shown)


releasenotes/notes/config.yaml
```diff
@@ -0,0 +1,5 @@
+---
+encoding: utf8
+default_branch: main
+unreleased_version_title: "Upcoming Qiskit Serverless release"
+earliest_version: 0.26.0
```
Tansito, line 5: earliest version is 0.26.1, isn't it?

ElePT: I set it up like this to get it started, it doesn't really affect the outcome, as the tag is automatically pulled (see that 0.26.1 correctly shows up). I think that once we have 0.27.0 out, we should update it to say `0.27.0` so that it accurately reflects the earliest possible release notes.

Screenshot 2025-10-28 at 14 17 41

Tansito (approved):

PR #1750, opened 2025-10-30
Update logs endpoint

korgan00:

### DIFF
--- a/tests/docker/test_docker.py
+++ b/tests/docker/test_docker.py
@@ -54,7 +54,7 @@ class TestFunctionsDocker:
             assert set(entry.keys()).issubset(allowed_keys)
 
         assert job.status() == "DONE"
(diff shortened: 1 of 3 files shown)


gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -37,14 +39,22 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
         if job is None:
             raise NotFoundError(f"Job [{job_id}] not found")
 
-        # Case 1: Provider function - check provider access policy
-        if job.program and job.program.provider:
-            if ProviderAccessPolicy.can_access(user, job.program.provider):
-                return job.logs
+        if user != job.author:
```
Tansito: I think this should be an AccessPolicy in jobs like `can_read_result`, `can_read_logs`.

PR #1753, opened 2025-11-04
New logs system management

Tansito:
### Context

Check specification in [LOGS.md](https://github.com/Qiskit/qiskit-serverless/pull/1753/changes#diff-b23dfcc46c1106e0b1bfeb785e7cd9ee7f3edeae2eb1aa18b422db296e8dd771)

Tasks:
- [x] Client: get_logger() and get_provider_logger() to write logs with the prefixes [PUBLIC] and [PRIVATE]
- [x] Scheduler: upload logs to COS when job transition from running to terminal state
- [x] Gateway: endpoints /logs and /provider_logs should return the COS log for finished job (if exist), Ray console if job is running and as a fallback, return the logs from the db.

### Details and comments

**DON'T MERGE**

Co-authored-by: @avilches , @korgan00

### DIFF
--- a/tests/integration/test_experimental.py
+++ b/tests/integration/test_experimental.py
@@ -70,7 +70,7 @@ class TestExperimental:
 
         assert file_count > 0
 
-        serverless_client.file_delete("uploaded_file.tar", file_consumer_function)
+        serverless_client.file_delete("my_file.tar", file_consumer_function)
 
         assert (file_count - len(serverless_client.files(file_consumer_function))) == 1
 


gateway/.gitignore
```diff
@@ -135,3 +135,6 @@ GitHub.sublime-settings
 !.vscode/launch.json
 !.vscode/extensions.json
 .history
+
+tests/resources/fake_media/**/arguments/*
+!tests/resources/fake_media/**/logs/*.log
```
avilches:
```suggestion
```

This is not needed anymore because of [this PR](https://github.com/Qiskit/qiskit-serverless/pull/1839).

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -38,14 +44,31 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
+            working_dir=WorkingDir.USER_STORAGE,
+            function_title=job.program.title,
+            provider_name=job.program.provider.name if job.program.provider else None,
+        )
 
-        # Access denied for all other cases
-        raise ForbiddenError(f"You don't have access to job [{job_id}]")
+        logs = logs_storage.get(job_id)
+
+        # No logs stored.
```
avilches:
```suggestion
        # Logs stored in COS
```

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -38,14 +44,31 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
+        if logs:
+            return logs
+
+        # Get from Ray if it is already running.
+        if job.compute_resource and job.compute_resource.active:
+            job_handler = get_job_handler(job.compute_resource.host)
+            logs = job_handler.logs(job.ray_job_id)
+            logs = check_logs(logs, job)
+            logs = extract_public_logs(logs)
```
avilches: get_job_handler could fail if the job is running in the GPU cluster. Wrap and ignore. Thanks @Tansito for the heads up!

gateway/api/v1/views/jobs/provider_logs.py
```diff
@@ -0,0 +1,69 @@
...
+    Args:
+        request: The HTTP request object.
+        job_id: The UUID of the job (path parameter).
+
+    Returns:
+        Response containing the serialized job logs.
+    """
+    user = cast(AbstractUser, request.user)
+    logs = GetProviderJobLogsUseCase().execute(job_id, user)
+    return Response(serialize_output(logs))
```
avilches, line 69: Don't forget to add a test to /logs and /provider_logs. Cases: job with logs in the db, job with logs in COS and job with a (mocked) running image. 

Extra ball: a test with mocked ray client that raises an error, the endpoint should work as if nothing happened! Ideally, the mock should raise the exact same Exception as if it had actually failed. To discover that Exception, modify the host by adding an extra string so it fails when connecting in the docker tests.

gateway/api/domain/function/check_logs.py
```diff
@@ -27,35 +26,36 @@ def check_logs(logs: Union[str, None], job: Job) -> str:
...
 
         return logs
 
-    logs_size = sys.getsizeof(logs)
-    if logs_size == 0:
-        return logs
+    if not logs:
+        return ""
+
+    max_bytes = int(settings.FUNCTIONS_LOGS_SIZE_LIMIT)
```
avilches: I think you could remove the int conversion, and move the int in the settings.py:

From
```python
FUNCTIONS_LOGS_SIZE_LIMIT = os.environ.get(...)
```
To:
```python
FUNCTIONS_LOGS_SIZE_LIMIT = int(os.environ.get(...)
```

By doing it this way, we'll spot a parsing error as soon as the application starts, not just when the `check_logs` function is used.

gateway/api/v1/views/jobs/provider_logs.py
avilches: Did you have a chance to update the tests?

gateway/api/use_cases/jobs/get_logs.py
avilches: Have you tried to add the try/except?

gateway/api/domain/function/check_logs.py
korgan00: https://github.com/Qiskit/qiskit-serverless/pull/1753/changes/baceb749c8c12110c2b9d6fb68566728c5e97c57

gateway/api/v1/views/jobs/provider_logs.py
korgan00: Yes. I think there are enough tests

gateway/api/access_policies/jobs.py
```diff
@@ -68,6 +69,28 @@ def can_read_result(user: type[AbstractUser], job: Job) -> bool:
...
+
+        Args:
+            user: Django user from the request
+            job: Job instance against to check the permission
+
+        Returns:
+            bool: True or False in case the user has permissions
+        """
+
+        has_access = user.id == job.author.id
```
avilches: With this, the provider will never be able to call `job.logs()`, unless they are the author, but they don't have to be in a provider job executed by another user.

avilches: I mean, if could add this so the provider jobs
```
    if job.program and job.program.provider:
        return ProviderAccessPolicy.can_access(user, job.program.provider)
```

Or maybe we can leave as it is, but then the provider will only have access to `job.provider_logs()` for jobs created by other users. @Tansito is this what we want?

avilches:
I can't use Github to review the tests, some changes don't appear. So I'm reviewing them in my local computer and adding comments here...

gateway/api/v1/views/jobs/provider_logs.py
avilches: I thinks these tests are still missing....
- `test_job_provider_logs_not_fount_empty` -> the /logs version of this test is missing
- `test_job_logs_by_author_for_function_without_provider` -> the /provider-logs version of this test is missing
- `test_job_logs_in_ray` -> the /provider-logs version of this test is missing

avilches: The test `def test_job_logs_by_author_for_function_with_provider(self)` has an incorrect name and comment:
My suggestion is
```python
def test_job_logs_by_non_author_for_function_with_provider(self):
    """Tests that a user who is not the author cannot access logs of a provider job."""
```

For this test, the /provider-logs version is missing

gateway/api/use_cases/jobs/provider_logs.py
```diff
@@ -0,0 +1,75 @@
...
+
+        # Get from Ray if it is already running.
+        if job.compute_resource and job.compute_resource.active:
+            logs = GetComputeResourceLogsUseCase().execute(job)
+            return logs.full_logs
+
+        logs = job.logs
+
+        if not logs or logs == NO_LOGS_MSG or logs == NO_LOGS_MSG_2:
+            raise NotFoundError(f"Logs for job[{job_id}] are not found")
```
avilches: This is not correct, we already removed it in the logs use case but we have to do the same in this usecase too.

gateway/tests/api/test_job.py
```diff
@@ -578,47 +582,314 @@ def test_stop_job(self):
...
+                    args=[job_id],
+                ),
+                format="json",
+            )
+
+            self.assertEqual(jobs_response.status_code, status.HTTP_404_NOT_FOUND)
+            self.assertEqual(
+                jobs_response.data.get("message"),
+                f"Logs for job[{job_id}] are not found",
+            )
```
avilches: It should returns an empty string or "No logs yet", it doesn't matter. Same behavior as /jobs endpoint is doing

gateway/tests/api/test_job.py
```diff
@@ -578,47 +582,314 @@ def test_stop_job(self):
...
+        with self.settings(MEDIA_ROOT=self.MEDIA_ROOT):
+            user = models.User.objects.get(username="test_user_2")
+            self.client.force_authenticate(user=user)
+
+            jobs_response = self.client.get(
+                reverse("v1:jobs-logs", args=["1a7947f9-6ae8-4e3d-ac1e-e7d608deec85"]),
+                format="json",
+            )
+            self.assertEqual(jobs_response.status_code, status.HTTP_200_OK)
+            self.assertEqual(jobs_response.data.get("logs"), "log entry 1")
```
avilches: This test works because the `1a7947f9-6ae8-4e3d-ac1e-e7d608deec85` job has the `test_user_2` as author. But right now, it should test a job where the author is different then the provider user.

avilches: To be honest, I don't like at all loading data with fixtures and reusing them to test specific cases. For this case, I had to open the fixtures and check the author. Test should have methods where it's crystal clear the data we want to create and test, like `def create_job(author: str, provider_author: str):`

korgan00: Yeah, you are totally right. As I told you also, I think we should rework all the Tests.

1. First of all, have several test files to test the `views` with the current method (django client calls) and mock the use-cases. This way we can test the serializers and data pre-processing.
2. Then we should test the use-cases as we usually test functions. This will make testing and reviewing easier.

My organization proposal is one test per endpoint group or even per endpoint and same for use-cases.

There are a lot of things to improve here but i don't want to have bigger and bigger PR, just try to keep smaller as possible

gateway/api/access_policies/jobs.py
Tansito: It's a good question, from my perspective I would say those logs are for the user not for the provider but how they write them I'm not fully sure about the scope. Let's double check with @pandasa123 , can you open a thread in Slack pinging him? I would like to have his opinion here.

gateway/tests/api/test_job.py
avilches: Sure, let's finish all the tests first, then we can refactor to make them simpler and put them in different files

gateway/api/v1/views/jobs/provider_logs.py
korgan00: > * test_job_logs_in_ray

there is already one: test_job_provider_logs_in_ray

gateway/tests/api/test_job.py
avilches: Hi. It's completely impossible for me to follow the code using hardcoded pks, it forces me to memorize that the `...ec85` job has as author and provider `test_user_2`, while `...ec86` job has author `test_user` but provider admin is `test_user_2`...

The code is hard to follow and it makes the review difficult. Some tests might be working because the data is wrong, but it's hard to tell in a review. I know it's not your fault (the tests were created using fixtures, you're just using them as they are) and I know you wanted to refactor this in the future to avoid bigger and bigger PR with refactors... but I think it really worth it, no matter what it takes, if it helps us ensure that the code is actually tested... I know that what I'm asking is a pain in the neck and time consuming, so I've done the refactor myself in this [commit](https://github.com/Qiskit/qiskit-serverless/pull/1753/commits/7bd93299f646e4f1bd8e76cfbe19db77def641e8). I really hope this helps you develop tests in an easier way (and it will help me review it better!) without worrying about memorizing pks.

As a general rule, I would avoid using fixtures except for the bare minimum necessary. Each test should create its own data, and this data should be explicit to make it easy to review. Django already handles a database transaction for each test, so each test can create whatever it needs without affecting the next one.

avilches: That said, it should now be easier to write these tests that are still missing:
- Check if calling /logs and /provider-logs in a user job by a non-author fails with a 403
- Provider job, request made the author: /logs -> 200 and /provider-logs -> 403

gateway/api/management/commands/free_resources.py
```diff
@@ -82,3 +100,76 @@ def remove_compute_resource(self, compute_resource):
...
+        Save the logs in the corresponding storages.
+
+        Args:
+            job: Job
+        """
+
+        print(f"save_logs_to_storage:: Job: {job.id}")
+        job_handler = get_job_handler(job.compute_resource.host)
+        logs = job_handler.logs(job.ray_job_id)
+        logs = check_logs(logs, job)
```
avilches: Wrap this in a try/except because the job could be in a different cluster.

LOGS.md
```diff
@@ -0,0 +1,93 @@
...
+Unprefixed logs (regular `print()` statements or other loggers from the user or 3rd party libreries) that appear in the console are treated as unclassified content.
+In provider jobs, they are hidden from users to protect intellectual property.
+
+- No validation of `get_provider_logger()` in user Jobs
+If a user uses `get_provider_logger()` by mistake in a user job, it will work as usual (adding the `[PRIVATE]` prefix). Since user jobs show all logs unfiltered, the user will see the content (with the prefix removed).
+
+- Provider access to `/logs`
+Providers can call `/logs` to see how the output appears from the user's perspective, which is useful for debugging.
+
+
```
avilches: @korgan00 hi, let's use this doc as a source of truth. I know it's no updated with the latest changes from @Tansito, we will update it (and implement the feature changes) when we finish all the tests.

I mean, the plan should be:
- first finish all the tests, with the old behavior
- when everything is working and tested, let's do apply the new changes

gateway/api/domain/function/filter_logs.py
```diff
@@ -6,7 +6,7 @@
 private_pattern = re.compile(r"^\[PRIVATE\]", re.IGNORECASE)
 
 
-def extract_public_logs(text: str) -> str:
+def log_filter_provider_job_public(text: str) -> str:
```
korgan00: This is a matter of logs, not jobs or providers, why mixing here so many concepts? Also, the name is a bit confusing for me.

Maybe: `filter_logs_with_public_tag`

The thing about filter is that I don't use to like it because it can be filter-in or filter-out thats why i used extract but I'm not totally against it so we can use filter.

gateway/api/domain/function/filter_logs.py
```diff
@@ -20,7 +20,7 @@ def extract_public_logs(text: str) -> str:
     return "\n".join(lines) + "\n" if lines else ""
 
 
-def remove_prefix_tags_in_logs(text: str) -> str:
+def log_filter_user_job(text: str) -> str:
```
korgan00: I won't use `filter` word here. In programming filter is usually delete some lines or create a selection of elements of a bigger group, so can be confusing to use it here.
Asking AI for better name describing the solution it proposed:

 - remove_log_prefixes
 - strip_log_prefix
 - normalize_log_lines
 - normalize_logs
 - clean_logs

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -72,11 +72,16 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
+
+            # Public logs from a user job
+            return log_filter_user_job(logs)
 
         # Legacy: Get from db.
-        if not job.program.provider:
+        if job.program.provider:
+            # Public logs from a provider job
+            return "No logs available."
+        else:
```
korgan00: This else is not needed.

gateway/api/domain/function/filter_logs.py
avilches: We have two functions, and we're going to need another one when we filter the private logs for the provider. My idea was that each function would represent the type of log it creates, without going so far as to define what it does, since this can change. 

- log_filter_provider_job_public
- log_filter_provider_job_private #(this is the one that would be created later)
- log_filter_user_job

So the name has provider job or user job, then public or private. If you don't like the name, you can use a different one, no worries at all.

gateway/api/use_cases/jobs/get_logs.py
avilches: I know. I always prefer to be more explicit than implicit. If someone in a hurry decides to add code at the end of the function to do something, without the else they maybe not realize there was an if. With the else they will.

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -37,14 +46,42 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
-            if ProviderAccessPolicy.can_access(user, job.program.provider):
-                return job.logs
+        if not JobAccessPolicies.can_read_logs(user, job):
+            raise ForbiddenError(f"You don't have access to job [{job_id}]")
 
-        # Case 2: User is the author of the job
-        elif user == job.author:
-            return job.logs
+        logs_storage = LogsStorage(
+            username=user.username,
```
avilches: The user should be job.author, not user, because the user is the endpoint caller. This is a a bug that cannot be reproduced in the tests. If a service/class/artifact is mocked, its behavior should not depend on the constructor input parameters. If the error is caused by an incorrect input data, but the class methods are mocked, then we will never discover the error except in integration tests or in production.

I refactored the LogStorage class in this [commit](https://github.com/Qiskit/qiskit-serverless/pull/1753/commits/d2d3799189e170436cbf399b11d6613e22e27432), now it doesn't receive the parameters to create the path. Instead, it receives the Job, so we can create a unit test for this class individually and mock it in the rest of the test without risk.

client/qiskit_serverless/core/job.py
```diff
@@ -103,6 +103,10 @@ def result(self, job_id: str) -> Any:
     def logs(self, job_id: str) -> str:
         """Return logs."""
 
+    @abstractmethod
+    def provider_logs(self, job_id: str) -> str:
+        """Return logs."""
```
avilches:
```suggestion
        """Return provider logs."""
```

gateway/api/management/commands/update_jobs_statuses.py
```diff
@@ -54,17 +60,18 @@ def update_job_status(job: Job):
         )
         status_has_changed = True
         job.status = job_new_status
-        # cleanup env vars
+        # cleanup env vars and save logs when job reaches terminal state
         if job.in_terminal_state():
             job.sub_status = None
             job.env_vars = "{}"
+            if job_handler:
+                save_logs_to_storage(job, job_handler)
```
avilches: Are the job.logs empty here? @korgan00

gateway/scheduler/management/commands/update_jobs_statuses.py
```diff
@@ -54,17 +60,18 @@ def update_job_status(job: Job):
...
+            if job_handler:
+                save_logs_to_storage(job, job_handler)
 
     if job_handler:
         logs = job_handler.logs(job.ray_job_id)
-        job.logs = check_logs(logs, job)
         # check if job is resource constrained
         no_resources_log = "No available node types can fulfill resource request"
-        if no_resources_log in job.logs:
+        if no_resources_log in logs:
```
avilches, line 89: I guess we should save the logs to storage too... (and clean the job.logs)

tests/integration/test_jobs.py
```diff
@@ -399,3 +411,24 @@ def test_wrong_function_name(self, serverless_client: ServerlessClient):
             serverless_client.function("wrong-title")
 
         assert str(exc_info.value) == expected_message
+
+    def test_provider_logs(self, serverless_client: ServerlessClient):
```
avilches, line 361: Could add to this test a check to see if the provider can see the provider logs?

tests/docker/test_docker.py
```diff
@@ -377,6 +363,32 @@ def test_get_filtered_jobs(self, serverless_client: ServerlessClient):
...
+            working_dir=resources_path,
+        )
+        function = serverless_client.upload(function)
+        job = function.run()
+
+        while not job.in_terminal_state():
+            sleep(1)
+
+        assert job.logs().endswith("""INFO:user: User log
+INFO:user: User multiline
```
avilches: I think we can get rid of the "user" and "private" part in the `_create_logger("user", "[PUBLIC]")`. The error level should be enough.

tests/integration/test_jobs.py
avilches: I mean, in a provider job we can check:
1. [ ] User can see the public logs filtered
2. [ ] User can't see the private logs
2. [ ] Provider can't see the public logs
4. [ ] Provider can see the private logs filtered

There is only test for the case 2. I think it's easy to add the checks because you can create more clients and re-use the same job.

tests/docker/test_docker.py
avilches: Add more asserts:
1. [ ] User can see the public logs
2. [ ] User can't see the private logs
3. [ ] Provider can't see the public and private logs

There is only test for the case 1. I think it's easy to add the checks because you can create more clients and re-use the same job.

gateway/api/management/commands/update_jobs_statuses.py
korgan00: I dont think so. I assume that we should empty them after save

gateway/scheduler/management/commands/update_jobs_statuses.py
korgan00: I think we should have an small meeting to rethink this methods, but i think we should move the save point.

tests/docker/test_docker.py
korgan00: > I think we can get rid of the "user" and "private" part in the `_create_logger("user", "[PUBLIC]")`. The error level should be enough.

we can get rid in the message but not in the function, we need the "user" as an ID for the logger.

gateway/scheduler/management/commands/update_jobs_statuses.py
korgan00: I think this is a better approach https://github.com/Qiskit/qiskit-serverless/pull/1753/commits/07ef39f35ea473b9a380a4890a0819ef88cb7a5b

gateway/api/management/commands/update_jobs_statuses.py
korgan00: take a look to this: https://github.com/Qiskit/qiskit-serverless/pull/1753/commits/07ef39f35ea473b9a380a4890a0819ef88cb7a5b

avilches: Nope, this will upload the logs in every transition. The original place where this code was is only executed when there is a state transition, and the final state is terminal, so the job is finished.

avilches: `test_update_jobs_statuses_filters_logs_user_function` test should assert job.logs is ""

Tansito:
@korgan00 can you also create a release note for this PR like we were doing in other PRs?

avilches (approved):

PR #1754, opened 2025-11-05
Limit username string in the pod

Tansito:
### Summary

This PR adds a simple sanity check for the pod's name generator. It cuts the username to 20 as maximum.


ElePT (approved):
LGTM. As mentioned in the internal discussion, using ids instead of names might be a good future-facing solution.

PR #1757, opened 2025-11-06
Update GitHub Action to release to pip

Tansito:
### Summary

This PR updates the current GitHub action that releases to pip as the old one it's not working. We took inspiration from different addons repositories as they make use of the action: `pypa/gh-action-pypi-publish`, to do the release.


ElePT (approved):
Looks good, let's see if it does the trick.

Tansito:
Well, let's see if it works 🤞

PR #1762, opened 2025-11-07
Provider logs endpoint

korgan00:
### Summary

An endpoint similar to job logs endpoint but specific for provider functions.

### Details and comments

- URL: /:job_id/provider_logs
- Only the provider of the Function can access to these logs
- This end-point will return the logs from the Logs Storage in the Provider WorkDir.
- If there is no logs in the Provider WorkDir we should return the logs from the database when the logs stored in the DB don't contain the default string: "No logs yet." (ref [here](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/models.py#L215))
- Error management:
  - If the user doesn't have access to the provider return a ForbiddenError and log a warning
  - If the user is a provider of the Function but there is no log file and in the DB there is the default message we should return a "No logs available." with a 404.


gateway/api/access_policies/providers.py
```diff
@@ -29,7 +29,7 @@ def can_access(user, provider: Provider) -> bool:
         """
 
         user_groups = set(user.groups.all())
-        admin_groups = set(provider.admin_groups.all() if provider else [])
+        admin_groups = set(provider.admin_groups.all())
```
avilches, line 32: Why did you change this? :)

gateway/api/use_cases/jobs/provider_logs.py
```diff
@@ -0,0 +1,64 @@
...
+            raise ForbiddenError(f"You don't have access to job [{job_id}]")
+
+        logs_storage = LogsStorage(
+            username=user.username,
+            working_dir=WorkingDir.PROVIDER_STORAGE,
+            function_title=job.program.title,
+            provider_name=job.program.provider.name,
+        )
+
+        logs = logs_storage.get(job_id)
```
avilches, line 56: Why this method is not throwing an error if the logs fails? I know it's not part of this PR, but the get method is hiding any error in this line:
https://github.com/Qiskit/qiskit-serverless/blob/00e717ded0caf96f6740951598bdc84241fe9e8d/gateway/api/services/storage/logs_storage.py#L86

I think the get and the execute use case should return the error to the user... it will be less confusing throwing an error than showing no logs. But maybe there is a specific reason to do it in this way, so I'm just asking :)

avilches (approved):
LGTM

gateway/api/access_policies/providers.py
korgan00: To match the method definition. `Provider` is not `Optional`, so it always should be `Truish`.

gateway/api/use_cases/jobs/provider_logs.py
korgan00: You are right. After talking with David, we can have a conversation about this and create a new PR to fix all the storages at the same time because this is a wider problem.

PR #1763, opened 2025-11-10
Add host to IBMServerlessClient client

avilches:
### Summary

`IBMServerlessClient` always uses the IBM production host and it doesn't allow to change it.
This PR defines a new parameter host that can be used by the user to connect the client to a different url.


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -680,6 +680,8 @@ def __init__(
         name: Optional[str] = None,
         instance: Optional[str] = None,
         channel: str = Channel.IBM_QUANTUM_PLATFORM.value,
+        *,
```
Tansito, line 683: Just curious, @avilches, for what is this `*` in the arguments?

avilches: Before the `*`, all the parameters must be used as keyword arguments and not as positional arguments. It was necessary to introduce it because otherwise the linter would complain with:

```
> qiskit_serverless/core/clients/serverless_client.py:677:4: R0917: Too many positional arguments (6/5) (too-many-positional-arguments) 
``` 
The __init__ had self + 5 positional parameters (host, token, name, instance, channel), and the `too-many-positional-arguments` check has a default limit of 5 positional parameters. The only way to avoid this is making the new arguments non-positional with the `*` operator.

Tansito: The way we were doing this until now was with a `pylint-disable`: https://github.com/Qiskit/qiskit-serverless/blob/768873754cf32d0322f239f0a42329671d873e3e/client/qiskit_serverless/core/clients/serverless_client.py#L98

But it's true that I didn't know about that solution. I would like to know @ElePT opinion here.

avilches: I think positional parameter shouldn't be more than 2-3, especially if all of them are optional. Imagine a call to `IBMServerlessClient(None, None, None, None, None, "http://...")`...

Tansito: Oh, I see. These are not considered positional arguments then:

```py
client = IBMServerlessClient(
            token=use_token, instance=use_instance, channel=use_channel, host=use_host
        )
```

We can continue using this approach.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -711,7 +714,7 @@ def __init__(
             channel=self.account.channel,
             token=self.account.token,
             instance=self.account.instance,
-            host=IBM_SERVERLESS_HOST_URL,
+            host=host if host is not None else IBM_SERVERLESS_HOST_URL,
```
korgan00:
```suggestion
            host=host if host else IBM_SERVERLESS_HOST_URL,
```

Maybe a wider condition is better here. If host is `None` or `False` or _empty string_ then, use the default value.

korgan00 (changes requested):

client/qiskit_serverless/core/clients/serverless_client.py
avilches: Changed

Tansito (approved):
It looks good to me @avilches thank you! Let's see what @korgan00 says also 🫡

PR #1766, opened 2025-11-11
Replaced `python3.11-3.11.11` by just `python3.11` in Dockerfile

avilches:
### Summary

Building the gateway Dockerfile throws an error:

```bash
  #11 1.223 No match for argument: python3.11-3.11.11
  #11 1.224 No match for argument: python3.11-devel-3.11.11
  #11 1.234 Error: Unable to find a match: python3.11-3.11.11 python3.11-devel-3.11.11
```

It looks like the current pinned version `registry.access.redhat.com/ubi9:9.6-1758184894` doesn't have this version anymore:

```bash
docker run --rm registry.access.redhat.com/ubi9:9.6-1758184894 dnf list available 'python3.11*'
```

Returns:
```
docker run --rm registry.access.redhat.com/ubi9:9.6-1758184894 dnf list available 'python3.11*'
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Red Hat Universal Base Image 9 (RPMs) - BaseOS  2.1 MB/s | 435 kB     00:00    
Red Hat Universal Base Image 9 (RPMs) - AppStre  19 MB/s | 2.2 MB     00:00    
Red Hat Universal Base Image 9 (RPMs) - CodeRea 1.7 MB/s | 174 kB     00:00    
Available Packages
python3.11.aarch64                   3.11.13-3.el9  ubi-9-appstream-rpms        
python3.11-devel.aarch64             3.11.13-3.el9  ubi-9-appstream-rpms
```

So, the version we should use is `python3.11.13-3.el9` instead of `python3.11-3.11.11`

This PR changes the version to a less strict one: just `python3.11`.

It will install the 3.11.13 version as expected:
```
docker build --no-cache -f gateway/Dockerfile -t test-gateway-build .
docker run --rm --entrypoint python test-gateway-build --version
Python 3.11.13
```

I'm open for comments

### DIFF
--- a/Dockerfile-ray-node
+++ b/Dockerfile-ray-node
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi9-minimal:9.6-1758184547
+FROM registry.access.redhat.com/ubi9-minimal:9.6-1760515502
 
 ARG DYNAMIC_DEPENDENCIES_FILE
 
--- a/gateway/Dockerfile
+++ b/gateway/Dockerfile
@@ -2,11 +2,11 @@ ARG MICRO_IMAGE_DIR=/ubi-micro-img
 
 # BASE image using UBI 9 micro where the
 # application and requirements will be installed
-FROM registry.access.redhat.com/ubi9-micro:9.6-1754467928 AS base
+FROM registry.access.redhat.com/ubi9-micro:9.6-1762316544 AS base
 
 # BUILD image using UBI 9 where the dependencies that
 # require installing with a package manager will be installed
-FROM registry.access.redhat.com/ubi9:9.6-1758184894 AS build
+FROM registry.access.redhat.com/ubi9:9.6-1760340943 AS build
 ARG MICRO_IMAGE_DIR
 
 # Copy the BASE image into the BUILD image


tests/docker/conftest.py
```diff
@@ -34,12 +34,12 @@ def set_up_serverless_client():
     compose = DockerCompose(
         resources_path,
         compose_file_name="../../../docker-compose-dev.yaml",
-        pull=True,
```
Tansito, line 37: I suppose we want to do `pull=True` here, don't we?

tests/docker/conftest.py
```diff
@@ -34,12 +34,12 @@ def set_up_serverless_client():
...
+        pull=False,
     )
     compose.start()
 
     connection_url = "http://localhost:8000"
-    compose.wait_for(f"{connection_url}/backoffice")
```
Tansito, line 42: it looks like from the Integration tests don't like the new value.

tests/docker/conftest.py
avilches: With false, the tests will use the images built in the previous step right? Which is what we want...

avilches:
> It looks good to me, @avilches , thank you! Should we also update the base images to the newest ones?
> 
> * Update registry.access.redhat.com/ubi9 Docker tag to v9.6-1760340943
> * Update registry.access.redhat.com/ubi9-micro Docker tag to v9.6-1762316544
> * Update registry.access.redhat.com/ubi9-minimal Docker tag to v9.6-1760515502

done

PR #1767, opened 2025-11-13
Put provider logs endpoint

korgan00:
### Summary

⚠️ Merge after https://github.com/Qiskit/qiskit-serverless/pull/1762/

Store logs generated by a Function in its specific log file

### Details and comments

- URL: PUT /:job_id:/provider_logs
- Only the author of the job can store new logs
- Only jobs with status `RUNNING` will be able to store logs
- Check the size of the logs.
- Error management:
  - If the user is not the owner return a ForbiddenError and log a warning about it
  - If the user doesn't have access to the provider return a ForbiddenError and log a warning
  - If the file doesn't exist we will create it.

### DIFF
--- a/gateway/api/v1/views/jobs_new/provider_logs.py
+++ b/gateway/api/v1/views/jobs_new/provider_logs.py
@@ -15,6 +15,7 @@ from rest_framework.request import Request
 from rest_framework.response import Response
 
 from api.use_cases.jobs.provider_logs import GetProviderJobLogsUseCase
+
 from api.use_cases.jobs.put_provider_logs import PutProviderJobLogsUseCase
 from api.v1.endpoint_decorator import endpoint
 from api.v1.endpoint_handle_exceptions import endpoint_handle_exceptions


gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,43 @@ def get(self, job_id: str) -> Optional[str]:
...
+    def put(self, job: Job, log: str) -> None:
+        """
+        Write a log in a file for the given job id
+
+        Args:
+            job (Job): the job to get the logs
+            log (str): the job to get the logs
+
+        Returns:
+            Optional[str]: content of the file
```
avilches:
```suggestion
            None
```

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,43 @@ def get(self, job_id: str) -> Optional[str]:
...
             )
             return None
+
+    def put(self, job: Job, log: str) -> None:
+        """
+        Write a log in a file for the given job id
+
+        Args:
+            job (Job): the job to get the logs
+            log (str): the job to get the logs
```
avilches:
```suggestion
            job (Job): the job to write the logs
            log (str): the job to write the logs
```

gateway/api/use_cases/jobs/put_provider_logs.py
```diff
@@ -0,0 +1,63 @@
...
+from api.services.storage.enums.working_dir import WorkingDir
+from api.services.storage.logs_storage import LogsStorage
+
+
+NO_LOGS_MSG: Final[str] = "No available logs"
+NO_LOGS_MSG_2: Final[str] = "No logs yet."
+
+
+class PutProviderJobLogsUseCase:
+    """Use case for retrieving job logs."""
```
avilches:
```suggestion
    """Use case for writing job logs."""
```

gateway/api/use_cases/jobs/put_provider_logs.py
```diff
@@ -0,0 +1,63 @@
...
+
+    jobs_repository = JobsRepository()
+
+    def execute(self, job_id: UUID, user: AbstractUser, log: str) -> str:
```
avilches:
```suggestion
    def execute(self, job_id: UUID, user: AbstractUser, log: str) -> None:
```

gateway/api/use_cases/jobs/put_provider_logs.py
```diff
@@ -0,0 +1,63 @@
...
+
+        Args:
+            job_id (str): Unique identifier of the job.
+            user (AbstractUser): User requesting the logs.
+
+        Raises:
+            NotFoundError: If the job does not exist.
+
+        Returns:
+            str: Job logs if accessible, otherwise a message indicating no logs are available.
```
avilches:
```suggestion
            None
```

gateway/api/use_cases/jobs/put_provider_logs.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        job = self.jobs_repository.get_job_by_id(job_id)
+        if job is None:
+            raise NotFoundError(f"Job [{job_id}] not found 1")
```
avilches:
```suggestion
            raise NotFoundError(f"Job [{job_id}] not found")
```

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,43 @@ def get(self, job_id: str) -> Optional[str]:
...
+
+                    if len(logs) != 0:
+                        logs += os.linesep
+                    logs += log
+
+                    logs = check_logs(logs, job)
+
+                    log_file.seek(0)
+                    log_file.truncate()
+                    log_file.write(logs)
```
avilches: It's not necessary to read the whole file, append the log in memory and write the content again (imagine read 5Mb just to append a line and write it again...). You can just check the limit before, reading the file size only + the log size to avoid write over the limit with something like:
```python
if os.path.getsize(log_path) + len((log + os.linesep).encode(self.ENCODING)) > max_bytes:
     # fail
```

Then, if everything was ok, just append the line to the file.
```python
    with FileLock(lock_path, timeout=5):  # thread safe
        with open(log_path, "a+", encoding=self.ENCODING) as log_file:
            log_file.write(log + os.linesep)
```
I think the problem was trying to reuse the `check_logs` function, which was designed for when logs were added to the database (and therefore, it is necessary to read the field, append the text at the end and save the whole thing again with sql). The only purpose of `check_logs` is set a default message if job is broken and truncate the logs.

avilches: I would just use RotatingFileHandler to write logs that already has log rotation capabilities. I would set a limit per file of 1Mb for example and there would be no general limit. The logs endpoint could return the last log file content.

avilches:
```python
import logging
from logging.handlers import RotatingFileHandler
logger = logging.getLogger('mi_app')
handler = RotatingFileHandler(
    filename='app.log',
    maxBytes=1_048_576,        # 1 MB en bytes
    backupCount=1'             # Keep only the last 10 files, so 10Mb in total
)
formatter = logging.Formatter('%(asctime)s - %(levelname)s - %(message)s')
handler.setFormatter(formatter)
logger.addHandler(handler)
logger.info(logs)
```

If the endpoint accepts the log level ("info", "debug", "error"), we could use this to call to logger.info, logger.debug and so on...

avilches: Sorry, after talking with @Tansito, he told me the log writing is done directly in a mounted volume in COS. I'm checking the docs, and in mounted volumes we have to follow some rules, like no renaming (rename in a COS is a read and write process, which is costly) and only append, never rewrite files (it will be very costly). 

The RotatingFileHandler I suggested above uses rename, so we can't use it. After reviewing some libraries, `loguru` is the perfect candidate:

- It doesn't use rename if you configure it with `counter=True`, because every file generated is unique, so it doesn't need to rename the most recent file like RotatingFIlehandler does.
- It has a queue system, where calls are not blocking. With `enqueue=True`, the write to the file (actually, COS), is enqued in a background process. This is a recommended setting for mounted volumes.

```python
from loguru import logger

logger.add(
    "filename.log",
    rotation=1024*1024*1,        # rotate every 1MB
    retention=10,                # keep only the last 10 files, so 10MB
    compression=None,            # needed in mounted volumes
    enqueue=True,                # needed in mounted para S3 Fuse
)
```

korgan00: Thank you, thats a lot of good information!

korgan00: Ok, I was trying to implement this and there are some key points that i would like to discuss.

Rotation and retention are very interesting because we can keep the latest logs instead of the first that use to be better. The cons are that we dont reuse the check_logs that is a minor thing in my opinion and also the get should be reimplemented to read all the log files.

Should we do that in this PR?

Another option is just have one file and use it similar to what you said before without rotation and retention:

```python
logger.add(
    "filename.log",
    compression=None,            # needed in mounted volumes
    enqueue=True,                # needed in mounted para S3 Fuse
)

if os.path.getsize(log_path) + len((log + os.linesep).encode(self.ENCODING)) > max_bytes:
     # truncate log
logger.info(log)
```

avilches: To me, it's fine to do it in this PR or in a different one: the feature is isolated in the `feat/logs-refactor` branch, so I think it's safe to merge this PR with or without it.

The `check_logs` function is not working well. It's supposed to truncate the text from the beginning, but it does it from the end, which if we're going to use a logging library isn't necessary.

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,40 @@ def get(self, job_id: str) -> Optional[str]:
...
+        Returns:
+            None
+        """
+        log_path = self._get_logs_path(job.id)
+        max_mb = int(settings.FUNCTIONS_LOGS_SIZE_LIMIT)
+        file_size = 1024 * 1024 * 1
+
+        try:
+            file_logger.remove()
+            file_logger.add(
```
Tansito, line 116: As we commented, if we follow this approach we need to ensure that `add` method doesn't add a new `Handler` to a global instance. Each instance should write into independent files instead to write in all `Handlers`.

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,40 @@ def get(self, job_id: str) -> Optional[str]:
...
+                log_path,
+                rotation=file_size,
+                retention=math.trunc(max_mb / file_size),
+                compression=None,  # needed in mounted volumes
+                enqueue=True,  # needed in mounted for S3 Fuse
+                format="({level})[{time}] {message}",
+            )
+            file_logger.info(log)
```
Tansito, line 124: It would be important to accept as input the level of log to be able to log `info`, `warning` and `error` as minimum.

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,40 @@ def get(self, job_id: str) -> Optional[str]:
...
+        log_path = self._get_logs_path(job.id)
+        max_mb = int(settings.FUNCTIONS_LOGS_SIZE_LIMIT)
+        file_size = 1024 * 1024 * 1
+
+        try:
+            file_logger.remove()
+            file_logger.add(
+                log_path,
+                rotation=file_size,
+                retention=math.trunc(max_mb / file_size),
```
Tansito, line 119: We need to study what configuration we want to use here taking into account the next use cases:
- If we create several files to retrieve those logs could be a bit of a pain (access to different files)
- What happens if a log is bigger than the retention / rotation policy.

gateway/api/services/storage/logs_storage.py
```diff
@@ -90,3 +95,40 @@ def get(self, job_id: str) -> Optional[str]:
...
+                compression=None,  # needed in mounted volumes
+                enqueue=True,  # needed in mounted for S3 Fuse
+                format="({level})[{time}] {message}",
```
Tansito, line 122: We need to review what format we want to use here. Probably something similar to IBM Cloud it would be the best.

gateway/api/v1/views/jobs_new/provider_logs.py
```diff
@@ -64,6 +79,18 @@ def provider_logs(request: Request, job_id: UUID) -> Response:
     Returns:
         Response containing the serialized job logs.
     """
+
     user = cast(AbstractUser, request.user)
-    logs = GetProviderJobLogsUseCase().execute(job_id, user)
-    return Response(serialize_output(logs))
+
```
Tansito, line 84: It looks like Django doesn't accept different HTTP methods per end-point 🤦 

cc:/ @ElePT : for the review that I left you

gateway/requirements.txt
```diff
@@ -25,3 +25,4 @@ django-cors-headers>=4.4.0, <5
 whitenoise>=6.7.0, <7
 objsize>=0.7.1, <1
 ibm-platform-services>=0.59.1, <1
+loguru>=0.7.3, <1
```
Tansito, line 28: It's an interesting library, we can consider to use it as our standard way to log in our application 🤔

gateway/api/services/storage/logs_storage.py
avilches: `logger.remove()` removes all the handlers in the system, so this call should be done only once at the beginning (in the `__init__` constructor for instance). This will ensure there are no handlers at all, including the default one which prints the logs in the console (we don't want this).

Every file_logger should be independent from any other. That means one file_logger per job to ensure the log files doesn't merge between them. Since `add` (and `remove`) are global, the official docs suggest creating a logger context with `bind` or `contextualize`, adding the `job.id` as a variable in the context, and filtering in a lambda with something like:

```python
try:
    handler_id = logger.add(
        filter=lambda record: record["extra"].get("job_id") == job.id,
    )
    logger.bind(job_id=job.id).info() # or debug or error...
finally:
    logger.remove(handler_id) # check None... you know... :) 
```

(In a future PR, this could be improved by adding the logger when the job starts and removing it when it ends...)

Tansito: Yes, this is exactly what @korgan00 and I saw today. He is going to try to implement it before merge and write a test to ensure that we fulfill this use case.

gateway/api/v1/views/jobs_new/provider_logs.py
avilches: Yes, it is possible: https://docs.djangoproject.com/en/5.2/topics/class-based-views/

It will be something like:

```python
class ProviderLogsView(APIView):
    permission_classes = [IsAuthenticated]

    def get(self, request: Request, job_id: UUID) -> Response:
        user = cast(AbstractUser, request.user)
        logs = GetProviderJobLogsUseCase().execute(job_id, user)
        return Response(serialize_output(logs))

    def put(self, request: Request, job_id: UUID) -> Response:
        user = cast(AbstractUser, request.user)
        serializer = ProviderLogInputSerializer(data=request.data)
        serializer.is_valid(raise_exception=True)
        PutProviderJobLogsUseCase().execute(job_id, user, log)
        return Response(status=status.HTTP_204_NO_CONTENT)
```

Tansito: yes, but that would be through the APIView class and we are using at this moment the decorator method for an easy future migration to django-ninja:
https://www.django-rest-framework.org/api-guide/views/#api_view

PR #1977, opened 2026-04-02
Helm secrets refactor for testing purpose

Tansito:
### Summary

Gateway secrets were previously defined with default values in `values.yaml` and created by templates in the gateway subchart. The setup of these secrets could generate errors as they can be easily incorrectly configured as by default they were being created.

### Changes

- Secret templates moved to root chart, gated on `platform=kind`

The three secret templates (secret-key, secret-psql, secret-superuser) have been moved from `charts/gateway/templates/` to the root `charts/qiskit-serverless/templates/`, following the same pattern already used by `pvcs.yaml`. They now render only when `platform=kind`, which is the only environment where secrets need to be created by the chart — production deployments rely on an external secret manager.

This also resolves a Helm subchart scoping limitation: the platform value is only accessible at the root chart level.

- CI and local deploy scripts updated

The kind-deploy environment in `tests/tox.ini` and `docs/custom_images/custom_function/local_cluster/deploy.sh` now inject secret values via `--set` flags alongside `--set platform=kind`, which triggers secret creation automatically without any extra flags.


Tansito:
This PR requires a follow-up in our internal repository when we merge this.

docs/custom_images/custom_function/local_cluster/deploy.sh
```diff
@@ -48,6 +48,15 @@ helm install qs \
 --set gateway.application.debug=1 \
 --set gateway.application.limits.keepClusterOnComplete=false \
 --set gateway.application.authMockproviderRegistry=test \
+--set gateway.secrets.secretKey.value=CHANGE_ME \
```
ElePT, line 51: Is this an oversight? (the "CHANGE_ME")

tests/tox.ini
```diff
@@ -104,6 +104,15 @@ commands =
     --set gateway.application.debug=1 \
     --set gateway.application.limits.keepClusterOnComplete=false \
     --set gateway.application.authMockproviderRegistry=test \
+    --set gateway.secrets.secretKey.value=CHANGE_ME \
```
ElePT, line 107: Same question as above.

ElePT (commented):
I have taken a look, but I am not familiar enough with the helm secret configuration to fully assess if there's anything missing. Other team members may be able to give a better assessment here.

docs/custom_images/custom_function/local_cluster/deploy.sh
Tansito: Nah, it's just a default value. It can be CHANGE_ME or any other value you want.

tests/tox.ini
Tansito: Same answer than before. It's just a default value. It can be CHANGE_ME or any other value you want.

Tansito:
Moving this to draft until the new release as we are not adding it to that one

Tansito:
Reopening the issue!

avilches (approved):

PR #1990, opened 2026-04-14
Add CodeEngineProject model and runner fields for Fleets support 

marceloamaral:
### Summary

This PR introduces the data model foundation required to support IBM Code Engine Fleets as an alternative execution backend alongside the existing Ray runner.   

### Details and comments

- Add CodeEngineProject model with project identifiers, region, resource group, subnet pool, PDS names, and active flag
- Add runner field to Program and Job models (ray/fleets choices, default ray)
- Add fleet_id and code_engine_project FK to Job model
- Add migration 0045: auto-generated by Django, covers all of the above schema changes.
- Add unit tests for CodeEngineProject and runner fields

### What this PR does NOT include (follow-up PRs):                                                      
  - COS/logging fields on CodeEngineProject
  - FleetsRunner implementation                                                                       
  - Scheduler changes for Fleets job scheduling             
  - get_runner() factory update                                                                       
  - Client-side runner field on QiskitFunction


gateway/tests/api/test_code_engine_models.py
```diff
@@ -0,0 +1,125 @@
...
+
+# CodeEngineProject Tests
+
+
+def test_code_engine_project_creation(ce_project):
+    """Test creating a CodeEngineProject."""
+    assert ce_project.project_id == "ce-test-123"
+    assert ce_project.project_name == "test-project"
+    assert ce_project.region == "us-east"
+    assert ce_project.active is True
```
korgan00: We shouldnt test django APIs so, I think it doesn't make sense to test a db insertion.

gateway/tests/api/test_code_engine_models.py
```diff
@@ -0,0 +1,125 @@
...
+            subnet_pool_id="subnet-456",
+            pds_name_state="pds-state-2",
+            pds_name_users="pds-users-2",
+            pds_name_providers="pds-providers-2",
+        )
+
+
+def test_code_engine_project_str_representation(ce_project):
+    """Test the string representation."""
+    assert str(ce_project) == "test-project (us-east)"
```
korgan00: Is this important for us?

gateway/tests/api/test_code_engine_models.py
```diff
@@ -0,0 +1,125 @@
...
+    job = Job.objects.create(
+        author=user,
+        program=program,
+        code_engine_project=ce_project,
+        fleet_id="fleet-abc-123",
+    )
+    assert job.code_engine_project == ce_project
+    assert job.fleet_id == "fleet-abc-123"
+    assert job.compute_resource is None
+    assert job.ray_job_id is None
```
korgan00: Same as the other, I think we should not test django.

korgan00 (dismissed):
I think that maybe we are overtesting. Should we test the model configurations? I feel like we are testing django. We didn't tested any until now.
What do you think @ElePT , @avilches

ElePT (commented):
Agreed @korgan00, this was part of @avilches's original feedback on #1968 regarding testing django that I never got to apply. We can focus the testing on future PRs that use the models in the API.

My other 2 comments are:

1. (general comment) we should keep an eye on these models as the integration progresses and make sure that all the fields contained here are useful
2. in principle this wouldn't affect any execution, but I don't love the idea of including this code in the 0.31.0 release planned for next week. I think we should hold off from merging this PR to main until the release is out. EDIT: After discussing it with @avilches we should merge on main as we go to be able to verify that the changes at least don't break the current state of the codebase.

gateway/core/models.py
```diff
@@ -117,6 +125,12 @@ class Program(ExportModelOperationsMixin("program"), models.Model):
     image = models.CharField(max_length=511, null=True, blank=True)
     env_vars = models.TextField(null=False, blank=True, default="{}")
     dependencies = models.TextField(null=False, blank=True, default="[]")
+    runner = models.CharField(
+        max_length=20,
+        choices=RUNNER_CHOICES,
+        default=RAY,
+        help_text="Execution backend for this program",
+    )
```
avilches, line 133: This flag should be added in the job model too. And it should be copied during job execution in the `RunJobSerializer.create` func, located in the `gateway/api/serializers.py`.

avilches (commented):
And the `ServerlessClient` should accept the runner flag during function creation. In this way, we can create integration tests with fleets.

gateway/tests/api/test_code_engine_models.py
avilches: These tests have already been deleted in @ElePT's original PR. The goal of this PR should be exclusively to add the "runner" flag to the creation of programs from the client, to the program and to the job. @marceloamaral

ElePT:
I didn't mention this in my initial review to avoid adding noise, but given that @avilches has now added #1968 to the merge queue, I think it makes sense to adjust the approach here.

Since the bulk of the changes in this PR overlap with #1968, I'd suggest rebasing once #1968 merges to main and centering this PR on the runner flag implementation, including:

- The runner flag addition throughout the codebase
- Client-side changes to support the flag
- Branching logic in the scheduler based on the flag
- Unit tests covering the new functionality

I think it would be a logical next step and enable a bunch of testing capabilities.

gateway/core/models.py
marceloamaral: Done! Now the runner field was added to Job and inherited from the program in `RunJobSerializer.create()`

marceloamaral:
Thanks all for the comments.
                                                                                                      
  @korgan00 Agreed and removed the Django ORM tests. We can add meaningful ones in another PRs once
  there's actual logic worth testing.                                                                 
                  
  @ElePT @avilches Rebased onto main after #1968 merged and refocused the PR around the runner flag,  
  as suggested. The changes now cover:
  - runner field on Program and Job (with auto-generated migration)                                   
  - Runner branching in the scheduler (`schedule_queued_jobs`, `update_jobs_statuses`, `schedule`)          
  - Runner support in the API serializers and use cases (`get_logs`, `provider_logs`, `stop`). But `provider_logs` will change later with `FleetsRunner` implementation.
  - runner field in the client QiskitFunction and upload functions                                    
                                                                                                      
  `FleetsRunner` implementation will come in a follow-up PR.                                            
                                                                                                      
  Please take another look when you get a chance, thanks!

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -850,6 +851,7 @@ def _upload_with_artifact(  # pylint:  disable=too-many-positional-arguments, to
                         "title": program.title,
                         "provider": program.provider,
                         "entrypoint": program.entrypoint,
+                        "runner": program.runner,
```
ElePT, line 872: We will need a unit test for this change.

client/qiskit_serverless/core/function.py
```diff
@@ -74,6 +74,7 @@ class QiskitFunction:  # pylint: disable=too-many-instance-attributes
     tags: Optional[List[str]] = None
     raw_data: Optional[Dict[str, Any]] = None
     image: Optional[str] = None
+    runner: str = None
```
ElePT, line 77: I think that the runner should have a default value and have it be "ray" until the transition is over

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -50,16 +50,18 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
         if logs:
             return logs
 
-        # Get from Ray if it is already running. Then filter
-        if job.compute_resource and job.compute_resource.active:
+        # Get from runner if job is active (Ray: compute_resource must be active, Fleets: fleet_id must exist)
+        should_try_runner = (job.compute_resource and job.compute_resource.active) or job.fleet_id
+        if should_try_runner:
```
ElePT: There is an ongoing discussion in #1991 regarding where the "is_active" check should be. I think that it makes sense to move the logic to the runner and add an "is_active" method to the fleets runner as well. The comment applies to all of these use cases.

gateway/api/migrations/0046_alter_config_options_alter_jobevent_options_and_more.py
```diff
@@ -0,0 +1,31 @@
...
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='config',
+            options={'ordering': ['name'], 'verbose_name': 'Configuration', 'verbose_name_plural': 'Config values'},
+        ),
+        migrations.AlterModelOptions(
+            name='jobevent',
+            options={'ordering': ('-created',)},
+        ),
```
ElePT, line 20: are any of these 2 migrations required? I don't see the job event being changed in the code. I am not the most knowledgeable in this front but would it be possible to simply extend 0045_codeengineproject_job_fleet_fields to include the runner changes? the change hasn't been deployed anyways.

gateway/core/services/runners/__init__.py
```diff
@@ -16,8 +16,12 @@ def get_runner(job: Job) -> AbstractRunner:
...
-    # if job.fleets:
-    #     return FleetsRunner(job)
-    return RayRunner(job)
+    Raises:
+        RunnerError: If runner type is invalid or not supported
+    """
+    if job.runner == Program.FLEETS:
+        raise RunnerError(f"Fleets runner is not supported yet: {job.runner}")
+    if job.runner == Program.RAY:
+        return RayRunner(job)
```
ElePT, line 26: we can already add a unit test for this

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -102,6 +112,8 @@ def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches
                 backup_logs = job.logs
                 backup_resource = job.compute_resource
                 backup_ray_job_id = job.ray_job_id
+                backup_fleet_id = job.fleet_id
+                backup_fleet_id = job.fleet_id
```
ElePT: Why is this duplicated?

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -41,9 +41,18 @@ def run(self):
             return
 
         self._update_job_status_counts_metric()
+        self._schedule_fleets_jobs()
```
ElePT, line 44: Do you think it makes sense to add the scheduler business logic changes now or would it make more sense to group these changes with the fleets runner? My initial suggestion was only referring to the changes in `gateway/core/services/runners/__init__.py`, I wasn't thinking of including the scheduler changes as part of this PR, but I am happy to hear your perspective, in the end is a matter of finding a balance between number of PRs and "reviewability". The only point is that if you think it makes sense here, then we need unit tests for all of the scheduler changes you added in this and the following files.

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -41,9 +41,18 @@ def run(self):
...
         self._schedule_cpu_jobs()
         self._schedule_gpu_jobs()
 
+    def _schedule_fleets_jobs(self):
+        """Schedule Fleets jobs (Code Engine). These don't use Ray clusters."""
+        max_fleets = settings.LIMITS_MAX_FLEETS
+        running_fleets = Job.objects.filter(status__in=Job.RUNNING_STATUSES, runner=Program.FLEETS).count()
+        self._schedule_jobs_if_slots_available(
+            max_fleets, running_fleets, gpu_job=False, runner=Program.FLEETS, max_limit=max_fleets
```
ElePT, line 53: I understand that this is a refactoring you introduced to add a potential slot limit to match the maximum number of allowed fleets per project. I think that's a good idea 👍

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -58,28 +67,29 @@ def _schedule_gpu_jobs(self):
 
     def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches
         self,
-        max_ray_clusters_possible,
-        number_of_clusters_running,
+        max_slots_possible,
+        number_of_slots_running,
         gpu_job,
+        runner=Program.RAY,
+        max_limit=100,
```
ElePT, line 74: Why did you add max_limit as an input here? Do you expect this limit to be different for ray and fleets? In the end the database is the same, so I don't think it should make a difference.

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -137,6 +149,8 @@ def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches
                         job.logs = backup_logs
                         job.compute_resource = backup_resource
                         job.ray_job_id = backup_ray_job_id
+                        job.fleet_id = backup_fleet_id
+                        job.fleet_id = backup_fleet_id
```
ElePT: same?

gateway/scheduler/tasks/update_jobs_statuses.py
ElePT, line 210: If the logic is the same in both cases can't we keep it without filtering and just add a separate counter for fleets and ray jobs for logging purposes? I might be missing something. On another note, I think that the comment about sequential jobs is valid, it's a limitation we have commented but not yet reached in production.

ElePT (commented):
Thanks a lot @marceloamaral! I think this PR is on the right track. I left a series of comments where I think we should clarify the logic a bit better. And as a general suggestion, try to think of where unit tests will be needed when you are working on a feature. Another general suggestion is to keep an eye on the CI checks, they will tell you what you need to fix in terms of formatting tests (for example, black is failing at the moment).

gateway/api/migrations/0046_alter_config_options_alter_jobevent_options_and_more.py
avilches: These migrations don't generate any SQL, but for some reason, Django generates them automatically because the internal domain model changed. They can be deleted, but they will reappear in the next migration. They're harmless...

client/qiskit_serverless/core/function.py
avilches: I agree. Anyway, the default value should actually be specified when we read it from the database. In other words, any job to be executed that comes from a program without a runner should be "ray".

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -50,16 +50,18 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
             try:
                 logs = get_runner(job).logs()
             except RunnerError:
                 return "Logs not available for this job during execution."
 
             logs = check_logs(logs, job)
 
-            logger.info("Getting logs from ray job [%s] job_id=%s", job.ray_job_id, job.id)
+            job_id_type = "ray_job_id" if job.ray_job_id else "fleet_id"
+            logger.info("Getting logs from %s [%s] job_id=%s", job_id_type, job.ray_job_id or job.fleet_id, job.id)
```
avilches, line 63: Why not just print the `job.runner` string in the logs?

gateway/api/use_cases/jobs/stop.py
```diff
@@ -110,13 +110,17 @@ def _cancel_runtime_job(self, job_instance, job_id_str):
...
                 if was_running:
                     self.status_messages.append("Serverless job was running and has been stopped.")
                 else:
                     self.status_messages.append("Serverless job was already not running.")
             except RunnerError:
-                logger.warning("Serverless job was not accessible from: %s", job.compute_resource)
+                if job.compute_resource:
+                    logger.warning("Serverless job was not accessible from: %s", job.compute_resource)
+                else:
+                    logger.warning("Serverless job was not accessible: fleet_id=%s", job.fleet_id)
```
avilches, line 125: We have a lot of places in the code where we are printing the job.compute_resource in the logs (the str(compute_resource) returns the title, which is the host url). I would add a new method in the runner like runner.host_info() or something like that, that returns a string. For Ray, it could return the compute_resource.title. For fleets, it could return whatever you deem necessary to debug an error.

gateway/scheduler/schedule.py
```diff
@@ -37,12 +37,19 @@ def execute_job(job: Job) -> Job:
...
+            if compute_resource is not None:
+                compute_resource.save()
+                job.compute_resource = compute_resource
+                span.set_attribute("job.clustername", compute_resource.title)
+
+            if job.runner == Program.FLEETS:
+                job.fleet_id = job_id
+            else:
+                job.ray_job_id = job_id
+
```
avilches: Can we postpone this change? the `submit()` method is not correct right now because it returns a ComputeResource, which only makes sense with Ray and forces you to have this code afterwards. The submit() method should handle (in Ray) creating the ComputeResource and in Fleets, assigning the fleet_id.

I know it's not your fault having to fix the submit I made, so I can take care of this change.

avilches: The fix: https://github.com/Qiskit/qiskit-serverless/pull/1997

gateway/api/use_cases/jobs/get_logs.py
marceloamaral: Alright, let’s merge #1991 first. After than, I can rebase and add the logic (`(job.compute_resource and job.compute_resource.active) or job.fleet_id`) there.

gateway/scheduler/tasks/update_jobs_statuses.py
marceloamaral: That makes sense, merging into a single query reduces the number of DB queries to maintain. 
I will updated the code accordingly.

gateway/api/use_cases/jobs/get_logs.py
marceloamaral: Good point! I simplified it to use `job.runner` directly.

gateway/api/use_cases/jobs/stop.py
marceloamaral: Interesting, let's add the `host_info()` method to the runner later

gateway/scheduler/schedule.py
marceloamaral: Let's merge #1997 and then I can follow the logic here...

gateway/scheduler/tasks/schedule_queued_jobs.py
marceloamaral: I think it makes sense to keep the scheduler changes in this PR.

The runner branching logic and the scheduler changes are closely related, so splitting them would make both PRs harder to review on their own. It really only makes full sense when you look at both together.

Also, the fact that this came up in the discussion feels like a sign that it belongs here :) @avilches has already created a new PR #1997 to improve the logic.

gateway/scheduler/tasks/schedule_queued_jobs.py
marceloamaral: Fleets will have very high concurrency, LIMITS_MAX_FLEETS (1000+), but since `max_limit` (default 100) caps the number of jobs queried per scheduling tick, a default of 100 means any job beyond that are silently skipped, so jobs never get scheduled. 

For Ray, 100 is fine because LIMITS_MAX_CLUSTERS (6) will only allow to run a small number of jobs at once.

marceloamaral: Actually, `max_limit` should be derived from `free_slots` since there's no point querying more jobs than available slots. I've removed it as a parameter and compute it internally as `min(free_slots, settings.LIMITS_MAX_FLEETS)` as a DB safety cap, applied consistently to both Ray and Fleets. And for Ray, `free_slots` will always be small enough that the min with `LIMITS_MAX_FLEETS` won't matter in practice.

client/qiskit_serverless/core/function.py
marceloamaral: @avilches is right, the gateway serializer already handles this at `api/serializers.py:153` with `validated_data.get('runner', Program.RAY)`, so any program uploaded without a runner is set to default to  `ray` on the server side.

marceloamaral: Although it is redundant, I will also add it here

client/qiskit_serverless/core/clients/serverless_client.py
marceloamaral: Makes sense! Added!

gateway/scheduler/tasks/schedule_queued_jobs.py
avilches: I'm ok adding new methods here where they do their own stuff. This way we can delete the Ray methods easier later. In fact @marceloamaral feel free to build a new task `schedule_fleets_jobs`.

gateway/scheduler/tasks/update_jobs_statuses.py
```diff
@@ -94,21 +103,23 @@ def update_job_status(self, job: Job):
...
-                    logs = runner.logs() or ""
-                except RunnerError:
-                    logs = ""
-                save_logs_to_storage(job, logs)
+                if not is_fleets_job:
+                    try:
+                        logs = runner.logs() or ""
+                    except RunnerError:
+                        logs = ""
+                    save_logs_to_storage(job, logs)
```
avilches, line 111: Why skip this part for fleets? Can't you just make logs() return nothing for fleets, so when we have it later on, it'll already be working?

avilches (dismissed):

gateway/scheduler/tasks/schedule_queued_jobs.py
marceloamaral: Sounds good, but let's keep the `schedule_fleets_jobs` simple now.
In the future, we might want to add more complexities, such as the resource availability logic, specially if we have multiple zones.

gateway/scheduler/tasks/update_jobs_statuses.py
marceloamaral: This is getting the log from Ray and saving it to COS, right?

In Code Engine Fleets, the logs are already stored in COS, so the Gateway does not need to manage that part. That is why we skip it here.

marceloamaral:
rebased with https://github.com/Qiskit/qiskit-serverless/pull/1998

client/qiskit_serverless/core/function.py
ElePT: Reminder

gateway/api/use_cases/jobs/get_logs.py
ElePT: did the change get implemented?

gateway/scheduler/tasks/update_jobs_statuses.py
ElePT: Did the code get updated?

ElePT (commented):
Mostly LGTM, I just added a few reminders from the past review of comments I didn't see implemented.

gateway/api/use_cases/jobs/get_logs.py
marceloamaral: Alright, I lost some updates after rebasing...
Let me redo it :)

gateway/scheduler/tasks/update_jobs_statuses.py
marceloamaral: @ElePT the code was updated correctly now, could you please have a look again?

avilches (dismissed):

gateway/scheduler/tasks/update_jobs_statuses.py
avilches: That's correct, ok then

avilches (approved):
LGTM!!

PR #1991, opened 2026-04-14
add is_active function

korgan00:
### Summary

Add this is active function to simplify some parts of the fleets code


ElePT (dismissed):
The change is straightforward enough, LGTM. @marceloamaral, this should help a bit with the `FleetsRunner` integration.

gateway/core/models.py
```diff
@@ -301,6 +301,10 @@ def in_terminal_state(self):
         """Returns true if job is in terminal state."""
         return self.status in self.TERMINAL_STATUSES
 
+    def is_active(self) -> bool:
+        """Returns true if job has any active pod."""
+        return self.compute_resource and self.compute_resource.active
```
avilches, line 358: Shouldn't we redirect to the Runner here, which is the one that knows how to distinguish whether the job belongs to fleets or Ray, and act accordingly? I think the `is_active` should be part of the runner instead of the Job, right?

But I'm afraid we're dealing with a very similar problem that we already have in the update sub_status method, do you remember Goyo? that we can't trust the job status just by looking at the database (because the Scheduler always has a delay of several seconds), and that the only way to know if the job is active (in Ray or Fleets), is by accessing the container url. This is the reason in the sub_status endpoint we have a call to the update_status scheduler function. So, I think we should move the is_active to the runner, and check the url. Then, we could use this is_active here, in the stop and get_logs endpoint, but also in update sub_status, and this way we can remove the call to update_job_status (which is, after all, a coupling of the Gateway to a scheduler method!). What do you think? :)

ElePT: That's a good point @avilches, although I was assuming this would be done as part of the new runner integration work. As you mention, there are a couple of places outside of the runner where we are assuming that a compute resource exists and this won't hold when we start introducing fleets jobs, to mention a few: `gateway/core/services/job_status.py`, `gateway/scheduler/tasks/schedule_queued_jobs.py `and `gateway/scheduler/tasks/update_jobs_statuses.py`. I am in favor of moving the compute-resource dependent logic to the runner, I don't have a strong opinion on whether to do it now or in follow-ups, I think both could work. cc: @marceloamaral.

korgan00: I dont know well how fleets work, but in the case of ray and compute resource, we change de CR to not active just after deleting it. So, the delay should not affect here because the job is finished but the CR is not free and we can access it until the `scheduler/free_resources` delete it.
Anyway, your idea i think it make sense. Just to double check to see if I understood correctly, you are thinking in something like:
```python
# usage
if get_runner(job).is_active():
```
...or the real case...
```python
# usage
job_runner = get_runner(job)
if job_runner.is_active():
    try:
        logs = job_runner.logs()
```
Am i right?

ElePT: @korgan00 you should take a look at #1990 for context. The PR is already touching on these points where I mentioned that we added a lot of branching accessing the compute resource. I also want to make sure that the work is coordinated with @marceloamaral.

avilches: > I dont know well how fleets work, but in the case of ray and compute resource, we change de CR to not active just after deleting it. So, the delay should not affect here because the job is finished but the CR is not free and we can access it until the `scheduler/free_resources` delete it. Anyway, your idea i think it make 

You are right Goyo! and it's very interesting because you've hit on a different nuance between knowing if the job has finished or if it's running, and one has a delay and the other doesn't. If we want to know if the job has stopped (to go to COS for the logs for example), we can simply check if the compute resource is not active, and there's no delay because it's set to False precisely when the cluster dies, just as you said. 

However, if we want to know whether the job is running and accepting requests (can be stopped, get the logs, get the state, whatever), we cannot rely simply on the status field in the database to check if it is RUNNING, as that status does have a delay because the scheduler...

> sense. Just to double check to see if I understood correctly, you are thinking in something like:
> 
> ```python
> # usage
> if get_runner(job).is_active():
> ```
> 
> ...or the real case...
> 
> ```python
> # usage
> job_runner = get_runner(job)
> if job_runner.is_active():
>     try:
>         logs = job_runner.logs()
> ```
> 
> Am i right?

Yes, that's correct. The implementation of this is_active() could be tricky. If we check only for ComputeResource.active = true, maybe the job has finished but the cluster is still alive. We can discuss it later if you don't like it, my proposal is to create two methods: instead of is_active() only, have 1) is_alive() is a fast check that returns true if the container is still alive (the ComputeResource flag or any other flag for Fleets could work) 2) is_running() to check that the job is still running. This method should indeed require that the container responds to requests (in the case of Ray, ask it for the status and have it respond that it's running, which is what the scheduler does).

So, for this PR, the is_alive() will be enough in the get log endpoint, and is_running will be needed for the stop() and the update sub_status endpoints.

marceloamaral: I might be missing something, but do we really need both `is_alive()` and `is_running()`?

The goal here is simply to determine whether it is still active so we know if we can fetch the logs in Ray, right?

In Code Engine, we do not need to overcomplicate this. We only need to know whether a fleet was running to decide whether logs can be fetched from COS. Log retrieval can then wait for the logs to appear in COS, handling any delay in between.

marceloamaral (approved):
Looks good to me.

Let’s merge this, and I can rebase and continue working on top of it in #1990.

gateway/core/models.py
avilches: We need to distinguish between active and alive because we need to use these methods for the update sub_status, stop, and get logs endpoints. Next time please don't merge a PR while we're still discussing the changes...

marceloamaral: Sorry @avilches, I thought we were aligned on adding `is_active()` first and then extending it later with `is_alive()` and `is_running()` for Ray, and I wanted to move forward with the other PR.

But we can revert the merge and continue the discussion.

avilches: It's ok. I know you needed this PR merged to move forward with the other PR on the fleets integration. The current code is totally valid, we can perfectly make the changes we want in another PR.

PR #1997, opened 2026-04-15
Decouple ComputeResource from the runner interface

avilches:
### Summary

ComputeResource was originally part of the AbstractRunner.submit() method, returned as a tuple with the job ray id. We expected that both Fleets and Ray runners would use the ComputeResource. Then we decided we don't need that ComputeResource in fleets.

### Details and comments

This PR removes the return value from submit(), it now assigns compute_resource and ray_job_id directly onto self._job instead of returning them. The caller no longer needs to unpack the result or wire the compute resource back to the job. Tests have been updated accordingly.

The Fleets submit() version should take care of the fleets_id.


marceloamaral (approved):
Alright, it looks good!

Let's merge it and I can rebase this logic into #1990

PR #1998, opened 2026-04-16
Refactor `is_active()`

avilches:
### Summary

The `job.is_active()` only checks if ComputeResource.active. We need a better way to:
- Ensure that if the ComputeResource is active, the host responds, so we can safely call the `logs()` or `stop()` methods (we can't rely on the job.status field because it may not be updated).
- Move this behavior in a Runner, so Fleets can have its own is_active() implementation.

### DIFF
--- a/gateway/core/services/runners/fleets_runner.py
+++ b/gateway/core/services/runners/fleets_runner.py
@@ -14,6 +14,10 @@ class FleetsRunner(AbstractRunner):
     def disconnect(self) -> None:
         """Disconnect from CodeEngine?? (remove the SDK)"""
 
+    def is_alive(self) -> bool:
+        """Check if the Fleets engine is alive."""
+        raise NotImplementedError("FleetsRunner not yet implemented")
+
     def submit(self) -> None:
         """
         Submit the job to Fleets.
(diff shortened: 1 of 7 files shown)


marceloamaral (dismissed):
We could have kept `is_active`, but `is_alive` is also fine.

The reason is that a Code Engine Fleet may already have been terminated, but if it was not deleted, it can still be considered active but maybe not alive.

That said, this is a very small semantic difference, and the PR looks good to me.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -14,6 +14,10 @@ def connect(self) -> None:
     def disconnect(self) -> None:
         """Disconnect from CodeEngine?? (remove the SDK)"""
 
+    def is_alive(self) -> bool:
```
marceloamaral: Maybe the best option would be `is_ready()`

- for Ray, “ready” can mean reachable by ping
- for Fleet, “ready” can mean it exists and is usable

avilches: Ready sounds to me like you can run a job there. But actually in Ray is the opposite... the container could be alive (you can get the logs or the state), but the job could completed or failed...

avilches: I can rename it to match both worlds

marceloamaral: What about
- `is_accessible()`
- `is_reachable()`

avilches: Done

marceloamaral (approved):
LGTM

PR #2001, opened 2026-04-17
Refactor queue jobs

avilches:
### Summary

This PR tries to reduce the time drastically between when a job is launched and the database is updaed.


gateway/core/services/job_status.py
```diff
@@ -58,9 +56,6 @@ def update_job_status(job):
             job.sub_status = None
             job.env_vars = "{}"
 
-    if logs:
-        job.logs = check_logs(logs, job)
-
```
avilches, line 63: Dead code, we don't use this field anymore...

gateway/core/services/runners/ray_runner.py
```diff
@@ -145,8 +152,19 @@ def submit(self) -> None:
                     ray_job_id,
                     title,
                 )
-                self._job.compute_resource = compute_resource
-                self._job.ray_job_id = ray_job_id
+
+                # we have some cases where functions executed and try to send error events or update substates,
+                # but the jobs table still doesn't reflect the data, doing this now mitigates the error
+                self._job.status = self.status()
+                self._job.save(update_fields=["status"])
```
avilches:
```suggestion
```

gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -98,62 +97,16 @@ def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches
...
-                        retries,
-                    )
-                else:
-                    logger.warning(
-                        "job_id=%s Job save failed after %s tries (%.2fs)",
-                        job.id,
-                        retries,
-                        time.monotonic() - t1,
-                    )
+                JobEvent.objects.add_status_event(
```
marceloamaral, line 107: So, was the race condition caused by this while loop? or something else?

avilches: Actually it was a totally different reason. This code is because Django uses optimistic locking, and it could fail if the job read in a previous step has a different version than when you save it. So the loop tries to solve this. It's not very efficient. We solve this by not using save() and persisting only specific fields in other place.

ElePT (commented):
I'm in favor of grouping the pending/running statuses, but I have noticed that the kubernetes tests are consistently timing out (I reran the, 3 times), which makes me thing that some tests might be getting stuck on the scheduler because of this change. What do you think @avilches @korgan00?

avilches:
I splitted this PR into two PRs: 
- This new PR has the fix only: https://github.com/Qiskit/qiskit-serverless/pull/2006
- And now this PR adds the refactor in the scheduler speed on top of the previous PR's branch.

avilches:
Replaced by this shorter version: https://github.com/Qiskit/qiskit-serverless/pull/2106/changes

PR #2006, opened 2026-04-20
Race condition with send error

avilches:
### Summary

Fix a race condition where jobs submitted to Ray are executed too fast, and the job state is still not updated in the db. The send event and update sub status required the job status to be running. Now we accept that this may not be the case.

### Details and comments

This PR removes the check that state is RUNNING and accept QUEUED and PENDING too


ElePT (approved):
This is consistent with our high-level discussions but I'd like to get @Tansito or @korgan00's opinions here before merging.

ElePT (commented):
Merging in an attempt to unblock the release, we can continue discussing improvements on follow-ups.

PR #2008, opened 2026-04-21
Remove duplicated code

avilches:
### Summary

During the access policies review, I discovered a duplicated method. This PR removes it and replaces it with a call to `ProviderAccessPolicy.can_access`


ElePT (approved):
LGTM

PR #2028, opened 2026-04-22
Add business_model to job table + migration

avilches:
### Summary

Adds a `business_model` field to the `Job` model as part of the new instance-based access control system.
- Now, this  `business_model` field is set on creation time based on the already existing `trial` field. If `trial` is true, `business_model` is "trial". If it `trial` is false, `business_model` will be "subsidized". 

In the future, we will get the business model from Runtime API authorization endpoint. 

- Existing production jobs are updated in a migration using the same rule.

Migration occurs in three steps:
- 0047 Adds the field as nullable
- 0048 Update the field -> business_model = trial ? "TRIAL" : "SUBSIDIZED"
- 0049 Sets the field as no nullable

There are around 23800 jobs in the production database. The query in the `0048` migration will update all of them in one statement. Tested in local with a 25000 fake jobs and it takes **well under one second**. No downtime, no table lock, no noticeable impact on concurrent reads or writes.


gateway/core/models.py
```diff
@@ -347,6 +347,13 @@ class Job(models.Model):
         (POST_PROCESSING, "Post-processing"),
     ]
 
+    BUSINESS_MODEL_TRIAL = "TRIAL"
+    BUSINESS_MODEL_SUBSIDIZED = "SUBSIDIZED"
+    BUSINESS_MODELS = [
+        (BUSINESS_MODEL_TRIAL, "Trial"),
+        (BUSINESS_MODEL_SUBSIDIZED, "Subsidized"),
+    ]
```
riley2012, line 357: Does it make sense to include `(BUSINESS_MODEL_CONSUMPTION, "Consumption"),` since that will be a valid value returned from the Runtime `GET /functions` API? Even though it won't be used now in the migration in this PR, it might make sense to include it as a valid value since it's included in both [[Epic] Access through instances](https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1423) and [[User Story]: Functions access through instances](https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1302)

avilches: I didn't want to it now, because it's not going to be used yet, but I can add it sure

avilches, line 357:
```suggestion
    BUSINESS_MODEL_TRIAL = "TRIAL"
    BUSINESS_MODEL_SUBSIDIZED = "SUBSIDIZED"
    BUSINESS_MODEL_CONSUMPTION = "CONSUMPTION"
    BUSINESS_MODELS = [
        (BUSINESS_MODEL_TRIAL, "Trial"),
        (BUSINESS_MODEL_SUBSIDIZED, "Subsidized"),
        (BUSINESS_MODEL_CONSUMPTION, "Consumption"),
    ]
```

gateway/core/models.py
```diff
@@ -347,6 +347,17 @@ class Job(models.Model):
         (POST_PROCESSING, "Post-processing"),
     ]
 
+    BUSINESS_MODEL_TRIAL = "TRIAL"
+    BUSINESS_MODEL_SUBSIDIZED = "SUBSIDIZED"
+    BUSINESS_MODEL_TRIAL = "TRIAL"
+    BUSINESS_MODEL_SUBSIDIZED = "SUBSIDIZED"
```
riley2012:
```suggestion
    BUSINESS_MODEL_TRIAL = "TRIAL"
    BUSINESS_MODEL_SUBSIDIZED = "SUBSIDIZED"
```
Do you want to remove the duplicated constants?

avilches: Ops, I think I applied the changes twice from the Github UI

riley2012 (approved):
LGTM

PR #2029, opened 2026-04-22
Add derived field `platform_id` to easy lookup functions from runtime api

avilches:
### Summary

Adds a `platform_id` field to the `Program` model as part of the new instance-based access control system.

- `platform_id` is a stable, unique identifier for provider functions, calculated automatically as `provider.name + "." + program.title`. It is `null` for private functions (no provider).
- The field is set on every `save()` via a model override, so it stays consistent regardless of the code path used to create or update a program (API, admin, management commands).

In the future, `platform_id` will be used as the key to identify functions against the external `FunctionAccessClient` (Phase 2).

- Existing production programs with a provider are updated in the migration using the same rule.

Migration occurs in two steps in the same 0047 migration:
- Adds the field as nullable + unique
- Update programs rows 

The update query updates all provider programs in one SQL statement. There are only 32 programs to update in production, so the expected update time is almost zero.


avilches:
Draft until #2028 is merged because the migrations need to be sequential...

gateway/core/models.py
```diff
@@ -157,6 +158,13 @@ class Meta:
         app_label = "api"
         permissions = ((RUN_PROGRAM_PERMISSION, "Can run function"),)
 
+    def save(self, *args, **kwargs):
+        if self.provider_id is not None:
```
ismaelRozasRamallal: We need to add an additional condition here, if platform_id exists, we don't want to change it. If not, each time the title changes, the id changes and we will be invalidating the current instance configurations

gateway/tests/api/test_v1_program.py
```diff
@@ -572,7 +576,7 @@ def test_upload_provider_function_with_title(self):
             programs_response = self.client.post(
                 "/api/v1/programs/upload/",
                 data={
-                    "title": "default/Provider Function",
+                    "title": "default/provider-function",
```
ismaelRozasRamallal, line 585: One doubt, which is the difference between this test and the previous one? both of them have title, the only thing is this concrete one prefixing it with the provider (which I don't think it's needed?)

gateway/core/models.py
avilches: Renaming doesn't happen very often, but product and architecture team are aware of the implications of renaming: basically, permission will stop working. We suggested using uuid because of this, and we all accept the risk because of the low chances of it happening.

gateway/tests/api/test_v1_program.py
```diff
@@ -473,6 +473,8 @@ def test_upload_private_function(self):
             )
             assert programs_response.status_code == status.HTTP_200_OK
             assert programs_response.data.get("provider") is None
+            program = Program.objects.get(title="Private function", author__username="test_user_2", provider=None)
+            assert program.platform_id is None
 
     def test_upload_custom_image_without_provider(self):
```
ismaelRozasRamallal, line 479: Is there any test about uploading a custom image succesfully? I don't see it anywhere and it could be also good to test they doesn't add a platform_id

gateway/tests/api/test_v1_program.py
avilches: It detects the `/` character and splits it into two parts. The first part becomes the provider name and the second one the function name.

gateway/tests/api/test_v1_program.py
avilches: Why shouldn't uploading a function result in a platform_id update?

ismaelRozasRamallal (commented):
Overall looks good! I just leave some comments we need to solve before merging 😄

avilches:
Closed as unnecessary. This approach is equally valid and doesn't require an additional field: https://github.com/Qiskit/qiskit-serverless/pull/2035
Thanks @ismaelRozasRamallal for the idea.

avilches:
> Closed as unnecessary. This approach is equally valid and doesn't require an additional field: #2035 Thanks @ismaelRozasRamallal for the idea.

Reopen: now, the platform_id will never change /cc @ismaelRozasRamallal @riley2012

riley2012:
### Proposal: non-nullable `platform_id`

Currently `platform_id` is nullable because it was designed only for provider functions. However, every function has either a `provider` or an `author`, so we can always construct a unique identifier. I'd like to propose making `platform_id` non-nullable:

- For provider functions: `{provider.name}.{program.title}`
- For private functions: `{author_id}.{program.title}`

The unique constraints from PR #2035 already guarantee both combinations are unique at the DB level, so this is safe.

Benefits:

- Every function has a stable, human-readable cross-service identifier from creation
- Eliminates null checks wherever `platform_id` is consumed
- Opens the door for human-readable identifiers in future use cases beyond the Runtime API

The migration impact is negligible — `api_program` has only 676 rows.

For private functions, `author_id` is an integer and the function is only accessible to its author, so there are no privacy concerns with using it as part of the identifier.

PR #2031, opened 2026-04-22
Add compute profile field

ElePT:
### Summary
This PR implements a new field through the client and gateway code to be able to specify a compute profile for a specific function execution. Note that the code is currently not used yet (will be used in the `FleetsRunner`) so the key for this PR is to verify that its addition doesn't break the current implementation.

### DIFF
--- a/gateway/api/serializers.py
+++ b/gateway/api/serializers.py
@@ -8,7 +8,6 @@ Version serializers inherit from the different serializers.
 
 import json
 import logging
-import re
 from typing import Tuple, Union
 
 from django.conf import settings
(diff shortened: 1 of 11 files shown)


gateway/api/serializers.py
```diff
@@ -219,11 +220,35 @@ def create(self, validated_data):
...
+
+    def to_representation(self, instance):
+        """
+        Include compute_profile in the response by reading it from the Job instance.
+        This is necessary because compute_profile is set during create() and needs
+        to be explicitly included in the serialized output.
+        """
+        representation = super().to_representation(instance)
+        representation["compute_profile"] = instance.compute_profile
+        return representation
```
ElePT: @avilches I was not able to find a solid alternative to defining this to be able to pass along compute_profile. From my understanding this fits "Django standards" but let me know if that's not the case.

avilches (approved):

gateway/api/serializers.py
avilches: Fixed!

marceloamaral (approved):
LGTM

PR #2000, opened 2026-04-16
Add minimal ce-swagger-sdk files

eletonia:
### Summary
This PR adds the relevant `ce-swagger-sdk` files that are required to run the Fleets MVP in staging
`ce-swagger-sdk` was autogenerated with Swagger Codegen and these files are the ones that will be required by the `qf-fleets-client JobHandler`

### Details and comments
Note: this library needs to be tested once the `qf-fleets-client JobHandler` has been merged

PR #2032, opened 2026-04-22
Add FleetHandler, JobWorkers, and JobCOS for Code Engine fleets (PR 3) 

marceloamaral:
### Summary
PR 3 of the IBM Cloud migration plan. Migrates JobHandler from `qf_fleets_client` into the gateway as FleetHandler, eliminating the external dependency for fleet lifecycle management.

### Details and comments
 - FleetHandler — submit, status, cancel, delete, poll with _wait_until_state (429 backoff, general primitive)                                      
  - JobWorkers — list, wait, resource allocation tracking (allocation ≠ utilization; Sysdig deferred to future PR)                                   
  - JobCOS — COS log retrieval with three-priority HMAC resolution: direct credentials → named CE secret → auto-discovery by key ID                  
  - fleet_utils — build_run_volume_mounts, build_run_env_variables, build_run_commands                                                               
  - ce-swagger-sdk trimmed to 17 models actually deserialised at runtime (down from 137)                                                             
  - pytest.ini / Dockerfile / .pylintrc — ce-swagger-sdk on PYTHONPATH for tests, runtime, and lint


gateway/core/ibm_cloud/code_engine/ce_client/api/secrets_and_configmaps_api.py
```diff
@@ -0,0 +1,1127 @@
...
+
+"""
+Code Engine
+
+REST API for Code Engine  # noqa: E501
+
+OpenAPI spec version: 2.0.0
+
+Generated by: https://github.com/swagger-api/swagger-codegen.git
+"""
```
ElePT, line 23: As mentioned in #2000, all files should contain the license header:

```
# This code is a Qiskit project.
#
# (C) Copyright IBM 2026.
#
# This code is licensed under the Apache License, Version 2.0. You may
# obtain a copy of this license in the LICENSE.txt file in the root directory
# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
#
# Any modifications or derivative works of this code must retain this
# copyright notice, and modified files need to carry a notice indicating
# that they have been altered from the originals.
```

gateway/core/services/ibm_cloud/code_engine/fleets/cos.py
```diff
@@ -17,73 +17,149 @@
...
+from typing import TYPE_CHECKING, Any
+
+from swagger_client.api.secrets_and_configmaps_api import SecretsAndConfigmapsApi
+from swagger_client.rest import ApiException
 
 from core.services.ibm_cloud.cos.cos_client import COSClient, CosHmacCredentials
 
 if TYPE_CHECKING:
-    from core.services.ibm_cloud.code_engine.fleets.job import JobHandler
+    from core.services.ibm_cloud.code_engine.fleets.fleet_handler import FleetHandler
```
ElePT: now that the file is added we don't need the `if TYPEC_CHECKING`, right?

gateway/core/services/ibm_cloud/code_engine/fleets/cos.py
```diff
@@ -17,73 +17,149 @@
...
+        elif hmac_secret_name:
+            # Priority 2: fetch both keys from the named CE secret.
+            logger.debug("Fetching HMAC credentials from CE secret %r", hmac_secret_name)
+            creds = self._get_hmac_from_ce_secret(hmac_secret_name)
+        elif hmac_access_key_id:
+            # Priority 3: key ID provided without secret — fetch from the CE secret
+            # whose name matches the key ID.
+            logger.debug(
+                "hmac_secret_access_key not provided — fetching from CE secret %r",
+                hmac_access_key_id,
```
ElePT: from priority 2 and 3, which one do we expect to use in the final version? I don't like having so many alternative paths to do the same thing. Given that we control this setup, I think that we should pick one and confirm it works.

gateway/core/ibm_cloud/code_engine/ce_client/api/secrets_and_configmaps_api.py
```diff
@@ -0,0 +1,1127 @@
+# coding: utf-8
```
ElePT, line 13:
```suggestion
# This code is a Qiskit project.
#
# (C) Copyright IBM 2026.
#
# This code is licensed under the Apache License, Version 2.0. You may
# obtain a copy of this license in the LICENSE.txt file in the root directory
# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
#
# Any modifications or derivative works of this code must retain this
# copyright notice, and modified files need to carry a notice indicating
# that they have been altered from the originals.

# coding: utf-8
```

gateway/core/services/ibm_cloud/code_engine/fleets/fleet_handler.py
```diff
@@ -0,0 +1,440 @@
...
+    )
+
+    # Fleet CRUD
+    handler.submit_job(name="my-fleet", ...)
+    handler.get_job_status("my-fleet")
+    handler.cancel_job("my-fleet", wait=True, delete=True)
+
+    # Workers (sub-manager)
+    handler.workers.list(fleet_id="...")
+    handler.workers.wait(fleet_id="...", timeout=120)
```
ElePT: If I'm not mistaken, we aren't really accessing the workers in the fleets runner, right? nor do we have a clear plan of when we'd access them. I'm not a bit fan of adding features without a plan, I'd rather keep some parts internal and add them on a need basis.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler_integration.py
```diff
@@ -0,0 +1,256 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Integration test for FleetHandler against a live IBM Cloud Code Engine project.
```
ElePT: As discussed offline, we don't have a good way to run these integration files yet on CI/CD, they are referring to env variables that don't exist, so I think we should stick to unit tests and tackle the integration tests in a separate effort. As additional context, if we were to add them, integration tests are placed under qiskit-serverless/tests and have their own configuration separate from the gateway/client tests, but I think this is out of scope for this PR.

ElePT (commented):
Thanks! I left some superficial comments about how the contribution should be structured, and I didn't get into the nitty gritty of the autogenerated code or the handler which uses it, because I think we can polish it in future iteration and the main requirement here is that it works. 

One general comment I have been thinking about since last week is that I find the file paths challenging to traverse and unnecessarily deep. I am going to make a suggestion which we don't need to follow strictly, but I would find it way more natural:

- for the swagger client, flatten the double nesting and rename to sth clearer like: `gateway/ce_client/` instead of `gateway/ce-swagger-sdk/swagger_client/` (the term swagger confuses a lot of people who think we are generating their swagger file)
- for the remaining classes, I think that `services/ibm_cloud/code_engine/` adds a lot of depth with little semantic value. My suggestion would be, instead of:

```
gateway/core/services/ibm_cloud/code_engine/fleets/
  ├── fleet_handler.py
  ├── fleet_workers.py
  ├── fleet_utils.py
  └── cos.py
```
to have the following:

```
gateway/core/fleets/
  ├── handler.py
  ├── workers.py
  ├── utils.py
  └── cos.py
```

This would allow to update the imports to a much simpler structure `from core.services.ibm_cloud.code_engine.fleets.fleet_handler -> from core.fleets.handler`

gateway/core/ibm_cloud/code_engine/ce_client/api/secrets_and_configmaps_api.py
korgan00: I was wondering if we can automatize this and asking AI it seems to be a plugin for the `pre-commit` hooks called [`license-header-checker`](https://github.com/lluissm/license-header-checker) with a config similar to this:

```yaml
- repo: https://github.com/lucasdirani/pre-commit-hooks
  rev: v1.1.0
  hooks:
    - id: insert-license
      args: [--license-filepath, license_header.txt, --comment-style, "#"]
```

ElePT: yes, that's a good idea, we also had a header checking workflow in other qiskit ecosystem repos: https://github.com/qiskit-community/qiskit-optimization/blob/main/tools/verify_headers.py (we could adapt it to not only verify but apply the header)

marceloamaral: Done!

gateway/core/services/ibm_cloud/code_engine/fleets/cos.py
marceloamaral: Not really...
```
ImportError: cannot import name 'FleetHandler' from partially initialized module 'core.ibm_cloud.code_engine.fleets.handler' (most likely due to a circular import)
```

gateway/core/services/ibm_cloud/code_engine/fleets/cos.py
marceloamaral: Using `hmac_secret_name` only now

gateway/core/services/ibm_cloud/code_engine/fleets/fleet_handler.py
marceloamaral: Well, the code was already in `add-fleets-jobhandler`, but I agree with you!

Do you think we should remove the worker code entirely, or only take it out of the example?

My understanding is that the plan was to keep it for scheduling purposes, so we can estimate the job cost. Is that correct?

Actually, I will remove the worker code from this PR!

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler_integration.py
marceloamaral: I see, I agree that wiring these into the automated pipeline is out of scope for this PR, and I understand the concern about env variables that don't exist in CI.                                                                                          

That said, I'd like to keep `test_fleet_handler_integration.py` in this PR for a specific reason: it gives us a simple way to manually validate the gateway's integration with Code Engine without needing to run the full gateway in a local cluster. It's **skipped by default** and only runs with `RUN_INTEGRATION_TESTS=1`), so it **doesn't affect CI at all**. 

We can move/remove this test  in a follow-up PR. For now, keeping it here makes the integration validation accessible to anyone working on the Fleets code without requiring a full deployment.

marceloamaral:
@ElePT, thanks for the detailed review! I've addressed all the comments:                                                                                   
                                                                                                                                                     
- `Workers removed` — JobWorkers and all references are gone due to no clear plan for when it would be used.
- `HMAC credentials simplified` — removed `hmac_access_key_id` and` hmac_secret_access_key` options. The only way to configure COS credentials is now via `hmac_secret_name`, which resolves from a named Code Engine secret. 
- Path restructuring — I agree that very long paths add unnecessary depth, but I also think some context in the path is valuable so we immediately know what package we're looking at. I went for something in between the original and your suggestion:                                            
```
  gateway/core/ibm_cloud/
    ├── clients.py
    ├── cos/
    └── code_engine/
        ├── ce_client/    ← renamed from ce-swagger-sdk/swagger_client/
        └── fleets/                                   
```
  - ce_client is now a regular Python package directly under `code_engine/` — no `sys.path` **hacks**, no `PYTHONPATH` env var needed.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler_integration.py
ElePT: My concern isn't about CI impact, you're right that a skipped test doesn't break anything. My point is that this file doesn't belong in the open-source repository. It references environment variables that only exist in our internal infrastructure, which means it's not usable or reproducible by external contributors. I agree that it can be useful for manual validation, but then the right place for it is the internal branch alongside the other demo and integration files we keep there. I'd prefer not to merge it here.

.gitignore
```diff
@@ -104,4 +104,5 @@ docs/_build/
 .virtual_documents/
 
 # any logs
-*.log
\ No newline at end of file
+*.log
+gateway/media/
```
ElePT: Good catch! These additional files should not be generated by the tests, we just merged a PR to fix that #2037 so this change is no longer required. The rule of thumb here for future contributions is: if you see an unwanted artifact, it's probably a bug introduced somewhere, and the preferred action is tackling the root cause, not adding it to gitignore.

```suggestion
```

gateway/tests/conftest.py
```diff
@@ -11,3 +23,16 @@ def media_root_tmp(tmp_path, settings):
...
+    """Register project-wide custom markers.
+
+    Args:
+        config: The pytest configuration object.
+    """
+    config.addinivalue_line(
+        "markers",
+        "integration: marks tests that require live IBM Cloud credentials "
+        "(skipped unless RUN_INTEGRATION_TESTS=1)",
+    )
```
ElePT: I hate to insist but I am not a fan at all of modifying the conftest to accomodate for a file that shouldn't be run. This change should be reverted and the testing file should be removed. 

```suggestion
```

ElePT (changes requested):
Thanks for the fast implementation! All of the code changes look good, the new paths too, my only final comments are the removal of the gitignore change and the integration test. I don't think its addition to the open source repo is justified and I don't like having to change our pytest configuration for it. I think that the best approach is to commit the file in the internal branch with the other "unofficial' testing files. Once that is done I am ready to merge.

Oh, one final comment, this is totally my bad, I handled a merge conflict manually on [c9f20d1](https://github.com/Qiskit/qiskit-serverless/pull/2032/commits/c9f20d18f4c9537b8004a2dd873571f6854292a0) and probably introduced an unwanted whitespace, so black fails. I didn't fix it because I requested changes on the conftest file anyways and black can get fixed there.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler_integration.py
marceloamaral: Alright, let me remove it then!

gateway/tests/conftest.py
marceloamaral: Right, since will remove the integration test, this will also be updated!

ElePT (dismissed):
LGTM! Thanks for the quick updates!

gateway/tests/conftest.py
```diff
@@ -1,13 +0,0 @@
...
-
-
-@pytest.fixture(autouse=True)
-def media_root_tmp(tmp_path, settings):
-    """Redirect MEDIA_ROOT to a temp directory for every test.
-
-    Prevents PathBuilder.absolute_path() from creating directories inside
-    the source tree (gateway/media/) during test runs.
-    """
-    settings.MEDIA_ROOT = str(tmp_path)
```
ElePT, line 13: sorry, a last-minute review comment: can you restore this file?

ElePT (approved):
approving again

PR #2033, opened 2026-04-23
Add FleetsRunner implementation

marceloamaral:
<!--
⚠️ Also, please add a release note file using reno if the change needs to be
  documented in the release notes.

- [ ] I have added the tests to cover my changes.
- [ ] I have read the CONTRIBUTING document.
-->

### Summary
Implements FleetsRunner, completing the four-PR migration from `qf_fleets_client` to the gateway. This PR depends on https://github.com/Qiskit/qiskit-serverless/pull/2032.

The runner handles the full job lifecycle,  artifact upload, PDS-backed dual-log execution, status polling, log and result retrieval from COS — using the FleetHandler, JobCOS, and fleet_utils modules from PR his PR depends on https://github.com/Qiskit/qiskit-serverless/pull/2032.


### Details and comments

     
  FleetsRunner
  - Implements `AbstractRunner`: connect, submit, status, logs, provider_logs, get_result_from_cos, stop, free_resources                               
  - `submit()` uploads arguments and artifact tar to COS, mounts user/provider PDS volumes, and wires the [public]-filter dual-log wrapper             
  - `status()` maps CE fleet statuses to Job.STATUS; returns None on 429 so the scheduler retries without failing the job                              
  - `get_runner()` factory updated — Program.FLEETS now returns FleetsRunner instead of raising                                                        
                                                                                                                                                                                                                                                 
  compute_profile_parser.py                                                                                                                          
  - New module: parses CE compute profile strings (e.g. gx3d-24x120x1a100p) to extract CPU, memory, and GPU config for submit_job


ElePT (commented):
I'm going to hold off from reviewing this PR because I think it's better to wait until #2032 and #2031 are merged. This way we can reduce this PR to essentially a single file + test. I'll also note that I think that the code is not up to date with the latest status of the MVP1 branch, it got some updates last Monday and for example the compute profile parser is no longer there. So it will be easier if we merge #2032 and #2031, rebase the MVP1 branch, and then see what is left for this contribution.

gateway/api/use_cases/jobs/provider_logs.py
```diff
@@ -52,18 +52,18 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
                 job.id,
                 user.id,
                 job.program.runner,
             )
 
+            try:
+                logs = runner.provider_logs()
+            except RunnerError:
+                return "Logs not available for this job during execution."
+
```
avilches: To be honest, I think the log should be inside the try, just after the provider_logs() call, right?

gateway/core/services/runners/abstract_runner.py
```diff
@@ -135,6 +136,23 @@ def logs(self) -> Optional[str]:
...
+        falls back to logs(), which is appropriate for engines that have no
+        distinction between user and provider logs (e.g. Ray).
+
+        Returns:
+            Provider log content or None
+
+        Raises:
+            RunnerError: If unable to get logs
+        """
+        return self.logs()
```
avilches: I would raise NotImplementedError error, and make the method abstract, like the others. This way, every runner will have the real implementation in the same class..

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,765 @@
...
 
     def is_active(self) -> bool:
-        """Check if the Fleets engine is alive."""
-        raise NotImplementedError("FleetsRunner not yet implemented")
+        """Return ``True`` if a fleet was submitted for this job.
+
+        Returns:
+            ``True`` when ``job.fleet_id`` is set.
+        """
+        return bool(self.job.fleet_id)
```
avilches: We did something similar in Ray, but it didn't work well... we can't trust that the job is active just by checking a flag in the db (in this case, in memory). The scheduler will always lag a bit when updating this id, and it will only update it in the database (actually, is the scheduler deleting the fleet_id when job finishes? I think it shouldn't). Even assuming it had updated it, the in-memory job object wouldn't have it and we'd need to do a job.refresh_from_db(). Anyway, the ideal thing here would be to open a connection to ping the server...

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,765 @@
...
+        back to the environment variable of the same name.  Returns ``None``
+        when the project fields or the secret name are not configured.
+
+        Returns:
+            COS config dict or ``None``.
+        """
+        if not self._project or not self._is_cos_configured():
+            return None
+
+        hmac_secret_name = getattr(settings, "CE_HMAC_SECRET_NAME", None) or os.environ.get("CE_HMAC_SECRET_NAME")
```
avilches: Can we move the CE_HMAC_SECRET_NAME, IBM_CLOUD_API_KEY and CE_ICR_PULL_SECRET to settings? The Django settings are easy to mock in tests and all the config will be in the same place.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,765 @@
...
+            Image reference string.
+
+        Raises:
+            RunnerError: If no image is configured and no default is set.
+        """
+        if not self.job.program:
+            raise RunnerError("Job has no program assigned")
+        if self.job.program.image:
+            return self.job.program.image
+        default = getattr(settings, "FLEETS_DEFAULT_IMAGE", None)
```
avilches: Why not settings.FLEETS_DEFAULT_IMAGE? The field in settings should have its own default value if the environment variable is not configured (or fail if it was mandatory)

gateway/core/services/runners/fleets_runner.py
ElePT: Agreed, I am not sure how troublesome the connection would be, but we are not really removing the id so the current check isn't very effective.

gateway/core/services/runners/fleets_runner.py
ElePT: I think that at least some of them are, and the implementation needs to catch up (but should be confirmed).

gateway/api/use_cases/jobs/provider_logs.py
marceloamaral: Moved the log after the `provider_logs()` call and added job/user context to the error path.

gateway/core/services/runners/abstract_runner.py
marceloamaral: Made provider_logs() abstract!

gateway/core/services/runners/fleets_runner.py
marceloamaral: Agreed, I changed to call `get_job_status()` via the CE API to verify the fleet actually exists. Returns False on 404 or connection errors. No more trusting the only in-memory flag.

gateway/core/services/runners/fleets_runner.py
marceloamaral: Moved `CE_HMAC_SECRET_NAME`, `IBM_CLOUD_API_KEY`, and `CE_ICR_PULL_SECRET` to Django settings.

gateway/core/services/runners/fleets_runner.py
marceloamaral: Added `FLEETS_DEFAULT_IMAGE` to settings.

gateway/core/services/compute_profile_parser.py
```diff
@@ -0,0 +1,142 @@
...
+    "l40s": "l40s",
+    "h100": "h100",
+    "h200": "h200",
+}
+
+# Pattern: [type]-[cpu]x[memory] or [type]-[cpu]x[memory]x[gpu_count][gpu_type]
+_PROFILE_RE = re.compile(r"^([a-z]+\d+[a-z]?)-(\d+)x(\d+)(?:x(\d+)([a-z0-9]+p?))?$")
+
+
+class ComputeProfileParser:
```
ElePT: As mentioned in another comment, the demo currently doesn't use this code. I think it's safer to update this PR to match the demo first. In the future, we will have to add more compute profile management capabilities, but they will probably look different to what's there now. The profile is likely going to be be abstracted away in some way. This means that this file will not be used as-is, and it's better to approach it in a clean manner and add the new capabilities when they are needed.

gateway/main/settings.py
```diff
@@ -286,6 +286,11 @@
 LIMITS_MEMORY_PER_TASK = int(os.environ.get("LIMITS_MEMORY_PER_TASK", "8"))
 # Compute profile settings for Fleets runner
 DEFAULT_COMPUTE_PROFILE = os.environ.get("DEFAULT_COMPUTE_PROFILE", "cx3d-4x16")  # 4 CPU, 16GB RAM
+FLEETS_DEFAULT_IMAGE = os.environ.get("FLEETS_DEFAULT_IMAGE", "python:3.11-slim")
```
ElePT: we can do this in a follow-up, but we will have to update the default image to sth more functional.

gateway/tests/api/test_v1_program.py
```diff
@@ -973,3 +973,27 @@ def test_upload_invalid_version_returns_error(self):
             # Error message should mention invalid version
             errors = json.dumps(response.data)
             assert "Invalid version" in errors
+
+    def test_upload_with_runner_field(self):
+        """Tests that the runner field is persisted on upload."""
```
ElePT, line 978: Nice one!

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
```diff
@@ -107,3 +107,139 @@ def test_cos_delete_object() -> None:
     mock_cos.delete_object.assert_called_once_with(
         bucket="my-bucket", key="some/key", wait=True, timeout_seconds=60, poll_interval=2
     )
+
+
+def test_cos_upload_fileobj() -> None:
+    """upload_fileobj() delegates to COSClient.upload_fileobj."""
+    import io  # pylint: disable=import-outside-toplevel
```
ElePT: why not have this import at the top level?

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
```diff
@@ -107,3 +107,139 @@ def test_cos_delete_object() -> None:
...
+    result = job_cos.get_object_bytes(bucket_name="my-bucket", key="some/key")
+
+    assert result == b"result data"
+    mock_cos.get_object_bytes.assert_called_once_with(bucket="my-bucket", key="some/key")
+
+
+def test_cos_property_priority2_secret_name() -> None:
+    """_cos fetches credentials from a named CE secret when hmac_secret_name is set."""
+    from unittest.mock import patch as _patch  # pylint: disable=import-outside-toplevel
+    from core.ibm_cloud.cos.cos_client import CosHmacCredentials  # pylint: disable=import-outside-toplevel
```
ElePT: same here, all imports should be at the top unless there's clear justification for it (like a circular dependency upon import).

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
```diff
@@ -107,3 +107,139 @@ def test_cos_delete_object() -> None:
...
+
+
+def test_cos_property_raises_when_no_cos_config() -> None:
+    """_cos raises ValueError when cos_config is None."""
+    mock_job = MagicMock()
+    mock_job.cos_config = None
+    job_cos = JobCOS(mock_job)
+
+    with pytest.raises(ValueError, match="COS not configured"):
+        _ = job_cos._cos  # pylint: disable=protected-access
```
ElePT: why make it a private property if you expect it to be accessed?

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
```diff
@@ -107,3 +107,139 @@ def test_cos_delete_object() -> None:
...
+    mock_job = MagicMock()
+    mock_job.project_id = "proj-id"
+    job_cos = JobCOS(mock_job)
+
+    mock_secret = MagicMock()
+    mock_secret.data = {"access_key_id": "ak123", "secret_access_key": "sk456"}
+
+    with patch("core.ibm_cloud.code_engine.fleets.cos.SecretsAndConfigmapsApi") as mock_api_cls:
+        mock_api_cls.return_value.get_secret.return_value = mock_secret
+        creds = job_cos._get_hmac_from_ce_secret("my-secret")  # pylint: disable=protected-access
```
ElePT: we haven't been too strict on this, but fyi, we shouldn't be testing protected attributes in unit tests, pylint's complaint is valid. If we expect the attribute to be used externally it should be public, if not, there should be a "public path" to test it. Again, this hasn't been enforced too well so we can let it slide, but I try to avoid it when possible.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
```diff
@@ -107,3 +107,139 @@ def test_cos_delete_object() -> None:
...
+
+    assert creds.access_key_id == "ak123"
+    assert creds.secret_access_key == "sk456"
+
+
+def test_cos_get_hmac_from_ce_secret_not_found() -> None:
+    """_get_hmac_from_ce_secret() raises ValueError on 404."""
+    from unittest.mock import MagicMock, patch  # pylint: disable=import-outside-toplevel
+    from core.ibm_cloud.code_engine.ce_client.rest import ApiException  # pylint: disable=import-outside-toplevel
+    from core.ibm_cloud.code_engine.fleets.cos import JobCOS  # pylint: disable=import-outside-toplevel
```
ElePT: imports!

gateway/tests/core/services/runners/test_fleets_runner.py
```diff
@@ -0,0 +1,702 @@
...
+def test_is_active_false_on_connection_error():
+    """is_active() returns False when connection to CE fails."""
+    runner, _ = _make_runner(fleet_id="fleet-123")
+    runner._connected = False  # pylint: disable=protected-access
+    runner._handler = None  # pylint: disable=protected-access
+    with patch.object(runner, "connect", side_effect=Exception("connection failed")):
+        assert runner.is_active() is False
+
+
+# status
```
ElePT: let's get rid of these separators.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,774 @@
...
+            self._get_handler().get_job_status(self.job.fleet_id)
+            return True
+        except ApiException as ex:
+            if ex.status == 404:
+                return False
+            logger.warning("CE API error checking fleet [%s]: status=%s", self.job.fleet_id, ex.status)
+            return False
+        except Exception:  # pylint: disable=broad-exception-caught
+            logger.warning("Unable to verify fleet [%s] existence", self.job.fleet_id)
+            return False
```
ElePT, line 160: we'll have to test how much overhead this api call would add. It's hard to tell at first glance, but if we see things running slowly once we start testing, we could look into it again.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,774 @@
...
+            logger.debug("No fleet_id to clean up for job [%s]", self.job.id)
+            return False
+
+        try:
+            self._get_handler().delete_job(self.job.fleet_id)
+            logger.info("Deleted fleet [%s]", self.job.fleet_id)
+            return True
+        except Exception as ex:  # pylint: disable=broad-exception-caught
+            logger.warning("Failed to delete fleet [%s]: %s", self.job.fleet_id, ex)
+            return False
```
ElePT: We may need to disable this temporarily for testing the MVP, because it would make it hard to check that the fleet was created (in case some other step failed, like getting logs or results).

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,774 @@
...
+        """
+        if self.job.code_engine_project:
+            project = self.job.code_engine_project
+            if not project.active:
+                raise RunnerError(f"Code Engine project '{project.project_name}' is not active")
+            return project
+
+        project = CodeEngineProject.objects.filter(active=True).first()
+        if not project:
+            raise RunnerError("No active Code Engine project available")
```
ElePT, line 463: we will have to update this at some point to a more sophisticated project assignment logic, but for the time being we can keep it.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,774 @@
...
+            raise RunnerError("Job has no program assigned")
+        if self.job.program.image:
+            return self.job.program.image
+        return settings.FLEETS_DEFAULT_IMAGE
+
+    def _get_cpu_limit(self) -> str:
+        """Return the CPU limit for fleet workers.
+
+        Priority: ``compute_profile`` → ``job.config.cpu_limit`` →
+        ``settings.FLEETS_DEFAULT_CPU_LIMIT`` (default ``"1"``).
```
ElePT: let's remember to test these limits with practical examples, they might need to be refined.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -1,44 +1,774 @@
...
+            return ComputeProfileParser.get_gpu_config(self.job.compute_profile) or {}
+
+        if not self.job.gpu:
+            return {}
+
+        logger.warning(
+            "Job [%s] uses legacy gpu=True without compute_profile — using default V100 config",
+            self.job.id,
+        )
+        return {"scale_gpu": {"preferences": [{"family": "v100", "allocation": "1"}]}}
```
ElePT: the latest version of the demo doesn't use the parser because the exact profiles weren't being found. As mentioned in an earlier comment, we will revisit this in the future, but I think it's best if this iteration matches the demo source code better.

ElePT (commented):
Thanks a lot @marceloamaral!! I have done a first pass, left some actionable comments and some reminders for the future. My main comment is to review the latest status of the mvp1 demo to make sure that this PR matches it where relevant. The main point I saw that needs to be checked is the use of the compute profile. 

As discussed offline, this PR is a first iteration and we know that some points will be refactored, like the COS handling, but it should also be in a reasonable shape to merge, and I think it's close! We are only missing a couple of integration tests for the COS paths.

I leave it to you to decide to include integration tests we discussed as part of this PR or as a follow-up, but I would like you to at least be able to test the integration tests locally and verify that this code works before doing the final merge of this PR.

gateway/core/services/compute_profile_parser.py
marceloamaral: I thought I have addressed that before, sorry... So, I removed `compute_profile_parser.py` and updated the runner to match the demo. The `compute_profile` string is now passed directly to Code Engine as                        `scale_preferred_worker_profile` in `extra_fields`, and CE handles the CPU/memory/GPU mapping from the profile.

gateway/main/settings.py
marceloamaral: Updated the default image to the ICR `helloworld` image. Added a note to update it with the default custom function ICR image that we might create.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
marceloamaral: Moved all imports to the top level in both test files.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
marceloamaral: Done for all private properties in all tests!

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_job_cos_unit.py
marceloamaral: That makes sense, I removed the protected-access from both `test_job_cos_unit.py`  and `test_fleets_runner.py`.

gateway/tests/core/services/runners/test_fleets_runner.py
marceloamaral: Done!

gateway/core/services/runners/fleets_runner.py
marceloamaral: This makes a lot of sense!

We can already be defensive here and add a lightweight class-level `TTLCache` to the fleets runner for `is_active()`.

With this, cached results are returned within the TTL window, avoiding unnecessary API calls. Also, the oldest entries are evicted when the cache is full, so it stays memory safe.

gateway/core/services/runners/fleets_runner.py
marceloamaral: Good point.

While checking this, I also found that `runner.stop()` was using `delete_job()` instead of `cancel_job()`, and then update it!

gateway/core/services/runners/fleets_runner.py
marceloamaral: Agreed, the current single-project assignment is sufficient for now. Plan for multi-project we will need to review this. I added a Note for it in the code to remind us.

gateway/core/services/runners/fleets_runner.py
marceloamaral: Good point, also added a Note to review it later.

gateway/core/services/runners/fleets_runner.py
marceloamaral: I removed this for now, but will add it back once you update this in the demo!

marceloamaral:
After reviewing the PR #2061, I noticed `CodeEngineProject` creation was missing from the startup flow, we were manually creating it in the demo as a workaround.

I added a `sync_ce_project` management command that creates a `CodeEngineProject` from `CE_*` env vars. It runs after `migrate_with_lock` in both entrypoints, following Django best practices (no DB queries in `AppConfig.ready()`).

Changes:
* `gateway/main/settings.py` — added 12 `CE_*` env vars (project ID, name, region, resource group, subnet pool, PDS names, COS buckets/credentials)
* `gateway/api/management/commands/sync_ce_project.py` — new command: creates or updates the `CodeEngineProject` row when `CE_PROJECT_ID` is set, skips silently when unset, validates all required fields before writing
* `gateway/entrypoint-gateway.sh` / `entrypoint-scheduler.sh` — added `sync_ce_project` after `migrate_with_lock`
* `gateway/examples/` — updated setup script and demo to use the project created by `sync_ce_project` instead of creating it inline

Pipeline: `migrate_with_lock → sync_ce_project → gunicorn/scheduler`

marceloamaral:
> I leave it to you to decide to include integration tests we discussed as part of this PR or as a follow-up, but I would like you to at least be able to test the integration tests locally and verify that this code works before doing the final merge of this PR.

The local integration test worked well:
```
DEBUG 2026-04-29 17:04:11,982 [FleetsRunner] fleets_runner.py:(277) Fleet [1be0d933-0f7f-4235-872b-e6eac37d58fb] status: successful → SUCCEEDED
    [113s] status=SUCCEEDED

    Fleet succeeded in 113s

[4] Retrieving user logs ([public] filtered)...
INFO 2026-04-29 17:04:11,983 [FleetsRunner] fleets_runner.py:(617) Retrieving user logs for fleet [1be0d933-0f7f-4235-872b-e6eac37d58fb]: bucket=[qiskit-functions-ce-user-data-bucket-staging] key=[users/3/provider_functions/default/fleets-custom-0163a21a/jobs/af22f4ee-98b0-4044-9e8a-afee50f520b9/logs.log]
INFO 2026-04-29 17:04:12,160 [FleetsRunner] fleets_runner.py:(634) Retrieved user logs for fleet [1be0d933-0f7f-4235-872b-e6eac37d58fb]

--- user logs ---
[public] Running on backend: ibm_sherbrooke
[public] Molecule atoms: ['H', 'H']
[public] Results saved

-----------------

[5] Retrieving provider logs (all output)...
INFO 2026-04-29 17:04:12,160 [FleetsRunner] fleets_runner.py:(617) Retrieving provider logs for fleet [1be0d933-0f7f-4235-872b-e6eac37d58fb]: bucket=[qiskit-functions-ce-provider-data-bucket-staging] key=[providers/default/fleets-custom-0163a21a/jobs/af22f4ee-98b0-4044-9e8a-afee50f520b9/logs.log]
INFO 2026-04-29 17:04:12,336 [FleetsRunner] fleets_runner.py:(634) Retrieved provider logs for fleet [1be0d933-0f7f-4235-872b-e6eac37d58fb]

--- provider logs ---
[public] Running on backend: ibm_sherbrooke
[public] Molecule atoms: ['H', 'H']
Loading user file: {'description': 'sample user file'}
Processing with backend ibm_sherbrooke
[public] Results saved

--------------------

[6] Retrieving results from COS...
DEBUG 2026-04-29 17:04:12,336 [FleetsRunner] fleets_runner.py:(340) Retrieving results for job [af22f4ee-98b0-4044-9e8a-afee50f520b9] from qiskit-functions-ce-user-data-bucket-staging/users/3/provider_functions/default/fleets-custom-0163a21a/jobs/af22f4ee-98b0-4044-9e8a-afee50f520b9/results.json
INFO 2026-04-29 17:04:12,465 [FleetsRunner] fleets_runner.py:(344) Retrieved results for job [af22f4ee-98b0-4044-9e8a-afee50f520b9] (142 bytes)
    results: {
  "backend_name": "ibm_sherbrooke",
  "molecule": {
    "atoms": [
      "H",
      "H"
    ]
  },
  "user_file": {
    "description": "sample user file"
  },
  "status": "completed"
}

[7] Validating user logs...

[8] Validating provider logs...

[cleanup] Deleting fleet 1be0d933-0f7f-4235-872b-e6eac37d58fb...
```

avilches (approved):
LGTM, any minor improvements can be done in further PRs. I would wait until @ElePT finally approves it though.

ElePT (approved):
Alright, the demo introduced a lot of last-minute changes that slightly diverge from this PR, so we will have to review the profile parsing and project startup scripts. That being said, I think it will be easier if we merge this first approximation and analyze the difference in follow-ups. I'm just noting here that there will be differences.

PR #2034, opened 2026-04-23
Update TTL of running job

ItamarGoldman:
### Summary
Change TTL of a job that run on the quatum computer per the discussion.

### Details and comments
To integrate the change in the TTL the following needs to be done:

- [x] Add the check (via `JobEvent` table) for the TTL in the scheduler.
- [x] Change the default setting for `PROGRAM_TIMEOUT` to 12 hours per the discussion.
- [x] Add test, making a job canceled by exceeding the time limit.
- [x] Fix older tests.
- [x] Add release notes.
- [x] Update documentation?


Tansito:
Converting it to draft as the PR is WIP 😄

gateway/scheduler/schedule.py
```diff
@@ -100,17 +100,21 @@ def get_jobs_to_schedule_fair_share(slots: int, gpu: bool, runner: str = Program
...
-        job.logs += "\nMaximum job runtime reached. Stopping the job."
-        logger.warning(
-            "job_id=%s timeout_days=%s Job reached maximum runtime, stopping",
-            job.id,
-            timeout,
-        )
-        return True
+    if job.status in Job.RUNNING_STATUSES:
+        timeout = settings.PROGRAM_TIMEOUT
+        latest_job_event = JobEvent.objects.filter(job=job).first()
```
avilches: Order by date descending

ItamarGoldman: Doesn't the class Meta of model `JobEvent` force it to be ordered by creation time? ([source](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/models.py#L492-L494))

```python
class JobEvent(models.Model):
    """Events history for jobs."""
    ...
       
    class Meta:
        app_label = "api"
        ordering = ("-created",)
```

gateway/tests/api/test_logs.py
```diff
@@ -24,21 +25,19 @@ def create_job(author: str, provider_admin: Optional[str] = None) -> Job:
...
-        provider = Provider.objects.create(name=provider_admin)
-        admin_group, _ = Group.objects.get_or_create(name=provider_admin)
-        admin_user, _ = User.objects.get_or_create(username=provider_admin)
-        admin_user.groups.add(admin_group)
-        provider.admin_groups.add(admin_group)
-
-    program = Program.objects.create(
-        title=f"{author_user}-{provider_admin or 'custom'}",
+        # admin_group = TestUtils.create_group_with_user(provider_admin, provider_admin)
+        # provider = TestUtils.get_or_create_provider(provider_admin, admin_group)
```
avilches: You have to create the group with TestUtils.add_user_to_group...

gateway/main/settings.py
```diff
@@ -316,7 +316,7 @@
     "ibm-cloud.kubernetes.io/worker-pool-name: gpu-workers",
 )
 
-PROGRAM_TIMEOUT = int(os.environ.get("PROGRAM_TIMEOUT", "14"))
+PROGRAM_TIMEOUT = int(os.environ.get("PROGRAM_TIMEOUT", "1"))  # Time units are days
```
avilches: Are hours right?

gateway/scheduler/schedule.py
avilches: keep in mind that latest_job_event could be None (boundary case, but never assume the data in the database is correct)

gateway/main/settings.py
ItamarGoldman: From what I saw in the internal repo (Issue 1395) it is days (that is why I added a comment there)

gateway/scheduler/schedule.py
avilches: > Doesn't the class Meta of model `JobEvent` force it to be ordered by creation time? ([source](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/models.py#L492-L494))
> 
> class JobEvent(models.Model):
>     """Events history for jobs."""
>     ...
>        
>     class Meta:
>         app_label = "api"
>         ordering = ("-created",)

You are right, anyway I would add it either way, just in case somebody change this in the future. Being explicit instead of relying on defaults. But it's ok if you don't.

gateway/tests/api/test_logs.py
ItamarGoldman: True. I duplicated the method that was my bad. This is works but I don't understand why doesn't it work with `TestUtils.get_or_create_provider...` since it creates a provider with an admin group.

gateway/scheduler/schedule.py
ItamarGoldman: I will do it as a safe fall back as you say, I just didn't want the database to be sorted again unnecessarily but it is better to be safe than sorry.
In addition, I will check if `latest_job_event` is `None`. Should I raise an error for this case? (since this can't happen as when job is created also a `JobEvent` associate with its creation is added)

ItamarGoldman: I changed the code as suggested but what should I do in case `latest_job_event` is `None`?
```python
latest_job_event = JobEvent.objects.filter(job=job).order_by('-created').first()
        if not latest_job_event:
            # Raising an error?
            pass
```

gateway/tests/api/test_logs.py
ItamarGoldman: Everything works now :)

gateway/main/settings.py
ItamarGoldman: [Here](https://github.com/Qiskit/qiskit-serverless/blob/main/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml#L410-L411) is where we can see this is days. Should we change it to hours?

avilches: You are right, but the code is using this value as hours in the `check_job_timeout` with `atest_job_event.created + timedelta(hours=timeout)`. Could you rename the `programTimeoutDays` in yaml to `programTimeoutHours`?

gateway/scheduler/schedule.py
```diff
@@ -100,17 +100,24 @@ def get_jobs_to_schedule_fair_share(slots: int, gpu: bool, runner: str = Program
...
+    if job.status in Job.RUNNING_STATUSES:
+        timeout = settings.PROGRAM_TIMEOUT
+        latest_job_event = JobEvent.objects.filter(job=job).order_by("-created").first()
+        if not latest_job_event:
+            # This should never happen since every job should have at least 1 JobEvent (on creation).
+            pass
+        endtime = latest_job_event.created + timedelta(hours=timeout)
+        now = datetime.now(tz=endtime.tzinfo)
+        if endtime < now:
+            job.logs += "\nMaximum job runtime reached. Stopping the job."
```
avilches:
```suggestion
```
No need to write anything in the logs field, this column will be disappear soon...

gateway/scheduler/schedule.py
```diff
@@ -100,17 +100,24 @@ def get_jobs_to_schedule_fair_share(slots: int, gpu: bool, runner: str = Program
...
-            job.id,
-            timeout,
-        )
-        return True
+    if job.status in Job.RUNNING_STATUSES:
+        timeout = settings.PROGRAM_TIMEOUT
+        latest_job_event = JobEvent.objects.filter(job=job).order_by("-created").first()
+        if not latest_job_event:
+            # This should never happen since every job should have at least 1 JobEvent (on creation).
+            pass
```
avilches: Even the `pass` is correct (because it's a boundary case), you could use the `job.created` date...

gateway/scheduler/schedule.py
```diff
@@ -100,17 +100,24 @@ def get_jobs_to_schedule_fair_share(slots: int, gpu: bool, runner: str = Program
...
+        endtime = latest_job_event.created + timedelta(hours=timeout)
+        now = datetime.now(tz=endtime.tzinfo)
+        if endtime < now:
+            job.logs += "\nMaximum job runtime reached. Stopping the job."
+            logger.warning(
+                "Job [%s] reached maximum runtime [%s] days and stopped.",
+                job.id,
+                timeout,
+            )
```
avilches, line 118:
```suggestion
            logger.warning(
                "job_id=%s cluster=%s timeout=%s hours: job stopped",
                job.id,
                timeout,
            )
```

gateway/main/settings.py
ItamarGoldman: Changed as suggested.

gateway/scheduler/schedule.py
ItamarGoldman: Changed as suggested

gateway/scheduler/schedule.py
ItamarGoldman: This is a good idea as it at least takes into consideration the most primitive time info we have.
Changed as suggested

gateway/scheduler/schedule.py
ItamarGoldman: Changed as suggested but kept the dot (`.`) at the end of the sentence.

gateway/scheduler/schedule.py
ItamarGoldman: Taking `job.created` as fallback.

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -410,7 +410,7 @@ spec:
             - name: OTEL_EXPORTER_OTLP_TRACES_INSECURE
               value: {{ .Values.application.ray.openTelemetryCollector.insecure | quote }}
             - name: PROGRAM_TIMEOUT
-              value: {{ .Values.application.limits.programTimeoutDays | quote }}
+              value: {{ .Values.application.limits.TimeoutHours | quote }}
```
avilches: typo: `programTimeoutHours`.

avilches:
```
__________ TestProgramApi.test_upload_without_runner_defaults_to_ray ___________
tests/api/test_v1_program.py:1006: in test_upload_without_runner_defaults_to_ray
    TestUtils.authorize_client(username="test_user_2", client=self.client)
E   TypeError: TestUtils.authorize_client() got an unexpected keyword argument 'username'
``` 

It looks like there are still one place where the username was not renamed to user...

gateway/tests/utils.py
```diff
@@ -13,20 +13,244 @@
...
 
 class TestUtils:
-    """Utility class to provide helper methods for gateway testing."""
+    # pylint: disable=line-too-long
+    """Utility class to provide helper methods for gateway testing.
+
+    This class provides static methods to simplify test setup by creating and managing
+    Django models (User, Group, Permission, Provider, Program, Job, JobConfig, JobEvent).
+
+    ## Public Methods Summary (With Minimal Input)
```
ElePT: Adding docstrings is always helpful, but we shouldn't overpower the class with these kind of summaries. if you want to list the methods, you should follow the formatting used in qiskit (It's called google formatting: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html). We only should allow to do "# pylint: disable=line-too-long" if it's really really necessary, we shouldn't systematically disable pylint warnings.

gateway/tests/utils.py
```diff
@@ -13,20 +13,244 @@
...
+        program_title="My Program",
+        author="testuser",
+        provider="ibm",
+        instances=["premium_users"]
+    )
+
+    # Authenticate test client
+    user = TestUtils.authorize_client(user="testuser", client=self.client)
+    ```
+    """
```
ElePT: No need for an example for a testing util only we use. This is fine for user-facing docstrinngs.

gateway/tests/utils.py
```diff
@@ -13,20 +13,244 @@
...
     ) -> tuple[User, str]:
-        """Helper to normalize author input into (User object, username string)."""
+        """Helper to normalize author input into (User object, username string).
+
+        - **Minimal creates**: User (if author is str)
+        - **Optional creates**: None
+        - Behavior:
+          * If `author` is User: returns (author, author.username)
+          * If `author` is str: gets or creates User with username=author
+        """
```
ElePT: Again, method docstrings should follow the google formatting. This applies to the whole class.

ElePT (commented):
I have some last-minute docstring comments I don't want to block the PR merge for, but that should be addressed at least in a follow-up.

gateway/tests/utils.py
ItamarGoldman: Changed as suggested.

gateway/tests/utils.py
ItamarGoldman: Changed as suggested.

gateway/tests/utils.py
ItamarGoldman: Changed as suggested.

ElePT (approved):
Thanks a lot!! The changes look very good.

PR #2035, opened 2026-04-23
Add unique index over program with provider and author

avilches:
## Context

- There are no database-level protection against creating two functions with the same name under the same provider or the same author in the database.
- The application layer handled this with "upsert" logic in the upload view, but nothing stopped duplicates at the DB level.

## Summary

Add unique constraints on the `api_program` table to prevent duplicate function names.

## Details

Why two indexes instead of one?

A simple `UNIQUE(provider_id, title)` does not work because `provider_id` is nullable. In PostgreSQL, `NULL != NULL`, so the constraint would allow unlimited duplicate rows with `(NULL, "same-title")`. Programs without a provider are private (user-owned), and their uniqueness is per author.

So we need two partial indexes:

1. **Provider functions**: unique `(provider_id, title)` where provider is not null
2. **Private functions**: unique `(author_id, title)` where provider is null

## Resulting SQL in PostgreSQL

```sql
CREATE UNIQUE INDEX unique_provider_title
    ON api_program (provider_id, title)
    WHERE provider_id IS NOT NULL;

CREATE UNIQUE INDEX unique_author_title_no_provider
    ON api_program (author_id, title)
    WHERE provider_id IS NULL;
```


gateway/api/migrations/0050_unique_program_constraints.py
```diff
@@ -0,0 +1,44 @@
...
+            field=models.CharField(
+                choices=[
+                    ("TRIAL", "Trial"),
+                    ("SUBSIDIZED", "Subsidized"),
+                    ("CONSUMPTION", "Consumption"),
+                ],
+                default="SUBSIDIZED",
+                max_length=50,
+            ),
+        ),
```
avilches, line 27: This migration is a no-op for the database... It only updates the internal Django model's metadata. We could remove it... but it will appear again in the next makemigrations...

gateway/api/serializers.py
```diff
@@ -46,6 +47,12 @@ class UploadProgramSerializer(serializers.ModelSerializer):
     class Meta:
         model = Program
 
+    def get_validators(self):
+        """Exclude UniqueConstraint validators.
+        Uniqueness is enforced at DB level; the upload view handles
+        upsert logic (find-or-create) before saving."""
+        return [v for v in super().get_validators() if not isinstance(v, validators_module.UniqueTogetherValidator)]
+
```
avilches, line 55: Django generates a `UniqueTogetherValidator` when it sees `UniqueConstraint` in the model, which is ok: a validation on the app and a unique index in the db. So far so good.
But Django doesn't understand partial unique indexes and it treats all constraint fields as required, which makes the provider mandatory in the request breaking the existing api where provider is optional. It also rejects updates to existing programs because it sees the combination as a duplicate, even though the upload view intentionally does upsert (find-or-create). For this reason, it's totally ok remove them.

avilches:
To check if the database is consistent before merging this and deploy it, I executed these 2 queries in staging and production, and both of them returned 0 elements. So the PR with the indexes are safe to deploy

```sql
SELECT provider_id, title, COUNT(*) FROM api_program WHERE provider_id 
IS NOT NULL GROUP BY provider_id, title HAVING COUNT(*) > 1;

SELECT author_id, title, COUNT(*) FROM api_program WHERE provider_id 
IS NULL GROUP BY author_id, title HAVING COUNT(*) > 1;  
```

riley2012:
> This could replace the `platform_id` approach from the https://github.com/Qiskit/qiskit-serverless/pull/2029, which added a derived field (`provider_name.title`) with a unique constraint. That approach required:
> - A new `platform_id` column on every program row
> - Overriding `save()` to compute `platform_id` on every write
> - A data migration to backfill existing rows
> 
> Instead, we can achieve the same goal with two partial unique indexes directly on the existing columns. No new fields needed

@avilches can this Appendix be removed from the Description?
Based on today's discussions, the team concluded that the constraints in this PR and the `platform_id` serve different purposes and both are valid. 
- unique constraints (this PR) prevent two functions from having the same name under the same provider or author. 
- `platform_id`(PR #2029)  is a stable, human-readable unique identifier shared across services to look up function permissions.

riley2012 (approved):
LGTM! 
If #2029 merges first, will the migration file in this PR need to be regenerated?

avilches:
> LGTM! If #2029 merges first, will the migration file in this PR need to be regenerated?

Yes! this one is about to be merged first, so I will have to change the #2029

avilches:
> @avilches can this Appendix be removed from the Description? Based on today's discussions, the team concluded that the constraints in this PR and the `platform_id` serve different purposes and both are valid.

Sure, deleted!

Issue #1380, opened 2024-06-18
Entrypoint is not being correctly checked


Tansito:
### What is the current behavior?

The logic that checks if the entrypoint exists is not working properly:
https://github.com/Qiskit/qiskit-serverless/blob/main/client/qiskit_serverless/core/job.py#L787

The next qiskit-function can be uploaded:
```py
QiskitFunction(title="my-title", working_dir="./", entrypoint="/User/me/source/my_code.py")
```

The reason because it pass the validation is because the file exists in the workind_dir, but that path is not going to exist when we submit the job to run it.

### What is the expected behavior?

We are going to need to validate if the path is absolute or maybe even only accept filenames.

PR #2038, opened 2026-04-24
Integrate telemetry in the API and configure it locally

ismaelRozasRamallal:
### Summary

We added end-to-end distributed tracing to Qiskit Serverless using OpenTelemetry. Every HTTP request hitting the gateway now produces a trace in Jaeger, broken down into child spans for each database query executed during that request. Traces carry meaningful attributes — the HTTP method, route, response status, SQL statement, database name, and user — giving a complete picture of what happens inside the service for any given request. Locally, the observability stack runs as a Docker Compose `full` profile and includes Jaeger all-in-one (trace storage and UI on port 16686, OTLP gRPC collector on port 4317), Prometheus (metrics), Loki and Promtail (log aggregation), and Grafana (unified dashboards on port 3000).

### Details and comments

We enabled two OTel auto-instrumentation libraries: `DjangoInstrumentor` for HTTP request spans and `Psycopg2Instrumentor` for PostgreSQL query spans. Both are zero-code integrations — they monkey-patch Django's request handling and psycopg2's `connect` respectively, so no changes to individual views or queries were needed. They are activated together in a centralised `setup_gateway_tracing()` function in `gateway/main/tracing.py`, which initialises an OTLP gRPC exporter pointed at the collector endpoint and registers a global `TracerProvider` with the service name. On top of auto-instrumentation, key business logic paths (such as program listing and job execution) are wrapped with custom named spans using a `@trace` decorator, providing a semantic layer above the raw HTTP and SQL spans. We also standardised environment variable naming from legacy `OT_JAEGER_*` prefixes to the `OTEL_*` standard, and consolidated duplicate OTLP setup that was previously spread across `manage.py` and a view file. For local development, a Jaeger all-in-one instance with its embedded OTLP collector is available via `docker compose --profile full up`, matching the same OTLP protocol used in production

The original setup call lived in `manage.py`, which only runs in Django's StatReloader parent process. The child process that actually handles requests never called it, so the psycopg2 wrapper was never active there and DB span attributes were empty. Moving the call to `ApiConfig.ready()` ensures it executes in every Django process — parent and child alike — so instrumentation is always in place when requests arrive.

🔧 Related #1380 

In future work, we will be setting up the environment for staging to test if everything connects as expected


docker-compose-dev.yaml
```diff
@@ -10,6 +10,12 @@ services:
       "env", "RAY_LOG_TO_STDERR=1", "ray", "start", "--head", "--port=6379",
       "--dashboard-host=0.0.0.0", "--block"
     ]
+    environment:
+      - OTEL_ENABLED=1
+      - OTEL_EXPORTER_OTLP_HOST=http://jaeger
+      - OTEL_EXPORTER_OTLP_PORT=4317
+      - OTEL_EXPORTER_OTLP_INSECURE=1
+      - QS_OT_RAY_TRACER=1
```
ElePT, line 18: I'm curious, why didn't you apply the changes to `docker-compose.yaml` as well? Asking because the two often get out of sync and I prefer to keep the parity. Even if the env variables can't get used because they require a deployment, we may forget in the future.

ElePT (dismissed):
LGTM, I just left a very minor comment.

docker-compose-dev.yaml
ismaelRozasRamallal: Good point, in fact I think I don't fully understand why we have `docker-compose.yaml` and `docker-compose-dev.yaml` when the file nature is purely development/testing 😄 which is the main difference between them?

ElePT: @ismaelRozasRamallal  `docker-compose.yaml` provides a local path to run the deployed ray node, gateway and scheduler. `docker-compose-dev.yaml` builds them from source. These files precede me but from what I know, while `docker-compose.yaml` is not really useful for our development of the serverless codebase, it is useful for local function source code testing, and I believe it is used by many function developers in their CI pipelines. It's a way they can "deploy" their function locally.

ismaelRozasRamallal: okokkk added then to keep changes consistent!

ElePT (approved):
LGTM

PR #2039, opened 2026-04-24
Add FunctionAccessClient and PLATFORM_ACTION_* constants

avilches:
## Summary
First step of the instance-based authorization. This PR doesn't change existing code, it only add classes.

## Details and comments

- Add `PLATFORM_PERMISSION_*` constants in `core/models.py`.
- `FunctionAccessClient` it's the client to connect to IBM Runtime. Not implemented but we can use in tests mocking the methods. Returns `FunctionAccessResult` which has a boolean flag has_response and a list of `FunctionAccessResult` with the function id, the array of actions and the business_model.
It exposes `get_function`, `has_action_for_provider`, `get_functions_by_provider` methods that will be needed during the auth.

### Why the has_response flag?
We need a way to distinguish between having access to 0 functions and a not configured crn. When `has_response=False` callers fall back to the existing Django groups system unchanged.

### DIFF
--- a/gateway/api/clients/function_access_client.py
+++ b/gateway/api/clients/function_access_client.py
@@ -4,11 +4,7 @@ from api.domain.authorization.function_access_result import FunctionAccessResult
 
 
 class FunctionAccessClient:
-    """Client for retrieving accessible functions for a given instance CRN.
-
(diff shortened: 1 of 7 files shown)


Tansito (approved):
LGTM!

PR #2040, opened 2026-04-26
Access policies and model manager uses new`FunctionAccessClient`

avilches:
## Summary

Prepares the access layer to support external authorization via `FunctionAccessResult`. It keeps backwards compatibility with the existing Django ACLs.

- The `FunctionAccessResult` and the client `FunctionAccessClient` are created in : #2039

## Details

Today, all access decisions (who can run a function, who can see provider jobs, who can manage provider files) are made purely via Django groups. This PR wires in a second path: when a `FunctionAccessResult` is available (coming from an external instance-level client introduced in #2039 , the access checks use it instead of hitting the database for group membership.

Both paths coexist through a fallback design: if no external result is provided, every check behaves exactly as before. This means this PR can be merged  without breaking anything becasue the callers that will actually supply the external result come in subsequent PRs.

The `FunctionAccessEntry` and `FunctionAccessResult` domain objects are moved from `api/` to `core/` so they can be used by the model manager layer without violating the import contract between packages.


avilches:
Wrong branch. Crated #2041 instead

PR #2041, opened 2026-04-26
Access policies using FunctionAccessResult and access manager

avilches:
## Summary

Prepares the access layer to support external authorization via `FunctionAccessResult`. It keeps backwards compatibility with the existing Django ACLs.

- The `FunctionAccessResult` and the client `FunctionAccessClient` are created in : #2039

## Details and comments

Today, all access decisions (who can run a function, who can see provider jobs, who can manage provider files) are made purely via Django ACLS. This PR creates a second path only if `FunctionAccessResult` is available (coming from an external instance-level client introduced in #2039), the access checks use it instead of the legacy database  group. Also `JobAccessPolicies` and `ProviderAccessPolicy` are fully implemented + tests, but the new implementation is dead code because the calls don't receive the `FunctionAccessResult`, which is obtained by calling the Runtime API and whose data is needed to grant access.

Both paths coexist through a fallback design: if no external result is provided, every check behaves exactly as before.


gateway/api/access_policies/providers.py
```diff
@@ -3,40 +3,125 @@
...
+    """Core provider access logic shared by all named methods.
+
+    When accessible_functions.has_response=True: checks the specific function entry (granular).
+    Otherwise falls back to Django admin_groups.
+    """
+    if accessible_functions is not None and accessible_functions.has_response:
+        entry = accessible_functions.get_function(provider.name, function_title)
+        return entry is not None and permission in entry.permissions
+    user_groups = set(user.groups.all())
+    return bool(user_groups.intersection(set(provider.admin_groups.all())))
```
avilches, line 38:
```suggestion
    # Runtime instances API has function granularity: user needs to have permission per function
    if accessible_functions is not None and accessible_functions.has_response:
        entry = accessible_functions.get_function(provider.name, function_title)
        return entry is not None and permission in entry.permissions
    # Legacy Django auth has provider granularity: user needs to be a provider admin    
    user_groups = set(user.groups.all())
    return bool(user_groups.intersection(set(provider.admin_groups.all())))
```

gateway/api/access_policies/providers.py
```diff
@@ -3,40 +3,125 @@
...
+        return has_access
 
+    @staticmethod
+    def can_manage_files(
+        user,
+        provider: Provider,
+        function_title: str,
+        accessible_functions: Optional[FunctionAccessResult] = None,
+    ) -> bool:
+        """Checks if the user can manage files for this provider."""
```
avilches:
```suggestion
        """Checks if the user can manage provider files for this provider."""
```

gateway/api/v1/views/jobs/provider_list.py
```diff
@@ -190,7 +192,12 @@ def get_provider_jobs(request: Request) -> Response:
     filters = JobFilters(**serializer.validated_data)
     user = cast(AbstractUser, request.user)
 
-    jobs, total = JobsProviderListUseCase().execute(user=user, filters=filters)
+    crn = getattr(request.auth, "instance", None)
+    accessible = (
+        FunctionAccessClient().get_accessible_functions(crn) if crn else FunctionAccessResult(has_response=False)
```
avilches: This is actually the only places where the new auth branch with Runtime API instances is already called.

avilches:
Created smaller PRs based on this, to make this PR easier to review.
- https://github.com/Qiskit/qiskit-serverless/pull/2058
- https://github.com/Qiskit/qiskit-serverless/pull/2059
- https://github.com/Qiskit/qiskit-serverless/pull/2060

Tansito (approved):
Good work here @avilches thank you!

PR #2058, opened 2026-04-29
Rename permission_name to legacy_permission_name in FunctionsQuerySet

avilches:
## Summary

Renames the `permission_name` parameter to `legacy_permission_name` in `FunctionsQuerySet.with_permission()` and `get_function_by_permission()`.

All callers in `api/views/programs.py` and `api/use_cases/files/` are updated accordingly.

### Why?

This clarifies that the parameter refers to a Django permission codename (legacy auth path), preparing for the upcoming Runtime Instances API integration which uses a different permission model and need a `permission` parameter in the same call.

### DIFF
--- a/gateway/api/use_cases/files/delete.py
+++ b/gateway/api/use_cases/files/delete.py
@@ -34,7 +34,7 @@ class FilesDeleteUseCase:
         """
         function = Function.objects.get_function_by_permission(
             user=user,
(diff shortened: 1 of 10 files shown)


ismaelRozasRamallal (approved):

Tansito (approved):
LGTM!

PR #2059, opened 2026-04-29
Move authorization domain from api to core

avilches:
## Summary

Moves `function_access_entry.py` and `function_access_result.py` from `api/domain/authorization/` to `core/domain/authorization/`. These domain classes are used by core model managers and don't depend on the API layer, so they belong in `core`.

All imports are updated accordingly. No logic changes.


ismaelRozasRamallal (approved):

Tansito:
After commenting with @avilches internally we considered that it doesn't have sense to move this logic outside the API scope. Authorization and authentication is very API-aligned logic and scheduler could even have another logic so that's the reason we considered to close this one and maintain the logic in its current domain.

PR #2061, opened 2026-04-29
Replace cos vol with client

korgan00:


gateway/core/services/storage/cos_client.py
```diff
@@ -0,0 +1,47 @@
+"""
```
marceloamaral, line 1: @korgan00 have you seen the cos clients in https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/ibm_cloud/cos/cos_client.py?

marceloamaral: The Fleets API also uses the cos clients here: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/ibm_cloud/code_engine/fleets/cos.py

marceloamaral: It will be great if we can align the implementations.

gateway/core/services/storage/fleets_cos_client.py
```diff
@@ -0,0 +1,114 @@
+"""COS client for Fleets-based jobs."""
```
marceloamaral, line 1: Again, we already have the client here: https://github.com/Qiskit/qiskit-serverless/tree/main/gateway/core/ibm_cloud/cos

I believe that having the a dedicated package for `ibm_cloud` sdks helps to make the code easier to follow later on.

gateway/core/services/storage/cos_client.py
korgan00: ok, i was copying the example that we had in other branch, i will work on alignment

gateway/core/services/storage/ray_cos_client.py
```diff
@@ -0,0 +1,86 @@
+"""COS client for Ray-based jobs."""
```
marceloamaral, line 1: Does it really need a separate COS client for Ray and Fleets?

My understanding was that the main difference is just how the COS client is initialized, for example, with the right HMAC secret, COS instance details, and related config. After that, the runners would mainly differ in the paths they use, meaning the COS object keys.

korgan00: I think is better to have them separated, less branching and easier to delete code when is not needed and also easier to add new COS in a future

gateway/main/settings.py
```diff
@@ -368,6 +365,22 @@
 # Functions logs size limite in Bytes
 FUNCTIONS_LOGS_SIZE_LIMIT = int(os.environ.get("FUNCTIONS_LOGS_SIZE_LIMIT", "52428800"))
 
+# COS
+RAY_COS_ENDPOINT = os.environ.get("COS_ENDPOINT", None)
+RAY_COS_BUCKET = os.environ.get("COS_BUCKET", None)
+RAY_COS_ACCESS_KEY = os.environ.get("COS_ACCESS_KEY", None)
+RAY_COS_SECRET_KEY = os.environ.get("COS_SECRET_KEY", None)
+
+FLEETS_USERS_COS_ENDPOINT = os.environ.get("FLEETS_USERS_COS_ENDPOINT", None)
```
marceloamaral, line 399: We also have few definitions for FLEETS variables that will be merged in the following PR:
https://github.com/Qiskit/qiskit-serverless/pull/2033/changes#diff-d3fab2ab912c6eb5c879651a1476f64f0551b747bf618b4ab13ddb4602dd45c1R299

gateway/core/services/storage/fleets_cos_client.py
marceloamaral: If we have this file here, we probably will not need the file in https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/ibm_cloud/code_engine/fleets/cos.py

gateway/core/services/storage/cos_client.py
marceloamaral: Also, feel free to refactor function names as needed

Tansito:
From internal conversations we are going to try to divide in smaller PRs this PR to try to merge some concepts.

Tansito:
I think we can just remove this one as it was added in separated PRs from all storages.

PR #2062, opened 2026-04-29
Make program title read-only on edit in admin

avilches:
## Summary

Prevents editing the `title` field of a Program from the Django admin once it has been created. The field remains editable only during creation.


riley2012 (approved):
LGTM!

Tansito:
Django never dissapoints 😂

PR #2063, opened 2026-04-29
Enable Instances auth: jobs/provider-list endpoint

avilches:
## Summary

- `FunctionAccessClient` now is used during the authorization. The functions with permission are stored in `CustomTokenBackend`, so any view could access to it.
- In this PR, the `jobs/provider-list` endpoint uses these permissions to gran or revoke access. This endpoint returns all jobs executed on the provider's functions, from any user. The caller must be an admin of the provider (or hold PLATFORM_PERMISSION_PROVIDER_JOBS via the runtime instances API.

## Details and comments

Inside the use case, `_apply_access_scope` implements the dual logic:
- If `accessible_functions.has_response=True`: access is granted per-function using `get_functions_by_provider(PLATFORM_PERMISSION_PROVIDER_JOBS)`, and the resulting set is used to filter the job queryset. No Django group check is performed.
- If `has_response=False` (no CRN, client down, feature flag off): falls back to `is_provider_admin` exactly as before.

`FunctionsQuerySet.with_permission` and `get_function_by_permission` are also updated with a `filter_function_names` parameter that supports the same dual-path pattern, used by the function listing endpoints (coming in later PRs).

### DIFF
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -52,6 +52,12 @@ class ProgramAdmin(admin.ModelAdmin):
         "disabled",
     ]
 
+    def get_readonly_fields(self, request, obj=None):
+        readonly_fields = list(super().get_readonly_fields(request, obj))
+        if obj:
+            readonly_fields.append("title")
+        return readonly_fields
+
     def get_urls(self):
         """Add program history url to the available urls."""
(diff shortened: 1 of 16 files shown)


avilches:
Wrong branch...

PR #2064, opened 2026-04-30
Enable Instances auth: jobs/provider-list endpoint

avilches:
## Summary

- `FunctionAccessClient` now is used during the authorization. The functions with permission are stored in `CustomTokenBackend`, so any view could access to it.
- In this PR, the `jobs/provider-list` endpoint uses these permissions to gran or revoke access. This endpoint returns all jobs executed on the provider's functions, from any user. The caller must be 1) an admin of the provider for legacy Django groups database or 2) hold PLATFORM_PERMISSION_PROVIDER_JOBS via the runtime instances API.

## Details and comments

Inside the use case, `_apply_access_scope` implements the dual logic:
- If `accessible_functions.has_response=True`: access is granted per-function using `get_functions_by_provider(PLATFORM_PERMISSION_PROVIDER_JOBS)`, and the resulting set is used to filter the job queryset. No Django group check is performed.
- If `has_response=False` (no CRN, client down, feature flag off): falls back to `is_provider_admin` exactly as before.

`FunctionsQuerySet.with_permission` and `get_function_by_permission` are also updated with a `filter_function_names` parameter that supports the same dual-path pattern, used by the function listing endpoints (coming in later PRs).


gateway/api/domain/authorization/function_access_result.py
```diff
@@ -37,3 +38,7 @@ def get_functions_by_provider(self, permission: str) -> Dict[str, Set[str]]:
             if permission in e.permissions:
                 by_provider[e.provider_name].add(e.function_title)
         return dict(by_provider)
+
+    def __str__(self) -> str:
+        functions_str = ", ".join(f"{e.provider_name}.{e.function_title}" for e in self.functions)
+        return f"has_response={self.has_response}, message={self.message!r}, functions=[{functions_str}]"
```
avilches, line 44: I added a message because every endpoint will print this information in the logs on the same log line along with the job_id and the action. I think this is better than printing the line alone in the auth method (it won't have the job_id)

gateway/api/use_cases/jobs/provider_list.py
```diff
@@ -33,17 +35,38 @@ def execute(
...
-            function = Function.objects.get_function(
-                function_title=filters.function,
-                provider_name=filters.provider,
-            )
-
-            if not function:
-                raise FunctionNotFoundException(function=filters.function, provider=filters.provider)
+        function_titles = self._apply_access_scope(user, provider, filters, accessible_functions)
+        if function_titles is not None:
+            filters.functions = function_titles
```
avilches, line 43: The original code allows filtering by function. Now, it also allows this, but it will also filter by all the functions the user has access to (which wasn't possible before)

Tansito (approved):
Reviewed, applied initial feedback and ready to merge! 🚀

PR #2068, opened 2026-05-02
Reduce redundant run() calls in integration tests

avilches:
Some integration tests were running duplicate jobs unnecessarily, adding execution time without additional coverage:

- `test_event_send_error` and `test_event_wrong_type` both uploaded the same function and called `run()` separately. Merged into a single `test_events` that reuses the same job for both assertions.
- `test_provider_logs` verified that `provider_logs()` raises a 403 error, but `test_logs` already covers this exact assertion. Removed as redundant.
- `test_dependencies.py` removed (was already staged).

Reduces active `run()` calls in `test_jobs.py` from 12 to 10.


ElePT (approved):
Amazing. LGTM.

PR #2069, opened 2026-05-04
Support COS private endpoint configuration for FleetRunner

ElePT:
### Summary
During the demo deployment I found that COS client was set up to go over the public internet, but our deployment is configured to stay within the VPC and use private endpoints. This PR adds a `CE_COS_ENDPOINT_URL` env var that, when set, overrides the default region-derived endpoint, so we can point it at the private endpoint and keep traffic internal.

The URL is threaded through the whole stack: settings → FleetsRunner → JobCOS → COSClient → ibm_boto3.

The client cache key was also updated from `(region, access_key_id)` to `(endpoint_url, access_key_id)` so public and private clients don't accidentally share a cached instance.

### DIFF
--- a/gateway/core/ibm_cloud/clients.py
+++ b/gateway/core/ibm_cloud/clients.py
@@ -211,34 +211,38 @@ class IBMCloudClientProvider:
         access_key_id: str,
         secret_access_key: str,
         bucket_region: str | None = None,
+        endpoint_url: str | None = None,
     ) -> Any:
         """
         Return an S3-compatible IBM COS client using HMAC keys.
 
         Use for object operations: uploads, downloads, streaming.
-        Cached by ``(bucket_region, access_key_id)``.
+        Cached by ``(endpoint_url, access_key_id)``.
 
         Args:
             access_key_id: HMAC access key id from a COS service key.
             secret_access_key: HMAC secret access key from a COS service key.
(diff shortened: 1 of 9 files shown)


marceloamaral (commented):
Good catch!

But since we will always use the private endpoint why not change the `COS_URL_TEMPLATE` directly?

So, instead of using:
`COS_URL_TEMPLATE = "https://s3.{region}.cloud-object-storage.appdomain.cloud"`

We can use:
`COS_URL_TEMPLATE = "https://s3.private.{region}.cloud-object-storage.appdomain.cloud"`

ElePT:
> Good catch!
> 
> But since we will always use the private endpoint why not change the `COS_URL_TEMPLATE` directly?
> 
> So, instead of using: `COS_URL_TEMPLATE = "https://s3.{region}.cloud-object-storage.appdomain.cloud"`
> 
> We can use: `COS_URL_TEMPLATE = "https://s3.private.{region}.cloud-object-storage.appdomain.cloud"`

I didn't go that route to allow the COS client to be tested and potentially used from outside of the staging cluster, for example, with a local deployment. What we could do is invert the order of priorities and have the "main" path to follow the private url, and have a secondary path for local testing, what do you think?

marceloamaral (dismissed):
/LGTM

isma-roza (approved):

ItamarGoldman (approved):

PR #2070, opened 2026-05-04
Small fix: handle US spelling of canceled in fleet status mapping

ElePT:
### Summary
CE can return either "cancelled" (UK) or "canceled" (US) depending on the API version, we were only handling the UK spelling, so US-spelled statuses were silently falling through to PENDING instead of STOPPED.

One line fix + test case added to the parametrized status mapping suite.


marceloamaral (approved):
LGTM

PR #2072, opened 2026-05-04
Expose fleet_id on Job and allow status polling via get_job_data

ElePT:
### Summary
Depends on: #2071 (fleet_id in job API response)

Two small additions to the Job client to support Fleets jobs:

- `Job.fleet_id` — property that returns the Code Engine fleet ID for
  the job, or None for Ray jobs
- `JobService.get_job_data()` — optional hook (default: None) that
  subclasses can implement to return the full job dict in one call
- `ServerlessClient.get_job_data()` — implementation that fetches
  `/jobs/{id}/` with `with_result=false` to avoid pulling large results
- `Job.status()` — now calls `get_job_data()` first; if it returns a
  dict it uses that for the status (and caches the rest of the fields
  including `fleet_id`), otherwise falls back to the dedicated status
  endpoint as before

The net effect is that after any `status()` call, `job.fleet_id` is
populated without a second API request.

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -350,6 +350,19 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
 
         return Job(job_id, job_service=self, compute_profile=response_data.get("compute_profile"))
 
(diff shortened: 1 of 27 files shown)


avilches (approved):

avilches:
I will wait until #2071 is merged, update the branch and merge this one

PR #2073, opened 2026-05-04
Enable Instances auth: programs endpoints

avilches:
## Summary

- The five `programs` endpoints (`list`, `upload`, `run`, `get_by_title`, `get_jobs`) now use the runtime instances API for authorization, following the same dual-path pattern established in PR #2064 for `jobs/provider-list`.
- `RunJobSerializer.create()` accepts an external `business_model` from the access entry, so jobs created via the runtime instances path use the correct billing model instead of computing it from Django trial groups.

## Details and comments

Each endpoint resolves access through the same dual logic:

- **`list()`** — splits on `use_legacy_authorization` first, then on `filter`:
  - Legacy: uses Django groups (`RUN_PROGRAM_PERMISSION` / `VIEW_PROGRAM_PERMISSION`).
  - Runtime: calls `accessible_functions.get_functions_by_provider(PLATFORM_PERMISSION_RUN/READ)` to build the queryset filter.
- **`upload()`** — delegates to `ProviderAccessPolicy.can_upload_function`, which already handles the dual path (added in PR2). The view now passes `accessible_functions`.
- **`run()`** — logic extracted into `_resolve_function(author, title, provider, accessible_functions) → (function, business_model)`:
  - Legacy: `get_function_by_permission` with Django groups fallback.
  - Runtime: loads the function from DB first, then checks `has_permission_for_function(PLATFORM_PERMISSION_RUN)`. If the entry carries a `business_model`, it is forwarded to the serializer and stored on the job directly, bypassing `is_trial()`.
- **`get_by_title()`** — passes `filter_function_names` (from `get_functions_by_provider(PLATFORM_PERMISSION_READ)`) to `get_function_by_permission`.
- **`get_jobs()`** — passes `accessible_functions` to `ProviderAccessPolicy.can_list_jobs`.


Tansito:
After almost 3-4 hours of review the PR is almost approved! 😂

Tansito (approved):
🚀 we finally got this! Really good work @avilches

PR #2074, opened 2026-05-04
Pass job env vars to fleet containers

marceloamaral:
### Summary
`FleetsRunner.submit()` now forwards decrypted `job.env_vars` to the CE container, matching the Ray runner path.

### Details and comments
- Decrypts and passes all env vars from `job.env_vars` (gateway token, Qiskit Runtime credentials, trial mode, etc.)
- Adds `QISKIT_IBM_RUNTIME_CUSTOM_CLIENT_APP_HEADER` for IBM Runtime attribution
- Removes redundant explicit `DATA_PATH` (already in `job.env_vars`)
- Enables `save_result()` calls from within fleet containers via the gateway API

### DIFF
--- a/gateway/core/services/runners/fleets_runner.py
+++ b/gateway/core/services/runners/fleets_runner.py
@@ -27,6 +27,7 @@ from core.ibm_cloud.code_engine.ce_client.rest import ApiException
 
 from core.models import Job, CodeEngineProject
 from core.services.runners.abstract_runner import AbstractRunner, RunnerError
+from core.utils import decrypt_env_vars
 from core.ibm_cloud.clients import IBMCloudClientProvider
 from core.ibm_cloud.code_engine.fleets.handler import FleetHandler
 from core.ibm_cloud.code_engine.fleets.utils import (
@@ -218,12 +219,12 @@ class FleetsRunner(AbstractRunner):
                     secondary_log_filename=LOG_FILENAME,
                     secondary_log_filter_key=LOG_FILTER_KEY,
                 )
-                run_env_variables.extend(
-                    [
-                        {"type": "literal", "name": "JOB_ID_GATEWAY", "value": job_id},
-                        {"type": "literal", "name": "DATA_PATH", "value": paths["user_mount_path"]},
-                    ]
+                run_env_variables.append(
+                    {"type": "literal", "name": "JOB_ID_GATEWAY", "value": job_id},
                 )
+
+                gateway_env = self._build_gateway_env_vars()
+                run_env_variables.extend(gateway_env)
                 run_commands = build_run_commands(
                     app_run_commands=["python", f"{paths['provider_mount_path']}/{self.job.program.entrypoint}"],
                     secondary_log_filter_key=LOG_FILTER_KEY,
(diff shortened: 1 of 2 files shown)


gateway/core/services/runners/fleets_runner.py
```diff
@@ -218,12 +219,12 @@ def submit(self) -> None:
                     secondary_log_filename=LOG_FILENAME,
                     secondary_log_filter_key=LOG_FILTER_KEY,
                 )
-                run_env_variables.extend(
-                    [
-                        {"type": "literal", "name": "JOB_ID_GATEWAY", "value": job_id},
-                        {"type": "literal", "name": "DATA_PATH", "value": paths["user_mount_path"]},
-                    ]
```
ElePT, line 225: was the DATA_PATH variable dead code?

gateway/core/services/runners/fleets_runner.py
```diff
@@ -505,6 +506,27 @@ def _get_handler(self) -> FleetHandler:
...
         return self._handler
 
+    def _build_gateway_env_vars(self) -> list[dict[str, str]]:
+        """Extract job env vars so the container can call save_result() and use Qiskit Runtime."""
+        try:
+            env = json.loads(self.job.env_vars)
+            env = decrypt_env_vars(env)
+        except Exception:  # pylint: disable=broad-exception-caught
+            logger.warning("Could not decrypt env_vars for job [%s]", self.job.id)
+            return []
```
ElePT: I don't think this try/except +  is log really necessary, `decrypt_env_vars` already has a try-except. If anything, what could fail is the json loading.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -505,6 +506,27 @@ def _get_handler(self) -> FleetHandler:
...
+
+        token = env.get("ENV_JOB_GATEWAY_TOKEN", "")
+        if token:
+            env["QISKIT_IBM_RUNTIME_CUSTOM_CLIENT_APP_HEADER"] = (
+                "middleware_job_id/" + str(self.job.id) + "," + token + "/"
+            )
```
ElePT: afaik this code isn't used anywhere, and should be cleaned up right @Tansito?

ElePT (commented):
Thanks for the change! I think that it's missing unit tests and a couple of small comments I left in the code.

gateway/core/services/runners/fleets_runner.py
marceloamaral: DATA_PATH is already in job.env_vars, so we do not need to add it here anymore.

gateway/core/services/runners/fleets_runner.py
marceloamaral: @ElePT removed this, we can add it back in future if needed!

ElePT (commented):
The changes look good, but as a reminder: the PR is missing the unit tests

marceloamaral:
@ElePT added the unit tests!

avilches (approved):

ElePT (approved):
LGTM

PR #2075, opened 2026-05-05
Add zone based CE project selection

ElePT:
### Summary
Adds `FLEETS_PROFILE_ZONE_MAP` — a JSON env var that maps compute profiles to availability zones, so Fleets jobs are routed to the right Code Engine project based on the hardware they need.

Before this, `_get_or_assign_project()` always picked the first active project. Now it checks the profile against the map and filters by zone. Profiles absent from the map (or mapped to `"any"`) fall back to the multi-zone project (zone null/blank), keeping existing behavior for CPU jobs.

### Details and comments
I will need to do a follow-up to revamp the docker-compose settings too, but I'd rather leave it to a follow-up.


gateway/main/settings.py
```diff
@@ -309,6 +310,10 @@
 CE_PDS_NAME_STATE = os.environ.get("CE_PDS_NAME_STATE")
 CE_PDS_NAME_USERS = os.environ.get("CE_PDS_NAME_USERS")
 CE_PDS_NAME_PROVIDERS = os.environ.get("CE_PDS_NAME_PROVIDERS")
+# Maps compute profiles to their availability zone. Populated at deploy time via FLEETS_PROFILE_ZONE_MAP
+# JSON env var, e.g. '{"gx2-8x64x1l40s": "us-east-1", "gx3d-24x120x1a100p": "us-east-2"}'.
+# Profiles absent from the map (or mapped to "any") fall back to the multi-zone project.
+FLEETS_PROFILE_ZONE_MAP: dict = json.loads(os.environ.get("FLEETS_PROFILE_ZONE_MAP", "{}"))  # type: ignore[assignment]
```
marceloamaral, line 317: This is a workaround for the demo, right? The Code Engine team said they expect to support automatic zone selection by the end of May. 

So, maybe we should add a `TODO:` here stating that this workaround should be deprecated and reviewed in June.

ElePT: Can add the TODO, but it's more than a workaround for a demo, it's the only way we can reliably provision resources today, that's why I decided to add it.

marceloamaral (changes requested):
@ElePT can you rebase with main and add the TODO?
After that, this PR is ready to be merged!

gateway/main/settings.py
```diff
@@ -309,6 +310,10 @@
 CE_PDS_NAME_STATE = os.environ.get("CE_PDS_NAME_STATE")
 CE_PDS_NAME_USERS = os.environ.get("CE_PDS_NAME_USERS")
 CE_PDS_NAME_PROVIDERS = os.environ.get("CE_PDS_NAME_PROVIDERS")
+# Maps compute profiles to their availability zone. Populated at deploy time via FLEETS_PROFILE_ZONE_MAP
+# JSON env var, e.g. '{"gx2-8x64x1l40s": "us-east-1", "gx3d-24x120x1a100p": "us-east-2"}'.
+# Profiles absent from the map (or mapped to "any") fall back to the multi-zone project.
```
ElePT, line 315:
```suggestion
# Profiles absent from the map (or mapped to "any") fall back to the multi-zone project.
# TODO: review once automatic zone selection is supported in CE
```

ElePT:
Done! Although I couldn't actually use the word TODO (it breaks the linter https://github.com/Qiskit/qiskit-serverless/actions/runs/25389292027/job/74459085297#step:6:27), I added it as a reminder. I don't think there is any risk of actually forgetting this one. Another note, @marceloamaral, is that there is no need to rebase unless there are conflicts with the main branch. The merge queue handles the branch update automatically.

gateway/main/settings.py
```diff
@@ -309,6 +310,11 @@
...
+# TODO: review once automatic zone selection is supported in CE
```
ElePT:
```suggestion
# Reminder: review once automatic zone selection is supported in CE
```

marceloamaral (dismissed):
/LGTM

gateway/core/services/runners/fleets_runner.py
```diff
@@ -458,7 +459,15 @@ def _get_or_assign_project(self) -> CodeEngineProject:
                 raise RunnerError(f"Code Engine project '{project.project_name}' is not active")
             return project
 
-        project = CodeEngineProject.objects.filter(active=True).first()
+        profile_zone_map: dict = getattr(settings, "FLEETS_PROFILE_ZONE_MAP", {})
```
avilches:
```suggestion
        profile_zone_map: dict = settings.FLEETS_PROFILE_ZONE_MAP
```

default value should be located in the settings.py...

avilches (dismissed):

gateway/core/services/runners/fleets_runner.py
avilches: Not a big deal actually, approved

isma-roza (approved):

ItamarGoldman (dismissed):

ItamarGoldman (approved):

PR #2076, opened 2026-05-05
Fleet scheduling updates: fleet_id retry preservation and COS result retrieval

ElePT:
### Summary
Two scheduler fixes for Fleets jobs discovered during demo deployment.

**`fleet_id` lost on RecordModifiedError retry** (`schedule_queued_jobs.py`) -> `FleetsRunner.submit()` sets `job.fleet_id` in memory, but the field isn't persisted until `job.save()`. If a concurrent write causes `RecordModifiedError`, `refresh_from_db()` reloads the DB state and wipes the in-memory `fleet_id`. The retry then saves the job without it, permanently losing the fleet reference. Fix: back up and restore `fleet_id` alongside the existing `ray_job_id` backup.

**Fleets results not saved to database** (`update_jobs_statuses.py`) ->  When a Fleets job reaches a terminal state, the scheduler was only fetching and persisting logs for Ray jobs. Fleets jobs have their logs written directly to COS by the PDS shell wrapper, but results (return values) still need to be pulled from COS and saved to `job.result`. Added the `else` branch that calls `runner.get_result_from_cos()` on terminal transition, with a logged warning on failure so a COS error doesn't fail the job status update.

Also updated the `save_logs_to_storage` docstring to clarify it is Ray-only.

### DIFF
--- a/gateway/scheduler/tasks/schedule_queued_jobs.py
+++ b/gateway/scheduler/tasks/schedule_queued_jobs.py
@@ -109,6 +109,7 @@ class ScheduleQueuedJobs(SchedulerTask):
                 backup_logs = job.logs
                 backup_resource = job.compute_resource
                 backup_ray_job_id = job.ray_job_id
+                backup_fleet_id = job.fleet_id
 
                 succeed = False
                 attempts = settings.RAY_SETUP_MAX_RETRIES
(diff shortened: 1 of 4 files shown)


marceloamaral (approved):
LGTM

PR #2090, opened 2026-05-06
Rename AbstractRunner to Runner and move the factory func to a static method

avilches:
### Summary

Rename AbstractRunner to Runner and move the factory func to a static method

### Details and comments

No behavior change, just add more clarity. IMHO it's better to import classes and then use their methods, rather than importing individual functions. When calling a static method of a class, you have to put the class name in front of it, so it acts as a namespace that adds semantics to the call.


gateway/core/services/runners/runner.py
```diff
@@ -30,9 +30,35 @@ def __str__(self) -> str:
...
+        from core.models import Program  # pylint: disable=import-outside-toplevel
+        from core.services.runners.fleets_runner import FleetsRunner  # pylint: disable=import-outside-toplevel
+        from core.services.runners.ray_runner import RayRunner  # pylint: disable=import-outside-toplevel
+
+        if job.runner == Program.FLEETS:
+            return FleetsRunner(job)
+        if job.runner == Program.RAY:
+            return RayRunner(job)
+        raise RunnerError(f"Unknown runner type: {job.runner}")
+
```
ElePT, line 61: I see why you are making this proposal, but to be honest I find this circular dependency kind of awkward. I need to think if there's a way we could make it look nicer.

avilches: Closing, nobody, including me, likes the local imports...

PR #2096, opened 2026-05-07
Update permission names

ismaelRozasRamallal:
### Summary
This PR is intended to align more on permissions naming with IBM Cloud conventions.

### Details and comments
TBD

### DIFF
--- a/gateway/core/models.py
+++ b/gateway/core/models.py
@@ -25,13 +25,13 @@ RUN_PROGRAM_PERMISSION = "run_program"
 # Platform permissions (Runtime API instances access client)
 PLATFORM_PERMISSION_READ = "function.read"  # see function in catalog and retrieve its metadata
 PLATFORM_PERMISSION_RUN = "function.run"  # execute a new job of this function
-PLATFORM_PERMISSION_JOB_READ = "function.job.read"  # retrieve a specific job from this function (non-author access)
-PLATFORM_PERMISSION_USER_FILES = "function.files"  # list, download, upload and delete files in user space
+PLATFORM_PERMISSION_JOB_READ = "function-job.read"  # retrieve a specific job from this function (non-author access)
+PLATFORM_PERMISSION_USER_FILES = "function.edit"  # list, download, upload and delete files in user space
 # Provider admin permissions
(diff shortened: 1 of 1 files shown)


ismaelRozasRamallal:
@Tansito @avilches what do you think about this new action names? does it make sense for you? (feel free to close the PR if not 👍 )

gateway/core/models.py
```diff
@@ -25,13 +25,13 @@
...
+PLATFORM_PERMISSION_USER_FILES = "function.edit"  # list, download, upload and delete files in user space
 # Provider admin permissions
-PLATFORM_PERMISSION_PROVIDER_UPLOAD = "function.provider.upload"  # create or update this function's code
-PLATFORM_PERMISSION_PROVIDER_JOBS = "function.provider.jobs"  # list jobs from all users of this function
-PLATFORM_PERMISSION_PROVIDER_LOGS = "function.provider.logs"  # read provider-side logs of jobs from this function
-PLATFORM_PERMISSION_PROVIDER_FILES = "function.provider.files"  # list, download, upload, delete files in provider space
+PLATFORM_PERMISSION_PROVIDER_UPLOAD = "function.upload"  # create or update this function's code
+PLATFORM_PERMISSION_PROVIDER_JOBS = "function-job.list"  # list jobs from all users of this function
+PLATFORM_PERMISSION_PROVIDER_LOGS = "function-log.read"  # read provider-side logs of jobs from this function
+PLATFORM_PERMISSION_PROVIDER_FILES = "function.manage"  # list, download, upload, delete files in provider space
```
avilches, line 34: What criteria did you choose to change these name? Each rename should have the reason to change. We have to keep in mind we have two use cases only: partners and users, each one has access to a specific part of the data. With your proposal, this boundary is blurry: you proposed `function-job.read` and `function-job.list` where sound similar but do things very different.

What is the permission naming style from IBM Cloud? Maybe we can take a look together and try to refine them.

avilches: After some time reviewing the PR, I think I get the point: you want `item.action` or `multiple-word-item.action`, where the action is a verb. But we dots in the middle of the item, and sometimes permissions don't end in an action. I will meet with @Tansito this afternoon to improve this.

ismaelRozasRamallal:
Finally solved in another PR, we can close this one.

PR #2099, opened 2026-05-07
Add ARGUMENTS_PATH env var for runner-controlled argument file location

marceloamaral:
<!--
⚠️ Also, please add a release note file using reno if the change needs to be
  documented in the release notes.

- [X] I have added the tests to cover my changes.
- [ ] I have read the CONTRIBUTING document.
-->

### Summary

 - Introduces `ARGUMENTS_PATH` environment variable so each runner controls where arguments are stored                                                    
  - SDK `get_arguments()` reads directly from this path instead of computing it from `DATA_PATH` + `ENV_JOB_ID_GATEWAY`                                    
  - Fleets uses `/data/arguments.json` (flat file at job root); 
  - Ray uses `/data/arguments/{job_id}.json`                                                   


### Details and comments

The SDK previously hardcoded the arguments path pattern (`{DATA_PATH}/arguments/{job_id}.json`), coupling path layout to the SDK. With job-level mount isolation in Fleets, there's no need for a `{job_id}.json` filename — a flat `arguments.json` is simpler and sufficient.
                                                                                                                                                           
  Changes:

  | File | Change |
  |------|--------|
  | `client/.../constants.py` | Add `ARGUMENTS_PATH` constant |                                                                                            
  | `client/.../program_serializers.py` | Rewrite `get_arguments()` to use `ARGUMENTS_PATH` directly |                                                     
  | `gateway/api/utils.py` | Set `ARGUMENTS_PATH` in `build_env_variables()` for Ray |                                                                     
  | `gateway/.../fleets_runner.py` | Use `username` (was `user_id`) in COS paths; simplify arguments key to `arguments.json`; replace `JOB_ID_GATEWAY` with `ARGUMENTS_PATH` in fleet env vars |                                                                                                                    
  | `gateway/tests/` | Explicit full-path assertions for both custom and provider function COS keys |


gateway/core/services/runners/fleets_runner.py
```diff
@@ -225,7 +224,11 @@ def submit(self) -> None:
                     secondary_log_filter_key=LOG_FILTER_KEY,
                 )
                 run_env_variables.append(
-                    {"type": "literal", "name": "JOB_ID_GATEWAY", "value": job_id},
+                    {
+                        "type": "literal",
+                        "name": "ARGUMENTS_PATH",
+                        "value": f"{paths['user_mount_path']}/arguments.json",
+                    },
```
avilches, line 235: You have to add the ARGUMENTS_PATH in the Ray part too.

marceloamaral: Yes, so `get_arguments()` now always uses `ARGUMENTS_PATH`, and both Ray and Fleets expose their own paths to the container through that environment variable.

client/qiskit_serverless/serializers/program_serializers.py
```diff
@@ -79,29 +79,25 @@ def object_hook(self, obj: Any) -> Any:
...
-            "Error getting arguments: JOB_ID_GATEWAY environment variable is missing or empty"
+            "Error getting arguments: ARGUMENTS_PATH environment variable is missing or empty"
         )
-    # DATA_PATH is just used in tests and local development.
-    # In k8 we always want to use the default "/data"
-    data_path = os.environ.get(DATA_PATH, "/data")
-    arguments_dir = f"{data_path}/arguments"
-    arguments_file_path = f"{arguments_dir}/{job_id_gateway}.json"
-
-    os.makedirs(arguments_dir, exist_ok=True)
```
avilches, line 97: I think we should keep as a fallback for Ray. DATA_PATH is needed for tests, where the mounted volume in Docker compose is the root, the k8s mounted volume as /data is the user folder).

avilches: Nevermind, we can use ARGUMENTS_PATH for Ray and docker compose, but we have to map a different folder in docker-compose than in k8s.

marceloamaral: Alright, so `ARGUMENTS_PATH` is already derived from `data_path` in `gateway/api/utils.py:121`, so it follows the existing docker-compose/K8s behavior.

`data_path` already handles those cases in `gateway/api/utils.py:95-111`:
```
# Local (docker-compose)
data_path = /data/{username}

# K8s
data_path = /data
```

Then `ARGUMENTS_PATH` is set from that value:
```
"ARGUMENTS_PATH": f"{data_path}/arguments/{job.id}.json",
```
The unit tests in `tests/api/test_utils.py` cover all three variants: K8s, local, and local with provider.

ElePT, line 97: Why is `os.makedirs(arguments_dir, exist_ok=True) `removed instead of adjusted to work with the new variable names?

client/qiskit_serverless/core/constants.py
```diff
@@ -42,6 +42,7 @@
 QISKIT_IBM_CHANNEL = "QISKIT_IBM_CHANNEL"
 
 DATA_PATH = "DATA_PATH"
+ARGUMENTS_PATH = "ARGUMENTS_PATH"
```
ElePT, line 45: What is the vision here regarding DATA_PATH? I had understood that we would essentially be replacing it with the specific paths (ARGUMENTS_PATH, RESULTS_PATH, LOGS_PATH...). Is this still the case? I see that data_path is used later to construct arguments path, which sounds like taking 2 steps to do sth that could be done in 1 step?

ElePT (commented):
Overall the change looks good to me. I left a couple of questions to understand a bit more how we see the changes progressing but I think this fits what we had discussed :)

client/qiskit_serverless/serializers/program_serializers.py
marceloamaral: Good question. I removed it because `get_arguments()` should only read the arguments file, not create directories.

In the expected flow, the gateway first saves the arguments through `ArgumentsStorage.save()`, which uses `PathBuilder.absolute_path()` to create the directory and write the file to the shared volume. Only after that, the gateway submits the job to Ray. When the Ray node calls `get_arguments()`, the file should already exist on the mounted volume.

So creating the directory again in `get_arguments()` is redundant. If the directory or file is missing at that point, the next `os.path.isfile()` check will fail anyway, because it means the arguments were not properly written or are not visible to the worker.

The old `os.makedirs` was redundant with the writer side and could not make the read succeed if the file itself was missing.

Or maybe I am missing something...

client/qiskit_serverless/core/constants.py
marceloamaral: Wouldn’t `DATA_PATH` also be used by users to access the other uploaded files?

I am not sure if we should completely remove the `DATA_PATH`?

ElePT: It's a choice we need to make, I would get rid of DATA_PATH and instead I would add a "FILES_PATH" for this specific case, but we can discuss it offline.

client/qiskit_serverless/serializers/program_serializers.py
ElePT: I am not 100% sure, but in my experience these changes need to be thoroughly tested. My take is: if you can't verify this doesn't break anything, it's better not to change it.

avilches: I'm with Marcelo, I don't think doing `os.makedirs(arguments_dir, exist_ok=True)` on a mounted volume in the COS will have any effect at all...

gateway/tests/core/services/runners/test_fleets_runner.py
```diff
@@ -242,30 +242,52 @@ def test_stop_returns_false_when_already_terminal():
...
+    runner.job.id = "job-aaa-111"
 
-    Verifies that user/provider function prefixes and job-level paths follow
-    the convention used by fleets_runner._build_cos_paths() and expected by
-    the entrypoint running inside the container.
-    """
+    paths = runner._build_cos_paths()  # pylint: disable=protected-access
+
+    assert paths["user_function_prefix"] == "users/alice/provider_functions/default/hello-world"
+    assert paths["user_job_prefix"] == "users/alice/provider_functions/default/hello-world/jobs/job-aaa-111"
```
avilches, line 256: Are we going to keep the /user prefix?

marceloamaral: I think we agreed to keep it because it preserves the semantic meaning of the levels

client/qiskit_serverless/serializers/program_serializers.py
marceloamaral: Well, I added it back...

avilches (approved):

PR #2100, opened 2026-05-07
Improve backoffice speed with auto complete long combo box

avilches:
### Summary

This PR changes some fields with autocomplete, which improves the speed. Without the autocomplete, all the data must be loaded in the combo boxes. The autocomplete allows you to type the id, using ajax to load the combo box.


ElePT (approved):
I am approving because improving the speed of backoffice would be glorious, but I am not the most proficient with it so I may be missing some details someone else might be able to review too.

avilches:
> I am approving because improving the speed of backoffice would be glorious, but I am not the most proficient with it so I may be missing some details someone else might be able to review too.

I tested locally and it's worth every penny...!

PR #2101, opened 2026-05-07
Stream Ray job logs to limit memory usage

avilches:
## Problem

`RayRunner.logs()` called `JobSubmissionClient.get_job_logs()`, which loads the full log into a single Python string before truncation. With GB-scale logs this causes OOM or HTTP timeouts.

## Solution

Stream the Ray dashboard endpoint (`/api/jobs/{id}/logs`) in 64 KB chunks, parse the JSON body line by line on the fly, and keep lines in a `deque[str]` rolling buffer bounded by `FUNCTIONS_LOGS_SIZE_LIMIT`. Memory stays ~50 MB regardless of log size.

### Scheduler 

The Ray logs must be uploaded to COS when job finishes.

Before:
1. `UpdateRayJobsStatuses` calls to `ray_runner.logs()`, which calls `get_job_logs()` and load full string in memory
2. `check_logs()` truncates
3. `save_logs_to_storage()` filter public/private prefixes and writes to COS

After:
1. `runner.logs()` streams into `deque[str]`, truncation happens inside the loop
2. `save_logs_to_storage()` receives the deque, filters directly (`Iterable[str]`) and writes to COS

### `/logs` and `/provider-logs`
Return the logs for a running Ray job

Before:
1. `runner.logs()` / `runner.provider_logs()` returns full string via `get_job_logs()`
2. `check_logs()` truncates
3. `filter_logs` splits and filters

After:
1. `runner.logs()` / `runner.provider_logs()` returns bounded `deque[str]`
2. `filter_logs` functions filter directly (`Iterable[str]`, no `splitlines()` round-trip)

## Other changes

- Dead `logs()` / `provider_logs()` stubs removed from `AbstractRunner` and `FleetRunner`
- `test_large_logs` integration test added, disabled by default (`LARGE_LOGS_TEST=1` to run, `LOGS_SIZE_MB` controls volume, default 500)


gateway/scheduler/tasks/update_ray_jobs_statuses.py
```diff
@@ -93,7 +90,7 @@ def update_job_status(self, job: Job):
                 try:
                     lines = runner.logs()
                 except RunnerError:
```
avilches, line 96: add logs

gateway/scheduler/tasks/update_ray_jobs_statuses.py
```diff
@@ -93,7 +90,7 @@ def update_job_status(self, job: Job):
...
-                    lines = deque()
+                    lines = FilteredLogs(public_logs=deque(), private_logs=None)
```
avilches: Log the error in the logs

gateway/tests/core/test_ray.py
```diff
@@ -119,8 +119,14 @@ def test_job_status(self):
 
     def test_job_logs(self):
         """Tests job logs."""
```
avilches, line 121: ????

Tansito (approved):
Awesome work @avilches ! Let's merge it and test this as soon as we can in staging 🚀

PR #2102, opened 2026-05-07
Support different accesses for Arguments storage

Tansito:
### Summary

This PR prepares the main structure to be able to inject different storages in the platform. The idea behind the refactor was the next:
- Unify inputs through ArgumentsStorage (and potentially among others storages like in this case LogsStorage).
- Create a factory to abstract the usage of the specific implementation for each kind of storage.

### Details and comments

- ArgumentsStorage was divided in two: one for ray and another one for fleets.
- ArgumentsStorage now uses Job as inpit simplifying the instantiation and unifying interfaces with LogsStorage.

This PR works as a first iteration from a bigger work here: #2061 

Co-authored-by: @korgan00

### DIFF
--- a/gateway/api/serializers.py
+++ b/gateway/api/serializers.py
@@ -18,7 +18,7 @@ from rest_framework import validators as validators_module
 from api.utils import build_env_variables, sanitize_name
 from core.domain.business_models import BusinessModel
 from core.model_managers.job_events import JobEventContext, JobEventOrigin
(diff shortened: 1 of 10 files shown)


ElePT (approved):
The changes look good to me, it's mostly the refactoring we agreed + using the job input directly, which is correctly checked in the tests.

avilches (approved):

PR #2104, opened 2026-05-08
Scheduler logs improvements

ElePT:
### Summary

While debugging recent scheduler incidents, I found that `runner.logs()` is called twice per scheduler iteration when a Ray job transitions to a terminal state: once in the terminal-state handler to save the logs (line 109 of `update_jobs_statuses.py`), and again unconditionally for the "no resources" string check (line 126).

It looks like this oversight was introduced in #1942, when the `if job_handler:` was replaced with a`runner.logs()` call. And well, it is a redundant fetch.

Also fixes a display bug that printed raw byte counts labelled as "MB" (e.g. `1157317193 MB > 52428800 MB` instead of `~1103.6 MB > 50.0 MB`).

### Details and comments

Added regression tests and tests for log utilities:

- `TestRayJobStatusUpdate::test_logs_fetched_once_on_terminal_state`: regression tests that asserts `runner.logs()` is called exactly once for a terminal-state transition. Fails against the unfixed code (`call_count == 2`).
- `TestRayJobStatusUpdate::test_logs_fetched_for_running_job`: ensures the "no resources" check still runs for non-terminal jobs.
- `TestCheckLogs`: new unit tests for `core.utils.check_logs`, including a regression test for the MB display fix.

### DIFF
--- a/gateway/scheduler/tasks/update_jobs_statuses.py
+++ b/gateway/scheduler/tasks/update_jobs_statuses.py
@@ -123,7 +123,7 @@ class UpdateJobsStatuses(SchedulerTask):
                 if job.status == Job.SUCCEEDED:
                     self._record_execution_duration(job)
 
-        if not is_fleets_job:
+        if not is_fleets_job and not job.in_terminal_state():
             try:
                 logs = runner.logs()
             except RunnerError:
--- a/gateway/core/utils.py
+++ b/gateway/core/utils.py
@@ -223,10 +223,10 @@ def check_logs(logs: Union[str, None], job) -> str:
 
     if logs_size > max_bytes:
         logger.warning(
-            "Job %s is exceeding the maximum size for logs %s MB > %s MB.",
+            "Job %s is exceeding the maximum size for logs %.1f MB > %.1f MB.",
             job.id,
-            logs_size,
-            max_bytes,
+            logs_size / (1024**2),
+            max_bytes / (1024**2),
         )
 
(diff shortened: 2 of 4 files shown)


gateway/scheduler/tasks/update_jobs_statuses.py
```diff
@@ -123,7 +123,7 @@ def update_job_status(self, job: Job):
                 if job.status == Job.SUCCEEDED:
                     self._record_execution_duration(job)
 
-        if not is_fleets_job:
+        if not is_fleets_job and not job.in_terminal_state():
```
ismaelRozasRamallal, line 126: When I saw this block of code yesterday, I was about to suggest a refactor. Why?

There is several levels chaining if clauses. It means, a decission tree with multiple depth levels which makes harder to know what happens and when happens. Chaining if clauses is useful when having a low level of complexity in the logic, but when it gets more complex, it's more useful to align explicit full condition if clauses which help to flatten the decision tree and makes much more easy to understand the implications and actions under each group of conditions.

I am not saying to do it in this PR, but could be an awesome improvement...

ismaelRozasRamallal (approved):
LGTM!

gateway/scheduler/tasks/update_jobs_statuses.py
ElePT: I think that's a good point, we can take it into account for a follow-up.

avilches (approved):

PR #2105, opened 2026-05-08
Functions trough instances: fixes

avilches:
### Summary

1. Adds the api key to connect to Runtime API
2. Bugfix: the author of a provider function will only be able to see it if they have permissions, regardless of whether they are the author or not.

### Details

`with_permission()` used `Q(author=author)` as an unconditional filter in the runtime instances path, which matched both custom functions and provider functions authored by the requesting user. This caused a user to see provider functions they had uploaded regardless of whether their instance had `function.read` permission. The fix narrows that criterion to Q(author=author, provider=None): custom functions are always visible (they are not instance-gated), but provider functions are always subject to instance permissions, even when the user is their author.

### DIFF
--- a/gateway/api/authentication.py
+++ b/gateway/api/authentication.py
@@ -69,7 +69,7 @@ class CustomTokenBackend(authentication.BaseAuthentication):
             public_access=public_access,
         ).execute()
 
-        accessible_functions = FunctionAccessClient().get_accessible_functions(crn)
+        accessible_functions = FunctionAccessClient().get_accessible_functions(crn, authorization_token)
         return quantum_user, CustomAuthentication(
             channel=channel,
             token=authorization_token.encode(),
--- a/gateway/api/clients/function_access_client.py
+++ b/gateway/api/clients/function_access_client.py
@@ -1,5 +1,6 @@
 """FunctionAccessClient."""
 
+import hashlib
 import logging
 
 import requests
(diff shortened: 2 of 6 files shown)


riley2012 (dismissed):
LGTM!

riley2012:
@avilches is this an issue?

One thing worth flagging: the cache key doesn't include the api key. Since the Runtime API
returns user-specific permissions, two users sharing the same CRN but with different api keys
could get each other's cached permission set. 

Maybe something like this:
` cache_key = f"accesible_functions:{instance_crn}:{hash(api_key)}"`

avilches:
> @avilches is this an issue?
> 
> One thing worth flagging: the cache key doesn't include the api key. Since the Runtime API returns user-specific permissions, two users sharing the same CRN but with different api keys could get each other's cached permission set.
> 
> Maybe something like this: ` cache_key = f"accesible_functions:{instance_crn}:{hash(api_key)}"`

You are right! Thanks, I will add it!

avilches:
Added in this commit: https://github.com/Qiskit/qiskit-serverless/pull/2105/changes/6c5895d8f4d4c33ec11f7d5ca27002b017dc46c3 /cc @riley2012 could you take a look again?

avilches:
It is worth explaining that permissions per function are unique by Instance crn. What can be different is the access to the /functions endpoint, which may be blocked for some users and not for others (in theory, in practice everyone should have access), but the response content only depends on the instance crn. :)

riley2012 (approved):

PR #2106, opened 2026-05-09
Replace retry loop with save(update_fields) in ScheduleQueuedJobs

avilches:
### Summary

`ScheduleQueuedJobs` uses `job.save()` inside a retry loop that handles `RecordModifiedError` because the [optimistic locking](https://django-concurrency.readthedocs.io/en/latest/).

### Details and comments

This PR replaces it with `save(update_fields=[...])`, which bypasses the django Optimistic locking version check, making the retry loop with the sleep(1) unnecessary.

### DIFF
--- a/gateway/scheduler/tasks/schedule_queued_jobs.py
+++ b/gateway/scheduler/tasks/schedule_queued_jobs.py
@@ -8,8 +8,6 @@ from datetime import datetime, timezone
 from django.conf import settings
 from django.db.models import Count
 
-from concurrency.exceptions import RecordModifiedError
-
 from opentelemetry import trace
 from opentelemetry.trace.propagation.tracecontext import TraceContextTextMapPropagator
 from core.config_key import ConfigKey
@@ -63,7 +61,7 @@ class ScheduleQueuedJobs(SchedulerTask):
         running_clusters = ComputeResource.objects.filter(active=True, gpu=True).count()
         self._schedule_jobs_if_slots_available(max_clusters, running_clusters, gpu_job=True)
 
-    def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches, too-many-locals
+    def _schedule_jobs_if_slots_available(
         self,
         max_slots_possible,
         number_of_slots_running,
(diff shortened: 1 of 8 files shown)


gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -105,64 +103,25 @@ def _schedule_jobs_if_slots_available(  # pylint: disable=too-many-branches, too
...
-                        retries,
-                    )
-                else:
-                    logger.warning(
-                        "job_id=%s Job save failed after %s tries (%.2fs)",
-                        job.id,
-                        retries,
-                        time.monotonic() - t1,
-                    )
+                job.save(update_fields=["status", "ray_job_id", "compute_resource", "fleet_id"])
```
avilches: The logs field was removed because is not needed anymore....

gateway/tests/scheduler/test_schedule_queued_jobs.py
```diff
@@ -28,40 +14,28 @@ def _make_task():
     return ScheduleQueuedJobs(kill_signal=kill_signal, metrics=MagicMock())
 
 
-def test_fleet_id_preserved_on_record_modified_error_retry():
-    """fleet_id set by execute_job is restored after RecordModifiedError + refresh_from_db."""
+def test_job_saved_with_update_fields_no_logs():
+    """Job is saved with explicit update_fields — logs is excluded (deprecated)."""
```
avilches:
```suggestion
    """Job is saved with explicit update_fields"""
```

Tansito (dismissed):
Commented with @avilches , we need to ensure in staging and also keep an eye in production because this kind of race conditions are difficult to detect without enough traffic or big loads.

avilches:
I merged the branch from https://github.com/Qiskit/qiskit-serverless/pull/2111 so the CI check passes

gateway/tests/scheduler/test_schedule_queued_jobs.py
```diff
@@ -1,21 +1,9 @@
...
-# (C) IBM 2026
-#
-# This code is licensed under the Apache License, Version 2.0. You may
-# obtain a copy of this license in the LICENSE.txt file in the root directory
-# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
-#
-# Any modifications or derivative works of this code must retain this
-# copyright notice, and modified files need to carry a notice indicating
-# that they have been altered from the originals.
-
```
ElePT, line 12: why remove this? It's not used widely in the tests but in Qiskit we keep the headers.

ElePT (approved):
I'm with @Tansito, I feel like this is something we need to test and keep an eye on, but in principle I don't have any comments for the code, makes sense.

gateway/tests/scheduler/test_schedule_queued_jobs.py
avilches: Should we add this to all the classes? I only see this header in a few test from fleet classes...

PR #2107, opened 2026-05-09
Extract job status metric into a dedicated task

avilches:
### Summary

`ScheduleQueuedJobs` owns the `_update_job_status_counts_metric` function, which counts all jobs by status and provider regardless of runner type.

This PR moves that logic into a dedicated `UpdateJobStatusCounts` scheduler task, giving it a single clear responsibility.


ElePT (approved):
LGTM! Makes a lot of sense

PR #2108, opened 2026-05-11
Refactor: decouple COS and auth from FleetHandler

ElePT:
### Summary
`FleetHandler` was doing too much: auth setup, COS credential resolution, and CE API calls. This PR splits those responsibilities apart. `build_ce_auth` is a new shared factory in `clients.py` that returns a ready `(ApiClient, IBMCloudClientProvider)`. On the COS side, `build_cos_client(project)` is a standalone factory that fetches HMAC credentials from the CE secret and returns a ready `JobCOS`. `FleetsRunner` initializes the two independently via `_get_handler()` and `_get_cos()`.

As part of this PR work I found that the swagger `Configuration` class used a shallow-copy singleton, so multiple `FleetHandler` instances were mutating each other's IAM token dict. This is now fixed by assigning fresh dicts instead of mutating. `FleetHandler` now just takes the pre-built `ApiClient` and stays a pure CE API wrapper

### Details and comments
I am leaving the PR as a draft because I am having issues testing it on the demo branch, I will make it ready for review once I have verified that it works.

The goal of this and the follow-up PR is to enable the following flow at job creation time (rough implementation) without unnecessarily calling the `FleetsRunner` or the `FleetHandler`:

```
code_engine_project = select_ce_project(compute_profile)
cos = build_cos_client(code_engine_project)
cos.upload_fileobj(fileobj=..., bucket_name=code_engine_project.cos_bucket_user_data_name, key=f"jobs/{job.id}/arguments.json")
```


Tansito (changes requested):
Thanks for the PR @ElePT in essence I think it's the correct path to follow! A couple of minor comments that doesn't change the implementation but I think they are worth it:
- We are building factories in the `__init__` files to avoid local imports like the one we have in `build_ce_auth`. You can use [get_runner](https://github.com/Qiskit/qiskit-serverless/blob/4fc1291424451589c1a57eec01419f0f3778b7dc/gateway/core/services/runners/__init__.py#L10) or [get_arguments_storage](https://github.com/Qiskit/qiskit-serverless/blob/4fc1291424451589c1a57eec01419f0f3778b7dc/gateway/core/services/storage/__init__.py#L8) as an example.
- I would like to propose change `build_ce_auth` and `build_cos_client` names by `get_*` to maintain consistency along all our factories.
- Maybe not for this PR, totally up to you here, I saw that in `build_ce_auth` we are returning a `Tuple`. Maybe it's better to return an object in this case as it can be confusing that a Factory returns a tuple here (it's also something we can do later if we want to move forward with this, no problem, it almost a refactor).

ElePT:
Suggestions applied @Tansito, including the refactoring you suggested (returning a `CEAuth` object instead of a tuple). Let me know what you think!

Tansito (approved):
Pretty cool thank you much @ElePT 🎉

PR #2109, opened 2026-05-11
Refactor: move CE project selection to job creation time

ElePT:
### Summary
~This PR is built on top of the changes from #2108 and is currently blocked by it.~

Up until now the gateway selected a Code Engine project the first time the scheduler touched a job, deep inside `FleetsRunner`. This PR moves that decision to `RunJobSerializer.create()`, the moment the job is written to the database.

The zone-lookup logic (reading `FLEETS_PROFILE_ZONE_MAP` and querying `CodeEngineProject` by zone) moves into a new `select_ce_project()` helper in `core/model_managers/`. The serializer calls it for every Fleets job and writes the chosen project onto the job record in the same save, so the runner never has to figure this out. If no suitable project exists the API returns a 400 right away instead of the job sitting in the queue and failing later. 

On the runner side, `_get_or_assign_project()` is replaced by a much simpler `_get_project()` that just reads and validates whatever the serializer already put there.

### Details and comments
This PR also fixes the`sync_ce_project` management command to define two separate initialization paths:

- a single_project path that uses the existing CE_PROJECT_ID variable. This path assumes automatic zone selection is available for resource provisioning (which is not the case currently, but should be in the future).
- a multi-project path with 2 new variables:`CE_ZONE` lets single-project deployments pin their project to an availability zone, and `CE_PROJECTS` accepts a JSON array for multi-zone setups where each zone has its own CE project.


Tansito (commented):
Initially the code here looks good to me but I would like to take a better look as soon as we merge #2108 . With less changes it will be quite easier.

gateway/core/model_managers/code_engine_projects.py
```diff
@@ -0,0 +1,57 @@
...
+import logging
+
+from django.conf import settings
+
+from core.models import CodeEngineProject
+
+logger = logging.getLogger("core.model_managers.code_engine_projects")
+
+
+def select_ce_project(compute_profile: str | None):
```
Tansito: @korgan00 do you mind to review this with @ElePT Monday? I think our model managers are being use with a QuerySet. I don't know if it has sense or not, if you can take a look before our weekly so we can comment it there it would be nice.

Tansito: Also, if you agree with the proposal feel free to implement it before and I can approve it when I get up.

Tansito (approved):
Lovely @ElePT , thank you so much! 🫡

PR #2124, opened 2026-05-13
Adding RESULT_PATH env var

ItamarGoldman:
### Summary
Make changes for supporting `RESULT_PATH` env variable for directing interacting with the results.

### Details and comments
- [x] Add `RESULTS_PATH` to `constants.py`.
- [x] Add correct path to the cos build path.
- [x] Update `save_result` method - replace the call to the gateway that is currently in save_result with a local save. For ray, the files would be saved to the mounted path while for fleets they would use the COS client (similar to the local read that takes place in get_arguments). Add fallback call to the old `save_result` for backwards compatibility.
- [x] Add tests to cover changes.


client/qiskit_serverless/core/job.py
```diff
@@ -352,6 +353,23 @@ def send_error(code: Union[str, int], message: str, exception: str, args: Option
...
+        raise QiskitServerlessException(
+            "Error getting arguments: RESULTS_PATH environment variable is missing or empty"
+        )
+
+    os.makedirs(os.path.dirname(results_path), exist_ok=True)
+
+    if not os.path.isfile(results_path):
+        raise QiskitServerlessException(f"Error saving results: {results_path} is not a file or doesn't exist")
+
+    return results_path
```
avilches, line 367: what are you testing here? I think asserting the string will be enough

ItamarGoldman: Here I check that the path exists. @korgan00 suggested to remove and make a test instead but I thought it is a minimal check that let us control the error.
If an error is raised is through the `try:.... except....` in `save_result(...)`. I can change it for string assertion instead if you think this is an over the top check.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -227,12 +227,19 @@ def submit(self) -> None:
...
+                    [
+                        {
+                            "type": "literal",
+                            "name": "ARGUMENTS_PATH",
+                            "value": f"{paths['user_mount_path']}/arguments.json",
+                        },
+                        {
+                            "type": "literal",
+                            "name": "RESULTS_PATH",
+                            "value": f"{paths['user_mount_path']}/results.json",
```
marceloamaral: we should use `user_job_prefix`, right?
```
"value": f"{paths['user_job_prefix']}/results.json"
```

The arguments and results will be in different paths.

gateway/tests/api/test_utils.py
```diff
@@ -48,6 +48,7 @@ def test_ibm_cloud_env_var_build(self, settings):
             "ENV_ACCESS_TRIAL": "False",
             "DATA_PATH": "/data",
             "ARGUMENTS_PATH": "/data/arguments/42.json",
+            "RESULTS_PATH": "/data/results/42.json",
```
marceloamaral: can you define the full result path in the test?
```
user_function_prefix = f"users/{username}/provider_functions/{provider_name}/{program_title}"
user_job_prefix = f"{user_function_prefix}/jobs/{job_id}"
```

gateway/tests/api/test_utils.py
```diff
@@ -86,6 +87,7 @@ def test_ibm_cloud_local_env_var_build(self, settings):
             "ENV_ACCESS_TRIAL": "False",
             "DATA_PATH": "/data/IBMid-691000IC75",
             "ARGUMENTS_PATH": "/data/IBMid-691000IC75/arguments/42.json",
+            "RESULTS_PATH": "/data/IBMid-691000IC75/results/42.json",
```
marceloamaral, line 142: again, try to use the full result path in the test

gateway/core/services/runners/fleets_runner.py
ItamarGoldman: I will change as suggested but there is something that is bugging me. this path doesn;t hold for custom functions.
In addition, all files (arguments, logs and results files) are all stored at `users/{username}/custom_functions/{function-title}/jobs/{job-uuid}/`. From my understanding, the `arguments.json` and `results.json` *should be in the same folder*. From that I need to use the same path as arguments: `paths['user_mount_path']}` but that is contradicting what you are saying as the path will be 
`users/{username}/provider_functions/{provider_name}/{program_title}/jobs/{job_id}` which is always a provider function. 

To summery what I don't understand:

- why `arguments` and `results` don't have the same path.
- why is the arguments path is `\data\`
- why we are using only the provider function path? what about the custom functions?

ElePT: Hey @ItamarGoldman! I just merged the latest refactor from #2160 to main, I think that the changes in that refactor answer a lot of your questions, it makes it more clear to see when it's a custom function vs. provider function and where each path points to.

gateway/tests/api/test_utils.py
ElePT: Building up on my previous answer, I recommend looking into the latest tests after the refactor and following their design.

gateway/tests/api/test_utils.py
ElePT: FYI this is the full result path, because all the function code sees is the mount, not the full key. The function doesn't know what /data maps to, that's on the gateway's side.

marceloamaral: That is right, @ItamarGoldman the RESULTS_PATH is good here.

gateway/core/services/runners/fleets_runner.py
marceloamaral: @ItamarGoldman as Elena mentioned, Aberto's PR made the logic easier to follow, see here: https://github.com/Qiskit/qiskit-serverless/pull/2160/changes#diff-a24cc84fea6615d8a5e2127615482c220bcd2db0336fe8b3a9561bc4da41e2bbR207

You can verify if a functions is from a provider or custom and then create the right path
```
self.job.program.provider
```

client/qiskit_serverless/core/job.py
```diff
@@ -352,6 +353,23 @@ def send_error(code: Union[str, int], message: str, exception: str, args: Option
...
+    results_path = os.environ.get(RESULTS_PATH)
+    if not results_path:
+        raise QiskitServerlessException(
+            "Error getting arguments: RESULTS_PATH environment variable is missing or empty"
+        )
+
+    os.makedirs(os.path.dirname(results_path), exist_ok=True)
+
+    if not os.path.isfile(results_path):
+        raise QiskitServerlessException(f"Error saving results: {results_path} is not a file or doesn't exist")
```
ElePT:
```suggestion
```
I tested this PR locally and I confirmed that you'll need to remove this check, @ItamarGoldman. The issue is that before writing the result the file will not exist, so the check will always fail. I think that checking the results path should be enough in this case.

ElePT (changes requested):

client/qiskit_serverless/core/job.py
ItamarGoldman: I thought that `os.makedirs(os.path.dirname(results_path), exist_ok=True)` makes sure it doesn't fail. I will remove it. Thanks!

gateway/tests/api/test_utils.py
marceloamaral: Sorry for the confusion; that was my fault. I commented in the wrong place.

If the path is being checked from within the container, using `/data` is correct, since that path is mounted into the container. If the test were verifying the COS path from the host side, then testing the full path would be better for visibility.

After looking at your tests again, I see that they are always testing the container path, so there’s no need to worry about the full COS path.

marceloamaral (commented):
LGTM

gateway/core/services/runners/fleets_runner.py
ItamarGoldman: This is irrelevant as the logic changed after introducing `build_job_paths` function.

ElePT (commented):
Hi @ItamarGoldman!!! The content of the PR looks good but I am missing one thing before approving, can you confirm where the RESULTS_PATH is being built for the fleets runner? I only see the ray path in the tests and source code.

ItamarGoldman:
> Hi @ItamarGoldman!!! The content of the PR looks good but I am missing one thing before approving, can you confirm where the RESULTS_PATH is being built for the fleets runner? I only see the ray path in the tests and source code.

For RAY it is easy to see that is is defined in `gateway/api/utils::build_env_variables`  ([source](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/api/utils.py#L113-L123)).
in the file `gateway/core/ibm_cloud/code_engine/fleets.utils` we can see the fleets assignment ([source](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/ibm_cloud/code_engine/fleets/utils.py#L181)):
```python
env.update(
        {
            "PUBLIC_LOG_PATH": paths.container_public_log_path,
            "ARGUMENTS_PATH": paths.container_arguments_path,
            "RESULTS_PATH": paths.container_result_path,
            "LOG_FLUSH_INTERVAL_SECONDS": str(flush_interval),
        }
    )
```
where paths is built using @avilches function `build_job_paths` in the same file ([source](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/ibm_cloud/code_engine/fleets/utils.py#L292-L296)).

Is this what you were looking for? @ElePT

ElePT (approved):
Great, so it's already implemented. With all the moving pieces I prefer to check. In that case, LGTM!

PR #2125, opened 2026-05-14
Scheduler: new task to submit Fleets jobs

avilches:
## Summary

The scheduler previously used a single `ScheduleQueuedJobs` task and a generic `execute_job` function for all runners. This PR separates Ray and Fleets scheduling end-to-end so each runner has its own task, its own execution function, and its own persistence logic.

## Comments and details

- `ScheduleQueuedJobs` is now `ScheduleRayJobs`, it filter jobs by runner=Ray.
- New `ScheduleFleetsJobs` fleets-only task to submit jobs.


gateway/scheduler/tasks/schedule_queued_jobs.py
```diff
@@ -121,7 +97,7 @@ def _schedule_jobs_if_slots_available(
                 if job.status == Job.PENDING:
                     self.add_queue_wait_time_metric(job)
 
-                logger.warning(
+                logger.info(
```
ElePT, line 100: Nice

gateway/scheduler/main.py
```diff
@@ -36,9 +37,11 @@ def __init__(self, metrics: SchedulerMetrics):
 
         self.tasks = [
             UpdateJobStatusCounts(self.kill_signal, self.metrics),
-            ScheduleQueuedJobs(self.kill_signal, self.metrics),
+            # submit jobs, status change from PENDING to QUEUED/FAILED
```
ElePT: The jobs don't change from pending to queued, right? it would be the other way round

gateway/scheduler/schedule.py
```diff
@@ -51,6 +53,54 @@ def execute_job(job: Job) -> Job:
     return job
 
 
+def execute_fleets(job: Job, ctx) -> Job:
```
ElePT: For consistency with the other function I would say:

```suggestion
def execute_fleets_job(job: Job, ctx) -> Job:
```

ElePT (commented):
LGTM! I just had a small comment on a comment and a renaming suggestion.

gateway/scheduler/main.py
avilches: submit QUEUED -> PENDING, fixed!

gateway/scheduler/schedule.py
avilches: Done!

ElePT (approved):
LGTM. Let's test it out in staging.

PR #2126, opened 2026-05-15
Fix COS endpoint URL: use direct instead of private

ElePT:
### Summary
The `COS_URL_TEMPLATE` was using the `s3.private` endpoint, which is incompatible with our network policies configured for the `s3.direct` endpoint. This caused fleet jobs to fail on staging after the latest merge from main.


ItamarGoldman (approved):
LGTM!

PR #2127, opened 2026-05-15
Configure connection params for telemetry in staging

ismaelRozasRamallal:
### Summary
- Replaces the split openTelemetry/openTelemetryCollector/tracing Helm values with a single unified tracing block consumed by all three templates (gateway, scheduler, Ray cluster).
- Fixes bool-to-string rendering, removes the console exporter from production, corrects the scheduler service name, and wires agentUrl directly as a full URL without host:port concatenation.

### Details and comments
This PR unifies the OpenTelemetry tracing configuration for the gateway Helm chart under a single top-level tracing values block, replacing the previous split between application.ray.openTelemetry (a boolean flag), application.ray.openTelemetryCollector (an object with host, port, insecure, enabled, and local fields), and a partial tracing block that had been introduced inconsistently. The old structure required environment-specific overrides to juggle two separate namespaces and used a host:port string concatenation in the template that was incompatible with full collector URLs including paths (e.g. http://collector:4318/v1/traces). The new tracing block consolidates all collector configuration — enabled, agentUrl, insecure, metricsPort, and a ray sub-object (ray.enabled, ray.local) — into one place that all three templates (deployment.yaml for the gateway container, deployment.yaml for the scheduler container, and rayclustertemplate.yaml for Ray worker nodes) read from exclusively.

  On the template side, OTEL_ENABLED and OTEL_EXPORTER_OTLP_TRACES_INSECURE are now rendered using Helm's ternary filter to correctly convert boolean values to the "1"/"0" strings the Python SDK expects, rather than quoting raw integers. OTEL_EXPORTER_OTLP_TRACES_ENDPOINT is set directly from tracing.agentUrl with no concatenation. OTEL_TRACES_EXPORTER is changed from console,otlp to otlp only, removing the console exporter that was adding trace noise to production stdout. A METRICS_PORT environment variable is introduced, following the same convention used across other platform services, wired to tracing.metricsPort. The scheduler container's OTEL_SERVICE_NAME is corrected from "Gateway" to "Scheduler" so spans from both processes are distinguishable in the collector. The staging values file (values-non-prod.yaml) is updated with the full tracing block pointing at the in-cluster OTLP HTTP collector endpoint, with enabled: true and insecure: true. Local development is unaffected — docker-compose sets OTEL_EXPORTER_OTLP_TRACES_ENDPOINT directly and does not go through Helm.

🔧 Related #1380


Tansito (approved):
LGTM Isma thank you! 🫡

PR #2128, opened 2026-05-15
Add pyproject.toml to gateway and tests

avilches:
#### Summary

Add minimal `pyproject.toml` files to `gateway/` and `tests/` to declare them as independent Python subprojects.

#### Comments and details

`gateway/` and `tests/` are already independent Python projects (each has its own virtual environment, dependencies, and tox configuration...) but nothing in the repository declared them as such. A `pyproject.toml` is the standard marker for a Python subproject boundary, and it is recognized by VSCode, PyCharm, linters, type checkers, and other tooling to automatically scope interpreter resolution to the correct subdirectory.

Both files are intentionally minimal (`name`, `version`, `requires-python` only). The existing development workflow is completely unaffected: tox continues to use `tox.ini` and `requirements.txt` as before. The `skipsdist = true` setting in both `tox.ini` files ensures tox does not attempt to install the package even with `pyproject.toml` present.

### DIFF
--- a/gateway/core/ibm_cloud/code_engine/ce_client/rest.py
+++ b/gateway/core/ibm_cloud/code_engine/ce_client/rest.py
@@ -104,7 +104,7 @@ class RESTClientObject(object):
                 cert_file=configuration.cert_file,
                 key_file=configuration.key_file,
                 proxy_url=configuration.proxy,
-                **addition_pool_args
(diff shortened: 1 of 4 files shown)


Tansito (approved):
Initially LGTM but I would also like a double check from @ElePT just in case 🫡

PR #2130, opened 2026-05-16
Add fleets integration tests with mock CE and MinIO

marceloamaral:
### Summary

- Add mock layer (`fleets_mock.py`) that patches IBM Cloud APIs with MinIO-backed dispatch                                                                        
- Add Docker Compose test infrastructure with gateway, scheduler, MinIO, and fleet-worker                                                                     
- Add 4 integration tests covering job submit, status, results, logs, upload, and failure path                                                                    
- Add CI workflow for automated fleet integration tests

### Details and comments

- Mock installs 9 patches at boot when `FLEETS_MOCK_ENABLED=1`                   
- Fleet-worker uses s3fs FUSE mounts to simulate CE volume mounts
- Status flow: worker writes to `task-store-bucket`, mock reads from it          
- Tests run via `tox -e py311-fleets`

### How to run

```
# 1. Start the stack
docker compose -f docker-compose-fleets-test.yaml up -d

# 2. Wait for readiness
curl --retry 30 --retry-delay 2 --retry-all-errors --fail http://localhost:8000/liveness/
timeout 60 bash -c 'until docker compose -f docker-compose-fleets-test.yaml logs fleet-worker | grep -q "s3fs mounts ready"; do sleep 2; done'

# 3. Provision CE project
docker compose -f docker-compose-fleets-test.yaml exec -T gateway python manage.py sync_ce_project

# 4. Run tests
tox -c tests/tox.ini -e py311-fleets
```


gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+    return os.environ["MINIO_ACCESS_KEY"]
+
+
+def _minio_secret_key() -> str:
+    return os.environ["MINIO_SECRET_KEY"]
+
+
+def _make_mock_s3_client():
+    """Create an ibm_boto3 S3 client pointing at MinIO for fleet-state operations."""
+    from ibm_boto3 import client as ibm_boto3_client  # pylint: disable=import-outside-toplevel
```
ElePT: Are these imports outside of the top level required? this shouldn't cause any conflict.

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+        """Return the fake JWT token."""
+        return self._token
+
+
+class _FakeIAMAuthenticator:
+    """Drop-in replacement for ``ibm_cloud_sdk_core.authenticators.IAMAuthenticator``.
+
+    Accepts both ``apikey`` (keyword used by the real SDK and
+    ``ibm_quantum_platform.py``) and a positional first arg (used by
+    ``IBMCloudClientProvider`` which passes the key positionally).
```
ElePT: Did you take into account any possible unwanted effects of patching `ibm_quantum_platform.py`? Right now, the `IBMQuantumPlatform` class is only used by the `AuthenticationUseCase` when we aren't doing local testing. For local testing we don't mock the IAM level but we bypass the authentication altogether. So I guess that whether it's patched or will not have a high impact in practice. I just want to make sure this is intentional/accounted for.

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+
+1. ``core.ibm_cloud.clients.IAMAuthenticator`` -- fake authenticator
+2. ``api.services.authentication.ibm_quantum_platform.IAMAuthenticator`` -- same, second import site
+3. ``IBMCloudClientProvider.get_cos_hmac_client`` -- MinIO-backed S3 client
+4. ``core.ibm_cloud.get_cos_client`` -- returns a JobCOS wired to MinIO
+5. ``core.services.runners.fleets_runner.get_cos_client`` -- same, bound import in runner
+6. ``FleetHandler.submit_job`` -- writes manifest to MinIO ``fleet-state/``
+7. ``FleetHandler.get_job_status`` -- reads status from MinIO ``task-store-bucket/``
+8. ``FleetHandler.cancel_job`` -- no-op
+9. ``FleetHandler.delete_job`` -- no-op
```
ElePT: I am not a big fan of these summaries at the beginning of the file, I think it's better to have each patch properly documented or a summary of the patches in the docstring of `install_mocks`. But I don't feel super strongly about this one in particular, I just think we should be weary of overcrowding the code with documentation we wouldn't naturally include but now is cheap to generate.

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+    """Bucket where the worker writes ``<fleet_id>.status``.
+
+    Mirrors CE's persistent task state. Sourced from the same env var the
+    gateway sets (``CE_COS_BUCKET_TASK_STORE_NAME``) so the compose file is
+    the single source of truth.
+    """
+    return os.environ["CE_COS_BUCKET_TASK_STORE_NAME"]
+
+
+_patches: list = []
```
ElePT, line 56: this is defined on line 62 but used for the first time on line 312. I think it should be moved to line ~304 or group all global variables at the beginning of the file (same for _MOCK_S3 below). I think the last one is probably better.

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def _task_store_bucket() -> str:
+    """Bucket where the worker writes ``<fleet_id>.status``.
+
+    Mirrors CE's persistent task state. Sourced from the same env var the
+    gateway sets (``CE_COS_BUCKET_TASK_STORE_NAME``) so the compose file is
+    the single source of truth.
+    """
+    return os.environ["CE_COS_BUCKET_TASK_STORE_NAME"]
```
ElePT: In general I find that these are defined in an unnatural order which makes the file difficult to follow, could you reorder them so they more or less get defined before they are used?

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+
+_MOCK_S3: tuple[int, object] | None = None
+
+
+def _get_mock_s3():
+    """Return a per-process MinIO S3 client for fleet-state operations.
+
+    Keyed on pid so gunicorn forks don't inherit the parent's socket.
+    """
+    global _MOCK_S3  # pylint: disable=global-statement
```
ElePT, line 115: curious about this one, isn't it already defined as a global variable before?

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,369 @@
...
+
+    def authenticate(self, req):
+        """No-op authentication."""
+
+
+def _mock_get_cos_hmac_client(
+    self, *, access_key_id, secret_access_key, bucket_region=None, endpoint_url=None  # pylint: disable=unused-argument
+):
+    """Return an ibm_boto3 S3 client pointing at MinIO for user/provider data buckets."""
+    from ibm_boto3 import client as ibm_boto3_client  # pylint: disable=import-outside-toplevel
```
ElePT: same comment about the imports (applies in general, only do the import outside toplevel if it's strictly required to avoid circular imports or has a good justification).

gateway/core/apps.py
```diff
@@ -25,6 +26,14 @@ def ready(self):
         # Check if all models has the "api_" prefix
         register(Tags.models)(check_model_labels)
 
+        mock_enabled = os.environ.get("FLEETS_MOCK_ENABLED") == "1"
+        if (settings.IS_GATEWAY or settings.IS_SCHEDULER) and mock_enabled:
+            from core.services.runners.fleets_mock import (  # pylint: disable=import-outside-toplevel
```
ElePT: same comment about imports.

tests/fleets/fleet-worker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.11-slim
```
ElePT, line 13: How are the tests passing the serverless import with python:3.11-slim? Side comment: I think that we should establish a base image for fleets runs and start replacing all the default image references with it

tests/fleets/source_files/other-file-by-user.json
```diff
@@ -0,0 +1 @@
+{"description": "Sample user data file for testing artifact upload"}
```
ElePT: not used

ElePT (commented):
Thanks a lot @marceloamaral!! I have just done a superficial pass at the code and tried the integration tests locally. I need to think through a couple of points but it looks like a good first approach.

gateway/core/services/runners/fleets_mock.py
marceloamaral: This patch is more defensive than actually required for the current test path.

With `SETTINGS_AUTH_MECHANISM=mock_token`, `IBMQuantumPlatform` should never be called, because auth is fully bypassed as you said. So, in the current test, this patch is not really triggered.

The reason I added it is mostly as a safety guard. If a future change, or some misconfiguration, accidentally reaches that import path, the test will fail using a clear mock instead of hanging or trying to make a real network call to IAM.

It is the same idea as patching `get_cos_client` in both places: the original source module and the already-bound import inside the runner. Maybe it is not needed, but it makes the test safer against import-site issues.

There should not have any side effect, since `_FakeIAMAuthenticator` implements the same interface we need here: `validate()`, `authenticate()`, and `token_manager.get_token()`.

gateway/core/services/runners/fleets_mock.py
marceloamaral: Fair point. I'll move the patch list to `install_mocks()` where it's actionable,  and minimize the module docstring to one sentence.

gateway/core/services/runners/fleets_mock.py
marceloamaral: Good point, I grouped `_patches` and `_MOCK_S3` together at the top with the other constants.

gateway/core/services/runners/fleets_mock.py
marceloamaral: good point, so  i reordered the file into logical sections: `env helpers -> S3 client -> auth fakes -> COS mocks -> FleetHandler mocks -> install_mocks()`

Everything is now defined before its first use...

gateway/core/services/runners/fleets_mock.py
marceloamaral: Yes, `_MOCK_S3` is defined at the module level, so the `global` is needed because this function reassigns it:

```python
_MOCK_S3 = (pid, ...)
```

Without `global`, Python would treat `_MOCK_S3` as a new local variable and shadow the module-level one. That is, reading it doesn’t need `global`, but updating/reassigning it does...

tests/fleets/fleet-worker/Dockerfile
marceloamaral: So, the fleet-worker Dockerfile installs the qiskit-serverless client from the local source tree:
 ```
 line 3: COPY client/ /tmp/client/ 
 line 4: pip install /tmp/client/ boto3
 ```
 
 Then, the `entrypoint.py` uses `get_arguments` and `save_result` from the client SDK, which is available in the container at runtime.

tests/fleets/source_files/other-file-by-user.json
marceloamaral: Removed!

gateway/core/apps.py
marceloamaral: I tried moving it to the top level, but it breaks with AppRegistryNotReady.

The import chain ends up loading Django models before the app registry is ready during app startup.

Django also recommends using deferred imports inside ready for this kind of case.
https://docs.djangoproject.com/en/5.1/ref/applications/#django.apps.AppConfig.ready

So I think it should be fine to keep it as a lazy import...

gateway/core/services/runners/fleets_mock.py
ElePT: I am not a big fan of overly defensive programming. Reading your explanation, I think the tests would be cleaner without these double patches. For someone who's not familiar with the code, they may assume that the patches are being used when they're not, which makes the behavior more difficult to track and follow. Could you look into this simplification?

ElePT (commented):
The PR mostly looks good to me, I left a final comment on the defensive mocking that I think should be addressed for future maintainability, and after this I think we can merge the PR and start experimenting with it.

gateway/core/apps.py
ElePT: ok, let's keep it if there's a strong reason to.

gateway/core/services/runners/fleets_mock.py
ElePT: hey @marceloamaral! This is the review comment I mentioned yesterday. GitHub is acting a bit flaky these days so if you already addressed it and it didn't show up let me know.

marceloamaral:
@ElePT I've updated the PR, addressed your review comments and refactored the test infrastructure to align with the latest changes from main, i.e. COS-backed log storage, wrapper templates, new storage modules, etc..

gateway/core/services/runners/fleets_mock.py
```diff
@@ -168,8 +166,7 @@ def get_token(self) -> str:
 class _FakeIAMAuthenticator:
     """Drop-in replacement for ``ibm_cloud_sdk_core.authenticators.IAMAuthenticator``.
 
-    Accepts both ``apikey`` (keyword used by the real SDK and
-    ``ibm_quantum_platform.py``) and a positional first arg (used by
```
ElePT: Taking a new look at the PR and I am a bit confused, why didn't you remove the `_FakeIAMAuthenticator` class altogether @marceloamaral? I still see a module-level patch using it that probably should be removed, right? Am I missing something?

gateway/core/services/runners/fleets_mock.py
```diff
@@ -142,8 +142,6 @@ def _make_fake_jwt() -> str:
     payload_data = {
         "iam_id": "iam-mock-fleets-test",
         "account": {"bss": "mock-account-id"},
-        "sub": "mock-subject",
-        "iss": "https://iam.mock.cloud.ibm.com",
```
ElePT: out of curiosity, why remove this?

gateway/main/settings.py
```diff
@@ -282,6 +282,7 @@
 # Fleets COS settings
 CE_COS_INSTANCE_NAME = os.environ.get("CE_COS_INSTANCE_NAME")
 CE_COS_KEY_NAME = os.environ.get("CE_COS_KEY_NAME")
+CE_COS_BUCKET_TASK_STORE_NAME = os.environ.get("CE_COS_BUCKET_TASK_STORE_NAME")
```
ElePT: This is one of the bugs I found too in my tests.

docker-compose-fleets-test.yaml
```diff
@@ -29,6 +29,7 @@ x-gateway-env: &gateway-env
   CE_COS_BUCKET_TASK_STORE_NAME: task-store-bucket
   CE_COS_INSTANCE_NAME: test-cos-instance
   CE_COS_KEY_NAME: test-cos-key
+  FLEETS_GATEWAY_HOST: http://gateway:8000
```
ElePT, line 26: One question, fleets wouldn't know how to access `http://gateway:8000` right? is this change intended to fix the mock or the actual deployment?

ElePT:
I would really appreciate it if @korgan00 can take a look at the PR too, he's got a lot of experience with integration tests refactoring and I think can have a good perspective on it.

gateway/core/services/runners/fleets_mock.py
ElePT: I was indeed missing something, we are still using the `IAMAuthenticator` for fleets so we still need the mock :thumbsup:

tests/fleets/fleet-worker/worker.py
```diff
@@ -0,0 +1,342 @@
+"""Fleet worker that polls for job manifests in MinIO and executes them."""
```
korgan00, line 13: Talking with @ElePT we understand that this is an S3 client and also run functions right?
May be we can try to split this file into 2 or 3 trying to separate responsibilities because the code  is a bit complicated to follow.

gateway/core/services/runners/fleets_mock.py
```diff
@@ -0,0 +1,391 @@
+# This code is part of a Qiskit project.
```
korgan00, line 1: Since this is test only code, we can put this code inside a test folder: `‎gateway/core/services/runners/test/fleets_mock.py` and we can not include it in the production deployment at some point to prevent having test code in production.

This kind of similar as web frameworks use to work. I know we are not following that convention but I think is better to mix conventions than have a mock file in a production folder without additional context.

korgan00 (dismissed):
Awesome work, I have a couple of comments that we can discuss

gateway/core/services/runners/fleets_mock.py
marceloamaral: Good point, but the `_FakeIAMAuthenticator` patch is still needed.

After the COS storage refactor in `main`, `_mock_get_cos_client` now creates an `IBMCloudClientProvider(api_key=..., region=...)` so it can build a real `COSClient` pointing at MinIO. That constructor calls `IAMAuthenticator(api_key, url=...)` internally and tries to fetch a token, so without the patch it would end up calling the real IAM service.

So I kept the fake authenticator, but removed the second patch for `ibm_quantum_platform.IAMAuthenticator`, since that module does not exist anymore in `main`.

gateway/core/services/runners/fleets_mock.py
marceloamaral: So, in fact, `IBMCloudClientProvider` only reads `iam_id` and `account.bss` from the JWT payload. The `sub` and `iss` fields were not being used anywhere, so I removed them to keep the fake JWT minimal...
But we can add it back in the future if needed...

docker-compose-fleets-test.yaml
marceloamaral: Good catch, the `FLEETS_GATEWAY_HOST` should be production/staging setting to define something different from `SITE_HOST`...

The actual issue was that `SITE_HOST` was only set in the `gateway` service, not in the shared `x-gateway-env` in my `docker-compose-fleets-test.yaml`, so the scheduler was defaulting to `localhost:8001`.

I fixed it in `docker-compose-fleets-test.yaml`

tests/fleets/fleet-worker/worker.py
marceloamaral: Good point!

I’ll split it into two files with different classes:

```text
storage.py
- StorageManager: S3 client, s3fs mounts, symlinks, and visibility checks

worker.py
- FleetWorker: poll loop and manifest execution
- receives StorageManager as a dependency
```

That should keep the storage logic separate from the worker flow and make the code easier to follow.

gateway/core/services/runners/fleets_mock.py
marceloamaral: Yeah, we discussed this at the beginning of the integration test proposal.

The reason I didn’t move it under `tests/` is that `.dockerignore` excludes `tests/` from the production image, which is used for the integration test. So if the mock lives there, it won’t be importable at runtime when the integration-test compose stack sets `FLEETS_MOCK_ENABLED=1`.

`apps.py:ready()` imports it conditionally at boot, and the env gate already prevents it from running in production:

```python
FLEETS_MOCK_ENABLED == "1"
```

So we could keeping it as it is now but add a good explanation in the file for now...
But I’m open to other suggestions if you have a preferred way...

marceloamaral:
The integration test is failing because it need the PR https://github.com/Qiskit/qiskit-serverless/pull/2185

tests/fleets/fleet-worker/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+RUN apt-get update && apt-get install -y s3fs fuse && rm -rf /var/lib/apt/lists/*
+COPY client/ /tmp/client/
+RUN pip install /tmp/client/ boto3
+COPY tests/fleets/fleet-worker/storage.py /app/storage.py
+COPY tests/fleets/fleet-worker/worker.py /app/worker.py
+COPY tests/fleets/fleet-worker/runner.py /runner/runner.py
+ENV PYTHONPATH=/runner
+WORKDIR /app
+CMD ["python", "worker.py"]
```
ElePT, line 22: This might be very naive but I am trying to work out the flow of information in these tests and it's kind of tricky. One question, when is the image defined in this dockerfile used? I see that it's built in the docker-compose-fleets-test, but then the image isn't referenced anywhere as far as I can see. The provider image tests below use `python:3.11-slim` directly in the image field.

tests/fleets/test_fleets_jobs.py
```diff
@@ -0,0 +1,309 @@
...
+        registers the function with image reference, and users run it. The gateway renders
+        main.tmpl (which imports Runner from the image) and uploads it + the wrapper to COS
+        at submission time. The fleet-worker mock image includes the Runner package.
+        """
+        fn = QiskitFunction(
+            title=unique_title,
+            provider=test_provider,
+            image="python:3.11-slim",
+            runner="fleets",
+        )
```
ElePT, line 220: I asked above, but is the fleet-worker mock image actually used here? how?

tests/fleets/fleet-worker/runner.py
```diff
@@ -0,0 +1,45 @@
...
+            arguments: Dict of job arguments from get_arguments().
+
+        Returns:
+            Result dict.
+        """
+        name = arguments.get("name", "world")
+        print("[public] Hello from fleets!", flush=True)
+        print(f"[public] Name: {name}", flush=True)
+        print("Processing internally", flush=True)
+        print("[public] Done", flush=True)
```
ElePT: Same question as above, I don't see any tests checking this code, and I am not sure if it's actually being run.

ElePT (commented):
Thanks for the updates! I see that you addressed our earlier review comments but still have a couple of questions about how the code is used. I hope I didn't miss anything obvious.

tests/fleets/fleet-worker/Dockerfile
marceloamaral: Thanks @ElePT for the review!

Alright, The `fleet-worker` is a background service in `docker-compose`; it is not referenced directly by the tests. It polls the `fleet-state` MinIO bucket every second for job manifests to `simulate` an execution...

The flow is:

1. The test submits a job through the gateway HTTP API.
2. `fleets_mock.py` writes a manifest to the `fleet-state` bucket instead of calling real Code Engine.
3. `fleet-worker` picks up the manifest, executes the job (within the fleet-worker container), and writes the results, logs, and status keys to COS.
4. The scheduler polls the status keys and updates the database.
5. The test polls `job.status()` until the job reaches a terminal state.

The `python:3.11-slim` image in the provider test is only the program image reference stored in the database. It is never actually pulled in mock mode. The `fleet-worker` executes all jobs, including both custom and provider jobs.

tests/fleets/test_fleets_jobs.py
marceloamaral: Just for the database, but this image is not used

tests/fleets/fleet-worker/runner.py
marceloamaral: This is executed by the provider function, the wrapper of the provider always calls this file.

avilches (dismissed):

avilches:
Minor comments:
1 Extract the queries in some helpers to make the tests less "noisy". 
2 The get_cos_client patch is done multiple times, can we try to do it only once? Maybe the get_cos_client func can just call to another internal slef._create_client(). Since this private func is never imported, you can patch it in one shot, instead of patching all the imported places...
3 You can use the get_logger() and get_provider_logger(), they will add the [PUBLIC] and [PRIVATE] prefiexes

avilches:
The get_cos_client mock is not patching the
- `core.services.storage.file_storage_fleets`
- `core.services.storage.job_file_explorer`

I know they are not used in these tests now, but in the future they will not work. That will be solved if we manage to patch everything in a single place just once.

avilches:
The real `FleetHandler.get_job_status` raises `ApiException(404)` for a missing fleet, and
`FleetsRunner.is_active()` returns `False` on that 404. The mock catches `BotoClientError` and
returns `status="pending"`, never raising, so `is_active()` always returns `True`. This ok in this PR's happy path, but it is a behavior different that would affect any future test relying on orphan/inactive-fleet detection.

avilches:
- **`tests/fleets/fleet-worker/worker.py:1353` and `:1439`** — the expression
  `status = "succeeded" if exit_code == 0 else "failed"` is duplicated in `_process_manifest`
  and `_report_status`; they can drift if a third status is added. Compute once and pass it
  down.
- **Three MinIO S3 client factories** (`storage.py`, `conftest.py`, `fleets_mock.py`) with the
  same endpoint/keys wiring and slightly different signatures (one sets `region_name`, another
  `use_path_request_style`). A connection-parameter change means editing 3 places; at minimum
  `worker.py` / `conftest.py` (both boto3) could share a helper.

avilches:
`test_cancel_running_job` can be flaky... proposing a fix

I think this test has a timing race. The RUNNING window is only around 4s (2s `FLEET_WORKER_STARTUP_DELAY_SEC` + 2s entrypoint sleep + 2s `_cleanup`), but the cancel has to propagate test -> gateway -> scheduler -> mock -> COS. If the job finishes before `stop()` re-reads the status, `_mock_get_job_status` matches `/succeeded/` before `/running/`, so `FleetsRunner.stop()` (which only cancels when the status is `running`/`pending`) becomes a no-op: no `canceled` key is written and the job terminates `DONE`, failing the `CANCELED` assertion. It rarely fires locally but it's not robustly guarded.

I don't think we should reorder the status priority, `succeeded` winning over `canceled` is the correct semantics when a job genuinely finished before the cancel was processed. The real fix is to make sure the cancel is processed **while the job is still running**.

Proposed fix:

1. Add a dedicated long-running entrypoint for this test (a sleep loop of around 120s)  so the job stays RUNNING well beyond the cancel propagation latency. The test  then reliably observes RUNNING and `stop()` always sees a cancellable state.

2. Make the worker poll for the `canceled` key during execution and terminate the subprocess when it sees it (switch `subprocess.run` → `Popen` + a wait/poll loop). This mirrors real Code Engine, which kills the workload on cancel, and it also prevents the worker from blocking its sequential poll loop on a long-running orphaned process.

As a bonus, (2) also addresses the related issue that the worker currently only checks for cancellation *after* the command has finished — with a long-running real job, cancel wouldn't be enforced until the command completed on its own.

ElePT (commented):
Hey @marceloamaral, fyi, CI is breaking.

marceloamaral:
@ElePT I fixed the issues and also added the documentation as you asked before ;)

ElePT (approved):
Alright, I think it's time to release these tests into the wild. LGTM

PR #2132, opened 2026-05-18
Add secret file

ItamarGoldman:
### Summary

Adding secret file for the public repo.


ElePT:
@ItamarGoldman this file is used in internal scans and doesn't really have a purpose in the public repo. I suggest that you close the PR.

PR #2133, opened 2026-05-19
Fleets runner fixes and API field additions from mvp branch

ElePT:
### Summary
This PR ports over the missing diff from the internal mvp branch (besides settings, helm charts, and the docker-compose-dev file). These include:

- Removes the dedicated `provider_logs` mount. Logs are now written under `provider_mount_path/jobs/<job_id>`, giving job-level isolation without the extra mount.
- Add `FLEETS_GATEWAY_HOST` setting: Decouples the gateway host used by Fleets jobs from `SITE_HOST`, allowing independent configuration per environment. Injected as `ENV_JOB_GATEWAY_HOST` into the job container. This fixes the use of save_result @marceloamaral.
- Adds a quick fix to avoid `ARGUMENTS_PATH` from gateway env vars passed twice to Fleets jobs (**this should be addressed in a proper refactoring later** cc: @marceloamaral).
- Expose `fleet_id` and runner in the API: Adds `fleet_id` to job retrieve serializers and runner to the program serializer (v1), making these fields visible to API consumers.
- And adds error event logging for Ray job submission failures.

### Details and comments
I am looking into adding a couple of new tests to verify these changes.


gateway/scheduler/schedule.py
```diff
@@ -48,6 +48,16 @@ def execute_ray_job(job: Job) -> Job:
...
+            job.save()
+            JobEvent.objects.add_error_event(
+                job_id=job.id,
+                origin=JobEventOrigin.SCHEDULER,
+                context=JobEventContext.SCHEDULE_JOBS,
+                code="SUBMISSION_ERROR",
+                message=str(ex),
+                exception=type(ex.__cause__ or ex).__name__,
+                args={},
+            )
```
ElePT: @avilches this was useful at some point during debugging but I am not sure if it belongs in main. What do you think?

avilches: There is a save later that saves only the necessary fields, so think you don't need the save, and we can get rid of the job.logs += of course...

Interestingly, we never call add_error_event; we let the users make this call when they want (and the function fails). For errors on the scheduler side, we just store a STATUS_CHANGE event which is done after the save:

https://github.com/Qiskit/qiskit-serverless/blob/6e819a5049e0503e8d872c3c7d1f2044d70ffe66/gateway/scheduler/tasks/schedule_ray_jobs.py#L80-L93

Is there any reason you wanted or needed to this?

avilches: My point is, unless you really need to add an error event in addition to the state change event to FAILED, all this code is unnecessary and you can remove it.

gateway/scheduler/schedule.py
```diff
@@ -48,6 +48,16 @@ def execute_ray_job(job: Job) -> Job:
             logger.error("job_id=%s error=%s Job set as FAILED: compute resource or submission error", job.id, ex)
             job.status = Job.FAILED
             job.logs += f"\nJob submission failed: {ex}"
```
avilches:
```suggestion
```

Dead code, it does absolutely nothing because the logs field is not saved.

gateway/scheduler/schedule.py
marceloamaral: So, would the user try to get the error not only the status?
I think this could be a possible scenario....

marceloamaral:
Looks good to me, we can move forward after addressing Alberto's comment.

Also, after this PR, I'll remove `ARGUMENTS_PATH` from `build_env_variables()` and let each runner set it independently in its submit flow.

gateway/scheduler/schedule.py
avilches: This is interesting. Now they can't: scheduler errors don't store error events. But this is something we can start doing, but then we will need to define our own set of error codes to avoid collision with the provider error codes. /cc @Tansito @pacomf @HuangJunye

ElePT: It's an interesting discussion, but I think this deserves to be its own effort instead of added as part of a small PR like this one, so I will revert the job status event change in this PR and make a note to discuss whether it makes sense to store scheduler errors as error events too (because what if there isn't an associated job? we've talked about this before, I think it can be tricky).

gateway/scheduler/schedule.py
ElePT: I removed this and the assertion in the unit test, as if this is not saved we would never be in a position to access this message.

marceloamaral (commented):
Looks good to me!
Let's wait @avilches final review

marceloamaral (approved):

ElePT:
Merging in the interest of time

PR #2134, opened 2026-05-19
Enable Instances auth: provider logs endpoint

avilches:
## Summary

Part of the Runtime Instances authorization epic. Enabled auth in the provider file endpoint `provider/logs`. It was not yet using the `accessible_functions` token from the Runtime API, falling back exclusively to legacy Django group checks.


Tansito (approved):
Reviewed and everything correct 👍

PR #2135, opened 2026-05-19
Arguments storage fleets

Tansito:
### Summary

Arguments persistence for Fleets jobs was happening inside `FleetsRunner.submit()` at run time, which is the wrong place — it should happen at job-creation time via the `get_arguments_storage` factory, consistent with how Ray jobs work.

- Implement **FleetsArgumentsStorage**: COS-backed `save()` and `get()` with the arguments key built from the job's author, provider, program title, and job ID. Raises loudly if the project or bucket is missing, since a Fleets job always has a `CodeEngineProject` assigned by serializer time.

- Wire the factory: `get_arguments_storage` now routes `Program.FLEETS` to `FleetsArgumentsStorage` instead of the `RayArgumentsStorage` placeholder.

- Remove upload logic from FleetsRunner: `_upload_arguments_to_cos` and its call in `submit()` are deleted — the runner no longer owns this responsibility.

- Add TestUtils.get_or_create_ce_project(): Reusable helper following the existing `get_or_create_*` convention, replacing repeated inline `CodeEngineProject` creation across test files.

Co-authored-by: @korgan00

### DIFF
--- a/gateway/core/services/storage/arguments_storage_fleets.py
+++ b/gateway/core/services/storage/arguments_storage_fleets.py
@@ -25,6 +25,7 @@ class FleetsArgumentsStorage(ArgumentsStorage):
             raise ValueError(f"Job '{job.id}' has no CodeEngineProject assigned")
 
         self._job_id = str(job.id)
+        self._user_id = job.author.id
         self._project = job.code_engine_project
 
         bucket = job.code_engine_project.cos_bucket_user_data_name
@@ -57,7 +58,8 @@ class FleetsArgumentsStorage(ArgumentsStorage):
             key=self._arguments_key,
         )
         logger.info(
-            "[save] job_id=%s bucket=%s key=%s Arguments saved to COS",
+            "[save] user_id=%s job_id=%s bucket=%s key=%s Arguments saved to COS",
+            self._user_id,
             self._job_id,
             self._bucket,
             self._arguments_key,
@@ -68,5 +70,10 @@ class FleetsArgumentsStorage(ArgumentsStorage):
             data = self._get_cos().get_object_bytes(bucket_name=self._bucket, key=self._arguments_key)
             return data.decode("utf-8") if data else None
         except Exception:  # pylint: disable=broad-exception-caught
-            logger.warning("[get] job_id=%s Could not retrieve arguments from COS", self._job_id)
+            logger.warning(
+                "[get] user_id=%s job_id=%s key=%s Could not retrieve arguments from COS",
+                self._user_id,
+                self._job_id,
+                self._arguments_key,
+            )
             return None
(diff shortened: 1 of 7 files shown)


gateway/core/services/storage/arguments_storage_fleets.py
```diff
@@ -1,19 +1,72 @@
...
+
         self._job_id = str(job.id)
+        self._project = job.code_engine_project
 
-    def get(self) -> Optional[str]:
-        raise NotImplementedError
+        bucket = job.code_engine_project.cos_bucket_user_data_name
+        if not bucket:
+            raise ValueError(
+                f"CodeEngineProject '{self._project.project_name}' has no cos_bucket_user_data_name configured"
```
Tansito, line 31: @avilches what format were we using for errors and logs? Just to ensure I use the same as you implemented in other parts.

gateway/tests/core/services/storage/test_arguments_storage_fleets.py
```diff
@@ -0,0 +1,68 @@
...
+from tests.utils import TestUtils
+
+
+@pytest.mark.django_db
+class TestFleetsArgumentsStorage:
+    """Tests for FleetsArgumentsStorage path generation."""
+
+    @pytest.fixture
+    def ce_project(self):
+        return TestUtils.get_or_create_ce_project(
```
Tansito, line 28: I also created a TestUtil for CEProject entity for future reuse following the pattern that @ItamarGoldman and @avilches created.

gateway/core/services/storage/arguments_storage_fleets.py
```diff
@@ -1,19 +1,72 @@
...
+            key=self._arguments_key,
+        )
+        logger.info(
+            "Arguments for job [%s] saved to COS at %s/%s",
+            self._job_id,
+            self._bucket,
+            self._arguments_key,
+        )
+
+    def get(self) -> Optional[str]:
```
Tansito: Happy to hear feedback here but from what I saw we are not using the `get` method anywhere in the API or de Scheduler. I'm thinking to return a simple `NotImplemented` exception by now.

marceloamaral: I agree with raise `NotImplementedError` for `get()` since after the changes this method will not be used anywhere.

marceloamaral (dismissed):
Looks good to me!

marceloamaral (dismissed):
LGTM, the `custom_functions`/`provider_functions` path aligns with what we planned for the full COS restructure

gateway/core/services/storage/arguments_storage_fleets.py
avilches: I always try to log the job_id and the user_id: `logger.info("[xxxx] user_id=%s job_id=%s |...`

avilches (approved):

PR #2148, opened 2026-05-21
Make Serverless error_message method backward compatible.

korgan00:
### Summary

The job.error_message() function work for both old jobs, displaying the result(), and new jobs, looking for the error in the events, in the same way that result() currently works.


client/qiskit_serverless/core/job.py
```diff
@@ -237,15 +237,23 @@ def filtered_logs(self, **kwargs) -> str:
...
 
-        if isinstance(error, str):
+        error_events = self.events(type="ERROR")
+        if len(error_events) > 0:
+            error_msg_list = [format_err_event(evt) for evt in error_events]
+            return "\n\n".join(error_msg_list)
+
+        error_result = self._job_service.result(self.job_id) if self.status() == "ERROR" else ""
+
+        if isinstance(error_result, str):
```
avilches, line 250: Can the result() reuse the error_message func (or a part of it, I don't know) somehow?

client/tests/core/test_job.py
```diff
@@ -144,6 +144,7 @@ def test_error_message(self):
         client.status = MagicMock(
             return_value="ERROR",
         )
+        client.events = MagicMock(return_value=[])
```
avilches, line 147: We should have some tests where the error is obtained from the events, shouldn't we?

avilches (dismissed):

avilches (approved):
LGTM!

PR #2151, opened 2026-05-22
Logs Storage for fleets

Tansito:
#### Splitting logs storage responsibilities
`LogsStorage` is now an abstract interface with four methods (`get_public_logs`, `save_public_logs`, `get_private_logs`, `save_private_logs`). The existing logic has been moved into RayLogsStorage, and FleetsLogsStorage has been added with the read methods implemented and the write methods as `NotImplementedError` — Fleets logs are written directly by the shell wrapper during execution, not by the gateway.

#### Reading Fleets logs from COS
`FleetsLogsStorage` reads logs directly from COS using `get_cos_client`, bypassing the runner entirely. Paths follow the same convention as `FleetsArgumentsStorage`: `custom_functions` for user jobs, `provider_functions` for provider jobs. Private logs are stored under `providers/` in the provider bucket. 404/NoSuchKey errors are treated as "log not yet available" and return `None`.

#### Simplifying FleetsRunner
With log reading moved to the storage layer, `logs()` and `provider_logs()` in the runner now raise `NotImplementedError`, and the `_get_logs_from_cos` private method that backed them is removed.

#### Factory and callers
`get_logs_storage(job)` is added to `core/services/storage/__init__.py`. The four existing callers (`get_logs`, `provider_logs`, `update_ray_jobs_statuses`, `migrate_old_job_logs`) are updated to use it instead of instantiating `LogsStorage` directly.

#### Tests
The runner logs tests are removed since there is no longer any logic to test there. A new `test_logs_storage_fleets.py` is added covering path generation, successful reads, 404 handling, and the runtime guard for non-provider jobs.


avilches (approved):
After this PR, it will be nice to have a unified PathTest to assert only the all the COS paths  in ArgumentStorate, FileStorage, LogStorage and ResultStorage. I mean, one test with the same user and the same job, and call to the public keys (the path) to see if they look as we expect. Just an idea. The PR loogs good.

ElePT (approved):
Agreed with @avilches's suggestion, we can cross-checks the paths once they are all in place. Thanks a lot @Tansito for the contribution, LGTM!

marceloamaral (changes requested):
The callers in `get_logs.py:48` and `provider_logs.py:48` still do `LogsStorage(job)`, but they need to be changed to `get_logs_storage(job)` to use `FleetsLogsStorage(job)` instead of the default `LogsStorage(job)`:

That is:
- `LogsStorage(job).get_public_logs()` always reads local filesystem, and then it will return None...
- we need to have `get_logs_storage(job).get_public_logs()`, which will route to `FleetsLogsStorage` for Fleets jobs

marceloamaral (approved):
Disregard my previous comment, I was comparing against the old code in my local laptop. After pulling the latest I can see `LogsStorage` is now abstract and the callers go through `get_logs_storage(job)`. Looks good.

PR #2152, opened 2026-05-22
Fix settings configuration

Tansito:
### Summary

In #2150 @avilches and I found an error in the environment variables configuration in `settings.py`:

```py
SETTINGS_AUTH_MECHANISM = os.environ.get("SETTINGS_AUTH_MECHANISM", "custom_token")
```

Basically this configuration doesn't set the variable to the default value when the value is "". To prevent this we changed the logic to:

```py
SETTINGS_AUTH_MECHANISM = os.environ.get("SETTINGS_AUTH_MECHANISM") or "custom_token"
```

Now the `or` keyword guarantees that `None` or `""` will always get the default value.

Along with these changes I removed `QUANTUM_PLATFORM_API_BASE_URL` and `SETTINGS_TOKEN_AUTH_VERIFICATION_FIELD` as they are not being used now (this will require also an update in the internal repository).


avilches (approved):
LGTM!

PR #2153, opened 2026-05-22
Enable Instances auth: user files endpoints

avilches:
## Summary

The user-files endpoints (`v1/files/`, `v1/files/download`, `v1/files/upload`, `v1/files/delete`) previously gated access only via `RUN_PROGRAM_PERMISSION` through Django groups. With this PR, when an instance authenticates with a CRN that resolves to a non-legacy `FunctionAccessResult`, access is gated per function:


ElePT (commented):
The changes in this PR look very logical. I do wonder (without having time to check), would these changes require any change in the Files clients? will it trigger the new path without any changes (I just want to make sure that the client isn't stuck in the legacy path).

Tansito:
@avilches can confirm better than me but initially no, the logic will be checked internally and I'm not expecting any change in the client interface

avilches:
@ElePT @Tansito No need to adapt the client, the PR doesn't affect how the client or the endpoints work (only how the auth is check)

ElePT (approved):
LGTM

PR #2155, opened 2026-05-26
Add release notes for user-facing features

ElePT:
### Summary
Documents user-facing changes for the next release:

- Job timeout behaviour (upgrade): clarifies that the time limit is now enforced while a job is PENDING or RUNNING, rather than applied at queue time
- error_message fix (fixes): notes that Job.error_message() now correctly handles both the events-based error format and the legacy string format.
- Runtime compatibility (upgrade): documents the new minimum qiskit-ibm-runtime<0.48.0 requirement.
- Telemetry (other): generic note on internal observability improvements.

### DIFF
--- a/releasenotes/notes/default-time-job-in-running_status-b1a6296c58d4dcb3.yaml
+++ b/releasenotes/notes/default-time-job-in-running_status-b1a6296c58d4dcb3.yaml
@@ -1,12 +1,14 @@
 ---
 upgrade:
   - |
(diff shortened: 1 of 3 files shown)


Tansito (approved):
LGTM Elena, thank you! 😄 

Question here, should we also mention the Python 3.9 deprecation? I don't know how we usually mention this in Qiskit.

ElePT:
yes! I added the release note directly to the deprecation PR, @Tansito, that's why it's not here

PR #2156, opened 2026-05-26
Add unit tests for public/private log filtering on Fleets path

ElePT:
### Summary
This PR adds specific unit tests to confirm the logs filtering logic through the Fleets code path. The Fleets container writes all output (including [public] lines) to the provider log file in COS. `GetProviderJobLogsUseCase` then applies `filter_logs_with_non_public_tags()` before returning, so [public] lines must be absent from the result and untagged / [private] lines must be present.

The tests confirm that:

- [public] tagged lines are excluded from provider logs
- Untagged lines are included
- [private] tagged lines are included with the prefix stripped

The existing tests only covered the COS-storage path, where `get_private_logs()` returns already-filtered content and the use case returns early, so `filter_logs_with_non_public_tags()` was never exercised by the test suite.

### DIFF
--- a/gateway/tests/api/use_cases/jobs/test_provider_logs.py
+++ b/gateway/tests/api/use_cases/jobs/test_provider_logs.py
@@ -1,6 +1,6 @@
 """Unit tests for GetProviderJobLogsUseCase."""
 
-from unittest.mock import patch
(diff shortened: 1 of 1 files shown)


ElePT:
Closing, as PR got stale after other refactorings.

PR #2157, opened 2026-05-26
Configure telemetry http

ismaelRozasRamallal:
### Summary
Switch OpenTelemetry OTLP trace exporters from gRPC to HTTP across gateway, client, and Helm chart configuration.

### Details and comments
This PR migrates the OpenTelemetry tracing setup in both the gateway service and the client library from the gRPC-based OTLP exporter (opentelemetry-exporter-otlp-proto-grpc) to the HTTP-based one (opentelemetry-exporter-otlp-proto-http). The change affects the exporter import, the endpoint construction (now using port 4318 with the /v1/traces path as per the OTLP/HTTP spec), and removes the insecure=parameter which is gRPC-specific — TLS is now controlled simply by using http:// vs https:// in the endpoint URL.

The Helm chart default agentUrl has been updated from http://otel-collector:4317 to http://otel-collector:4318/v1/traces, and the OTEL_EXPORTER_OTLP_TRACES_INSECURE environment variable and its corresponding insecure values field have been removed from both the gateway and scheduler container specs. Unit tests have been added for both tracing modules (gateway/tests/test_tracing.py and client/tests/core/test_tracing.py) covering the exporter source, endpoint format, and absence of the insecure parameter.


ElePT (approved):
I am going to trust you here because my knowledge of open telemetry is very limited. LGTM

PR #2159, opened 2026-05-26
Remove JobCOS.logs

Tansito:
### Summary

As a follow-up from #2151 after some internal conversations this PR removes logic that is not being used anymore in the logic from logs flow.


marceloamaral (approved):
LGTM

PR #2160, opened 2026-05-26
Fleet jobs: implement COS-backed log endpoints

avilches:
### Summary

This PR implements the log retrieval endpoints for Fleet jobs, which use a different architecture than Ray jobs.

#### Context
There are two types of logs, based on the function type.
- `/logs`: only accesible by the **job author**:
  - For customs jobs, print all the output, removing all the prefixes `[PRIVATE]` and `[PUBLIC]`.
  - For provider jobs, print only lines starting with `[PUBLIC]` without the prefix.
- `/provider-logs`: only accesible by **provider admin** for provider jobs: include lines starting with `[PRIVATE]` without the prefix, and the 3rd party logs (lines with no prefix). The lines starting with `[PUBLIC]` are removed.

**Why Fleet needs its own path:**
- Ray jobs allow the Gateway (/log and /provider-logs endpoints) to use the runner at request time and pull real time logs from the live container (and filtering the logs before serving them).
- Fleet jobs run in IBM Code Engine containers and there is no way to get the logs in realtime. So, 
the function wrapper filters the logs (separating public from private) and periodically uploads both files to COS. By the time a user calls `/logs`, the filtering is already done and the result is sitting in an object store.

So, for Fleets, the endpoints always reads directly from COS. If the object isn't there yet (the wrapper uploads every ~15s), it returns `"No logs yet."`. There is no runner fallback and no legacy DB path for Fleet jobs.

### Details and comments

`FleetsLogsStorage` reads from COS using the same path logic (`build_cos_paths`) that the runner uses to write, ensuring the two never get out of sync. It returns `None` on `ClientError` (object not yet uploaded) rather than propagating the error. `save_*` methods raise `NotImplementedError` because the gateway never writes Fleet logs, the container does.


gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -84,11 +90,11 @@ def build_run_volume_mounts(
...
-    primary_mount_path: str,
-    primary_log_filename: str = "logs.log",
-    secondary_mount_path: str | None = None,
-    secondary_log_filename: str = "logs.log",
-    secondary_log_filter_key: str | None = None,
+    public_mount_path: str,
+    public_log_filename: str = "logs.log",
+    private_mount_path: str | None = None,
+    private_log_filename: str = "logs.log",
+    flush_interval_seconds: int = 15,
```
avilches: Do we need this to be in different variables? To me, no need to have path and file per cos. @ElePT @marceloamaral I mean, I would reduce these 4 variables to 2: public_log_path and private_log_path

marceloamaral: Right, it should be fine to have the filename only...

gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -84,72 +92,53 @@ def build_run_volume_mounts(
...
 def build_run_env_variables(
     *,
-    primary_mount_path: str,
-    primary_log_filename: str = "logs.log",
-    secondary_mount_path: str | None = None,
-    secondary_log_filename: str = "logs.log",
-    secondary_log_filter_key: str | None = None,
+    public_log_path: str,
+    private_log_path: str | None = None,
+    flush_interval_seconds: int = 15,
```
ElePT: much easier to follow now

gateway/api/use_cases/jobs/provider_logs.py
```diff
@@ -59,6 +59,11 @@ def execute(
         if logs:
             return logs
 
+        if job.program.runner == Program.FLEETS:
+            # Fleets wrapper needs up to 15 seconds to start uploading a log to COS
```
ElePT, line 63: did you measure this number or is it an approximation? just curious of how accurate it is

gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -182,37 +173,91 @@ def build_run_commands(
...
+
+    Both user bucket (public logs + data) and provider bucket (entrypoint at
+    function level, private logs at job level) are included.
+
+    Args:
+        job: Job instance with a provider assigned.
+    """
+    username = job.author.username
+    provider_name = job.program.provider.name
+    program_title = job.program.title if job.program else "unknown"
```
ElePT: is this "unknown" case possible? shouldn't all programs have a title?

gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -182,37 +173,91 @@ def build_run_commands(
...
+        "provider_function_prefix": provider_function_prefix,
+        "provider_job_prefix": provider_job_prefix,
+        "provider_log_key": f"{provider_job_prefix}/{LOG_FILENAME}",
+        "provider_mount_path": FUNCTION_MOUNT_PATH,
+    }
+
+
+def build_cos_paths(job: Job) -> dict[str, str]:
+    """Dispatcher: returns custom or provider COS paths depending on job type."""
+    if job.program and job.program.provider:
```
ElePT: same here, can there be a job without a program?

gateway/core/services/runners/fleets_runner.py
```diff
@@ -245,7 +233,7 @@ def submit(self) -> None:
                     "COS configured for job [%s]: user_key=[%s] provider_key=[%s]",
                     self.job.id,
                     paths["user_log_key"],
-                    paths["provider_log_key"],
+                    paths["provider_log_key"] or "-",
```
ElePT: is - a valid key? I am asking for my own knowledge here. If the field is optional isn't it cleaner to have a None?

gateway/main/settings.py
```diff
@@ -275,6 +275,7 @@
     "FLEETS_DEFAULT_IMAGE",
     "private.icr.io/qc-qiskit-functions-ce-staging/helloworld:latest",
 )
+FLEETS_GATEWAY_HOST = os.environ.get("FLEETS_GATEWAY_HOST", os.environ.get("SITE_HOST", ""))
```
ElePT: are gateway host and site host equivalent in all situations?

ElePT (commented):
I started a review but I know you are making changes so I will submit my comments so far and do a new review when the PR is ready.

gateway/api/use_cases/jobs/provider_logs.py
avilches: The upload has a 15s delay before upload, so the log will be ready after this seconds and the time needed to upload the cos (I guess after only 15s the log will be very small). This is the same value `flush_interval_seconds` you have seen above.

gateway/core/ibm_cloud/code_engine/fleets/utils.py
avilches: No actually, a job will always have a program! I will remove it, thanks!!!!

gateway/core/ibm_cloud/code_engine/fleets/utils.py
avilches: Changed!

gateway/main/settings.py
avilches: Removed, dead code

gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -83,92 +148,75 @@ def build_run_volume_mounts(
...
+                "name": "PRIVATE_LOG_PATH",
+                "value": paths.container_private_log_path,
+            }
         )
 
-    return run_env_variables
+    # Add the job env vars without overwriting the system vars
+    system_names = {v["name"] for v in system_vars}
+    user_vars = [e for e in (extra or []) if e["name"] not in system_names]
+    return system_vars + user_vars
```
ElePT, line 201: @marceloamaral this overlaps with your change and already adds some improvement, but I don't think it fully replaces the refactoring we discussed. It's worth still looking into adding more transparency to the env. var building steps.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -423,7 +396,7 @@ def free_resources(self) -> bool:
         # NOTE: fleet deletion disabled to preserve fleets for post-run inspection.
         # Re-enable after the demo.
         # if not self.job.fleet_id:
-        #     logger.debug("No fleet_id to clean up for job [%s]", self.job.id)
+        #     logger.debug("No fleet_id to clean up for job_id=[%s]", self.job.id)
         #     return False
         #
         # try:
```
ElePT, line 402: I am making a note of this TODO in my list of tasks so we don't forget about it.

ElePT (approved):
I verified all the paths against the agreed upon COS storage structure and couldn't find any issue. The new structure makes it much easier to check and verify things. The tests are also super clear and handy as a reference for future tests. I will not merge until the new release is out but the PR LGTM.

ElePT:
Adding it to the merge queue as the release date has been pushed.

gateway/core/ibm_cloud/code_engine/fleets/utils.py
marceloamaral: @ElePT, yes, that is why I was waiting for this PR.

gateway/core/services/runners/fleets_runner.py
avilches: Good! Because we should change this in the whole codebase...

PR #2174, opened 2026-05-27
Result storage architecture

Tansito:
### Summary

Pretty similar PR as #2149 . Here we update the arquitecture for Result Storage to follow a similar approach as Arguments and Logs and also unifying the interface to use the Job instead to send the username and the job.id later.

Implementation for fleets will come in a next PR.


Tansito:
Maintaining this PR as draft as it we don't need it for the release tomorrow.

ElePT (approved):
LGTM. The change is quite straightforward and I think it's safe to merge before the release (although I am also fine holding off for a day).

PR #2177, opened 2026-05-28
Consolidate fleets env var construction into single function

marceloamaral:
### Summary
Consolidate all container env var assembly into build_run_env_variables().

### Details and comments

- Move gateway host override and system var overlay into the builder
- Remove _build_job_env_vars() from fleets_runner.py
- Use dict.update() for transparent override (no hidden collision filtering)
- Caller does parse + decrypt, builder does the rest
- Update tests to match new signature


avilches:
LGTM!

avilches (approved):

PR #2178, opened 2026-05-28
Enable Instances auth: /run and /upload for custom functions

avilches:
## Summary

The Runtime API `/functions` endpoint now returns a `custom_functions` field with permissions to create or execute functions without a provider (also known as "custom" or "user" functions, where only the user who created them can access them)

```json
"custom_functions": {
  "permissions": ["function-custom.create", "function-custom.run"]
}
```

Previously, any authenticated user could upload and run their own custom functions unconditionally. This PR wires the new field into the gateway so that `/upload` and `/run` respect the instance-level entitlements.

- Parses `custom_functions.permissions` from the Runtime API response into `FunctionAccessResult.custom_function_permissions`
- `/upload` without provider): checks `function-custom.create` before creating the function
- `/run` without provider: checks `function-custom.run` before running the function
- Legacy authorization (`use_legacy_authorization=True`) keeps the existing behavior: always allow
- When the field is absent in the API response, access is denied (empty permission set)


Tansito (approved):
Reviewed with Alberto and fixed the feedback from the review. Ready to merge!

PR #2179, opened 2026-05-28
Limit fleet log size to FUNCTIONS_LOGS_SIZE_LIMIT

avilches:
### Summary
Fleet jobs write their logs to `/tmp` before uploading them periodically to the COS-backed mount. Without a size cap, a long-running job producing many logs could fill `/tmp` (which in Kubernetes is usually `tmpfs`, i.e. pod RAM) and also upload arbitrarily large files to COS.

This PR adds the same log size limit that non-fleet jobs already enforce via `check_logs` in `core.utils`.

### Details and comments

Two new shell functions replace the plain `cp` calls used to upload logs to COS:

- `upload_log(src, dst, [size])`: if the source exceeds `LOG_SIZE_LIMIT_BYTES`, writes a truncation header + the last N bytes to the destination (keeping the most recent output, same semantics as `check_logs`). Accepts an optional pre-computed size to avoid a redundant `stat` call.

- `truncate_local(file)`: after each successful COS upload, shrinks the local `/tmp` file in-place to the last `LOG_SIZE_LIMIT_BYTES` bytes. Keeps the same inode so awk's open write fd remains valid (`truncate -s 0` + `cat >>`). This bounds `/tmp` usage to roughly `2×` the limit rather than letting it grow for the full duration of the job.


ElePT (commented):
The PR in principle looks good but the template code is quite hard to review, I think it would make sense to wait until the integration test infra is there to add an integration test here. What do you think?

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler.py
```diff
@@ -703,13 +703,20 @@ def test_build_run_env_variables_passes_through_stored_vars():
...
 
 def test_build_run_env_variables_overlays_system_vars():
-    """build_run_env_variables includes system vars derived from paths."""
+    """build_run_env_variables injects all required system vars with correct values.
+
+    Verifies that PUBLIC_LOG_PATH, ARGUMENTS_PATH, and RESULTS_PATH come from the
+    paths object; LOG_FLUSH_INTERVAL_SECONDS defaults to 15 when not overridden in
+    settings; LOG_SIZE_LIMIT_BYTES is present and numeric; PRIVATE_LOG_PATH is absent
+    when container_private_log_path is None.
+    """
```
avilches:
```suggestion
    """build_run_env_variables includes system vars derived from paths."""
```

avilches:
> The PR in principle looks good but the template code is quite hard to review, I think it would make sense to wait until the integration test infra is there to add an integration test here. What do you think?

I added a spec file with the explanation and I covered all the edge cases with unit tests that are serverless-agnostic: they don't care (because they don't know...) about results, arguments, they just assert the logs and the exit code: filtering, truncation, the header, ... have a look, now I'm more confident about this, but of course it will need testing in staging with real functions.

Integration tests (@marceloamaral please!) should assert Serverless stuff only: that the function can read the arguments in the mounted COS, uploaded with the `ArgumentStorage` and, after it finishes, the results and the logs are in the right COS folder and they can be read with the `LogStorage` and the `ResultStorage` made by @Tansito 
If the integration test can not access/use these `*Storage` services, ... then tests can assert the folders in COS using the `build_jobs_path`... don't assert with hardcoded folders!

avilches:
Alternative PR: https://github.com/Qiskit/qiskit-serverless/pull/2181

avilches:
We all agree to use the PR #2181 instead

PR #2173, opened 2026-05-27
Refactor entrypoint upload for custom/provider functions + COS variable management for Fleets

ElePT:
### Summary

~**THIS PR IS BASED ON #2185 and is going to be kept as a draft until said PR is merged to main and this branch is rebased.**~ -> Rebased and ready for review

This PR completes provider image function support for the Fleets runner, fixing COS path layout, log capture, and result storage. Both custom (artifact-based) and provider (image-based) functions now work end-to-end, tested on a local serverless deployment that accesses CE Fleets.

#### Changes

1. **4-mount COS path layout**: Replaces the flat `/data` + `/function_data` layout with four explicit PDS mounts that follow the newly defined COS conventions:

```
  USER_BUCKET
  └── users/{user}/
      ├── custom_functions/{title}/
      │   ├── data/          → FUNCTION_USER_DATA_PATH  (entrypoint, wrapper)
      │   └── jobs/{id}/     → JOB_USER_DATA_PATH       (arguments, logs, results)
      └── provider_functions/{provider}/{title}/
          ├── data/          → FUNCTION_USER_DATA_PATH
          └── jobs/{id}/     → JOB_USER_DATA_PATH

  PROVIDER_BUCKET
  └── providers/{provider}/{title}/
      ├── data/              → FUNCTION_PROVIDER_DATA_PATH  (entrypoint, wrapper)
      └── jobs/{id}/         → JOB_PROVIDER_DATA_PATH       (private logs)
```

  Custom functions get 2 mounts; provider functions get 4. The upload dispatcher correctly populates all paths at submit time.

~2. **Arguments and Results storage fixes**: Removed the blocking `FleetsArgumentsStorage.save()` call from the serializer request handler, it was triggering a live CE secrets API call on every /run request, causing 30s timeouts. Instead, `FleetsArgumentsStorage.save()` is called during the function upload phase, and arguments are uploaded to COS together with the function entrypoint and files.~ -> this change was a bug, reverted to the original code.

3. **Test refactor**: Moved all `utils.py` tests into a new `test_fleet_utils.py`.

4. **Wrapper execution and log fix**: Two fixes that complete the log pipeline:
  - shutil.copyfile() → open()+read()+write(): sendfile() fails silently on FUSE-backed PDS mounts, causing log files to always be 0 bytes.
  - Run wrapper from PDS mount directly: build_run_commands() now returns ["python", container_docker_entrypoint] instead of a base64-encoded inline script.
 
#### Offline testing

[...]

PR #2181, opened 2026-05-30
Fleet log wrappers: Python implementation (alternative to #2179)

avilches:
### Context

This is an alternative to #2179 that implements the same fleet log functionality (size limit, periodic COS upload, log filtering) using **Python instead of bash**.

Both PRs are open as drafts for comparison. Only one should be merged.

---

### Summary

Fleet job wrappers run inside Code Engine containers and are responsible for:
- Filtering `[PUBLIC]`/`[PRIVATE]` log lines into the right COS files
- Uploading logs to COS periodically (`LOG_FLUSH_INTERVAL_SECONDS`, default 15 s)
- Capping log size at `LOG_SIZE_LIMIT_BYTES` (default 50 MB), keeping the most recent output
- Propagating the real application exit code

This PR replaces the bash+awk wrappers with Python scripts rendered from Django templates and executed via `["python3", "-c", script]`.

---

### Changes

- `gateway/templates/fleet_custom_job_wrapper.tmpl` — rewritten in Python (`subprocess.Popen` + `threading.Thread` uploader)
- `gateway/templates/fleet_provider_job_wrapper.tmpl` — rewritten in Python; eliminates the `mkfifo`/`awk`/`LOGGER_PID` pattern
- `gateway/core/ibm_cloud/code_engine/fleets/utils.py` — `build_run_commands` returns `["python3", "-c", script]` and uses `json.dumps` for the app command
- `gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_handler.py` — updated assertions (`sh` → `python3`)
- `gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_template.py` — added Docker integration tests; uses `python:3-alpine` instead of `alpine:3`
- `specs/FLEETS_LOGS_UPLOAD.md` — full specification of the upload flow, memory analysis, and Python vs bash comparison

---

#### Pros of Python

| Area | Detail |
|---|---|
| **Provider wrapper** | Eliminates `mkfifo`, background `awk`, `LOGGER_PID`, and `wait "$LOGGER_PID"`. Line routing is a plain `if/elif` in a loop. |
| **Exit code** | `proc.wait()` returns the code directly — no sidecar file, no `$?` ambiguity through a pipeline. |
| **Signal handling** | `threading.Event` stop lets the uploader finish any in-progress upload before the final flush, avoiding the double-flush on SIGTERM present in the bash version. |
| **Subprocess safety** | `subprocess.Popen(list, shell=False)` — no shell injection risk. App command passed as a JSON list. |
| **No external tools** | No `awk`, `stat`, `truncate`, `mkfifo`, `tail` needed in the container image. Only Python stdlib. |
| **Error visibility** | Python exceptions and tracebacks instead of silent `2>/dev/null` shell failures. |

#### Cons of Python

| Area | Detail |
|---|---|
| **Interpreter overhead** | `python3` wrapper process: ~25–40 MB RAM vs ~2 MB for `sh`. Negligible for Qiskit workloads (several GB), but non-zero. |
| **Startup latency** | ~50–100 ms for interpreter init before the first wrapper line runs. Negligible for jobs lasting minutes. |
| **`python3` required** | Container image must have `python3`. Production images are guaranteed (they run Python functions). Test images need `python:3-alpine` instead of `alpine:3`. |

#### Memory (both implementations)

`/tmp` in containers is `tmpfs` (RAM-backed). Both bash and Python peak at **~2 × LOG_SIZE_LIMIT_BYTES** RAM during an over-limit upload:
- bash stored the tail in a `.upload` tmpfs file
- Python stores the tail in a heap `bytes` object

Neither is cheaper in total RAM for the upload step. See `specs/FLEETS_LOGS_UPLOAD.md` §5 for the full analysis.


gateway/templates/fleet_provider_job_wrapper.py
```diff
@@ -0,0 +1,190 @@
...
+                for line in self.user_process.stdout:
+                    out, cleaned = self.clean_line(line, pub, priv)
+                    # out could be private logs or public, based on the prefix (or the absence of)
+                    # [PRIVATE] or no prefix, return priv (removing the prefix)
+                    # [PUBLIC] returns pub (removing the prefix)
+                    out.write(cleaned)
+                    out.flush()
+            code = self.user_process.wait()
+        finally:
+            self.flush()
```
marceloamaral, line 186: Before the `finally`, user process could has already flushed to COS in line 139 (due to kill or Ctrl+C), but because line 141 `sys.exit()` raises `SystemExit`, which triggers the finally block, it will flush to COS again...

To fix this, don't call `flush` or `sys.exit()` into the `on_signal()` only save the exit code there, the `finally` should call `flush` and exit with the given code or 0.

gateway/templates/fleet_custom_job_wrapper.py
```diff
@@ -0,0 +1,165 @@
...
+                # Write (upload to COS) the truncation header + the log tail
+                with open(cos_path, 'wb') as f:
+                    f.write(TRUNCATION_HEADER)
+                    f.write(log_tail)
+                self.truncate_file(temporal_log_path, log_tail)
+            except OSError:
+                pass
+        else:
+            try:
+                shutil.copy(temporal_log_path, cos_path)
```
marceloamaral: `shutil.copy` copies the file content + permissions, it is better to just use `shutil.copyfile(temporal_log_path, cos_path)`

gateway/templates/fleet_custom_job_wrapper.py
```diff
@@ -0,0 +1,165 @@
...
+            # Listen to changes in the local temporal log file
+            with open(LOCAL_PUBLIC_LOG, 'a', buffering=1) as pub:
+                # Process changes line by line
+                for line in self.user_process.stdout:
+                    # custom job logs are public, and it contains all the logs, just remove the prefixes
+                    pub.write(self.clean_line(line))
+                    pub.flush()
+            code = self.user_process.wait()
+        finally:
+            self.flush()
```
marceloamaral, line 161: Same here

gateway/templates/fleet_custom_job_wrapper.py
avilches: Changed, permissions should be no-op by the s3fuse driver, but just in case, it's safer withe copyfile!

gateway/templates/fleet_provider_job_wrapper.py
avilches: I'm afraid you are right, fixed! thanks!

specs/FLEETS_LOGS_UPLOAD.md
```diff
@@ -0,0 +1,222 @@
+# Fleet Wrapper — COS Upload Specification
```
ElePT, line 1: I like the idea of having a `specs/` folder, it makes it easier to locate the files and doesn't "contaminate" the rest of the repo.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_template.py
```diff
@@ -0,0 +1,339 @@
...
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
+"""
+Integration tests that run the fleet wrapper scripts inside a real Docker container.
```
ElePT: I am not a big fan of mixing unit and integration tests with no way of identifying which is which. I think that we should find a new location for this test.

specs/FLEETS_LOGS_UPLOAD.md
```diff
@@ -0,0 +1,222 @@
...
+
+### 1. Periodic upload to COS
+
+Logs are uploaded to COS every `LOG_FLUSH_INTERVAL_SECONDS` seconds (default: 15).
+A final unconditional upload is guaranteed when the job ends, regardless of exit code.
+The `try/finally` block and SIGTERM/SIGINT signal handlers guarantee a final flush
+so no buffered output is lost when the container terminates.
+
+Uploads are skipped when the local file has not changed size since the last cycle,
+avoiding unnecessary COS PUTs when the application is silent.
```
ElePT, line 25: nice

ElePT (commented):
Thanks a lot @avilches! The PR looks quite good to me, I see that you addressed all the feedback from @marceloamaral, and the changes make sense.  My only comment here is about mixing the integration and the unit test in the same folder. Given that we are creating a dedicated integration test directory for fleets, shouldn't it go there? I know that this test in particular doesn't require que whole docker setup we are doing for testing fleets, but I still feel like it would fit there better than in the unit tests. Open to discussion.

gateway/tests/core/services/ibm_cloud/code_engine/fleets/test_fleet_template.py
avilches: I removed the "integration" term in the sentence, because in my opinion, the name "integration" is not totally accurate here (but I know the boundaries are blurry).
The real integration tests run jobs from a Serverless client through Gateway, use a real database and depend on the scheduler to trigger the container.
To me, this is a unit test because it focuses only on the template and everything it does: upload during execution, upload when process ends, filter, truncating, and nothing in the test knows anything about fleets, ray, scheduler, serverless....
It would be weird to me if we moved this test to the /tests/ folder with all the integration tests, but I'm open to suggestions and ideas. :)

ElePT (approved):
alright, I think that my latest tests showed a few small updates we should make to the code, but @avilches, given that we are basing all of our newest changes on this PR, I think that it would be easiest to merge it as-is and apply any follow-ups in #2173.

PR #2182, opened 2026-06-01
Refactor/file storage

korgan00:
### Summary
Refactor files storage using the other storages pattern.
Create the abstract files storage, ray files storage and fleets files storage.
Migrate ray storage to use it.
Let fleets as not implemented. (will be ready in a next PR)


gateway/tests/core/services/storage/test_file_storage.py
```diff
@@ -31,87 +32,156 @@ def function_with_provider():
...
     settings.MEDIA_ROOT = str(tmp_path)
-    storage = FileStorage(
+    storage = get_file_storage(
         username="user1",
-        working_dir=WorkingDir.USER_STORAGE,
         function=function_without_provider,
     )
 
-    assert storage.sub_path == "user1"
-    assert storage.absolute_path == f"{tmp_path}/user1"
```
Tansito, line 44: @korgan00 from the tests refactors I think we are missing this assert about the path and I think is interesting to maintain it to ensure the generated path is the correct one (let me know if we are just doing it and I didn't find the line)

gateway/tests/core/services/storage/test_file_storage.py
```diff
@@ -31,87 +32,156 @@ def function_with_provider():
...
     settings.MEDIA_ROOT = str(tmp_path)
-    storage = FileStorage(
+    storage = get_file_storage(
         username="user1",
-        working_dir=WorkingDir.USER_STORAGE,
         function=function_with_provider,
     )
 
-    assert storage.sub_path == "user1/provider1/myfun"
-    assert storage.absolute_path == f"{tmp_path}/user1/provider1/myfun"
```
Tansito, line 58: Similar to my previous comment:

@korgan00 from the tests refactors I think we are missing this assert about the path and I think is interesting to maintain it to ensure the generated path is the correct one (let me know if we are just doing it and I didn't find the line)

Tansito (dismissed):
@korgan00 I think the implementation is correct I just left you a couple of comments in two tests that I was not fully sure if we were analyzing these paths in the tests or not.

Tansito (approved):

PR #2183, opened 2026-06-01
Harden gateway security: secrets, authz, and input validation

quantova-inc:
## Summary

This PR addresses a set of security issues found while auditing the gateway, client, and deployment configuration. Each change is defensive and aims to fail closed.

### Configuration / secrets
- **Fail closed on missing secret key:** the app now refuses to boot when `DJANGO_SECRET_KEY` is unset and `DEBUG` is off, instead of silently using a public hardcoded key. A hardcoded fallback remains only for `DEBUG`/tests.
- **Secure defaults:** `DEBUG` now defaults to `0`; `ALLOWED_HOSTS` defaults to `localhost` (not `*`) in production.
- **Auth fail-closed:** an unknown `SETTINGS_AUTH_MECHANISM` now raises instead of silently falling back to the `mock_token` backend.
- **Token encryption:** the Fernet key is derived with SHA-256 (full entropy) instead of space-padding the secret; added an optional dedicated `TOKENS_ENCRYPTION_KEY`. `decrypt_env_vars` now drops a value it cannot decrypt rather than injecting ciphertext into the runtime environment.

### Authorization / data exposure
- **`get_jobs`:** provider listings no longer serialize other authors' private `result` field.
- **Base serializers:** the base `Job`/`Program` serializers now use explicit field allowlists so `env_vars` / `account_id` / `instance_crn` are never serialized implicitly.

### Input validation
- **Container image:** validate the image reference grammar and enforce a registry path boundary so a configured registry is not matched by `registry.attacker.com/...`.
- **Entrypoint:** validate it is a relative `.py` path with no `..` segments or shell characters before it is interpolated into the runner command.
- **Artifact tar → COS:** reject absolute / `..` traversal members when uploading the artifact, and drop the duplicate plaintext `ENV_JOB_ARGUMENTS` literal env var (arguments are delivered via COS at `ARGUMENTS_PATH`).

### Templates / deployment
- Move the extra template directory off world-writable `/tmp/templates` to `/etc/gateway/templates` (env-overridable via `GATEWAY_TEMPLATES_DIR`) and update the Helm chart mount.
- `docker-compose.yaml` now provides a local dev secret key and `ALLOWED_HOSTS` so the new fail-closed boot still works locally.

### Client
- Basename the download target so a crafted file name cannot escape the download directory.
- Type-validate the result-decoder payload before constructing objects.

## Testing
- `py_compile` passes on all changed Python files.
- Standalone unit checks of the new pure logic (image grammar + registry boundary, entrypoint rules, tar member normalization, Fernet round-trip) pass.
- The full `tox -e py311` suite was not run in my local environment (Python 3.11 + Ray/Qiskit not available); CI should exercise it.

## Notes
- Token env vars are still passed to Code Engine fleets as `literal` env vars; migrating them to `secret_key_ref` is left as a follow-up as it touches the fleet-submission flow more broadly.

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
@@ -330,7 +330,7 @@ spec:
           volumeMounts:
             - mountPath: "/usr/src/app/media/"
               name: gateway-pv-storage
-            - mountPath: "/tmp/templates/"
+            - mountPath: "/etc/gateway/templates/"
               name: ray-cluster-template
             - mountPath: "/tmp/gpujobs/"
               name: gpu-jobs
--- a/client/qiskit_serverless/core/files.py
+++ b/client/qiskit_serverless/core/files.py
@@ -98,7 +98,9 @@ class GatewayFilesClient:
             total_size_in_bytes = int(req.headers.get("content-length", 0))
             chunk_size = 8192
             progress_bar = tqdm(total=total_size_in_bytes, unit="iB", unit_scale=True)
-            file_name = target_name or f"downloaded_{str(uuid.uuid4())[:8]}_{file}"
+            # Strip any directory components so a crafted target_name/file
+            # (e.g. "../../etc/passwd") cannot escape download_location.
+            file_name = os.path.basename(target_name or f"downloaded_{str(uuid.uuid4())[:8]}_{file}")
             with open(os.path.join(download_location, file_name), "wb") as f:
                 for chunk in req.iter_content(chunk_size=chunk_size):
                     progress_bar.update(len(chunk))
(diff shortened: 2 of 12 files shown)


Tansito:
@quantova-inc thanks for your contribution, do you mind to sign the CLA?

avilches:
The PR has been divided into smaller PRs to allow for easier review:

- https://github.com/Qiskit/qiskit-serverless/pull/2300
- https://github.com/Qiskit/qiskit-serverless/pull/2301
- https://github.com/Qiskit/qiskit-serverless/pull/2302
- https://github.com/Qiskit/qiskit-serverless/pull/2303
- https://github.com/Qiskit/qiskit-serverless/pull/2304
- https://github.com/Qiskit/qiskit-serverless/pull/2307
- https://github.com/Qiskit/qiskit-serverless/pull/2308
- https://github.com/Qiskit/qiskit-serverless/pull/2310
- https://github.com/Qiskit/qiskit-serverless/pull/2311
- https://github.com/Qiskit/qiskit-serverless/pull/2312
- https://github.com/Qiskit/qiskit-serverless/pull/2313

PR #2185, opened 2026-06-02
Fix FleetsResultStorage to read results from COS

marceloamaral:
## Summary

- `FleetsResultStorage.get()` was `NotImplementedError` and the factory in `__init__.py` incorrectly returned `RayResultStorage` (reads from local filesystem) for fleets jobs
- Fleets jobs write results to COS via the `RESULTS_PATH` mount, so the gateway must read them back from COS — not from `MEDIA_ROOT`
- Implements `FleetsResultStorage.get()` following the same pattern as `FleetsLogsStorage` (uses `get_cos_client` + `build_job_paths`)

## What broke

PR #2174 (Result storage architecture) + PR #2124 (RESULT_PATH env var) changed `save_result()` in the SDK to write to a mounted file path instead of POSTing to the API. For fleets, this file is on the COS mount (`/data/results.json`). The `JobRetrieveUseCase` then calls `get_result_storage(job).get()`, which was using `RayResultStorage` — looking for the file at `MEDIA_ROOT/<user>/results/<job_id>.json` (gateway filesystem), where it doesn't exist.

## Changes

| File | Change |
|------|--------|
| `gateway/core/services/storage/result_storage_fleets.py` | Implement `get()` to read from COS |
| `gateway/core/services/storage/__init__.py` | Fix factory: return `FleetsResultStorage(job)` instead of `RayResultStorage(job)` |

## Test plan

- [x] Existing unit tests pass (`tests/core/services/storage/`, `tests/api/test_job.py`) — verified locally, 89 tests pass
- [x] Fleet integration tests can retrieve `job.result()` successfully

### DIFF
--- a/gateway/core/services/storage/result_storage.py
+++ b/gateway/core/services/storage/result_storage.py
@@ -1,16 +1,39 @@
+# This code is part of a Qiskit project.
+#
+# (C) IBM 2026
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
 """Abstract base class for result storage."""
 
 from abc import ABC, abstractmethod
-from typing import Optional
 
 
 class ResultStorage(ABC):
-    """Abstract interface for job result storage."""
+    """Abstract interface for job result storage.
+
+    Subclasses implement backend-specific logic for reading and writing
+    job results (e.g. local filesystem for Ray, COS for Fleets).
+    """
 
     @abstractmethod
-    def get(self) -> Optional[str]:
-        """Retrieve the result for the job."""
+    def get(self) -> str | None:
+        """Retrieve the result for the job.
+
+        Returns:
+            The result string, or None if the result is not available.
+        """
 
     @abstractmethod
     def save(self, result: str) -> None:
-        """Persist the result for the job."""
+        """Persist the result for the job.
+
+        Args:
+            result: The result string to store.
+        """
--- a/gateway/core/services/storage/result_storage_fleets.py
+++ b/gateway/core/services/storage/result_storage_fleets.py
@@ -1,23 +1,108 @@
+# This code is part of a Qiskit project.
+#
+# (C) IBM 2026
+#
+# This code is licensed under the Apache License, Version 2.0. You may
+# obtain a copy of this license in the LICENSE.txt file in the root directory
+# of this source tree or at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# Any modifications or derivative works of this code must retain this
+# copyright notice, and modified files need to carry a notice indicating
+# that they have been altered from the originals.
+
 """Fleets implementation of result storage."""
 
-from __future__ import annotations
+import logging
 
-from typing import Optional
+from ibm_botocore.exceptions import ClientError
 
+from core.ibm_cloud import get_cos_client
+from core.ibm_cloud.code_engine.fleets.utils import build_job_paths
 from core.models import Job
 from core.services.storage.result_storage import ResultStorage
 
+logger = logging.getLogger("core.FleetsResultStorage")
+
 
 class FleetsResultStorage(ResultStorage):
-    """Handles the storage and retrieval of user job results for Fleets jobs."""
+    """Handles the storage and retrieval of user job results for Fleets jobs.
+
+    Fleets jobs write results to COS via the RESULTS_PATH mount. This class
+    reads them back from the same COS location.
+    """
+
+    NOT_FOUND_CODES = {"404", "NoSuchKey", "NotFound"}
 
     def __init__(self, job: Job) -> None:
-        self._job_id = str(job.id)
+        """Initialize the storage with COS path information from the job.
 
-    def get(self) -> Optional[str]:
-        """Retrieve the result for this job."""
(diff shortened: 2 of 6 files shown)


Tansito (changes requested):
I didn't read yet the full implementation but I would like to see the tests validating the paths as we are doing here, for example: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/tests/core/services/storage/test_arguments_storage_fleets.py#L56

marceloamaral:
@Tansito adn @ElePT I've added the unit tests and also improved a bit the docstring of the files
So, could you please review this PR again?

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -2,22 +2,90 @@
...
 
-    def get(self) -> Optional[str]:
-        """Retrieve the result for this job."""
-        raise NotImplementedError
+        Returns:
+            The UTF-8 decoded result string, or None if the result is not
+            available (missing project, not-found, or COS error).
+        """
+        if not self._project or not self._user_bucket or not self._results_key:
+            return None
```
ElePT: There should be at least a log here explaining what caused the "return None"

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -2,22 +2,90 @@
...
+        """
         self._job_id = str(job.id)
+        self._project = job.code_engine_project
+        if self._project:
+            paths = build_job_paths(job)
+            self._results_key = paths.cos_results_key
+            self._user_bucket = self._project.cos_bucket_user_data_name
+        else:
+            self._results_key = None
+            self._user_bucket = None
```
ElePT: Why not error directly if there is no project?

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -2,22 +2,90 @@
...
+            return content_bytes.decode("utf-8")
+        except ClientError as e:
+            code = e.response.get("Error", {}).get("Code", "")
+            if code in self.NOT_FOUND_CODES:
+                logger.warning(
+                    "[get] job_id=%s | Result not found in COS at %s/%s",
+                    self._job_id,
+                    self._user_bucket,
+                    self._results_key,
+                )
```
ElePT, line 81: why not report the original code to give more info?

ElePT (dismissed):
Thanks a lot @marceloamaral! I could approve the PR now, although I found a few points where the code could be improved. I will leave it to you to decide to implement them now or in a follow-up.

gateway/core/services/storage/result_storage.py
```diff
@@ -1,16 +1,29 @@
 """Abstract base class for result storage."""
 
+from __future__ import annotations
```
Tansito: I can be wrong but I think this import is not needed 🤔

gateway/core/services/storage/result_storage_fleets.py
Tansito: @marceloamaral take a look to how we are doing it right now in the logs storage: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/services/storage/logs_storage_fleets.py#L59

gateway/core/services/storage/result_storage_fleets.py
Tansito: Correct me if I'm wrong but I would even say that this if it's not needed because these checks are in the instantiation of the class.

gateway/core/services/storage/result_storage_ray.py
```diff
@@ -1,8 +1,9 @@
 """Ray implementation of result storage."""
 
+from __future__ import annotations
```
Tansito: Similar to my previous comment, I have the feeling that this import is not needed.

Tansito (dismissed):

gateway/core/services/storage/result_storage.py
ElePT: I would also vote that, but sometimes there are some weird failures depending on how the annotations are defined. @marceloamaral try to remove it and if it fails, it's fine to keep it. This is actually a deprecated import and I think will be removed in Python3.13, it was supposed to go away I think with 3.10 but there were still some failures in the annotations or something like that.

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Good point, your are right, this guard is redundant since the `__init__` already handles the case where `_project` is `None`, let me update it and also add a log message `self._project == None`!

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Good point, we should raise `ValueError` to follow the same pattern used in `FleetsResultStorage`, I will update it.

gateway/core/services/storage/result_storage.py
marceloamaral: Alright, I removed `from __future__ import` annotations from both files, and it is not needed here...

gateway/core/services/storage/result_storage_ray.py
marceloamaral: That is right and I have updated them

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Yeap, it makes sense, so that we can see the which error it has returned!

marceloamaral:
@Tansito and @ElePT I have addressed all your comments!

ElePT (approved):
Thanks for the quick updates, the PR LGTM

Tansito (approved):
Cool, LGTM! We can always add more improvements later, thanks @marceloamaral !

PR #2186, opened 2026-06-03
Fleets logs: return presigned URL redirect instead of buffering content

avilches:
## Summary

Only for Fleet jobs, the /provider-logs and /logs endpoitns return a 302 redirect to a presigned COS url instead of reading the log file into memory and forwarding it as JSON (if the log object does not exist yet (job still running), it returns 204 No Content). Ray jobs are unchanged.

## Why

Downloading large log files from COS entirely into gateway memory causes the server to run out of resources and fail. Fleet is a new system so we are doing it right from the start. Fleet logs are also already pre-filtered by the in-container wrapper before being uploaded to COS, which makes the redirect straightforward: the client fetches the file directly from COS without any extra transformation needed on the gateway side.

This also opens the door to offering the same presigned URL approach for Ray jobs in a future PR, as an opt-in for new clients while keeping the existing JSON behaviour for older ones.

## Client

The Python client already handles the 2 response: 1) 302 is followed automatically and the text content is returned; (204 returns "No logs yet."); 2) the regular JSON response with the log inside.


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -423,25 +423,31 @@ def result(self, job_id: str):
...
+            f"{self.host}/api/{self.version}/jobs/{job_id}/logs/",
+            headers=get_headers(token=self.token, instance=self.instance, channel=self.channel),
+            timeout=REQUESTS_TIMEOUT,
+            allow_redirects=False,
         )
-        return response_data.get("logs")
+        if response.status_code == 204:
+            return "No logs yet."
+        if response.status_code == 302:
+            return requests.get(response.headers["Location"], timeout=REQUESTS_TIMEOUT).text
```
marceloamaral: If the COS fetch fails (transient network error, 5xx), this either raises an unhandled exception or returns an HTML error page as the 'logs' string. So, we should check the response status before returning `.text`, for example:
```python
if response.status_code == 302:
    cos_response = requests.get(response.headers["Location"],
timeout=REQUESTS_TIMEOUT)
    if cos_response.ok:
        return cos_response.text
    return None
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -423,25 +423,31 @@ def result(self, job_id: str):
...
+            f"{self.host}/api/{self.version}/jobs/{job_id}/provider-logs/",
+            headers=get_headers(token=self.token, instance=self.instance, channel=self.channel),
+            timeout=REQUESTS_TIMEOUT,
+            allow_redirects=False,
         )
-        return response_data.get("logs")
+        if response.status_code == 204:
+            return "No logs yet."
+        if response.status_code == 302:
+            return requests.get(response.headers["Location"], timeout=REQUESTS_TIMEOUT).text
```
marceloamaral: We should also check the response status here

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -46,37 +42,34 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
...
         # Logs stored in COS. They are already filtered
         logs_storage = get_logs_storage(job)
-        logs = logs_storage.get_public_logs()
-        if logs:
-            return logs
 
         if job.program.runner == Program.FLEETS:
-            # Fleets wrapper needs up to 15 seconds to start uploading a log to COS
-            return "No logs yet."
+            logs_storage = get_logs_storage(job)
```
marceloamaral, line 44: Do we need to call it again if line 44 already did it? Or maybe move the object constructor from line 44 to Ray part?

gateway/api/use_cases/jobs/logs_result.py
```diff
@@ -0,0 +1,15 @@
...
+
+from dataclasses import dataclass, field
+from typing import Optional
+
+
+@dataclass
+class LogsResult:
+    """Discriminated result from log use cases.
+
+    Exactly one of raw_log or redirect_url will be set, or neither (no logs yet).
```
marceloamaral, line 11: Minor comment, the docstring says '_exactly one of raw_log or redirect_url will be set, or neither_' but nothing prevents both being set. A `__post_init__` validation would catch bugs at construction time, for example:

````python
def __post_init__(self):
    if self.raw_log is not None and self.redirect_url is not None:
        raise ValueError("Cannot set both raw_log and redirect_url")
```

client/qiskit_serverless/core/clients/serverless_client.py
avilches: Changed. Also returning 'Error fetching logs.' instead of None on failure, consistent with the 'No logs yet.' pattern.

client/qiskit_serverless/core/clients/serverless_client.py
avilches: Changed.

gateway/api/use_cases/jobs/get_logs.py
avilches: Changed. Removed the duplicate call on line 44 and moved it to the Ray path only.

gateway/api/use_cases/jobs/logs_result.py
avilches: Changed.

marceloamaral (commented):
LGTM from my side, let's wait for @ElePT's review

ElePT (dismissed):
LGTM on my side too. I was able to locally test the changes of this PR on top of #2173 and verify that both custom and provider functions work with the change and logs are retrievable. You just need to fix the merge conflict @avilches and I can re-approve.

ElePT (approved):
LGTM

PR #2187, opened 2026-06-03
Functions client unification to runtime - part 1\n

ItamarGoldman:
Functions client unification to runtime - part 1\n

### Summary
validating instance access to Qiskit IBM Runtime client via authentication process.

### Details and comments

- [x] using runtime client as the service of the client `IBMServerlessClient`.
- [x] Add release notes - authentication to the cloud (instance) is done through `QiskitRuntimeService`


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -706,12 +705,12 @@ def __init__(
...
             channel: identifies the method to use to authenticate the user
         """
-        self.account = self._discover_account(
-            token=token,
-            instance=instance,
-            channel=channel,
-            name=name,
-        )
+
+        channel = channel or Channel.IBM_QUANTUM_PLATFORM.value
```
HuangJunye: why do we need `or Channel.IBM_QUANTUM_PLATFORM.value`?

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -23,11 +23,23 @@
...
+        mock_list_instances.return_value = [
+            {
+                "crn": "my_instance",
+                "plan": "test_plan",
+                "name": "my_instance",
+                "tags": "test_tags",
+                "pricing_type": "test_pricing_type",
+            }
+        ]
+
```
HuangJunye, line 42: I am not very familiar with this part. Is the `mock_list_instances` mocking the instances returned by calling `CloudAccount.list_instances`?

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -91,13 +163,27 @@ def test_ibm_quantum_channel(self):
         use_instance = "h/g/p"
         use_token = "save_token"
 
-        with pytest.raises(ValueError, match=r"Your channel value is not correct"):
+        # The error message miss a ' at the end but that is the message in qiskit ibm runtime
```
HuangJunye, line 166: Hmm, can you file an issue in qiskit-ibm-runtime? Perhaps we should make this more robust. Is it possible to not match exact phrase?

HuangJunye (commented):
This looks great! I left a few comments to understand the changes better.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: That is here because it was behavior `_discover_account`used previously. Essentially, it means that it can't be `None` (even if the user try to force `None` on it it will be set). We can remove it as it is just to keep the previous behavior.

ElePT: I vote for keeping it. When in doubt it's better to respect the former behavior. @ItamarGoldman maybe you could add a comment explaining that this is to maintain backwards compatibility?

client/tests/core/test_ibm_serverless_client.py
ElePT: yes, essentially, it's avoiding the API calls by `CloudAccount.list_instances`

client/tests/core/test_ibm_serverless_client.py
ElePT: I think that on our side this test is fine, this code would match the error even if they added the missing `'`

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Yes. The method retrieve all crns with the IBM Cloud Global Search API. I made a mock return value for it.

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -47,11 +59,71 @@ def test_init(self, mock_file_path, mock_verify_credentials):
...
+        https://cloud.ibm.com/docs/account?topic=account-crn#format-crn)
+        """
+
+        instance_dic = {
+            "crn": "crn:v1:bluemix:public:cloud-object-storage:global:a/59bcbfa6ea2f006b4ed7094c1a08dcdd:"
+            "1a0ec336-f391-4091-a6fb-5e084a4c56f4:bucket:mybucket",
+            "plan": "test_plan",
+            "name": "my_instance_crn",
+            "tags": "test_tags",
+            "pricing_type": "test_pricing_type",
```
ElePT: if you want to try sth more realistic, instead of a cloud object storage, I think that you should try with a quantum-ish CRN: `crn:v1:bluemix:public:quantum-computing:us-east:a/my-crn::`

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -47,11 +59,71 @@ def test_init(self, mock_file_path, mock_verify_credentials):
...
         """Test saving and loading accounts with the IBMServerlessClient."""
 
+        # Mock list of instance crns in the IBM Cloud Global
+        mock_list_instances.return_value = [
+            {
+                "crn": "dummy_hub/dummy_group/dummy_project",
+                "plan": "test_plan",
+                "name": "dummy_hub/dummy_group/dummy_project",
+                "tags": "test_tags",
+                "pricing_type": "test_pricing_type",
```
ElePT: the CRN should not be hub/group/project. this was the old way of doing things (if you find any other reference, you should remove it). The name should also be sth like "test_name" rather than H/G/P

ElePT (commented):
Mostly LGTM! I left some minor suggestions and comments to the earlier review.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: I can open an issue regarding the text.
About making the test more robust, I can check the a `ValueError` is raised but not check the message. Here I wanted to check there is `ValueError` exception and that it is regarding the channel value (and not for other values).
Maybe this test is redundant since it checks that `qiskit_ibm_runtime` raise this error. WDYT?

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: You are absolutely correct! Thanks!

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: I will change it. It was written like this because of prior test data. The reason is because when the CRN value doesn't start with `crn`, it search by name and then it extract the CRN value so:
```python
client.account.instance == use_instance
```
Essentially checks:
```python
mock_list_instances.return_value["crn"] == mock_list_instances.return_value["name"] 
```

ElePT (approved):
LGTM

PR #2188, opened 2026-06-03
Restore filter horizontal fields in backoffice

Tansito:
### Summary

This PR fixes a problem introduced in #2100 where we removed fields with `filter_horizontal` in favor of `autocomplete` fields. Here we basically restored those specifically fields letting the problematic ones as they are to not impact the improvement in the performance.


marceloamaral (approved):
looks like a basic fix, LGTM

Tansito:
It is haha

PR #2200, opened 2026-06-04
Move code_engine_project FK from Job to Program

marceloamaral:
## Summary
- Moves the `code_engine_project` foreign key from the `Job` model to the `Program` model, simplifying the architecture so CE project assignment is managed at the program level rather than per-job
- Adds `select_related("program__code_engine_project")` to job queries as a performance improvement to fetch the related program and CE project in a single DB query
- Removes the `CodeEngineProjectManager` and zone-routing settings (`CE_PROJECTS`, `CE_ZONE`, `FLEETS_PROFILE_ZONE_MAP`) that are no longer needed
- Fixes `get_result_storage` factory to correctly return `FleetsResultStorage` for Fleets jobs (previously returned `RayResultStorage`)

## Test plan
- [ ] Verify migration applies cleanly on a fresh database
- [ ] Verify `sync_ce_project` management command assigns CE project to Fleets programs
- [ ] Verify job submission still works for Fleets runner programs
- [ ] Verify result retrieval works for Fleets jobs
- [ ] Run full test suite (`tox -e unit`)


gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -37,7 +37,7 @@ def execute(self, job_id: UUID, user: AbstractUser) -> str:
             str: Job logs if accessible, otherwise a message indicating no logs are available.
         """
         try:
-            job = Job.objects.get(id=job_id)
+            job = Job.objects.select_related("program__code_engine_project").get(id=job_id)
```
marceloamaral: The `select_related("program__code_engine_project")` is just for performance optimization to fetch the related data in a single DB query...

ElePT (commented):
Let's discuss this today, but as you may have seen in my internal PR review, this links very much to the environment variables.

marceloamaral:
This PR depends on https://github.com/Qiskit/qiskit-serverless/pull/2173 to be fully tested

gateway/api/migrations/0053_remove_codeengineproject_zone_and_more.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("api", "0052_job_account_id_job_instance_crn"),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
Tansito: to be able to do a regression all our db changes and backward compatible at this moment and we try not to modify the database in a way we couldn't make use of a previous version in case we need to do a rollback I would keep the `AddField` by now and I wouldn't remove the other two fields. We can annotate them to do a clean-up of fields later on, we have an issue for it.

gateway/api/use_cases/jobs/get_logs.py
Tansito: Mmm.. I don't get this performance improvement, why is this faster than just getting the job by it's primary key? Also thinking that the code engine project is not the unique relationship, Job has plenty of them.

gateway/api/admin.py
```diff
@@ -169,7 +169,7 @@ class JobAdmin(admin.ModelAdmin):
     exclude = ["arguments", "env_vars", "logs", "result"]
     ordering = ["-created"]
     inlines = [JobEventInline]
-    autocomplete_fields = ["author", "program", "compute_resource", "code_engine_project", "config"]
+    autocomplete_fields = ["author", "program", "compute_resource", "config"]
```
Tansito, line 172: This doesn't apply to my comment from the database migration. Remove them from the admin panel I think it's fine 👍

gateway/core/models.py
```diff
@@ -187,6 +193,21 @@ class Meta:
             ),
         ]
 
+    def ensure_code_engine_project(self) -> None:
```
Tansito: I think what you want to do here is what we are defining as `model_manager`. An example: https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/model_managers/code_engine_projects.py

gateway/main/settings.py
```diff
@@ -268,6 +268,27 @@
...
+CE_HMAC_SECRET_NAME = os.environ.get("CE_HMAC_SECRET_NAME")
+# Set to "true" to use the public COS endpoint instead of the private VPC endpoint.
+# Only needed for local testing outside IBM Cloud (e.g. docker-compose).
+CE_COS_USE_PUBLIC_ENDPOINT = os.environ.get("CE_COS_USE_PUBLIC_ENDPOINT", "false").lower() == "true"
+# Code Engine project definitions: JSON array delivered via deployment manifest.
+# Each entry must contain: project_id, project_name, region, resource_group_id,
+# subnet_pool_id, pds_name_state, pds_name_users, pds_name_providers,
+# cos_instance_name, cos_key_name, cos_bucket_task_store_name,
+# cos_bucket_user_data_name, cos_bucket_provider_data_name.
+CE_PROJECTS: list = json.loads(os.environ.get("CE_PROJECTS", "[]"))  # type: ignore[assignment]
```
Tansito: I suppose this is not needed anymore with my PR.

gateway/api/use_cases/jobs/get_logs.py
```diff
@@ -33,7 +33,7 @@ def execute(self, job_id: UUID, user: AbstractUser) -> LogsResult:
             or LogsResult with raw_log set (Ray).
         """
         try:
-            job = Job.objects.get(id=job_id)
+            job = Job.objects.select_related("program__code_engine_project").get(id=job_id)
```
avilches: Yes, but why is the code_engine_project needed to get the logs? Same in the others use cases.

gateway/main/settings.py
```diff
@@ -375,34 +396,4 @@
...
-CE_COS_USE_PUBLIC_ENDPOINT = os.environ.get("CE_COS_USE_PUBLIC_ENDPOINT", "false").lower() == "true"
-# Maps compute profiles to their availability zone. Populated at deploy time via FLEETS_PROFILE_ZONE_MAP
-# JSON env var, e.g. '{"gx2-8x64x1l40s": "us-east-1", "gx3d-24x120x1a100p": "us-east-2"}'.
-# Profiles absent from the map fall back to the first active CodeEngineProject.
-#######
-### Note: Remove environment variable as it's not needed now
-#######
-FLEETS_PROFILE_ZONE_MAP: dict = json.loads(os.environ.get("FLEETS_PROFILE_ZONE_MAP", "{}"))  # type: ignore[assignment]
-# Multi-project configuration: JSON array of project dicts
-CE_PROJECTS: list = json.loads(os.environ.get("CE_PROJECTS", "[]"))  # type: ignore[assignment]
```
Tansito, line 406: Similar to my previous comment. I don't think you need to touch `settings`, Marcelo.

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -144,7 +144,9 @@ def run(self):
         # Note: with LIMITS_MAX_FLEETS potentially reaching 1000+ concurrent jobs, updating statuses
         # sequentially will become a bottleneck. This loop should be parallelized using multiple
         # threads or batched processing for performance reasons.
-        jobs = Job.objects.filter(status__in=Job.RUNNING_STATUSES, runner=Program.FLEETS)
+        jobs = Job.objects.select_related("program__code_engine_project").filter(
```
Tansito: Looking at how many changes `select_related` introduces in the PR I would open a new one only for this. I think it deserves the time if it's correct or not to use `select_related` in this specific scenario. Add it in all the queries I don't think it scales.

gateway/api/use_cases/jobs/get_logs.py
avilches: Nevermind, I see the storages get the bucket from cos_bucket_user_data_name

Tansito (changes requested):

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -73,11 +62,10 @@ def get(self) -> str | None:
             code = e.response.get("Error", {}).get("Code", "")
             if code in self.NOT_FOUND_CODES:
                 logger.warning(
-                    "[get] job_id=%s | Result not found in COS at %s/%s (code=%s)",
+                    "[get] job_id=%s | Result not found in COS at %s/%s",
```
avilches: code?

gateway/core/models.py
```diff
@@ -187,6 +193,21 @@ class Meta:
...
+
+        Uses the first active CE project as the default.
+        No-op if the program already has a CE project or is not a Fleets program.
+        Logs a warning if no active project is available.
+        """
+        if self.runner != Program.FLEETS:
+            return
+        if self.code_engine_project:
+            return
+        self.code_engine_project = CodeEngineProject.objects.filter(active=True).order_by("pk").first()
```
avilches: Is this what we want? The first? I would prefer put the id in a config or something similar, I don't know...

marceloamaral: Right, I was not sure about this part either. Let me think about how to improve it.

Maybe we can use the project name to find the default project...

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
marceloamaral: It makes sense!

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Will add it back, thanks!

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
Tansito: I don't think the change is wrong, it has sense but we need to look for a better way to implement it (maybe in the model manager or something like that).

gateway/api/migrations/0053_remove_codeengineproject_zone_and_more.py
marceloamaral: Alright, it makes sense, let me do this!

gateway/api/use_cases/jobs/get_logs.py
marceloamaral: Right, the `select_related` does not make the `get(id=job_id)` itself faster. That is still just a PK lookup.

The reason I added it is because later in this flow we access `job.program.code_engine_project`, so without `select_related` Django would do extra queries when those fields are accessed.

With `select_related("program__code_engine_project")`, Django already fetches those relations in the same query, so it will need only one query...

And yes, `Job` has more relations, but I only added this one because it is the one used in this code path

gateway/core/models.py
marceloamaral: Added `CE_DEFAULT_PROJECT_NAME` env var for config-based project selection

```
select_default() — looks up by CE_DEFAULT_PROJECT_NAME first, and raises ValueError if CE_DEFAULT_PROJECT_NAME not configured
```

gateway/core/models.py
marceloamaral: Alright, good point! So, I restored the `CodeEngineProjectQuerySet` model manager (previously only for zones) and moved the project selection logic there...

Now the serializer calls:

```python
CodeEngineProject.projects.assign_to_program(program)
```

instead of keeping this logic in the `Program` model.

gateway/main/settings.py
marceloamaral: Yeap, I rebased with your changes that were merged in main

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
marceloamaral: No worry, I agreed to revert the `select_related` additions from this PR. The FK move works fine without them (just triggers lazy loading in multiple queries). I'll open a separate PR later...

gateway/api/management/commands/sync_ce_project.py
```diff
@@ -37,53 +34,65 @@
...
+    def _assign_to_fleets_programs(self):
+        """Assign active CE project to Fleets programs that lack one."""
+        project = CodeEngineProject.objects.filter(active=True).order_by("pk").first()
+        if not project:
+            return
+        updated = Program.objects.filter(runner=Program.FLEETS, code_engine_project__isnull=True).update(
+            code_engine_project=project
+        )
+        if updated:
+            logger.info("Assigned CodeEngineProject [%s] to %d Fleets program(s)", project.project_name, updated)
```
ElePT: Up until now we have been assigning the project to a job at job creation time (`RunJobSerializer.create()`). For me, the equivalent change when moving to the program would be to move the project assignment to a program to the program creation time (`/upload`), not modifying this sync command. I don't think we need to touch this file at this point. Is there any important argument for this change I am missing here?

gateway/api/migrations/0053_add_code_engine_project_to_program_and_job_snapshots.py
```diff
@@ -0,0 +1,47 @@
...
+            name="zone",
+            field=models.CharField(
+                blank=True,
+                help_text="[DEPRECATED] Kept for rollback compatibility — not used by current code",
+                max_length=64,
+                null=True,
+                unique=True,
+            ),
+        ),
+        migrations.AlterField(
```
ElePT, line 48: I am not so familiar with Django migrations so I will defer this specific review to @Tansito or @avilches.

gateway/api/serializers.py
```diff
@@ -132,7 +132,10 @@ def create(self, validated_data):
         normalized_dependencies = [self._normalize_dependency(dep) for dep in raw_dependencies]
         validated_data["dependencies"] = json.dumps(normalized_dependencies)
 
-        return Program.objects.create(**validated_data)
+        program = Program(**validated_data)
+        CodeEngineProject.projects.assign_to_program(program)
+        program.save()
+        return program
```
ElePT, line 140: Following up on my earlier comment, you seem to be already assigning the project to a program in the serializer, so I am a bit confused by the change in `sync_ce_project.py`

gateway/api/serializers.py
```diff
@@ -355,13 +360,8 @@ def create(self, validated_data):  # pylint: disable=too-many-locals
...
 
-        code_engine_project = None
-        if program.runner == Program.FLEETS:
-            code_engine_project = CodeEngineProject.objects.select_for_profile(compute_profile)
-            if code_engine_project is None:
-                raise serializers.ValidationError(
-                    f"No active Code Engine project available for compute profile {compute_profile}"
-                )
+        if program.runner == Program.FLEETS and not program.code_engine_project:
+            raise serializers.ValidationError("Program has no Code Engine project assigned. Contact administrator.")
```
ElePT, line 366: Is the case of not `program.code_engine_project` possible if we are enforcing it upon program creation? Not saying we need to remove it upright, I just want to question when the code looks overly defensive. It might still make sense in this case.

gateway/api/serializers.py
ElePT, line 140: Should we add a verification here that the assigned project is not `None` so we don't have to check individually every time we access it from the program?

gateway/core/services/runners/fleets_runner.py
```diff
@@ -695,7 +697,7 @@ def _get_max_instances(self) -> int:
         """Return the maximum number of fleet instances.
 
         Uses ``job.config.workers`` if set, otherwise falls back to
-        ``settings.FLEETS_DEFAULT_MAX_INSTANCES`` (default ``1``).
+        ``settings.FLEETS_DEFAULT_MAX_INSTANCES``.
```
ElePT: any reason why the default was removed (did it change?)

gateway/core/services/storage/__init__.py
```diff
@@ -26,6 +26,8 @@ def get_arguments_storage(job: Job) -> ArgumentsStorage:
     Raises:
         ValueError: If the runner type is unknown.
     """
+    if not job.program or not job.program.runner:
+        raise ValueError(f"Job '{job.id}' has no program or runner assigned")
```
ElePT: See my earlier validation comment.

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -34,30 +22,31 @@ class FleetsResultStorage(ResultStorage):
...
-        self._results_key = paths.cos_results_key
-        self._user_bucket = self._project.cos_bucket_user_data_name
+        self._project = job.program.code_engine_project
+        if self._project:
+            paths = build_job_paths(job)
+            self._results_key = paths.cos_results_key
+            self._user_bucket = self._project.cos_bucket_user_data_name
+        else:
+            self._results_key = None
+            self._user_bucket = None
```
ElePT: I don't think we should do this silent assignment to None, but rather raise an error if self._project is not there (also see my earlier comments).

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -87,20 +76,10 @@ def get(self) -> str | None:
...
-                e,
-            )
-            return None
 
     def save(self, result: str) -> None:
         """Not implemented — Fleets results are written by the SDK via RESULTS_PATH.
 
-        Args:
-            result: Unused. Present only to satisfy the abstract interface.
-
```
ElePT, line 103: why remove the args from the docs?

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -87,20 +76,10 @@ def get(self) -> str | None:
                 e,
             )
             return None
-        except UnicodeDecodeError as e:
-            logger.error(
-                "[get] job_id=%s | Failed to decode result: %s",
-                self._job_id,
-                e,
-            )
-            return None
```
ElePT, line 96: Why remove this (was this redundant? could be)

gateway/examples/setup_fleets_env.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
ElePT: We should not be sharing this file.

gateway/main/settings.py
ElePT: It seems like the changes were put on top of the rebase. As david mentioned, we should not be using any of these anymore.

gateway/main/settings.py
ElePT: (still modified)

ElePT (dismissed):
Thanks a lot for this PR! I made a review of the main code with some points I think should be changed. If I missed anything in my analysis feel free to reach out.

Side comment: force pushes after a review make it a bit harder to follow the changes and see what was applied and what not from previous reviews. In general, I try to keep past history for traceability.

gateway/api/migrations/0053_add_code_engine_project_to_program_and_job_snapshots.py
avilches: These migrations are auto-generated by Django based on the changes between the model and the db, so we can always (kinf-of) trust them

avilches:
Now that the project is moving to the program, I wonder if it wouldn't be interesting to store certain project information at the time of job execution for historical purposes. For instances, now we are storing in the job the account_id from the user. I said that because project data fields can change over the time, and the programs can move from one project to other. So, maybe it's worth it, at least consider doing it if not in this PR then later on, to save the project name, the region and the zone. The change will be super simple: just add the fields and copy them when creating the job from the program.project.

gateway/main/settings.py
marceloamaral: Right, I am only adding now the new `CE_DEFAULT_PROJECT_NAME` to deterministically select the default CE Project and removing the `FLEETS_PROFILE_ZONE_MAP`

gateway/api/management/commands/sync_ce_project.py
marceloamaral: This is trick, you are right that the upload path already handles this with:

```python
CodeEngineProject.projects.assign_to_program(program)
```

But `_assign_to_fleets_programs` in `sync_ce_project.py` has a different purpose. It is a backfill for programs that were created before the CE project existed in the DB, for example during the first deploy, or if `CE_PROJECTS` was added after programs already existed.

Without it, those existing programs would still have `code_engine_project=None` and could fail on the next job submission.

I think we have two options:

1. Keep the backfill in `sync_ce_project`. It is idempotent and only runs at container startup.
2. Remove it, but then pre-existing programs would need manual intervention, or a one-off migration.

**So, should I remove this?**

**Are the old programs considered dead data?** If yes, then maybe we do not need to handle this backfill anymore.

gateway/api/migrations/0053_add_code_engine_project_to_program_and_job_snapshots.py
marceloamaral: Yes, I generated it with `tox`

gateway/api/serializers.py
marceloamaral: I left a comment there, the `_assign_to_fleets_programs` in `sync_ce_project.py` has a different purpose. It is a backfill for programs that were created before the CE project existed in the DB...

But we can manually do that once to fix the DB and remove the code from `sync_ce_project.py`, wdyt?

marceloamaral: Yes, good point!

gateway/api/serializers.py
marceloamaral: So, based on your previous comment, I added the upload validation now, and this check could be potentially redundant : )

But I think it still covers two edge cases:

1. Programs created before the migration, with no CE project assigned yet. Specially if we remove the logic from `sync_ce_project.py`
2. Programs where the CE project was deactivated after the upload.

Since both can happen during the rollout, I think it makes sense to keep this as a safety check. It is just a small guard at the system boundary.

gateway/core/services/runners/fleets_runner.py
marceloamaral: No good reason, I'm adding it back!

gateway/core/services/storage/__init__.py
marceloamaral: Fair point,  we have already validate at upload and at job creation, so we do not need it here!

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Alright, as you pointed before, validation at upload ensures `code_engine_project` is always set for Fleets programs. Adding a `ValueError` here is over-defensive... Let me remove it...

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: It was removed by mistake, adding back!

gateway/core/services/storage/result_storage_fleets.py
marceloamaral: Adding back!

gateway/examples/setup_fleets_env.sh
marceloamaral: yeap, removed it! Sorry :D

marceloamaral:
@avilches Good point. If a program is moved to a different CE project later, or if some project fields change, we lose the info about where the job actually ran! I think it is probably the same reason why we store `account_id` on `Job`.

Adding snapshots like `project_name`, and `region` when the job is created is straightforward. I added it so that we do not need to create another migration later.

gateway/api/serializers.py
marceloamaral: removed the filling in `sync_ce_project.py`

gateway/main/settings.py
```diff
@@ -391,17 +391,11 @@
 FLEETS_DEFAULT_MAX_INSTANCES = int(os.environ.get("FLEETS_DEFAULT_MAX_INSTANCES", "1"))
 
 CE_HMAC_SECRET_NAME = os.environ.get("CE_HMAC_SECRET_NAME", None)
+CE_DEFAULT_PROJECT_NAME = os.environ.get("CE_DEFAULT_PROJECT_NAME", "")
```
Tansito, line 394: Sorry @marceloamaral , why do we need a CE_DEFAULT_PROJECT_NAME? It sounds weird to me initially (like, the project name is a name, it's like if we have a default title for a function)

marceloamaral: @Tansito We were discussing a better way to select the default project (before we were getting the first one in the query), especially because in the future we may have multiple projects, for example one per provider.

We could select the project by name or by ID. I think both could work fine. Wdyt?

Tansito: I see, so this is thinking not in provider functions but in custom functions maybe, for example

marceloamaral: Note that the default project here will be the Custom Functions project once we start having one project per provider.

Tansito (approved):
Thanks @marceloamaral for this and overall address all the comments!

PR #2201, opened 2026-06-05
Update componente version GHA

Tansito:
### Summary

This PR updates the logic for the `update-component-version` GHA removing old references and adding new ones for the next time we need to run the GHA.


ElePT (approved):
Thanks for looking into it, I think it's good.

PR #2202, opened 2026-06-05
Add missing acceptance tests for custom function permissions

avilches:
Add acceptance tests that were missing for custom function permissions, plus `specs/INSTANCES_AUTH.md` documenting the instance-based authorization system.

**`TestNoPermissionsInstance`**
- `test_serverless_list_includes_custom_function`: user can list their own serverless functions without any permission (author-based access).

**`TestProviderInstance`** (no `function-custom.*` permissions)
- `test_upload_custom_function_raises_404`: upload blocked (404) when `function-custom.write` is absent.
- `test_run_custom_function_raises_404`: run blocked (404) when `function-custom.run` is absent.
- `test_serverless_list_includes_custom_function`: own custom function still visible in serverless list.

**`TestCombinedInstance`** (has `function-custom.*` permissions)
- `test_upload_custom_function_succeeds`: upload succeeds with `function-custom.write`.
- `test_run_custom_function_succeeds`: run creates a job with `function-custom.run`.
- `test_serverless_list_includes_custom_function`: own custom function visible in serverless list.


ElePT (approved):
We hadn't discussed explicitly the listing case but I agree with what is described in the specs and the description, authors should always be able to list functions, so I think this implementation looks good.

PR #2203, opened 2026-06-08
Fleets results: return presigned URL redirect instead of buffering content

avilches:
## Summary

Continuing the approach from #2186 with the Fleet logs. This PR adds `GET /jobs/{id}/result/` as a new endpoint for downloiad job results and updates the Python client to use it.

## New endpoint: `GET /jobs/{id}/result/`

Serves both Ray and Fleet jobs, but behaves differently by runner:
- Fleet: checks whether `results.json` exists in COS and returns 302 redirect to a presigned URL (result ready) or 204 No Content (not ready yet). The gateway never download the COS content.
- Ray: a redirect is not possible because Ray results live on the gateway's COS volume mounted, so the endpoint reads the file and returns the content inline as a 200 JSON response.

The existing `GET /jobs/{id}/` endpoint is unchanged and still supports `?with_result=true` for backward compatibility with older clients.

## Client: `result()`

- Before: called `GET /jobs/{id}/?with_result=true`.
- After: calls `GET /jobs/{id}/result/` and handles 302 (follows redirect to COS, decodes JSON from the COS response), 204 (returns `None`), and 200 (decodes `response["result"]` as before for Ray jobs).

## Client: `job()`

- Before: called `GET /jobs/{id}/` without an explicit `with_result` parameter. The server defaulted to `with_result=true`, so it read the result from COS or the local filesystem on every call.
- After: explicitly passes `with_result=false`. The server skips the storage read entirely and returns `result: null`. This does not break anything: `job()` only uses `id` and `compute_profile` from the response and never exposed the result field to the caller. The result has always been fetched separately via `result()`.


avilches:
The PR shares code from #2186, so it needs to be merged after it (and probably solve conflicts)

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -289,6 +289,7 @@ def job(self, job_id: str) -> Optional[Job]:
             request=lambda: requests.get(
                 url,
                 headers=get_headers(token=self.token, instance=self.instance, channel=self.channel),
+                params={"with_result": "false"},
```
avilches, line 291: Default for with_result is true, so now has been changed explicitly to false, so the Gateway skips the COS read and returns `result: null`. This does not break anything because job() don't use the result.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -412,15 +413,20 @@ def stop(self, job_id: str, service: Optional[QiskitRuntimeService] = None):
...
         )
-        return json.loads(response_data.get("result", "{}") or "{}", cls=QiskitObjectsDecoder)
+        if response.status_code == 204:
+            return None
+        # Not all redirects go to COS — HTTP→HTTPS redirects stay on the same host.
+        # Checking the hostname detects only redirects to an external host (COS/MinIO).
+        redirected_to_cos = urlparse(response.url).hostname != urlparse(gateway_url).hostname
+        if redirected_to_cos:
+            return json.loads(response.text, cls=QiskitObjectsDecoder) if response.ok else None
+        return json.loads(response.json().get("result", "{}") or "{}", cls=QiskitObjectsDecoder)
```
avilches, line 428: Use the new /result endpoint following the same approach than /logs and /provider-logs

gateway/api/use_cases/jobs/get_logs_response.py
```diff
@@ -5,7 +5,7 @@
 
 
 @dataclass
-class LogsResult:
+class GetLogsResponse:
```
avilches, line 7: LogsResult was a terrible name. Now, GetJobLogsUseCase -> returns GetJobLogsResponse and GetResultUseCase -> returns GetResultResponse

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -73,36 +57,41 @@ def get(self) -> str | None:
...
             logger.error(
-                "[get] job_id=%s | COS error %s: %s",
+                "[get] user_id=%s job_id=%s | COS error %s: %s",
+                self._user_id,
                 self._job_id,
                 code,
                 e,
             )
             return None
-        except UnicodeDecodeError as e:
```
marceloamaral, line 84: Any particular reason why dropping this?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -412,15 +413,20 @@ def stop(self, job_id: str, service: Optional[QiskitRuntimeService] = None):
...
-            )
+        gateway_url = f"{self.host}/api/{self.version}/jobs/{job_id}/result/"
+        response = requests.get(
+            gateway_url,
+            headers=get_headers(token=self.token, instance=self.instance, channel=self.channel),
+            timeout=REQUESTS_TIMEOUT,
         )
-        return json.loads(response_data.get("result", "{}") or "{}", cls=QiskitObjectsDecoder)
+        if response.status_code == 204:
+            return None
```
marceloamaral: The function should consistently return an empty dic `{}` instead of sometimes returning `None` and sometimes `{}`. 

We should define the return type as a dict, for example:
```python
from typing import Any

def result(self, job_id: str) -> dict[str, Any]:
```
and make sure the function always returns {} when there is no result...

gateway/api/use_cases/jobs/get_result_response.py
```diff
@@ -0,0 +1,20 @@
+"""Return type for result retrieval use cases."""
+
+from dataclasses import dataclass, field
+from typing import Optional
```
marceloamaral: We try to avoid using `Optional`, to use  ` str | None` as much as possible...

client/qiskit_serverless/core/clients/serverless_client.py
avilches: Changed

gateway/api/use_cases/jobs/get_result_response.py
avilches: Changed

gateway/core/services/storage/result_storage_fleets.py
avilches: Changed

gateway/core/services/storage/result_storage_fleets.py
```diff
@@ -12,6 +12,8 @@
 
 """Fleets implementation of result storage."""
 
+from __future__ import annotations
+
```
avilches:
```suggestion
```

gateway/api/v1/views/jobs/save_result.py
```diff
@@ -96,25 +98,40 @@ def serialize_output(job: Job):
...
         request: The HTTP request.
         job_id: Job identifier (UUID path parameter).
 
     Returns:
-        Response containing the updated serialized job.
+        Response containing the updated serialized job (POST) or the result (GET).
     """
+    user = cast(AbstractUser, request.user)
+
+    # GET for the Serverless client that submited the jo
```
avilches:
```suggestion
    # GET for the Serverless client that submited the job
```

gateway/api/v1/views/jobs/save_result.py
```diff
@@ -96,25 +98,40 @@ def serialize_output(job: Job):
     },
 )
 @endpoint("jobs/<uuid:job_id>/result", name="jobs-result")
-@api_view(["POST"])
+@api_view(["GET", "POST"])
 @permission_classes([permissions.IsAuthenticated])
 @endpoint_handle_exceptions
 def save_result(request: Request, job_id: UUID) -> Response:
```
avilches:
```suggestion
def jobs_result(request: Request, job_id: UUID) -> Response:
```

gateway/api/v1/views/jobs/save_result.py
Tansito, line 119: add a swagger auto schema for GET

Tansito (changes requested):

gateway/api/use_cases/jobs/retrieve.py
```diff
@@ -44,14 +44,27 @@ def execute(
...
 
-        can_read_result = JobAccessPolicies.can_read_result(user, job)
-
-        if with_result and can_read_result:
-            result_store = get_result_storage(job)
-            result = result_store.get()
-            if result is not None:
-                job.result = result
+        # with_result is legacy, new clients will use the new GET /:id/results endpoints
+        if with_result:
```
Tansito, line 48: We need to ensure then that this with_result is false. I think we need to change it in the serializer if I remember correctly.

Tansito (changes requested):

gateway/tests/core/services/storage/test_result_storage.py
```diff
@@ -10,7 +10,7 @@
 
 @pytest.mark.django_db
 class TestResultStorage:
```
avilches:
```suggestion
class TestRayResultStorage:
```

gateway/api/use_cases/jobs/retrieve.py
avilches: Confirmed: the default value was true. I changed to false in this [commit](https://github.com/Qiskit/qiskit-serverless/pull/2203/changes/69fcfe80082e74970244488aa3a46d0a04ca5fc7#diff-0e2ad92a2a49282a20cf382e317a31b5d7c6c611360a314533f946f01a43b047R36)

gateway/api/v1/views/jobs/save_result.py
avilches: Added in this [commit](https://github.com/Qiskit/qiskit-serverless/pull/2203/changes/69fcfe80082e74970244488aa3a46d0a04ca5fc7#diff-0e2ad92a2a49282a20cf382e317a31b5d7c6c611360a314533f946f01a43b047R36)

Tansito (approved):
We applied the feedback and everything looks good! Nice work here @avilches 🫡

PR #2207, opened 2026-06-09
Remove gateway /result fallback from save_result

avilches:
PR #2124 updated the client to always write results to the path set in `RESULTS_PATH`, which is mounted COS storage provided by the runner (Ray or Fleets). However, the `save_result` function kept a try/except that fell back to calling the gateway `/result` endpoint if the file write failed.

This PR removes that fallback. The gateway endpoint remains available for partner functions that have not yet migrated to the new client, but the client itself should only write to COS. Keeping the fallback defeats the purpose of the file-based approach and keeps a dependency on a deprecated code path.

The test for `save_result` is updated to match: it now sets `RESULTS_PATH` to a temp file and asserts the result is written there, instead of mocking an HTTP POST to the gateway.

### DIFF
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -398,58 +398,20 @@ def save_result(result: Dict[str, Any]):
         result: data that will be accessible from job handler `.result()` method.
     """
 
-    version = os.environ.get(ENV_GATEWAY_PROVIDER_VERSION)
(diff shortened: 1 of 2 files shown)


ElePT (approved):
I think this was a misinterpretation of the fallback also on my side (I reviewed the PR and didn't flag it). LGTM.

PR #2209, opened 2026-06-09
Emit job usage data through IBM Cloud Event Streams instances

ismaelRozasRamallal:
### Summary
 Fleets jobs now report their classical compute usage to downstream consumers in real time. From the moment a job starts running until it completes, consumers receive a continuous stream of usage events they can use to track how long the job has been executing and react when it finishes. This enables billing, monitoring, and any other business logic that depends on knowing a job's runtime to operate without polling the API.

One key point here is, Ray-based jobs are not affected by this change. Event publishing is scoped exclusively to Fleets (Code Engine) jobs.

🔨 related with #https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1507

### Details and comments

- Add job usage event publishing to IBM Cloud Event Streams for Fleets (Code Engine) jobs. 
- When a job transitions to `RUNNING`, a job_started event is published; subsequent scheduler iterations emit job_in_progress events with the elapsed classical time in nanoseconds; and when the job reaches a terminal state, a job_ended event is published. 
- Events follow the CloudEvents 1.0 spec and are published to a topic named `quantum.{ENVIRONMENT}.function-usage.v1`. Publishing is gated by an `EVENT_STREAMS_ENABLED` feature flag — when disabled, a no-op client logs at DEBUG level instead. 
- Publish failures block the DB status update and are retried on the next scheduler iteration.
- ⚠️ Consumers should handle duplicate events keyed on job_id + event_type, particularly for job_ended which can be re-emitted if the scheduler restarts between publish and DB commit. It should be that way anywhere since that's one principle for Event Driven architectures but just in case to take it into account
- There is a DB migration to add `running_started_at` field. We need to know when the job starts to run in order to keep it as base for the calculated delta we will be publishing as usage. As a consequence, each job will sabe a timestamp on when it changed to running status

### More context
Still missing other events to be published. This PR just covers the Kafka integration and the very basic publishing, let's take it as reference for next iterations.

It also misses adding an instance in docker-compose to make acceptance tests. Didn't want to add it at first to avoid make the PR even bigger.


marceloamaral (commented):
Fleets will also need the compute profile (actually the allocated resources, number of CPU, RAM and GPUs) to calculate the cost. Even though this might be out of scope for this PR, it would be good to start thinking about how this information will be added later...

ismaelRozasRamallal:
> Fleets will also need the compute profile (actually the allocated resources, number of CPU, RAM and GPUs) to calculate the cost. Even though this might be out of scope for this PR, it would be good to start thinking about how this information will be added later...

Thanks for the comment Marcelo!! you're right, there is a potential metric there which can be good to use. The compute profile can be also took as metric for complexity, but we need to figure out if we can use since it's easy the "cheat" (since the profile is decided when adding a function, we need to avoid somebody deciding to pick up the highest compute profile to force users paying more on their function execution)

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -0,0 +1,127 @@
...
+            "source": "qiskit-serverless/scheduler/fleets",
+            "type": self.topic,
+            "time": now.isoformat(),
+            "subject": str(job.id),
+            "datacontenttype": "application/json",
+            "data": {
+                "event_type": event_type,
+                "usage_nanoseconds": usage_nanoseconds,
+                "instance_crn": job.instance_crn,
+                "function_id": str(job.id),
```
ElePT, line 89: It's totally fine to keep the current data as-is as a first approximation, but we should iterate over what data we want to send as part of these events (i.e. compute profile, user id...)

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -152,8 +174,13 @@ def run(self):
...
                 return
-            if self.update_job_status(job):
-                counter += 1
+            try:
+                if self.update_job_status(job):
+                    counter += 1
+            except Exception:  # pylint: disable=broad-exception-caught
+                logger.exception(
+                    "job_id=%s Failed to publish event, skipping DB update — will retry next iteration", job.id
+                )
```
ElePT, line 185: We should keep an eye on how often this fails and how many retries take place, as every missed iteration adds one second of "paid" runtime for the user. I don't know how likely a kafka failure actually is, so for a first iteration it might not make sense to over-engineer this feature but we should be careful with these retries. If we see that they happen, we might need to keep track of the retries and add some early failure mechanism (right now the only safeguard is the 24h TTL).

ElePT (commented):
This LGTM as a first approach, we discussed offline adding more logs as well as the specific comments I left in the review. Are the logs something you'd like to implement as part of this PR @ismaelRozasRamallal? On another note, if we were to merge this to main, is there an environment variable we can use in our deployment to turn this messaging off?

gateway/core/ibm_cloud/event_streams/event_streams_client.py
```diff
@@ -0,0 +1,127 @@
...
+    def emit_job_in_progress(self, job) -> None:
+        """Log job_in_progress at DEBUG level (no-op)."""
+        logger.debug("job_id=%s [noop] emit_job_in_progress", job.id)
+
+    def emit_job_ended(self, job) -> None:
+        """Log job_ended at DEBUG level (no-op)."""
+        logger.debug("job_id=%s [noop] emit_job_ended", job.id)
+
+
+class IBMEventStreamsClient:
```
korgan00: We can create an ABC class to work as an interface so we can return that in `UpdateFleetsJobsStatuses.event_streams_client`. Also, it is a clean implementation in case that we should change the event streams client in a future.

gateway/core/ibm_cloud/event_streams/event_streams_client.py
```diff
@@ -0,0 +1,127 @@
...
+import json
+import logging
+import os
+import uuid
+from datetime import datetime, timezone
+
+try:
+    from confluent_kafka import Producer
+except ImportError:  # confluent-kafka is optional at import time
+    Producer = None  # type: ignore[assignment,misc]
```
korgan00: Can we do this inline inside the `IBMEventStreamsClient.__init__` ? To me it sounds cleaner

korgan00 (dismissed):

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
ismaelRozasRamallal: Completely agree on that. That's pointed as one of the next steps in the roadmap specified in [[Epic]: Billing MVP Part 1](https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1457#top). I plan to add it in a next PR 👍

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
ismaelRozasRamallal: Very good point, and also made me to realize what happens actually with stuck jobs before being stopped with the TTL...

Added as point to improve in the next cycle since by the moment it doesn't imply anything to the user because of being in the experimental phase, does it sound good for you?

gateway/core/ibm_cloud/event_streams/event_streams_client.py
ismaelRozasRamallal: I would challenge on this and just make the import just a regular hard dependency since it will be part of the core business logic in the short term 👍 (however, I don't see a case where the import should fail 😄 )

gateway/core/ibm_cloud/event_streams/event_streams_client.py
ismaelRozasRamallal: worth doing, good point!!

ismaelRozasRamallal:
> This LGTM as a first approach, we discussed offline adding more logs as well as the specific comments I left in the review. Are the logs something you'd like to implement as part of this PR @ismaelRozasRamallal? On another note, if we were to merge this to main, is there an environment variable we can use in our deployment to turn this messaging off?

Thanks for the fast review!!! regarding your points:
- I added an action point in the roadmap to analyze which is the best way to have more visibility on what's happening. Logs are one option but for me it seems a cleaner approach to use telemetry on that logic. I will take a look on that in the next iterations 👍 
- regarding the env variable, yes, we have an env variable which switches between the mock emitter (the one pushing to IBM Cloud logs instead) and the real one 👍 . You can check it here: https://github.com/Qiskit/qiskit-serverless/pull/2209/changes#diff-325735f50cc3d4bbcccbd9a4e1179f5d197d7130cc2f8a22ada6e59c2ffb73adR34

gateway/core/ibm_cloud/event_streams/event_streams_client.py
ismaelRozasRamallal: Done here! [bfa8ebe](https://github.com/Qiskit/qiskit-serverless/pull/2209/commits/bfa8ebe74623533154e14e1c12931d43a53f4120)

At the end, I separated them in different files in the same way we are doing with runners (abstract_runner.py, fleets_runner.py, ray_runner.py)

gateway/core/ibm_cloud/event_streams/event_streams_client.py
ElePT: I'd say this is a pretty common import handling pattern, on my side I am fine keeping it as-is.

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
ElePT: Sounds good.

ElePT (approved):
Thanks for the fast iteration!!! LGTM

PR #2210, opened 2026-06-09
Add mount invariant test for FleetJobPaths COS-to-container paths

avilches:
### Summary

Adds `test_cos_to_container_mount_invariant` to `test_fleet_utils.py` to verify that every `(cos_key, cos_prefix)` pair in `FleetJobPaths` resolves to the expected container path after PDS mounting.

The invariant: if a COS key is `{prefix}/{filename}` and `prefix` is mounted at `mount_point`, the container path must be `mount_point/{filename}`. Currently nothing catches a change to one side (e.g. renaming a COS prefix) without updating the other (the container path), so this test closes that gap.

The test covers all 9 relevant pairs across both custom jobs (4 cases) and provider jobs (5 cases, including the private log) in a single table-driven assertion:

```python
for label, cos_key, prefix, mount, container in cases:
    relative = cos_key.removeprefix(prefix + "/")
    assert container == f"{mount}/{relative}", f"mount invariant failed for '{label}'"
```

PR #2211, opened 2026-06-09
Derive container_* paths from cos_* via _paired helper

avilches:
### Summary

Refactors `build_custom_job_paths` and `build_provider_job_paths` so that `container_*` fields are no longer assigned independently from their corresponding `cos_*` fields.

Before, each pair was written by hand — it was easy to update a COS prefix and forget its container counterpart. Now a private helper `_paired` generates both from the same inputs:

```python
def _paired(cos_prefix: str, mount_point: str, filename: str) -> tuple[str, str]:
    """Return (cos_key, container_path) for a file at cos_prefix/filename mounted at mount_point."""
    return f"{cos_prefix}/{filename}", f"{mount_point}/{filename}"
```

Usage in the builders:

```python
cos_function_entrypoint, container_function_entrypoint = _paired(
    cos_user_function_data, FUNCTION_USER_DATA_PATH, job.program.entrypoint
)
```

Three inputs, two outputs: cos_key and container_path are two views of the same file, generated together so they can't diverge. The `FleetJobPaths` dataclass and all callers are unchanged.

The invariant test added in the base branch (`test_cos_to_container_mount_invariant`) now serves as a regression guard rather than a correctness check — inconsistency is impossible by construction.


avilches:
I'm not sure if this is an improvement or it just makes the code harder to read. I think I prefer the current version, hardcoding the container_* fields by our selves, and keep only the #2210 test only...

PR #2222, opened 2026-06-10
Fix scheduler marking jobs FAILED on CE API rate limit

marceloamaral:
## Summary

- When the CE API returns 429 (Too Many Requests), `FleetsRunner.status()` correctly returns `None` to signal "retry later"
- `UpdateFleetsJobsStatuses.update_job_status()` didn't handle `None` — it fell into the `else` branch which called `to_terminal(job, Job.FAILED)` with "Unknown new job status: None"
- This permanently killed jobs that were actually running fine in Code Engine

## Fix

Added a `None` guard before the status if/elif chain. When status is `None`, the job stays in its current state and will be retried on the next scheduler cycle.

## Test plan

- [x] Added `test_none_status_skips_update_and_returns_false` — RUNNING job stays RUNNING
- [x] Added `test_none_status_preserves_pending_state` — PENDING job stays PENDING
- [x] All 16 scheduler tests pass
- [x] Pylint 10/10, black clean, 89% coverage

### DIFF
--- a/gateway/scheduler/tasks/update_fleets_jobs_statuses.py
+++ b/gateway/scheduler/tasks/update_fleets_jobs_statuses.py
@@ -41,6 +41,10 @@ class UpdateFleetsJobsStatuses(SchedulerTask):
             self.to_terminal(job, Job.FAILED)
             return False
 
+        if new_status is None:
+            logger.debug("job_id=%s status poll returned None (rate-limited), skipping update", job.id)
+            return False
+
         if new_status == Job.SUCCEEDED:
             self.to_terminal(job, Job.SUCCEEDED)
             self._record_execution_duration(job)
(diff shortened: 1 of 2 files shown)


avilches (approved):

Tansito:
Let me convert this to draft as we agreed to try to have implemented #2223 in favor of this one

marceloamaral:
Actually, #2223 already has its own fix for this:

[https://github.com/Qiskit/qiskit-serverless/pull/2223/changes#diff-325735f50cc3d4bbcccbd9a4e1179f5d197d7130cc2f8a22ada6e59c2ffb73adR44](https://github.com/Qiskit/qiskit-serverless/pull/2223/changes#diff-325735f50cc3d4bbcccbd9a4e1179f5d197d7130cc2f8a22ada6e59c2ffb73adR44)

Since it also has a case where the status does not exist and returns `None`, it will need to handle this anyway.

So I think we can close this PR.

PR #2223, opened 2026-06-10
Replace CE API polling with COS-based fleet status detection

marceloamaral:
## Summary

- `FleetsRunner.status()` now reads task state from the PDS COS bucket where CE Fleets writes status objects (`queue/pending`, `queue/running`, `queue/succeeded`, `queue/failed`)
- Eliminates per-job CE API calls that caused 429 rate limits under concurrency
- COS `ListObjects` with `max_keys=5` returns in 2-5ms per job with no rate limit
- CE API retained as a 30s-interval fallback only for crash detection (containers that die before writing to COS)
- More precise status tracking: detects RUNNING state from COS (previously jobs often jumped from INITIALIZING straight to DONE/ERROR)

## Depends on

- #2222 (handles `None` from rate-limited CE API fallback)

## Changes

- `core/services/runners/fleets_runner.py` — rewrote `status()` with COS-first detection, CE API as throttled fallback
- `core/ibm_cloud/cos/cos_client.py` — added `max_keys` param to `list_keys()` for single-call mode
- `core/ibm_cloud/code_engine/fleets/cos.py` — passes `max_keys` through
- `main/settings.py` — added `FLEETS_CE_API_FALLBACK_INTERVAL` (default 30s)

## Test plan

- [x] 64 unit tests pass (10 new COS status tests)
- [x] Integration test (`submit_fleets_job.py custom`) — job reaches DONE via COS detection
- [x] Scheduler logs confirm: `Fleet [...] COS status: SUCCEEDED` with zero CE API status calls
- [x] RUNNING state correctly detected mid-execution (QUEUED → INITIALIZING → RUNNING → DONE)

### DIFF
--- a/gateway/core/services/runners/fleets_runner.py
+++ b/gateway/core/services/runners/fleets_runner.py
@@ -12,8 +12,6 @@
 
 """Runner for executing jobs on IBM Code Engine Fleets."""
 
-from __future__ import annotations
-
 import json
 import logging
 import re
@@ -24,6 +22,7 @@ from io import BytesIO
 
 from django.conf import settings
 from django.template.loader import get_template
+from ibm_botocore.exceptions import ClientError
 from core.ibm_cloud.code_engine.ce_client.rest import ApiException
 
 from core.models import Job, CodeEngineProject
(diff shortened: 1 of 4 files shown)


gateway/core/services/runners/fleets_runner.py
```diff
@@ -260,27 +259,94 @@ def submit(self) -> None:
...
+
+        Returns:
+            Job status constant or ``None`` if no state found yet.
+        """
+        if not self._project:
+            self._project = self._get_project()
+
+        bucket = self._project.cos_bucket_task_store_name
+        if not bucket:
+            return None
```
ElePT: We think this should raise an exception.

gateway/core/services/runners/fleets_runner.py
```diff
@@ -260,27 +259,94 @@ def submit(self) -> None:
...
+
+        prefix = f"ce/{self._project.project_id}/{self.job.fleet_id}/v2/queue/"
+        keys = self._list_task_state_keys(bucket, prefix)
+        if not keys:
+            return None
```
ElePT, line 300: keys should exist

gateway/core/services/runners/fleets_runner.py
```diff
@@ -260,27 +259,94 @@ def submit(self) -> None:
...
+
+        joined = " ".join(keys)
+        for pattern, status in self._COS_STATUS_PRIORITY:
+            if pattern in joined:
+                logger.debug("Fleet [%s] COS status: %s", self.job.fleet_id, status)
+                return status
+
+        return None
+
```
ElePT, line 309: error

gateway/main/settings.py
```diff
@@ -265,6 +265,7 @@
 LIMITS_MAX_CLUSTERS = int(os.environ.get("LIMITS_MAX_CLUSTERS", "6"))
 LIMITS_GPU_CLUSTERS = int(os.environ.get("LIMITS_MAX_GPU_CLUSTERS", "1"))
 LIMITS_MAX_FLEETS = int(os.environ.get("LIMITS_MAX_FLEETS", "1000"))  # Fleets Project limit
+FLEETS_CE_API_FALLBACK_INTERVAL = int(os.environ.get("FLEETS_CE_API_FALLBACK_INTERVAL", "30"))
```
avilches: This is not used in the code, you can delete it right?

gateway/core/services/runners/fleets_runner.py
```diff
@@ -261,44 +260,79 @@ def submit(self) -> None:
...
 
-        try:
-            fleet_status = self._get_handler().get_job_status(self.job.fleet_id)
-            raw = fleet_status.get("status")
-            mapped = self._map_fleet_status(raw) if raw else None
-            logger.debug("Fleet [%s] status: %s → %s", self.job.fleet_id, raw, mapped)
-            if mapped in (Job.FAILED, Job.STOPPED):
-                logger.warning("Fleet [%s] terminal status [%s] raw=[%s]", self.job.fleet_id, mapped, raw)
-            return mapped
+        return self._check_cos_task_state()
```
avilches: Why not merge the 2 funcs?

gateway/core/services/runners/fleets_runner.py
avilches: why resolved?

gateway/core/services/runners/fleets_runner.py
avilches: why resolved?

gateway/core/services/runners/fleets_runner.py
avilches: why resolved?

gateway/core/services/runners/fleets_runner.py
marceloamaral: Good point, after the code simplification, it does not make sense to have 2 funcs anymore

gateway/core/services/runners/fleets_runner.py
marceloamaral: Added the exception!

gateway/core/services/runners/fleets_runner.py
marceloamaral: Here we should return None since the CE could take 10-15s after fleet creation to write the first queue/ key, so in the beginning there is no status.

gateway/core/services/runners/fleets_runner.py
marceloamaral: It is already raising an error

gateway/core/services/runners/fleets_runner.py
Tansito: For now I think it's fine but @marceloamaral can you create an issue in our internal repo to keep an eye on this and review this logic to see if we can improve it?

I suppose we should be aware in case this data is never created and avoid the job waiting eternally for it (maybe just a TTL as we have in Ray keeping track of those jobs that can be blocked by X reason).

Tansito (approved):
Good work here @marceloamaral !

PR #2224, opened 2026-06-10
Fix COS presigned URLs to use public endpoint

avilches:
Presigned URLs generated by the gateway in `/log` and `/result` were embedding the private COS endpoint (`s3.direct.{region}...`), which is only reachable from within IBM Cloud. ServerlessClient can't access to this url.

`generate_presigned_url` now uses a dedicated S3 client (`_s3_presigned`) always configured with the public COS endpoint (`s3.{region}...`). The existing `_s3_hmac` client is unchanged and continues to use whichever endpoint is configured via `CE_COS_USE_PUBLIC_ENDPOINT` for internal gateway operations.


ElePT (approved):
LGTM. But would appreciate review from @marceloamaral too in case I missed sth.

marceloamaral (approved):
LGTM

PR #2225, opened 2026-06-10
Replace CE API-based is_active() with fleet id check

marceloamaral:
## Summary

- `is_active()` now calls `self.status()` instead of the CE API
- A fleet is active once it moves past pending (worker claimed task, logs may appear)
- No CE API calls for activity checks — reuses COS-based status detection
- Cache still prevents repeated COS calls within 30s

## Depends on

- #2223 (COS-based fleet status detection)

## Test plan

- [x] 53 unit tests pass
- [x] Tests cover: running→True, succeeded→True, pending→False, no keys→False, cache hit

### DIFF
--- a/gateway/core/services/runners/fleets_runner.py
+++ b/gateway/core/services/runners/fleets_runner.py
@@ -17,7 +17,6 @@ import logging
 import re
 import tarfile
 import time
-from collections import OrderedDict
 from io import BytesIO
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


marceloamaral:
This PR depends on #2223

gateway/core/services/runners/fleets_runner.py
```diff
@@ -129,13 +129,14 @@ def disconnect(self) -> None:
...
 
     def is_active(self) -> bool:
-        """Return ``True`` if the fleet exists in Code Engine.
+        """Return ``True`` if the fleet has moved past pending.
 
-        Results are cached for 30 seconds (class-level :class:`TTLCache`)
-        to avoid redundant API calls when the scheduler polls frequently.
+        A fleet is active once the worker has claimed the task — meaning
+        COS logs may start appearing. Uses ``status()`` which reads COS
+        task state without hitting the CE API.
```
ElePT: We need to be a bit careful with this behavior change, I think it's not the same to check if the fleet was created than to check if it's moved past pending, right? I am asking because we have use cases like `StopJobUseCase` that check if a runner is active before stopping the job. If we implement this change, would this mean that pending fleets can't be stopped? There might be other instances like this.

ElePT (commented):
Thanks for looking into this, I am in favor of the COS check, but I think we need to be careful and make sure we understand the implications in the behavior first.

gateway/core/services/runners/fleets_runner.py
marceloamaral: I have simplified the function as we discussed

ElePT (approved):
LGTM

PR #2227, opened 2026-06-11
Fix sync_ce_project crash with duplicate zone rows

marceloamaral:
## Summary

- `update_or_create` raised `MultipleObjectsReturned` in production when the DB had multiple `CodeEngineProject` rows with the same `project_id` but different `zone` values (legacy per-zone project configuration)
- Now deletes all existing rows for a `project_id` and recreates a single clean row without zone
- Zone field is deprecated and no longer used by any code path

## Test plan

- [x] Added 4 unit tests covering: create, update, duplicate removal, missing fields

### DIFF
--- a/gateway/api/management/commands/sync_ce_project.py
+++ b/gateway/api/management/commands/sync_ce_project.py
@@ -52,12 +52,12 @@ def _upsert_project(project_id: str, data: dict) -> bool:
     defaults = {k: data[k] for k in _REQUIRED_KEYS}
     defaults["active"] = True
 
-    _, created = CodeEngineProject.objects.update_or_create(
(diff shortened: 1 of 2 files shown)


Tansito:
we decided to do an update instead to a delete to have more control by now, tomorrow we will try to find the correct fix to this situation

Tansito (approved):

Tansito:
Moved to draft as I think we can deploy this in staging without this fix as this fix is only required in this specific script. So we can just remove it by now and tomorrow we can look at this calmly.

ElePT:
Closing as this contribution became stale after other changes.

PR #2230, opened 2026-06-11
fix: add FLEETS_GATEWAY_HOST to scheduler env

ElePT:
The scheduler's `{{- if .Values.ce }}` Helm env block was missing FLEETS_GATEWAY_HOST, so settings.FLEETS_GATEWAY_HOST fell back to SITE_HOST (http://localhost:8001) in the scheduler process. Since the scheduler calls build_run_env_variables when dispatching fleet jobs, this localhost address was baked into the fleet container env vars instead of the real gateway URL — causing the runtime job association callback to fail with a connection error.

Also wrap the requests.post call in associate_runtime_job_with_serverless_job with a RequestException handler so a misconfigured gateway host logs a warning and returns False instead of crashing the user's job container.


marceloamaral (approved):
LGTM

PR #2231, opened 2026-06-11
Fix list of files

Tansito:
### Summary

This PR is pretty simple, it just fix the output from the list and provider-list end-points for fleets to only return the name of the files not the full path. Example:

```
['providers/provider-name/function-title/data/file.txt'] 
```

to just:

```
['file.txt'] 
```


marceloamaral (approved):
LGTM

avilches (approved):
awesome

PR #2232, opened 2026-06-11
Add COS file explorer in Django Admin

avilches:
Adds a COS file browser accessible from each Job in Django Admin. When viewing a job, a new "Storage" section shows a link that opens a dedicated page listing all files associated with that job, grouped by category.

**What it shows:**
- Data Files: files the user uploaded for the function
- Results, Logs, Arguments: job-specific execution artifacts
- For provider functions: Provider Data and Private Logs sections

**Supported backends:**
- Fleets jobs: files listed from IBM COS via `list_objects_v2` (single API call per prefix, returns size + last modified)
- Ray jobs: files listed from the local mounted filesystem via `os.scandir` / `os.stat`

‼️ Read metadata only: only name, date and size. No download links.


ElePT (approved):
Love the idea. LGTM

PR #2233, opened 2026-06-12
Upload now returns the file name only

korgan00:
### Summary

Hide internal organization path in upload return data.


ElePT (approved):
LGTM, thanks for the quick fix.

PR #2234, opened 2026-06-12
Improve tests files for fleets

Tansito:
### Summary

This PR improves some tests in the fleets files storage tests. It just add a specific check for each path so now we are going to be able to identify easier looking the tests if any path changed unintentionally.

### DIFF
--- a/gateway/tests/core/services/storage/test_file_storage_fleets.py
+++ b/gateway/tests/core/services/storage/test_file_storage_fleets.py
@@ -56,6 +56,28 @@ class TestFileStorageFleets:
         program.save()
         return program
 
+    # ── path generation ────────────────────────────────────────────────────────
+
+    def test_public_folder_key_custom_function(self, function):
+        """_public_folder_key uses custom_functions path when program has no provider."""
+        storage = FileStorageFleets("alice", function)
+        assert storage._public_folder_key == "users/alice/custom_functions/my-program/data"
+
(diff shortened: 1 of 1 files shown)


gateway/tests/core/services/storage/test_file_storage_fleets.py
```diff
@@ -98,7 +120,7 @@ def test_get_public_files_returns_list(self, function):
         assert result == ["file1.txt", "file2.txt"]
         mock_cos.list_keys.assert_called_once_with(
             bucket_name="user-bucket",
-            prefix=storage._public_folder_key,
+            prefix="users/alice/custom_functions/my-program/data",
```
Tansito, line 123: I changed this test to ensure we are not getting the value from the variable because if we change it the test would pass and I want to be sure that if we change these tests to pass is because we are intentionally changing them.

gateway/tests/core/services/storage/test_file_storage_fleets.py
```diff
@@ -129,7 +151,7 @@ def test_get_private_files_returns_list(self, function_with_provider):
         assert result == ["private1.txt", "private2.txt"]
         mock_cos.list_keys.assert_called_once_with(
             bucket_name="provider-bucket",
-            prefix=storage._private_folder_key,
+            prefix="providers/good-partner/my-program/data",
```
Tansito, line 154: Similar to my previous comment.

avilches (approved):

PR #2235, opened 2026-06-12
Carbon Design System theme and Job admin list improvements

avilches:
Applies IBM Carbon Design System styles to the Django admin and improves the Job list view.

**Carbon theme**

Introduces a self-hosted CSS theme (`carbon_theme.css`) loaded via a custom `base_site.html` template. It replaces the stock Django admin look with Carbon White / Gray 100 tokens: IBM Plex Sans and Plex Mono fonts, filled text fields with a bottom border, square corners, a black UI shell header, and Carbon-style tables, buttons, paginator, filters, and sidebar. Dark mode follows the OS preference automatically.

Also tightens the horizontal padding on `<select>` elements so their text is not truncated, and adds a `cursor: pointer` on list rows to signal they are clickable.

**Job admin list**

- New `list_display` showing runner, author, program (`provider / title`), status, created, and last updated.
- `JobProgramFilter` sidebar filter that groups jobs by provider / program title; jobs whose program has no provider (or no program at all) are grouped under **Custom**.
- Clicking anywhere on a row navigates to the job detail page (via a small JS snippet loaded through `Media`), instead of only the first column being a link.


ElePT (approved):
Cannot comment too much on the content but happy to merge and try in staging.

PR #2236, opened 2026-06-15
refactor: split runtime_jobs endpoint into GET and POST handlers

avilches:
`runtime_jobs` was a single view function handling both GET and POST with an internal `if request.method ==` branch.

This PR separates them into two dedicated files (`get_runtime_jobs.py` and `associate_runtime_jobs.py`), matching the pattern used by every other endpoint in the module.

To support this, `@endpoint` now accepts a `method=` parameter and `RouteRegistry` accumulates per-method handlers, building a single combined dispatcher automatically at URL-load time. Swagger annotations are merged so the generated schema is unchanged.

Existing single-method endpoints are unaffected.

Validation: swagger schema before and after is identical, all 722 tests pass.


ElePT (dismissed):
I like this idea. This is something I tried to implement in the original runtime jobs PR (as discussed with @Tansito) and never found a good solution. So on my side, LGTM (I will not merge in case anyone else has a different opinion).

Tansito (approved):
LGTM

gateway/api/v1/route_registry.py
```diff
@@ -24,12 +27,55 @@ def register(cls, url_path: str, view_func: Callable, name: str):
...
 
-        route = path(clean_path, view_func, name=name)
-        cls._routes.append(route)
+        if method is None:
+            cls._routes.append(path(clean_path, view_func, name=name))
+            return
+
+        if clean_path not in cls._method_handlers:
+            cls._method_handlers[clean_path] = {}
+        cls._method_handlers[clean_path][method.upper()] = (view_func, name)
```
korgan00, line 36: We can check here if the combination url-method already exist and throw an exception to prevent some errors. This will break on start so tests will fail.

korgan00 (commented):
I'm a bit late, sorry. Very good PR! I just have a minor improvement to make development more confident.

PR #2237, opened 2026-06-16
Remove pyarrow

korgan00:
### Summary

Removing pyarrow since it doesn't seem to be used anymore and it is causing a PSIRT issue.
First time added in https://github.com/Qiskit/qiskit-serverless/pull/1538


ElePT (approved):
LGTM. @korgan00 could you add a PR description explaining why we removed it (linking to the original PR adding the dependency)

PR #2239, opened 2026-06-16
Serverless ibm client fail fast mechanics and runtime API expose

ItamarGoldman:
## Summary
This PR exposes IBM Quantum backend methods in IBMServerlessClient and adds comprehensive pre-flight validation to run() to provide early error detection and better user experience.

- Exposed service useful methods in the client for the user.
- Overrode `IBMServerlessClient.run()` with three pre-flight checks that fail fast before submitting to the gateway
- Added two env-overridable constants to control usage quota thresholds
- Added tests covering all new methods and edge cases
- Added release notes

## Motivation
Previously, calling `function.run()` against an instance with no backends, a restricted backend, or exhausted quota would succeed at submission but then hang silently waiting for a runtime slot that would never become available.
These checks surface those failures immediately on the client side with a clear error message.

## Checklist

- [x] `backends(**kwargs)` — lists backends via `QiskitRuntimeService.backends()`, scoped to `self.instance`, and warms a per-instance cache.
- [x] `run()` pre-flight: usage quota — raises when remaining runtime is at or below `USAGE_ZERO_EPSILON_SECONDS` (default 1s); warns when below `USAGE_LOW_THRESHOLD_SECONDS` (default 600s); degrades gracefully on transient errors.
- [x] run() pre-flight: backend access — when `arguments["backend_name"]` is set, validates via a single-backend lookup (`_service.backend(name)`), which is cheaper than a full listing and simultaneously confirms access.
- [x] `run()` pre-flight: instance has backends — when no `backend_name` is given, calls `backends()` (cached) to confirm the instance is not empty.
- [x] `_backends_cache` is instance-level and lazily populated, so no network call is added to construction 
- [x] Threshold constants (`USAGE_LOW_THRESHOLD_SECONDS`, `USAGE_ZERO_EPSILON_SECONDS`) tunable via environment variables

### Notes
`qiskit_ibm_runtime` v0.46+ does not expose a max-jobs or job-count quota API. The usage check is time-based (`usage_remaining_seconds` from `service.usage()`), which is the closest available proxy. The threshold defaults are provisional and can be tuned without a code change via environment variables.


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+        self._backends_cache = {}
+
+        for backend in backend_list:
+            self._backends_cache[backend.name] = backend
+
+        return backend_list
+
+    def backend(
+        self,
+        name: str,
```
HuangJunye: The [`backend` method](https://github.com/Qiskit/qiskit-ibm-runtime/blob/c6b352c9b6b7f684796c088ab64c6ba6a19b9938/qiskit_ibm_runtime/qiskit_runtime_service.py#L917-L922) in `qiskit-ibm-runtime` has `instance` as well. Can you please add that?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+            raise QiskitServerlessException(f"Failed to retrieve backend '{name}': {exc}") from exc
+
+        self._backends_cache[name] = backend
+        return backend
+
+    def least_busy(
+        self,
+        min_num_qubits: int | None = None,
+        instance: str | None = None,
+        filters: Callable[[IBMBackend], bool] | None = None,
```
HuangJunye, line 887: The [`least_busy` method](https://github.com/Qiskit/qiskit-ibm-runtime/blob/c6b352c9b6b7f684796c088ab64c6ba6a19b9938/qiskit_ibm_runtime/qiskit_runtime_service.py#L1281-L1288) in `qiskit-ibm-runtime` has `use_fractional_gates`, can you please add here as well?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+                min_num_qubits=min_num_qubits, instance=instance, filters=filters, **kwargs
+            )
+        except QiskitBackendNotFoundError as exc:
+            raise QiskitServerlessException(
+                f"No available backend matches the criteria with instance '{self.instance}'. "
+                f"Call client.backends() to list accessible backends."
+            ) from exc
+        except Exception as exc:
+            raise QiskitServerlessException(f"Failed to retrieve backend: {exc}") from exc
+        self._backends_cache[backend.name] = backend
```
HuangJunye, line 923: Is this to keep the cache more fresh/up to date?

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+                f"(remaining: {remaining:.2f}s). "
+                "The job would wait indefinitely for runtime resources. "
+                "Check your instance quota at https://quantum.cloud.ibm.com/instances."
+            )
+
+        if not supress_low_usage_warning and remaining is not None and remaining <= USAGE_LOW_THRESHOLD_SECONDS:
+            warnings.warn(
+                f"Instance '{self.instance}' has low remaining runtime quota "
+                f"({remaining:.0f}s remaining, threshold: {USAGE_LOW_THRESHOLD_SECONDS:.0f}s). "
+                "Consider checking your quota before submitting long-running jobs. "
```
HuangJunye, line 981: Now I remember, I think `qiskit-ibm-runtime` emits a warning for low usage. Perhaps propagate that instead of implementing our own. See here: https://github.com/Qiskit/qiskit-ibm-runtime/blob/c6b352c9b6b7f684796c088ab64c6ba6a19b9938/qiskit_ibm_runtime/qiskit_runtime_service.py#L1222-L1242

HuangJunye (commented):
First past of the review. I haven't reviewed the run method and tests yet. I will review throughout the day. Please take a look at the existing comment. Thank you!

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: I didn't want to use the instance since the caching of backends is depended on instance (what backends are available) so I thought that if someone would like to use 'instance', the user should either instantiate new client or use `qiskit_ibm_runtime` directly.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Yes. I thought we get it for free so why not.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: You are right that `qiskit-ibm-runbtime`. It just fire a warning in case the usage is 0. I can catch the warning and raise an exception if it fires but we would still need almost all the logic for the "low usage" warning.
Would it better to catch the warning if it fires and raise a 'QiskitServerlessException' and handle the low usage or to keep it as it is now?

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Done! :)

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: I think I'm with @ItamarGoldman here. Unlike `qiskit-ibm-runtime,` we require an instance for instantiating the client, so there isn't really flexibility to choose an instance. I think that allowing to query backends for a different instance would mostly bring confusion about what is accessible on the function side.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: My 2 cents: it's better to catch the runtime warning and expose it as a qiskit serverless exception to keep consistency on the errors raised on our side. But if the runtime warning is just for the usage 0 case, then we would directly blocking the job, right? So we still have the open question around raising a warning when a low remaining usage is left.

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: I'm not against adding the option for consistency, but the fractional gate support in functions is not very clear, as it depends on vendors implementing a path for it (because we can't serialize the backend directly).

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: The down side of catching the warning and not calculate ourselves is that there will be no way to check the epsilon usage.
If the user need a minimum of 2 min of QPU time to calculate something, but he has only 1 min usage time, we will not tell him before he send it for execution that he doesn't have enough usage time and in the better case get rejected from queue or worst case, will get only the usage time he has remaining and then fail leading to a waste of money.

korgan00 (dismissed):
LGTM
The serverless file is a big one, we should think how to make it smaller in a future refactor

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+        name: str | None = None,
+        min_num_qubits: int | None = None,
+        instance: str | None = None,
+        dynamic_circuits: bool | None = None,
+        filters: Callable[[IBMBackend], bool] | None = None,
+        *,
+        use_fractional_gates: bool | None = False,
+        calibration_id: str | None = None,
+        **kwargs: Any,
+    ) -> list[IBMBackend]:
```
ElePT, line 793: Sorry in advance because I am going to give conflicting feedback that may not match what we had discussed before: After looking at the PR again, I would only list as explicit args those that are useful for us. This is: `refresh_cache`, `name`, `min_num_qubits` and `filters`. I don't see the point in explicitly documenting inputs like "instance" when we can't effectively have more than 1 active CRN in this client, or dynamic_circuits (which is in fact a stale argument). If we document these, we are now responsible for maintaining them, which we don't necessarily want to do, and all methods would anyways be available through `**kwargs`, so not documenting them doesn't stop someone from using them if they really really want to. So my vote is to keep the bare minimum here.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+                    client.backends(filters=lambda b: b.max_shots > 50000)
+                    client.backends(filters=lambda x: "rz" in x.basis_gates)
+
+            use_fractional_gates: If ``True``, include backends with fractional gates.
+                Note that backends now support dynamic circuits and fractional gates
+                simultaneously. Control flow instructions are not removed when this
+                flag is set to ``True``. If ``None``, both fractional gates and
+                control flow operations are included.
+            calibration_id: The calibration ID for instantiating the backend. Should only
+                be used when selecting a single backend.
```
ElePT: I don't think we should be documenting these, as stated above.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+            **kwargs: Simple filters that require a specific value for an attribute in
+                backend configuration or status.
+                Examples::
+
+                    # Get operational real backends
+                    client.backends(simulator=False, operational=True)
```
ElePT: `simulator` is a stale argument, it can't be used now

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+
+                    # Get backends with at least 127 qubits
+                    client.backends(min_num_qubits=127)
+
+                    # Get backends that support OpenPulse
+                    client.backends(open_pulse=True)
```
ElePT: `open_pulse` is a stale argument, it can't be used now

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+
+                For the full list of backend attributes, see the `IBMBackend class documentation
+                <https://quantum.cloud.ibm.com/docs/api/qiskit-ibm-runtime>`_
```
ElePT: This is good, it covers all the things we don't want to document ourselves.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+        for backend in backend_list:
+            self._backends_cache[backend.name] = backend
+
+        return backend_list
+
+    def backend(
+        self,
+        name: str,
+        use_fractional_gates: bool | None = False,
+        calibration_id: str | None = None,
```
ElePT: same comment as above, I would replace these 2 with just **kwargs

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+        except Exception as exc:
+            raise QiskitServerlessException(f"Failed to retrieve backend '{name}': {exc}") from exc
+
+        self._backends_cache[name] = backend
+        return backend
+
+    def least_busy(
+        self,
+        min_num_qubits: int | None = None,
+        instance: str | None = None,
```
ElePT: I'd also remove instance here

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: I agree with the current implementation

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,320 @@ def save_account(
...
+    ) -> "Job":
+        """Run a Qiskit Function with pre-flight validation before submitting to the gateway.
+
+        Checks are performed in order before delegating to ``ServerlessClient.run``:
+
+        1. **Usage quota**: raises when the instance has no remaining runtime; warns when low.
+        2. **Backend access** (when ``arguments["backend_name"]`` is set): validates via a
+           single-backend lookup — cheaper than a full listing, and sufficient to confirm access.
+        3. **Instance has backends** (when no ``backend_name`` is given): calls ``backends()``
+           (cached after the first invocation) to ensure the instance is not empty.
```
ElePT: This part of the docstring looks focused to a developer, rather than users. There is no need to over-explain how things work, just what the behavior will be.

ElePT (commented):
Thanks a lot @ItamarGoldman!! Most of the implementation looks good to me, but I still have a couple of final comments regarding the new methods that were added. Once these are addressed, LGTM.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: I added `**kwargs` to the backend and removed the unwanted args. This will allow us to keep the functionality without the maintance.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: removed this example.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: aklso removed this example.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: changed as suggested.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: True. Change to the following:

```python
"""Run a Qiskit Function with pre-flight validation before submitting to the gateway.

        Checks are performed before delegating to ``ServerlessClient.run``. checks include  instance has no 
        remaining runtime; warns when low and check accessibility to the backed 
        Args:
        ...
        """
```

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Changed as suggested.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: I agree with how you are envisioning the client. We should only put things that benefit what we want to do with the client. For complex and detailed usage, the user should use `qiskit-ibm-runtime`.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+
+            **kwargs: Simple filters that require a specific value for an attribute in
+                backend configuration or status.
+                Examples::
+
+                    # Get backends with at least 127 qubits
+                    client.backends(min_num_qubits=127)
+
+                For the full list of backend attributes, see the `IBMBackend class documentation
+                <https://quantum.cloud.ibm.com/docs/api/qiskit-ibm-runtime>`_
```
HuangJunye: Add a link directly to the corresponding`.backeds` method in `QiskitRuntimeService`: https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#backends

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+        a targeted backend lookup. This is more efficient than listing all backends
+        and validates that the caller has access to the requested backend. The cache
+        is updated on every call to reflect current access permissions.
+
+        Args:
+            name: Name of the backend (e.g., ``"ibm_torino"``).
+            **kwargs: Simple filters that require a specific value for an attribute in
+                backend configuration or status.
+                For the full list of backend attributes, see the `IBMBackend class documentation
+                <https://quantum.cloud.ibm.com/docs/api/qiskit-ibm-runtime>`_
```
HuangJunye: Add a link directly to the corresponding`.backed` method in `QiskitRuntimeService`: https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#backends

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+                f"({remaining:.0f}s remaining, threshold: {USAGE_LOW_THRESHOLD_SECONDS:.0f}s). "
+                "Consider checking your quota before submitting long-running jobs. "
+                "Check https://quantum.cloud.ibm.com/instances for details."
+            )
+
+    def run(  # pylint: disable=too-many-positional-arguments
+        self,
+        program: Union[QiskitFunction, str],
+        arguments: Optional[Dict[str, Any]] = None,
+        config: Optional["Configuration"] = None,
```
HuangJunye:
```suggestion
        config: Optional[Configuration] = None,
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+        provider: Optional[str] = None,
+        *,
+        compute_profile: Optional[str] = None,
+        supress_low_usage_warning: bool = False,
```
HuangJunye:
```suggestion
        suppress_low_usage_warning: bool = False,
```

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+        Checks are performed before delegating to ``ServerlessClient.run``. Checks include instance has no
+        remaining runtime; warns when low and check accessibility to the backed
+        Args:
+            program: ``QiskitFunction`` object or function title string.
+            arguments: Runtime arguments. The optional ``backend_name`` key is inspected for
+                the backend access check but forwarded unchanged.
+            config: Optional execution configuration.
+            provider: Optional provider name override.
+            compute_profile: Optional compute-profile name.
+            supress_low_usage_warning: If ``True``, suppress the warning when remaining runtime
```
HuangJunye:
```suggestion
            suppress_low_usage_warning: If ``True``, suppress the warning when remaining runtime
```

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+    @patch(_CONFIG_FILE)
+    def test_get_backend_raises_on_not_found(self, mock_file_path, mock_verify, mock_list_instances):
+        """QiskitBackendNotFoundError is converted to QiskitServerlessException with a clear
+        message that mentions the backend name and instance."""
+        client = _make_client(mock_file_path, mock_verify, mock_list_instances)
+        client._service.backend = MagicMock(side_effect=QiskitBackendNotFoundError("not found"))
+
+        with pytest.raises(
+            QiskitServerlessException,
+            match="Backend 'nope' is not available or you do not have access",
```
HuangJunye:
```suggestion
            match="Backend 'ibm_valencia' is not available or you do not have access",
```

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+        ):
+            client.backend("nope")
```
HuangJunye:
```suggestion
            client.backend("ibm_valencia")
```

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            client._check_usage()
+
+    @patch(_LIST_INSTANCES)
+    @patch(_VERIFY_CREDS)
+    @patch(_CONFIG_FILE)
+    def test_check_usage_limit_reached_with_no_remaining_key(self, mock_file_path, mock_verify, mock_list_instances):
+        """usage_limit_reached=True but no remaining key → treated as zero → raises."""
```
HuangJunye: Improve description: no `remain_usage..` key in the self.usage() returned dict

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+    def test_run_with_valid_backend_calls_super(self, mock_super_run, mock_file_path, mock_verify, mock_list_instances):
+        """run(backend_name=X) validates the backend then delegates to ServerlessClient.run."""
+        client = self._setup_client(mock_file_path, mock_verify, mock_list_instances)
+        fake_backend = MagicMock()
+        client._service.backend = MagicMock(return_value=fake_backend)
+
+        fake_job = MagicMock()
+        mock_super_run.return_value = fake_job
+
+        result = client.run("my_function", arguments={"backend_name": "ibm_torino"})
```
HuangJunye:
```suggestion
        job = client.run("my_function", arguments={"backend_name": "ibm_torino"})
```

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+
+        client._service.backend.assert_called_once_with(name="ibm_torino")
+        mock_super_run.assert_called_once()
+        assert result is fake_job
```
HuangJunye:
```suggestion
        assert job is fake_job
```

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,414 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+        client._service.backend = MagicMock(return_value=MagicMock())
+        client._service.backends = MagicMock()
+        mock_super_run.return_value = MagicMock()
+
+        client.run("fn", arguments={"backend_name": "ibm_torino"})
+        client.run("fn", arguments={"backend_name": "ibm_torino"})
+        client.run("fn", arguments={"backend_name": "ibm_torino"})
+
+        # service.backend called once per run (refresh on each hit, per spec)
+        assert client._service.backend.call_count == 3
```
HuangJunye: No need to have this test as the previous test covers the same scenario.

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
```diff
@@ -0,0 +1,49 @@
+---
+features:
+  - |
+    ``IBMServerlessClient`` now exposes several methods that wrap the Qiskit IBM Runtime
+    ``QiskitRuntimeService`` API to interact with IBM Quantum backends:
+    
+    * ``backends(**kwargs)`` — Lists IBM Quantum backends accessible through the configured instance.
+      Results are cached per client instance, making repeated calls within a session inexpensive.
+      Use ``refresh_cache=True`` to force a refresh. Wraps ``QiskitRuntimeService.backends()``.
+    * ``backend(name, use_fractional_gates=False, calibration_id=None)`` — Fetches a single backend
```
HuangJunye: remove `use_fractional_gates` and `calibration_id`

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
```diff
@@ -0,0 +1,49 @@
...
+      by name and updates the cache. More efficient than listing all backends and validates access.
+      Wraps ``QiskitRuntimeService.backend()``.
+    * ``least_busy(min_num_qubits=None, instance=None, filters=None, **kwargs)`` — Returns the
```
HuangJunye: remove `instance`

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
```diff
@@ -0,0 +1,49 @@
...
+      backend with the fewest pending jobs that matches the specified criteria. Result is cached.
```
HuangJunye: Result is cached. is not accurate. Calling the method updates the `backends_cache` but it doesn't cache the least busy backend. Each call will query ibm runtime service to get the least busy backend.

HuangJunye (changes requested):
Looks great. Just some minor changes as discussed.

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
ItamarGoldman: Changed to:
```
* ``least_busy(min_num_qubits=None, filters=None, **kwargs)`` — Returns the
      backend with the fewest pending jobs that matches the specified criteria. This update the cache  
      with the retrieved backend. Wraps ``QiskitRuntimeService.least_busy()``.
```

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
ItamarGoldman: I removed `instance` from the exmaple.

releasenotes/notes/ibm-client-run-backend-validation-a9c0e2609474f9bd.yaml
ItamarGoldman: Deleted those options.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Deleted the test.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Done.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Done.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: changed to:
```python
"""usage_limit_reached=True but there is no `remain_usage` key in the return dict → treated as zero → raises."""
```

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Done.

client/tests/core/test_ibm_serverless_client.py
ItamarGoldman: Done.

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Done

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Done

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Done

client/qiskit_serverless/core/clients/serverless_client.py
ItamarGoldman: Done.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+
+            **kwargs: Simple filters that require a specific value for an attribute in
+                backend configuration or status.
+                Examples::
+
+                    # Get backends with at least 127 qubits
+                    client.backends(min_num_qubits=127)
+
+                For the full list of backend attributes, see the `IBMBackend class documentation
+                <https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#backends>`_
```
HuangJunye:
```suggestion
                <https://quantum.cloud.ibm.com/docs/api/qiskit-ibm-runtime/qiskit-runtime-service#backends>`_
```

Removing the language will redirect the page to the language set by the user which is preferred.

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -754,6 +763,281 @@ def save_account(
...
+        a targeted backend lookup. This is more efficient than listing all backends
+        and validates that the caller has access to the requested backend. The cache
+        is updated on every call to reflect current access permissions.
+
+        Args:
+            name: Name of the backend (e.g., ``"ibm_torino"``).
+            **kwargs: Simple filters that require a specific value for an attribute in
+                backend configuration or status.
+                For the full list of backend attributes, see the `IBMBackend class documentation
+                <https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime/qiskit-runtime-service#backends>`_
```
HuangJunye:
```suggestion
                <https://quantum.cloud.ibm.com/docs/api/qiskit-ibm-runtime/qiskit-runtime-service#backend>`_
```

Removing the language will redirect the page to the language set by the user which is preferred. And the link should be to `backend` not `backends`

client/tests/core/test_ibm_serverless_client.py
```diff
@@ -285,3 +318,391 @@ def test_backward_compatibility_with_explicit_ibm_quantum_platform(
...
+
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            client._check_usage()
+
+    @patch(_LIST_INSTANCES)
+    @patch(_VERIFY_CREDS)
+    @patch(_CONFIG_FILE)
+    def test_check_usage_limit_reached_with_no_remaining_key(self, mock_file_path, mock_verify, mock_list_instances):
+        """usage_limit_reached=True but there is no `remain_usage` key in the return dict → treated as zero → raises."""
```
HuangJunye:
```suggestion
        """usage_limit_reached=True but there is no `usage_remaining_seconds` key in the return dict → treated as zero → raises."""
```

The key should be `usage_remaining_seconds`

HuangJunye (changes requested):
Final review. I will apply the requested changes and approve.

HuangJunye (approved):

PR #2251, opened 2026-06-17
Add IBM Cloud Event Streams external secret

ismaelRozasRamallal:
### Summary

This PR adds the Helm infrastructure needed for the scheduler to publish job usage events to IBM Cloud Event Streams (Kafka), as introduced in #2209.

- New kafka-credentials ExternalSecret that fetches the IBM Event Streams service credential JSON blob from IBM Cloud Secret Manager (via the existing qiskit-serverless SecretStore) and transforms it into three discrete keys: api_key, bootstrap_servers, and user.
- Per-environment Secret Manager paths wired in values-non-prod.yaml (staging) and values-gpu-prod.yaml (production).
- EVENT_STREAMS_ENABLED, EVENT_STREAMS_BOOTSTRAP_SERVERS, EVENT_STREAMS_API_KEY, and EVENT_STREAMS_USER injected into the scheduler container. The flag is always present; the secret refs are gated on secrets.kafka being set so the scheduler starts cleanly in environments without Kafka (local, kind, pentest).

For that, we follow the recommended Infrastructure team pattern for external secrets:

- The ExternalSecret is placed in the outer qiskit-serverless chart (same level as gateway-gpu-production-secret-key.yaml) and is only rendered when gateway.secrets.kafka is set in values — keeping local and pentest environments unaffected.
- target.template with engineVersion: v2 and fromJson is used to destructure the credential blob; kafka_brokers_sasl | join "," converts the broker array to the comma-separated string that confluent-kafka expects.
- argocd.argoproj.io/sync-wave: "-1" ensures the K8s Secret exists before the scheduler deployment starts.
- The SecretStore referenced is qiskit-serverless (not billing-service), matching the one already in use for the Django secret key ExternalSecret.
- EVENT_STREAMS_USER is extracted from the credential even though IBMEventStreamsClient currently hardcodes "token" — this makes the secret forward-compatible without a chart change when that is wired up.


gateway/api/migrations/0054_job_running_started_at.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 5.2.14 on 2026-06-08 10:57
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
avilches, line 6: this is not related to this PR right?

avilches: Nevermind, I've seen the field in the models.py

charts/qiskit-serverless/templates/kafka-credentials.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.gateway.secrets.kafka -}}
+apiVersion: external-secrets.io/v1
+kind: ExternalSecret
+metadata:
+  name: kafka-credentials
+  annotations:
+    argocd.argoproj.io/sync-wave: "-1"
```
Tansito, line 7: what this option does, @ismaelRozasRamallal ? I don't know it

ismaelRozasRamallal: This option is to manage the order of the resources to be synched. With it, we ensure this kafka credentials are synched before the rest of project resources, which by default goes in the sync wave 0. It's some kind of pre-hook before synching the application 😄

gateway/core/models.py
```diff
@@ -454,6 +454,7 @@ class Job(models.Model):
 
     account_id = models.CharField(max_length=255, null=True, blank=True)
     instance_crn = models.CharField(max_length=255, null=True, blank=True)
+    running_started_at = models.DateTimeField(null=True, blank=True)
```
ElePT, line 457: As discussed offline @ismaelRozasRamallal, this new field could help the implementation of the running TTL, which is now technically using the time since creation.

ElePT (approved):
On my side I can't give too many comments on the templates themselves, but happy to approve to test in staging.

PR #2252, opened 2026-06-17
Refactor: remove @api_view decorator from all endpoints

avilches:
## Summary

Refactors the endpoint registration system in the gateway to eliminate the redundancy between `@api_view([METHOD])` and `method=` in `@endpoint`.

### Problem

Every endpoint had two HTTP method declarations: one in `@endpoint(..., method="GET")` and another in `@api_view(["GET"])`. This caused:
- Unnecessary redundancy in every view
- Risk of silent inconsistency if they didn't match (e.g. `method="POST"` with `@api_view(["GET"])` would produce a silent 405)

### Changes

**`endpoint_decorator.py`**: The `@endpoint` decorator now applies `api_view([method])` internally and `method` becomes a required parameter. Callers no longer need to add `@api_view`.

**`route_registry.py`**: Removed the `_routes` list and the `method is None` branch. `method` is required in `register`. Refactored `_build_combined_view` for readability: the `CombinedView` class it creates exists solely for DRF middleware introspection, not for actual dispatch (which is done by the `dispatch` function).

**`route_registry.py` - duplicate validation**: Registering the same method twice for the same path now raises `ValueError` instead of silently overwriting.

**21 views**: Removed `@api_view([...])` from all views. Added explicit `method=` to `@endpoint` in every case.

**`save_result.py`**: Resolved conflict with main, which added a GET endpoint. Split into two functions `get_result` (GET) and `save_result` (POST), following the pattern of `get_runtime_jobs.py` + `associate_runtime_jobs.py`.

### Result

Each view has a single decorator declaring the HTTP method:

```python
# Before
@endpoint("jobs/<uuid:job_id>/logs", method="GET", name="jobs-logs")
@api_view(["GET"])
@permission_classes([permissions.IsAuthenticated])
def get_logs(...):

# After
@endpoint("jobs/<uuid:job_id>/logs", method="GET", name="jobs-logs")
@permission_classes([permissions.IsAuthenticated])
def get_logs(...):
```


gateway/api/v1/route_registry.py
```diff
@@ -27,55 +26,66 @@ def register(cls, url_path: str, view_func: Callable, name: str, method: Optiona
...
-        all_methods = list(method_map.keys())
+
+        # method -> handler function, e.g. {"GET": get_func, "POST": post_func}
         handlers = {m: func for m, (func, _) in method_map.items()}
 
+        # Merge swagger schemas from all handlers so the combined URL exposes all of them
         merged_swagger: dict = {}
-        for _, (func, _) in method_map.items():
-            merged_swagger.update(getattr(func, "_swagger_auto_schema", {}))
+        for func, _ in method_map.values():
```
korgan00, line 52:
```suggestion
        for func in handlers.values():
```
minor suggestion.

korgan00 (approved):
LGTM

PR #2253, opened 2026-06-17
Migrate programs list and get_by_title to standalone function endpoints

avilches:
## Summary

Migrates two read endpoints of the `programs` resource from the DRF ViewSet to standalone function-based views, following the same pattern used in `api/v1/views/jobs/`.

- Extracts `GET /api/v1/programs/` into `api/v1/views/programs/list.py` backed by `ListFunctionsUseCase`
- Extracts `GET /api/v1/programs/get_by_title/<title>/` into `api/v1/views/programs/get_by_title.py` backed by `GetFunctionByTitleUseCase`
- The ViewSet stays alive with only `upload`, `run`, and `get_jobs` (covered in a follow-up PR)
- Adds `docs/superpowers/specs/VIEWS_REFACTOR.md` documenting the pattern for future migrations

No behavior changes. All existing integration tests in `test_v1_program.py` pass without modification — the URL paths and names are identical.


tests/integration/test_jobs.py
```diff
@@ -382,10 +382,7 @@ def test_wrong_function_name(self, serverless_client: ServerlessClient):
         )
 
         expected_message = (
-            "\n| Message: Http bad request.\n"
-            "| Code: 404\n"
-            "| Details: User program 'wrong-title' was not found or you do not "
-            "have permission to view it."
+            "\n| Message: Http bad request.\n| Code: 404\n| Details: Qiskit Function wrong-title doesn't exist."
```
avilches, line 385: This is the standard output when `FunctionNotFoundException(function=title, provider=provider)` is raised. The old format was hardcoded.

gateway/api/v1/views/programs/__init__.py
```diff
@@ -2,7 +2,6 @@
 Programs view api for V1.
```
avilches, line 2: Python doesn't allow having a `programs.py` inside a `programs` folder. On the other hand,t his `__init__.py` will be removed when the run and upload endpoints are migrated, so this ugly workaround is going to disappear very soon.

korgan00 (approved):
LGTM

PR #2254, opened 2026-06-17
Migrate upload endpoint to standalone use case and view

avilches:
Migrates the `POST /programs/upload/` endpoint from the `ProgramViewSet` to a standalone function-based view, following the same pattern used for the `list` and `get_by_title` endpoints.

**What changes:**
- Adds `UploadFunctionUseCase` in `api/use_cases/programs/upload.py`, which owns all business logic: serializer validation, provider/custom-function permission checks, create-vs-update detection, and persistence.
- Adds `upload_program` view in `api/v1/views/programs/upload.py` using the `@endpoint` decorator pattern; the view only extracts auth context, logs inputs, calls the use case, and returns the serialized result.
- Removes the `upload` method from the v1 `ProgramViewSet`.
- Adds 5 unit tests for `UploadFunctionUseCase` covering create, update, and access-denied cases.

This PR targets `program-refactor` (which contains the earlier list/get_by_title migration). It will be rebased onto `main` once that branch merges.


Tansito (approved):
Feedback applied and approved! Thanks for the refactors @avilches !

PR #2256, opened 2026-06-19
Add support for local builds in apple arm64

ElePT:
### Summary

The update to Ray 2.55 results in the scheduler hanging on Apple Silicon Macs (and potentially other architectures). On Apple Silicon Macs the amd64 ray-node image runs under QEMU emulation, which is too slow for Ray 2.55's hardcoded ~30s dashboard startup timeouts: the dashboard HTTP port responds but the jobs API never comes up, so locally-submitted jobs stay stuck in QUEUED. 

This PR proposes a path for running ray-head as a native arm64 image, which removes the emulation overhead so Ray's default timeouts are satisfied. There are no changes to the "official" image building path, as this runs on linux and doesn't experience the same issue with Ray.

### Details and comments

To be able to run a local build for arm64 I had to provide the following custom files

- `ray-node/Dockerfile.arm64`: arm64 variant mirroring ray-node/Dockerfile, with an optional Rust-toolchain block for ffsim.
- `ray-node/requirements-dynamic-dependencies-arm64.txt`: dynamic deps minus ffsim and qiskit-addon-aqc-tensor (ffsim 0.0.60 has no arm64 wheel).
- `docker-compose.arm64.yaml`: override running ray-head native arm64 while keeping gateway/scheduler/postgres on amd64 (no Ray code; HTTP-only, so arch-mixed is fine).
-` docs/custom_images/run_on_apple_silicon.rst`: explains the symptom, root cause and the build+run steps; registered in the custom_images toctree.

### DIFF
--- a/docker-compose.arm64.yaml
+++ b/docker-compose.arm64.yaml
@@ -1,0 +1,22 @@
+# Local-dev override for Apple Silicon (M-series) Macs.
+#
+# Use together with the base docker-compose.yaml:
+#
+#   docker build --platform linux/arm64 \
+#     -t custom-ray-node:0.32.0-arm64 -f ray-node/Dockerfile.arm64 .
+#   VERSION=0.32.0 docker compose -f docker-compose.yaml -f docker-compose.arm64.yaml up
+#
+# Runs ray-head as a NATIVE arm64 image (no QEMU) so Ray 2.55 startup is fast enough that its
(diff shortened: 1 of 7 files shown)


HuangJunye (approved):
I and another vendor confirmed that this new custom image for arm64 works on our devices.

PR #2258, opened 2026-06-19
Check that the logs file is created before erasing the job logs

korgan00:
### Summary

Modify logs migration script to include a file check


gateway/scheduler/management/commands/migrate_old_job_logs.py
```diff
@@ -31,6 +31,24 @@ def save_job_logs_to_storage(job: Job):
     logger.info("Logs saved to storage for job [%s]", job.id)
 
 
+def check_storage_logs(job: Job):
+    """
+    Check that te logs ar correctly storaged.
```
avilches:
```suggestion
    Check that the logs are correctly stored.
```

gateway/scheduler/management/commands/migrate_old_job_logs.py
avilches:
```suggestion
            default=1,
```

avilches: You don't need the `check_storage_logs` function is you validate right after the save:

```suggestion
if job.program.provider:
    logs_storage.save_private_logs(logs)
    assert logs_storage.get_private_logs() == logs
else:
    logs_storage.save_public_logs(logs)
    assert logs_storage.get_public_logs() == logs
```

korgan00: right

avilches (approved):

PR #2259, opened 2026-06-19
fix: make Ray result storage provider-aware

ElePT:
### Summary

Provider (custom-image) functions running on the Ray runner can complete successfully (status == DONE) but return an empty result from job.result(). The cause is a path mismatch between where the function writes its result and where the gateway reads it back:

```
┌──────────────────┬─────────────────────────────────────────────────────┬──────────────────────────────────────────────────┐
│       Side       │                        Path                         │                      Source                      │
├──────────────────┼─────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
│ Write (function, │                                                     │ build_env_variables() in gateway/api/utils.py    │
│  via             │ {username}/{provider}/{title}/results/{job_id}.json │ (provider-aware DATA_PATH)                       │
│ RESULTS_PATH)    │                                                     │                                                  │
├──────────────────┼─────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
│                  │                                                     │ RayResultStorage in                              │
│ Read (gateway)   │ {username}/results/{job_id}.json                    │ core/services/storage/result_storage_ray.py      │
│                  │                                                     │ (username-only)                                  │
└──────────────────┴─────────────────────────────────────────────────────┴──────────────────────────────────────────────────┘

```

For user (custom) functions the two paths are identical, so nothing breaks. For provider functions they diverge, the gateway looks in the wrong directory, finds no file, and returns an empty result.

`RayArgumentsStorage` is already provider-aware (it uses `PathBuilder` with `provider_name`), but it looks like the result storage was not. This PR makes `RayResultStorage` build its directory with the same `PathBuilder.absolute_path(WorkingDir.USER_STORAGE, …, provider_name=…)` call, so the read path matches the write path for provider functions and stays `{username}/results` for user functions.

### Details and comments

- Affects only jobs with `program.runner == RAY`. Fleets jobs use `FleetsResultStorage` (COS-based) and are unaffected.
- I have not had time to verify if current deployed provider images are affected. If they are, we should get the fix in a provide a release ASAP.

### How to reproduce

Deploy a custom provider image as specified in the documentation and check that result is empty.

### Why it wasn't caught earlier

The existing storage unit tests only used user functions (no provider), so the divergent branch was never exercised; and the integration tests mock `_submit_to_ray`, so no real function ever wrote a result file for the gateway to read back. The write↔read seam — the only place the mismatch is visible — was never tested for a provider function.

### DIFF
--- a/gateway/core/services/storage/result_storage_ray.py
+++ b/gateway/core/services/storage/result_storage_ray.py
@@ -15,10 +15,10 @@
 import logging
 import os
 
-from django.conf import settings
-
 from core.models import Job
 from core.services.storage.result_storage import ResultStorage
+from core.services.storage.path_builder import PathBuilder
+from core.services.storage.enums.working_dir import WorkingDir
 
 logger = logging.getLogger("core.RayResultStorage")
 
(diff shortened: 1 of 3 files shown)


korgan00 (approved):
Good catch LGTM

PR #2261, opened 2026-06-19
Sync charts folder

Tansito:
### Summary

This PR basically syncs the sync-ce-project as it is disabled internally so we have the same here also (not changes at all in fact).


ElePT (approved):
LGTM

PR #2262, opened 2026-06-23
Fix detect-secrets issues

ismaelRozasRamallal:
### Summary
This PR fixes couple unaudited secrets in the repository:
48 potential secrets in .secrets.baseline were reviewed. Found 0 live secrets, 2 unaudited secrets and 0 secrets that were audited as real.
Failed Condition    Secret Type     Filename                                                     Line
------------------  --------------  ---------------------------------------------------------  ------
Unaudited           Secret Keyword  charts/qiskit-serverless/templates/kafka-credentials.yaml      20
Unaudited           Secret Keyword  gateway/api/static/admin/css/carbon_theme.css                 513

Without that, the pipeline fails in the detect-secrets step.

### Details and comments
To solve that, we escaped the two false positives with `# pragma: allowlist secret` in order to be skipped when scanning.

### DIFF
--- a/charts/qiskit-serverless/templates/kafka-credentials.yaml
+++ b/charts/qiskit-serverless/templates/kafka-credentials.yaml
@@ -17,7 +17,7 @@ spec:
     template:
       engineVersion: v2
       data:
-        api_key: >-
+        api_key: >- # pragma: allowlist secret
           {{ printf "{{ (.credentials | fromJson).api_key }}" }}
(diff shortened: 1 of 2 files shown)


korgan00 (approved):

ElePT (changes requested):
@ismaelRozasRamallal I left instructions offline on how we usually fix detect-secrets. Did you take a look?

PR #2276, opened 2026-06-25
Replace qiskit-ibm-runtime private imports with public imports

diego-plan9:
### Summary

In the upcoming release of `qiskit-ibm-runtime`, the `.utils` module is rearranged a bit, which includes moving `.utils.json` ([here](https://github.com/Qiskit/qiskit-ibm-runtime/pull/2920)).

This PR replaces a couple of imports that are using "private" (per the definition of `qiskit-ibm-runtime`- anything that is not documented [here](https://quantum.cloud.ibm.com/docs/en/api/qiskit-ibm-runtime)) with their public equivalents:
* For `RuntimeDecoder` and `RuntimeEncoder`, it is needed as `qiskit_ibm_runtime.utils.json` will no longer be available (the top-level imports are kept, and still public)
* For `RuntimeJobV2`, there are no plans of moving the module, it is mostly for consistency

### Details and comments

Looking at other imports, the only remaining seems to be `.accounts.exceptions.InvalidAccountError` (plus some low-level mocking in the tests). While it is also not public, there are no plans of moving that module around ... and the top-level exceptions are technically not public either, so I think it is safe to bet on its stability. It might be worth considering adjusting to its parent `IBMAccountError` class (which is available at top-level) instead.

### DIFF
--- a/client/qiskit_serverless/serializers/program_serializers.py
+++ b/client/qiskit_serverless/serializers/program_serializers.py
@@ -32,7 +32,7 @@ import os
 from typing import Any, Dict
 from qiskit.primitives import SamplerResult, EstimatorResult
 from qiskit_ibm_runtime import QiskitRuntimeService
-from qiskit_ibm_runtime.utils.json import RuntimeDecoder, RuntimeEncoder
+from qiskit_ibm_runtime import RuntimeDecoder, RuntimeEncoder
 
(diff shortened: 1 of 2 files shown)


ElePT (approved):
Thanks a lot @diego-plan9! LGTM

PR #2278, opened 2026-06-25
Remove repository layer in favor of Django ORM and model managers

avilches:
The repository pattern is useful when you need to abstract a custom data layer, like writing raw SQL or using an ORM that does not follow Active Record. In those cases, a repository hides complexity and makes the data access layer swappable.

Django's ORM already follows Active Record: models are the domain objects, and querying through them is the intended way to work. Most of the repositories in this project were thin wrappers with no added value, just delegating straight to `objects.filter()` or `objects.create()`.

This PR removes them entirely:

- Simple one-liner repositories are inlined directly in the use cases and signals.
- Repositories that contained real query logic (`get_provider_by_name`, `get_by_name`) are moved to custom model managers (`ProviderQuerySet`), which is the Django-idiomatic equivalent.


gateway/api/use_cases/authentication.py
```diff
@@ -61,30 +58,61 @@ def execute(self) -> tuple[Optional[type[AbstractUser]], Optional[str]]:
...
             )
-            self.provider_repository.get_or_create_by_name(
+            provider, created = Provider.objects.get_or_create(
                 name="mockprovider",
                 registry=settings.SETTINGS_AUTH_MOCKPROVIDER_REGISTRY,
-                admin_groups=groups,
             )
+            if created:
+                for admin_group in groups:
+                    provider.admin_groups.add(admin_group)
```
korgan00, line 82: Why not keep this into the model manager?

gateway/api/use_cases/authentication.py
```diff
@@ -61,30 +58,61 @@ def execute(self) -> tuple[Optional[type[AbstractUser]], Optional[str]]:
...
+        self,
+        user: type[AbstractUser],
+        authentication_groups: List[AuthenticationGroup],
+        permission_names: List[str],
+    ) -> List[Group]:
+        new_groups = []
+
+        permissions = []
+        for permission_name in permission_names:
+            permissions.append(Permission.objects.get(codename=permission_name))
```
korgan00:
```suggestion
        permissions = Permission.objects.filter(codename__in=permission_names)
```

gateway/api/use_cases/authentication.py
```diff
@@ -61,30 +58,61 @@ def execute(self) -> tuple[Optional[type[AbstractUser]], Optional[str]]:
...
+
+        user.groups.clear()
+
+        for authentication_group in authentication_groups:
+            group, created = Group.objects.get_or_create(name=authentication_group.group_name)
+            if created:
+                for permission in permissions:
+                    group.permissions.add(permission)
```
korgan00:
```suggestion
                group.permissions.add(*permissions)
```

gateway/api/signals.py
korgan00, line 1: The whole function is duplicated, only one parameter in the `Program.objects.create' changes, maybe we can have an aux function with that value as parameter.

korgan00 (changes requested):
So clean, just some suggestions and a question :)

gateway/api/use_cases/authentication.py
avilches: This method contains authentication business logic, not query logic: it manages group creation, permission assignment and user-group linking as part of the auth flow. Model managers are meant for query/ORM operations on querysets, not for orchestrating this kind of side-effectful logic, so keeping it as a private method in the use case that owns it feels more appropriate.

Also worth noting: this method is expected to be deprecated once we complete the migration to auth through Runtime instances, so investing in a dedicated manager for it would be short-lived.

gateway/api/use_cases/authentication.py
avilches: Changed

gateway/api/use_cases/authentication.py
avilches: Changed

gateway/api/signals.py
avilches: Changed

korgan00 (approved):
LGTM

PR #2282, opened 2026-06-29
Instance permission acceptance tests via NTC and Runtime API

avilches:
## What this adds

Acceptance tests for instance-level function permissions driven through NTC, plus the fix that makes NTC writes actually propagate to the Runtime API.

Instead of relying on four pre-provisioned CRNs, the suite uses a SINGLE service instance and reconfigures its entitlements before each test through the NTC account and resource-controller APIs. Every permission level is driven the same way:

1. grant the function(s) and permissions on the account plan (kept at a superset so the broker accepts the instance write)
2. grant the same on the instance (resource-controller PATCH)
3. assert that GET /functions (the endpoint the gateway authorizes against) returns exactly that
4. exercise the gateway operations and assert each one is allowed or denied according to the granted permissions

## What is tested (schematic)

NONE
- account and instance set to functions=[]
- assert /functions returns 200 with an empty list (clean per-function deny, NOT a 204 legacy fallback)
- validate that every operation is denied

USER (business_model trial)
- add the function with user permissions (function.read/run, function-files.read/write) plus custom to the account
- same on the instance
- assert /functions
- validate that only user operations work and provider operations are denied

PROVIDER (business_model consumption)
- add the function with provider permissions (function.write, function-job.read, function-provider-logs.read, function-provider-files.read/write) to the account
- same on the instance
- assert /functions
- validate that the provider operations work

ALL / combined (business_model consumption, two functions)
- add both functions with the full permission set plus custom to the account
- same on the instance
- assert /functions
- validate that all operations work and the second function stays isolated where it should

CUSTOM
- grant the custom-function permissions (function-custom.write/run)
- assert /functions exposes them under custom_functions
- validate upload, run and list of the custom serverless function

Propagation account to instance (test_instance_propagation.py)
- account superset and instance at ALL: the function is usable
- narrow the account to a sibling function only: the function disappears from the instance (the account to instance sync only narrows)
- re-add the function to the account: it is NOT restored (the sync never re-adds)
- re-add the function to the instance: usable again
- and: the broker rejects an instance PATCH that exceeds the account plan grants

Runtime API ground truth (test_runtime_api.py)
- for each level above, read /functions directly (Service-CRN plus apikey, exactly as the gateway does) and assert the stored functions, permissions and custom permissions match what was written
- pins down the contract and the response shape independently of the gateway

## The propagation fix

The resource-controller PATCH that writes the instance entitlements now carries an advancing timestamp. That timestamp is the signal the Runtime API uses to invalidate its per-instance cache and re-sync; without it a PATCH that returns 200 and is stored was not reflected by /functions. set_instance_entitlements reads any timestamp already on the instance and writes one strictly greater (sent both at the top level and inside parameters), so a re-add wins the account narrow-sync last-write-wins comparison even when this machine's clock lags the server.

## Offline coverage

test_ntc_client.py and test_runtime_api_client.py are requests_mock unit tests for the two HTTP clients (account/instance configuration and the Runtime API read), so the client logic runs in CI without any network or credentials.


avilches:
Draft, don't review. Big refactor is coming.

avilches:
The tests have already been executed and we validated the feature. Tests are not part of the CI; they must be executed locally using an api key from a person with elevated privileges (like support team) to update accounts. I mean, this is historical code, not part of the test suite. Maybe this make it easier to review @ElePT @korgan00 @riley2012

ismaelRozasRamallal (approved):

PR #2295, opened 2026-07-01
IBM w3id SSO login for the backoffice

avilches:
This PR adds IBM w3id SSO (OIDC) login for the Django backoffice.

## What it does

- Adds a "Login IBM SSO" button on the backoffice login page, next to the normal username and password form.
- On first SSO login the user is created and keyed by their email. No flags are changed, so an administrator still decides what access they get.
- The feature is off by default. It only turns on when both the client id and the client secret are set. With only the id (or nothing), the button is hidden and the SSO entry view falls back to the normal login, so nothing changes for current users.

## Config

Three environment variables on the gateway:

- W3ID_SSO_CLIENT_ID
- W3ID_SSO_CLIENT_SECRET
- W3ID_SSO_BASE_URL (defaults to the test environment)

The Helm chart wires these into the gateway deployment, but only when the credentials Secret is provided. If it is not, no variable is added and the gateway starts as usual.

## Still open

How to provision the credentials Secret is not decided yet. The deployment already references the Secret by name, but there is no template that creates it. I will add a comment with the two options (ExternalSecret like Kafka, or a plain Secret like superuser) so we can agree on one before merging.


korgan00 (dismissed):
LGTM

ElePT (approved):
LGTM

PR #2300, opened 2026-07-02
Hide author-private job result on cross-author provider listings

avilches:
### Summary

The `GET /api/v1/programs/{id}/get_jobs/` endpoint lets a provider admin list every author's jobs for one of their functions.
Job results are author-private everywhere else in the API, but this listing returned the `result` field for every job, so a provider admin could read other authors' private results through it.

This PR hides `result` on the cross-author provider listing and keeps it on the regular own-jobs listing.

### Details and comments

Before, the endpoint serialized every job with `OutputSerializer`, which includes `result`, for both the provider cross-author listing and the own-jobs listing.
A provider admin therefore read other authors' results.

Now, `GetJobsUseCase` returns the job queryset plus a boolean that tells the view whether this is the cross-author provider listing.
The provider listing uses a new `OutputSerializerWithoutResult` that drops `result`, and the own-jobs listing keeps `OutputSerializer` with `result`.

Results are author-private, so a provider admin should be able to see that a job exists and its status without reading its private output.
This also matches the other job listing endpoints, which already use a result-less serializer variant for cross-author listings.

A provider admin listing jobs for their function no longer sees `result` for any job in that response, including their own jobs in that same listing.
They can still read their own results through the own-scoped endpoints.


ElePT (commented):
I totally see the security perspective of this change, but I think this is precisely the only mechanism our current providers have to see how many jobs were sent on their functions. Of course, they should not be able to access results, but if we block this path we should provide a different way of solving the provider's request. Maybe a new endpoint for providers that allows to list jobs without seeing results?

avilches:
> I totally see the security perspective of this change, but I think this is precisely the only mechanism our current providers have to see how many jobs were sent on their functions. Of course, they should not be able to access results, but if we block this path we should provide a different way of solving the provider's request. Maybe a new endpoint for providers that allows to list jobs without seeing results?

Oh, no worries! no new endpoint is needed... the provider path you describe is the GET `/jobs/provider`, which lists all jobs on a provider's functions with multiple and different authors, and it already never returned the results. So providers keep the "how many jobs ran on my function" view without seeing results.

This PR only fixes the old and kind of deprecated `programs/{id}/get_jobs` endpoint so it stops sending the result on the cross-author listing. A provider can still read the result of their own job through the per-job endpoint (GET /jobs/{id} with with_result, or GET /jobs/{id}/result/).

korgan00 (approved):
LGTM

ElePT (changes requested):
I'd like to think a bit about how to give an alternative to providers before merging this PR. I don't want to break their only source of data without an alternative.

ElePT (approved):
Sorry, @avilches, I missed your last comment. You are right, my concern doesn't apply here. LGTM

PR #2301, opened 2026-07-02
Move gateway ray template dir off world-writable /tmp

avilches:
### Summary

Move the extra Django template directory from /tmp/templates to /etc/gateway/templates.

The ray cluster template is delivered through a config map mount, and Django looks in that directory when it renders templates.
/tmp is world writable and shared, so another process on the same host could drop a malicious template into Django's search path.
Moving it to a path that is not world writable closes that gap.

### Details and comments

Two references are updated together and must stay in sync: the mount path in the gateway Helm chart deployment, and the TEMPLATES DIRS entry in gateway settings.
The template still resolves at runtime because both now point at the same new path.


ElePT (approved):
LGTM

PR #2302, opened 2026-07-02
Fail closed on unknown SETTINGS_AUTH_MECHANISM

avilches:
### Summary

Fail closed when SETTINGS_AUTH_MECHANISM is set to an unknown value, instead of silently falling back to the mock_token backend.

The old code used a dict lookup with mock_token as the default.
A typo or a removed mechanism name would quietly select MockTokenBackend, which authenticates anyone that presents a static shared token.
That turns a small config mistake into a full authentication bypass.
The app now raises ImproperlyConfigured at startup and lists the valid values.

### Details and comments

The default for SETTINGS_AUTH_MECHANISM stays custom_token, which is valid, so a correctly configured deployment and CI are not affected.
The new check only trips when the value is unknown, which is exactly the case we want to catch early instead of at request time.


avilches:
Safe to deploy, no production impact.

The auth mechanism is now checked against the known list. The value used in production is a valid mechanism, so it does not fail closed there. It only stops a boot when the mechanism name is a typo or was removed, which is the intended fail fast behavior.

ElePT (dismissed):
LGTM

ElePT (approved):
LGTM again (forgot to approve)

PR #2303, opened 2026-07-02
Enforce registry path boundary for custom images

avilches:
### Summary

Make the provider registry check accept a custom image only when the image really sits inside the provider registry.

The old check accepted an image if its name just started with the registry name, using a plain startswith. That is too loose and can be tricked. For example, if a provider registry is `docker.io`, then an image like `docker.io.attacker.com/evil` also starts with `docker.io`, so it passed the check even though it is hosted on a different, attacker owned host. This change requires the image to be the registry itself or to start with the registry followed by a slash, so a look alike host no longer slips through.

### Details and comments

The check now uses a small helper that treats the registry as a path prefix: the image passes only when it equals the registry or starts with registry + "/".

This is stricter than before, so an image that only matched by string luck is now rejected. The mock provider used in local development sets its registry to "test", and the docs and integration tests used a bare image name "test-local-provider-function", which only passed the old startswith by coincidence. That name is not actually inside the "test" registry, so it now returns a 400. To keep the documented local flow and the tests working, the image name was changed to be nested under the registry as "test/local-provider-function".

Changes in this PR:
- The registry check in the upload serializer.
- Integration test fixtures updated to the nested image name, plus a unit test that a look alike host like `docker.io/awesome.attacker.com/evil` is rejected.
- The custom image docs updated to build and reference the nested image name.


ElePT (commented):
I see 2 things here. On one hand, this will affect some of our partner's local testing paths FYI @HuangJunye, so we will have to inform them of the new conventions. On the other hand, given our process and the fact that our partners can't upload images themselves, I wonder how much of a risk this malformed path actually is.

avilches:
Two things here...

On the risk: you're right that it's not a big one. Only registered providers can set a custom image, so someone would need provider access to even hit this, and the registry check is a secondary allowlist on top of already-trusted providers, not the control that actually gates who can upload. But the plain `startswith` is still a bug: with registry `docker.io/awesome` it happily accepts `docker.io/awesome.attacker.com/evil`, which actually points to a different registry. The fix is just one line and doesn't change anything for normal `registry/...` images, so it feels worth keeping even if the way to abuse it is narrow.

On the partner side: agreed. Local testing that used an image name without the registry prefix (`test-local-provider-function` under registry `test`) now needs the nested form (`test/local-provider-function`). The docs and the integration fixtures in this PR already use that, so all that's left is giving partners a heads-up about the new naming.

So... what do you think, still good to merge? If so, I'd say the only follow-up is the partner notice about the naming change; the code and docs are ready.

korgan00 (dismissed):
LGTM

ElePT (changes requested):
I'd like to include in this PR a release note explaining the change required in the existing local flow @avilches, because we know that users don't read the docs and they are going to see that their build fails.

ElePT (approved):
LGTM

PR #2304, opened 2026-07-02
Fail closed on missing DJANGO_SECRET_KEY in production

avilches:
### Summary

The gateway used to fall back to a hardcoded secret key when DJANGO_SECRET_KEY was not set, and that key is public because it lives in this repository.

This change makes the gateway stop at startup when DEBUG is off and no key is provided, so it never signs anything with a key that everyone can read.

### Details and comments

SECRET_KEY is the value Django uses to sign session cookies, CSRF tokens and other signed data.

Before this change, if DJANGO_SECRET_KEY was missing the app booted with a fixed key that is committed in the repository, so anyone could forge those signed values.

Now the hardcoded key is only used when DEBUG is on or when the tests are running.

When DEBUG is off and no DJANGO_SECRET_KEY is set, the app raises an error at startup and does not boot.

The DEBUG line was moved above the key logic so the key logic can read it, and its default value is not changed.

Because the app now fails closed with DEBUG off, two local setups need a dev key so they keep working.

docker-compose.yaml sets a local only dev key for the two services that run with DEBUG off, so local compose still starts.

gateway/tox.ini sets a local only dev key too, because the lint tools import the Django settings module and would otherwise hit the new startup error.

Deployments that already set DJANGO_SECRET_KEY are not affected, since the key they use does not change.


avilches:
Safe to deploy, no production impact.

This only changes boot behavior when `DJANGO_SECRET_KEY` is missing. It does not change how secrets or any stored data are encrypted, and no data migration is needed.

Any deployment that already sets `DJANGO_SECRET_KEY` keeps working exactly as before, with no changes required. The only new behavior is that when `DEBUG` is off and no key is provided, the gateway now refuses to start until a real key is set, which is the intended fail closed behavior.

ElePT (approved):
LGTM, verified that DJANGO_SECRET_KEY exists in our deployment templates.

PR #2306, opened 2026-07-02
Add environment prefix variable

ismaelRozasRamallal:
### Summary
The Kafka topic name used by the gateway to publish function usage events includes an environment prefix — quantum.{prefix}.function-usage.v1 — that distinguishes staging traffic from production traffic on the same Event Streams instance. Originally, this prefix was sourced from a generic ENVIRONMENT environment variable that was not consistently defined across deployments so introducing a new EVENT_STREAMS_ENVIRONMENT_PREFIX specifically for this.

### Details and comments
We introduced a dedicated EVENT_STREAMS_ENVIRONMENT_PREFIX environment variable, injected into the gateway container via the Helm chart. A new environmentPrefix key was added to the eventStreams section of the gateway chart values, with environment-specific overrides set to `staging` in values-non-prod.yaml and "production" in values-gpu-prod.yaml. The KafkaEventStreamsClient was updated to read this new variable instead of ENVIRONMENT, and the unit tests were updated accordingly. This makes the Kafka topic configuration explicit, self-contained, and independently controllable from the rest of the deployment environment settings.


charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -510,6 +510,8 @@ spec:
             {{- end }}
             - name: EVENT_STREAMS_ENABLED
               value: {{ .Values.application.eventStreams.enabled | default false | quote }}
+            - name: EVENT_STREAMS_ENVIRONMENT_PREFIX
+              value: {{ .Values.application.eventStreams.environmentPrefix | default "" | quote }}
```
avilches: I think we can use the information about which environment is running the app for more features than Kafka in the future.
I mean, we can define a high-level variable in the root of the yaml specifying the environment, then you can use it in Kafka, but we can also use it in the scheduler or the gateway to change the behavior based on the environment, what do you think?

ismaelRozasRamallal: I can make it more generic, no problem!

gateway/core/ibm_cloud/event_streams/noop_event_streams_client.py
```diff
@@ -27,10 +27,10 @@ class NoOpEventStreamsClient(EventStreamsClient):
     """
 
     def emit_job_started(self, job) -> None:
-        logger.debug("job_id=%s [noop] emit_job_started", job.id)
+        logger.info("job_id=%s [noop] emit_job_started", job.id)
```
ismaelRozasRamallal, line 30: This is not related with the env variable, just an improvement to make easier the noops validation. INFO level is logged by default, DEBUG level needs explicit activation

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
ismaelRozasRamallal: Done!

korgan00 (approved):

PR #2307, opened 2026-07-02
Default DEBUG to off

avilches:
### Summary

Default DEBUG to off (0) instead of on (1).

Running with DEBUG makes Django return a full traceback page on any server error, which leaks source code, settings, and local variable values.

Off is the safe default.

Developers can still set DEBUG=1 explicitly for local work.

### Details and comments

The change is the default value, flipped from 1 to 0 in the DEBUG line and in the LOG_LEVEL line that reads it.

Deployments that already set DEBUG explicitly, for example through the Helm chart, are not affected.


ElePT (dismissed):
LGTM. All of our deployments (helm charts and local docker compose) already set DEBUG explicitly. This means that the change are not so impactful in practice but it's good to have the safeguard.

ElePT (dismissed):
LGTM

ElePT (approved):
LGTM

PR #2310, opened 2026-07-02
Basename client download target to prevent path traversal

avilches:
### Summary

Strip any directory part from the download file name in the client so a crafted name cannot escape the download location.

The download target name can come from the server or the caller.
Without stripping it, a value like "../../etc/passwd" would let the write land outside download_location.
Wrapping it in os.path.basename keeps only the file name, so the write always stays inside the intended folder.

### Details and comments

The fix is a single line added right before the file is opened.
It does not change the normal case where the name is a plain file name.

When target_name is not set, the name is built from a random value, so this only matters when a name is passed in or returned by the server.


ElePT (approved):
LGTM

PR #2311, opened 2026-07-02
Raise a clear error on malformed result decoder payloads

avilches:
### Summary

Make the client result decoder fail with a clear error when a decoded payload has an unexpected shape, instead of a confusing constructor error.

The decoder builds known Qiskit types by passing the "__value__" field as keyword arguments. When that value is not a well formed mapping of string keys, it now raises a clear ValueError up front.

This is a small hardening and error message improvement, not a full mitigation. For SamplerResult and EstimatorResult a bad value would already fail inside the constructor, so this mainly gives a clearer message. For QiskitRuntimeService a malicious gateway can still send a well formed mapping, and the values we have to accept are the same ones that would be dangerous, so a shape check alone cannot tell a safe payload from a dangerous one.

### Details and comments

The check is applied to the three known types (QiskitRuntimeService, SamplerResult, and EstimatorResult) through a small helper, so a well formed response behaves exactly as before.

A deeper look at which objects the encoder and decoder should support, and whether to explicitly disallow sending a QiskitRuntimeService through arguments, is worth a separate follow up. See the review discussion for context.


ElePT (commented):
I have a couple of comments on the value of this PR. 

The check you added is reasonable, but I don't think it protects against what the summary describes, and I don't think we can do much better at this layer (I'll explain why). We could merge it as an error message improvement, but I don't think it would change things much in practice. For `SamplerResult/EstimatorResult`, a bad `__value__` already fails in the constructor, so this is mainly a clearer error message. `QiskitRuntimeService` is a bit trickier, there is a higher risk, but we can't allow-list the "safe" kwargs because the values we have to accept are the same ones that would be dangerous if the gateway were malicious. A malformed-shape check doesn't separate the two.

Now, this made me think of when do we actually decode a payload with `QiskitRuntimeService`. The decoder practically runs in 3 scenarios: in `get_arguments()` (inside the function), as part of `stop()` (gateway), and in `job.result()` (user side). The first one, sending a service in arguments, as far as I know is not used by any of the provider functions, and I am wondering if it's something we should explicitly disallow. Something to discuss with @HuangJunye . The stop() case is relatively contained, as we are the ones controlling the `QiskitRuntimeService` construction. And the result() case is on the user side. 

Conclusion: the risk of a malformed payload I think is low, but I also think that checking values like this PR does doesn't erase it. I take from this review that we should do a more thorough analysis of what objects are we supporting in our encoder/decoder.

avilches:
> Conclusion: the risk of a malformed payload I think is low, but I also think that checking values like this PR does doesn't erase it. I take from this review that we should do a more thorough analysis of what objects are we supporting in our encoder/decoder.

Interesting...! Agreed, and thanks for the careful review. You are right that this doesn't close the threat the summary describes, sorry for that. I reviewed the code and... correct, for `SamplerResult/EstimatorResult` a bad `__value__` already fails in the constructor, so the check is really a clearer error there. And for `QiskitRuntimeService` the dangerous `kwargs` are the same ones we legitimately accept, so a shape check cannot tell them apart... I will update the PR title and description as a small check + clearer error, rather than a real mitigation... The deeper fix (auditing which objects the encoder/decoder support, and whether to explicitly disallow sending a `QiskitRuntimeService` through arguments in get_arguments) is worth its own issue, happy to open it and work with @HuangJunye. Thanks!!

ElePT (approved):
Awesome, thanks for the PR anyways. LGTM

PR #2312, opened 2026-07-02
Validate entrypoint is a safe relative .py path

avilches:
### Summary

Validate the uploaded function entrypoint before it is used.

The entrypoint is put into the runner execution command and into object storage paths.
Without a check, a value that is absolute, uses ".." to climb out of the function directory, or contains shell characters could be abused.
The serializer now only accepts a relative path to a .py file with no ".." segments and no shell characters.

### Details and comments

The rule is applied in the upload serializer as a validate_entrypoint method.
A normal entrypoint like main.py or a subfolder path like src/main.py keeps working, while unsafe values are rejected with a clear error.


gateway/api/v1/views/programs/upload.py
```diff
@@ -63,6 +64,29 @@ def validate_provider(self, value):
...
+                "Invalid entrypoint. It must be a relative path to a .py file "
+                "without '..' segments or shell characters."
+            )
+        segments = value.split("/")
+        if not re.fullmatch(r"[A-Za-z0-9_./-]+\.py", value) or value.startswith("/") or ".." in segments:
+            raise ValidationError(
+                "Invalid entrypoint. It must be a relative path to a .py file "
+                "without '..' segments or shell characters."
+            )
+        return value
```
ElePT: We could normalize the path first so it matches how the fleets runner (#2313) cleans archive members with normpath. Otherwise `./main.py` is accepted here but normalized to `main.py` there, the entrypoint match fails, and the job errors with "Entrypoint not found." Returning the normalized value keeps both sides aligned.

```suggestion
        normalized = posixpath.normpath(value)
        segments = normalized.split("/")
        if (
            not re.fullmatch(r"[A-Za-z0-9_./-]+\.py", normalized)
            or normalized.startswith("/")
            or ".." in segments
        ):
            raise ValidationError(
                "Invalid entrypoint. It must be a relative path to a .py file "
                "without '..' segments or shell characters."
            )
        return normalized
```

ElePT (commented):
overall LGTM, but I was cross-comparing this PR with #2313 and I found a very small inconsistency. I'm pointing it out in the code directly.

gateway/api/v1/views/programs/upload.py
avilches: Good catch, thanks. I applied the suggestion: the entrypoint is now normalized with posixpath.normpath before the checks and the normalized value is returned, so it lines up with how the fleets runner cleans archive members in #2313. I also added the missing import posixpath, since the suggestion did not include it.

So ./main.py now validates and comes back as main.py, which matches the runner side and avoids the "Entrypoint not found" mismatch. Paths that resolve inside the function dir like a/../b.py are accepted as their normalized form (b.py), while anything that escapes with .. or is absolute is still rejected.

ElePT (approved):
LGTM

PR #2313, opened 2026-07-02
Reject path traversal members in fleets artifact upload

avilches:
### Summary

Skip unsafe archive members when the fleets runner uploads a function artifact to object storage.

A crafted artifact could include member names like "../../other-user/results.json", which would place object storage keys outside this job's prefix.
The runner now cleans up each member name and skips any that is absolute or climbs out with "..", logging a warning.
Safe members use the cleaned name for the storage key and for the entrypoint match.

### Details and comments

Normal artifacts are not affected, because their member names stay inside the job prefix after they are cleaned up.


gateway/tests/core/services/runners/test_fleets_runner.py
```diff
@@ -793,3 +795,114 @@ def test_submit_uses_artifact_upload_when_artifact_set():
...
+    runner.job.id = "job-1"
+    runner._project.cos_bucket_user_data_name = "user-bucket"  # pylint: disable=protected-access
+    paths = build_job_paths(runner.job)
+    return runner, paths
+
+
+def test_upload_custom_image_entrypoint_skips_traversal_member():
+    """_upload_custom_image_entrypoint() skips a member that escapes via ".." and only uploads safe files."""
+    runner, paths = _make_custom_artifact_runner()
+    tar = _build_artifact_tar({"../evil.txt": b"malicious", "main.py": b"safe"})
```
ElePT, line 838: hahaha I like this evil.txt

ElePT (approved):
LGTM, I will group this and other security improvements into an internal issue to keep track of the changes. I'll also make a note to test in staging.

PR #2314, opened 2026-07-02
Drop duplicate ENV_JOB_ARGUMENTS literal env var for fleets

avilches:
### Summary

Stop sending the job arguments as a plaintext literal env var to Code Engine fleets.

The full JSON arguments already reach the container through object storage at ARGUMENTS_PATH, so the extra ENV_JOB_ARGUMENTS env var only shows the arguments in plaintext in the Code Engine control plane.

### Details and comments

The container still reads its arguments from the mounted file at ARGUMENTS_PATH, so runtime behavior does not change.

Only the extra plaintext copy in the env var is removed.


ElePT (approved):
LGTM, we'll have to verify in staging

PR #2315, opened 2026-07-03
improve migrations

korgan00:
### Summary

Since filtering by nullish is not working. If the result is null we change it to empty string


ElePT (commented):
Makes sense, but @korgan00 can you add a PR description for future reference?

korgan00:
> Makes sense, but @korgan00 can you add a PR description for future reference?

done

ismaelRozasRamallal (approved):

PR #2317, opened 2026-07-06
Cache the 204 response from the Runtime functions API

avilches:
## Summary

`FunctionAccessClient.get_accessible_functions` calls the Runtime functions API to resolve which functions an instance can access. When the API responds `204` (instance not yet configured, meaning the caller should fall back to legacy Django-based authorization), the result was returned immediately without going through the cache, unlike every other response path. This meant every request for an unconfigured instance hit the Runtime API again instead of reusing a cached fallback result.

This PR restructures the status-code handling so the `204` case builds its result and falls through to the same `cache.set(...)` call as the `200` and error paths, so it now benefits from `RUNTIME_API_CACHE_TTL` like the rest.

It also includes an unrelated small config change lowering `cacheTtl` from `60` to `1` in the gateway Helm chart values as default value (it will be overwritten per environment in the values.yaml)

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/values.yaml
+++ b/charts/qiskit-serverless/charts/gateway/values.yaml
@@ -52,7 +52,7 @@ application:
     url: "https://cloud.ibm.com"
   runtimeApi:
     url: "https://quantum.cloud.ibm.com"
-    cacheTtl: "60"
+    cacheTtl: "1"
   allowedHosts: "*"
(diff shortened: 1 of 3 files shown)


avilches:
TLDR: Basically, when I did the client, I only cached the NTC response for 200 status codes only. This PR ensures the 204 response is cached as well...

ElePT (approved):
LGTM

PR #2318, opened 2026-07-07
Fix Ray cluster pod security context

avilches:
### Summary

This locks down the security context of the Ray cluster pods that run user jobs, so job code runs with least privilege. Right now the head container and its TLS init containers run with the full default set of Linux capabilities, privilege escalation is allowed, no seccomp profile is set, and the pod runs without an explicit non-root group so the process keeps root group membership (`gid=0`). None of this is needed to run quantum circuits, and it makes in-container privilege escalation easier than it should be.

### Details and comments

Changes in the Ray cluster template:
- Pod level: run as a non-root user with an explicit non-root group (`runAsNonRoot`, `runAsUser: 1000`, `runAsGroup: 1000`) so the process no longer defaults to the root group, keep `fsGroup: 1000`, and apply a `RuntimeDefault` seccomp profile.
- Head container and both TLS init containers: drop all Linux capabilities and set `allowPrivilegeEscalation: false`. With privilege escalation blocked and no capabilities, setuid binaries in the image can no longer be used to gain privileges.

The Ray TLS private key was created world readable (`chmod +r`). Since the init container and the head container now run as the same user, the key is created with owner only permissions (`chmod 600`) instead.

This matches the Kubernetes restricted Pod Security Standard. Ray does not need any Linux capabilities to run user circuits, so dropping them all is safe.


ElePT (commented):
Thanks a lot for the changes! LGTM

For future reference, I verified on my side that:

- **The `600` (owner-only) permission is safe on the head path.** `gencert_head.sh` (in the `ray-head-tls` init container) writes `tls.key` to the shared `ray-tls` volume, and the `ray-head` container reads it from that same volume. Since the pod runs everything as `runAsUser: 1000` and no container overrides it, the writer and reader are the same user — so owner-only access works. Our images run as non-root, so that's fine too. 👍

- **The worker change is harmless because we don't actually run workers.** `gencert_worker.sh` also gets `chmod 600`, but our RayCluster template only defines a head group — there's no `workerGroupSpec`. The `workers` / `min_workers` / `max_workers` values passed to `render()` in `ray_runner.py` are never used in the template, so no worker pods get created and this script is effectively dead code. No risk today. 

- **Minor:** the cert-manager path (`gencert_cert_head.sh`) writes the key with `base64 -d` and no `chmod`, so its permissions just follow the container's umask. Not a problem here — could be a nice follow-up if we want full coverage.

ElePT (approved):

avilches:
> Thanks a lot for the changes! LGTM
> 
> For future reference, I verified on my side that:
> 
> * **The `600` (owner-only) permission is safe on the head path.** `gencert_head.sh` (in the `ray-head-tls` init container) writes `tls.key` to the shared `ray-tls` volume, and the `ray-head` container reads it from that same volume. Since the pod runs everything as `runAsUser: 1000` and no container overrides it, the writer and reader are the same user — so owner-only access works. Our images run as non-root, so that's fine too. 👍
> * **The worker change is harmless because we don't actually run workers.** `gencert_worker.sh` also gets `chmod 600`, but our RayCluster template only defines a head group — there's no `workerGroupSpec`. The `workers` / `min_workers` / `max_workers` values passed to `render()` in `ray_runner.py` are never used in the template, so no worker pods get created and this script is effectively dead code. No risk today.
> * **Minor:** the cert-manager path (`gencert_cert_head.sh`) writes the key with `base64 -d` and no `chmod`, so its permissions just follow the container's umask. Not a problem here — could be a nice follow-up if we want full coverage.

Thanks for the review! I will test it against the staging following the x-force team example.

PR #2327, opened 2026-07-08
Use w3id SSO credentials from Secrets Manager

avilches:
### Summary

Adds the Helm chart pieces to wire up the optional w3id SSO backoffice login credentials from IBM Secrets Manager. The application code is in PR #2295; this PR handles the secret provisioning side.

An ExternalSecret reads the connector credentials from Secrets Manager and creates a Kubernetes secret named `gateway-w3id-sso`. The gateway deployment picks up `W3ID_SSO_CLIENT_ID` and `W3ID_SSO_CLIENT_SECRET` from that secret. The feature stays off by default (`application.w3idSso.enabled: false`) until the flag is turned on in the environment values.

### Details and comments

`application.w3idSso.enabled` is the single flag that controls both the ExternalSecret and the env var injection in the deployment. Setting it to `false` skips both: no ExternalSecret is created and no secret reference is added to the pod spec, so Kubernetes does not require the secret to exist.

The ExternalSecret is conditioned on `gateway.application.w3idSso.enabled`. The secret name and keys are fixed (`gateway-w3id-sso`, `client-id`, `client-secret`) so the deployment does not need per-environment configuration for those.

The environment-specific path and `enabled: true` flag go in the environment values files.

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/values.yaml
+++ b/charts/qiskit-serverless/charts/gateway/values.yaml
@@ -68,6 +68,7 @@ application:
   # Optional IBM w3id SSO backoffice login. The provider base url is not
   # sensitive and lives here; credentials go in secrets.w3idSso (see below).
   w3idSso:
+    enabled: false
     baseUrl: "https://test.login.w3.ibm.com"
 
 tracing:
@@ -106,9 +107,6 @@ secrets:
       password: password
       email: email
   w3idSso: ""
-
-w3idSsoCredentials:
-  enabled: false
 
 image:
   repository: icr.io/quantum-public/qiskit-serverless/gateway
--- a/charts/qiskit-serverless/charts/gateway/values.yaml
+++ b/charts/qiskit-serverless/charts/gateway/values.yaml
@@ -107,6 +107,9 @@ secrets:
       email: email
   # w3idSso: ""  # path in IBM Secrets Manager, e.g. app_qiskit_serverless_staging/arbitrary/staging-w3id-sso
 
+w3idSsoCredentials:
+  enabled: false
+
 image:
   repository: icr.io/quantum-public/qiskit-serverless/gateway
   pullPolicy: IfNotPresent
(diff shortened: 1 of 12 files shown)


charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -101,6 +106,7 @@ secrets:
       name: name
       password: password
       email: email
+  w3idSso: ""
```
ItamarGoldman, line 109: What will we expect the line to be when is not empty?

charts/qiskit-serverless/templates/w3id-sso-credentials.yaml
```diff
@@ -0,0 +1,25 @@
...
+    remoteRef:
+      key: "{{ .Values.gateway.secrets.w3idSso }}"
+  secretStoreRef:
+    kind: SecretStore
+    name: qiskit-serverless-secret-store
+  target:
+    template:
+      engineVersion: v2
+      data:
+        client-id: >-
```
ItamarGoldman, line 19: is the transition from between `client_id` to `client-id` is intentional?

ItamarGoldman (dismissed):
LGTM. Asked a couple of questions that will help me to better understand this PR.

charts/qiskit-serverless/charts/gateway/values.yaml
avilches: I will answer by DM

charts/qiskit-serverless/templates/w3id-sso-credentials.yaml
avilches: Good eye!
`client_id` is how we store the secret as json: `{"client_id": "", "client_secret": ""}`
`client-id` is the yaml value. We use to use dashes for yaml and underscore for json. Different storage, different nomenclatures (like CamelCase in one place and kebab-for-others...)

ElePT (approved):
I am assuming that this PR is a prerequisite of #2295, so let's merge it first. LGTM

PR #2329, opened 2026-07-09
Remove trailing slashes in Serveless host

korgan00:
### Summary

If you try to connect using `ServerlessClient` with a `host` that has a trailing slash, connection will fail.
This PR remove the trailing slash to make the connection experience smoother.


avilches (dismissed):

avilches:
On hold until deploy in production

ElePT (changes requested):
Thanks @korgan00! can you create a quick release note for the change before merging it too?

ElePT (approved):
Great, thank you!

PR #2330, opened 2026-07-09
Arguments validations (1/3): add `arguments_schema` field and preserve optional fields on program re-upload

avilches:
### Summary
Add an `arguments_schema` field to programs so a function can declare a JSON Schema describing its valid arguments, and store it on upload. This is the storage layer for argument validation; the validation itself and the client SDK support land in follow-up PRs.

This also makes re-uploads non-destructive: re-uploading a function with only some fields no longer resets `runner`, `dependencies`, or `env_vars` to their defaults, and the `entrypoint`/`image` requirement now applies only when creating a function, not on update.

### Details and comments
- `arguments_schema` is a text field validated as JSON at the serializer level (schema semantics are checked later, when validating arguments against it).
- On re-upload, optional fields are preserved when omitted, using `None` as the sentinel in the upload input dataclass.
- New migration `0055_program_arguments_schema.py`.


gateway/api/migrations/0056_program_arguments_schema.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ("api", "0054_job_running_started_at"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="program",
+            name="arguments_schema",
+            field=models.TextField(blank=True, default="{}", null=True),
```
avilches, line 16: I make the filed nullable in the db (with `null=True`), so it will survive to a deployment rollback in production. /cc @Tansito @ElePT

avilches:
Draft until deployment in production works

avilches:
The PR is safe to merge: it adds a nullable field, so it is rollback compatible (old code inserting a program row just leaves the column NULL, which the read path treats as "no schema").

It also changes the /upload endpoint behaviour for the better. Before the PR, re-uploading a function to update it required sending entrypoint or image every time, and any optional field you left out (runner, dependencies, env_vars) was reset to its default. Now only the title is mandatory on update; every other field is preserved from the database when omitted. So to update a schema you only need to send the title plus the arguments_schema field.

avilches:
Safe to merge to main since pentest finished

gateway/api/use_cases/programs/upload.py
```diff
@@ -104,21 +109,23 @@ def _update(self, instance: Function, data: UploadFunctionInput, user) -> Functi
...
+            raw_deps = json.loads(data.dependencies)
+            instance.dependencies = json.dumps([_normalize_dependency(d) for d in raw_deps])
+        if data.env_vars is not None:
+            instance.env_vars = json.dumps(encrypt_env_vars(json.loads(data.env_vars)))
         instance.artifact = data.artifact
         instance.author = user
         instance.image = data.image
-        instance.runner = data.runner
+        if data.runner is not None:
+            instance.runner = data.runner
```
ElePT, line 123: I'm not convinced about preserving the runner value on re-upload. If someone omits runner, silently keeping fleets when they expected ray breaks job execution with no obvious error. The current SDK always sends runner anyway (it defaults to "ray"), so the only callers affected by this change are old SDK versions or direct API users, and for those, I think that falling back to ray is safer than preserving whatever was stored.

ElePT (commented):
The PR mostly looks good, I just find the runner change a bit strange. In practice I think this affects very few users but given how the runner has proven to be complicated to use before, I want to make sure we don't add even more hidden behaviors (what's a hidden behavior I guess is another discussion we can have).

gateway/api/use_cases/programs/upload.py
korgan00: Should we send ray by default in case of re-upload? I think someone can reupload a fleet function without specifying and have errors.

korgan00 (dismissed):
LGTM

gateway/api/use_cases/programs/upload.py
ElePT: We need to have a default because the runner flag is not public yet, so our users don't expect to have to set it.

avilches: It doesn't sound consistent to me: default values are only for creating. On updating, omitted fields should keep the current value: if it weren't like that, to update a scheme in a fleets function will require re-sending the runner every time.... Anyway, since Fleets is new and Ray is fortunately going to die soon, I'm open to however you want it to behave, but personally it seems odd to me...

ElePT: Ok, both options look suboptimal to me for different reasons, so let's go for consistency and preserve the runner on re-upload.

ElePT (approved):
LGTM

PR #2331, opened 2026-07-09
improve storage file read check on migrations

korgan00:
### Summary

improve storage file read check on migrations


ElePT (approved):
LGTM. I think that the integration test failure is unrelated. This one PR is safe to merge now as it is not making any changes to the client.

PR #2332, opened 2026-07-10
Make ray an optional dependency of the client

ElePT:
### Summary
Ray is now installed via the `ray` extra (`pip install qiskit-serverless[ray]`) instead of being a base requirement, so `import qiskit_serverless` and the non-Ray client surface work without Ray installed. Note that the ray node dockerfile now installs ray independently so this will not affect ray node images, it just opens the path for having non-ray images.

### Details and comments
The steps for this include: 

- Remove top-level ray imports from decorators.py and serializers.py; import ray lazily where it is actually used.
- Delete the dead `RuntimeEnv = ray.runtime_env.RuntimeEnv` alias in job.py.
- Ray-only helpers (distribute_task, get, put, get_refs_by_status, remote) raise a clear "pip install qiskit-serverless[ray]" error when ray is missing.
- Add [project.optional-dependencies] ray extra; drop ray from requirements.txt.
- Ray node images install the [ray] extra (behavior unchanged).
- Add tests covering the no-ray import surface and the clear-error behavior; guard the ray-cluster tests behind the extra.
- Add release note and README install note.


korgan00 (dismissed):
LGTM

HuangJunye (dismissed):

HuangJunye (approved):

PR #2333, opened 2026-07-10
Bump client qiskit version to 2.2 and officially drop symengine from build

ElePT:
### Summary

Since Qiskit 2.1, symengine is no longer a Qiskit dependency; it was only pulled in through the `qpy-compat` extra, which exists to deserialize the legacy QPY formats (10/11/12) that Qiskit 1.0-1.1 wrote with `use_symengine=True`. We don't need to read those legacy payloads.

### Details and comments

- client: change the requirement from `qiskit[qpy-compat]>=1.4,<3` to `qiskit>=2.2,<3`, so symengine is no longer installed. Modern QPY serialization (including parameterized circuits) is unaffected. Note that the `qiskit-ibm-runtime` dependency currently pins `qiskit` (current pin is 2.5) so workloads have already been running with qiskit 2.2+ in the base image for a while.
- gateway: remove the qpy-compat as well.
- ray-node images: drop the `cp` step that copied symengine between the UBI9 lib and lib64 site-packages dirs. It was a workaround that only applied when symengine was installed and would now fail on a missing source path.


korgan00 (dismissed):
LGTM

HuangJunye (approved):

PR #2334, opened 2026-07-13
Fix missing title field in Program admin form

avilches:
### Summary
Creating a Program (Qiskit Function) from the Django admin was not possible because the technical `title` field was missing from the add form. Since `title` is required, there was no way to set it when adding a new program.

### Details and comments
When the `Info` fieldset was introduced in `ProgramAdmin`, it listed `readable_title` but left out `title`. With explicit fieldsets in place, Django only renders the listed fields, so `title` no longer appeared in the form at all: it was absent from the add form, and on the change form the existing `get_readonly_fields` logic that marks it read-only had no effect since the field was not in any fieldset either.

This adds `title` back to the `Info` fieldset. It is editable when creating a program and stays read-only when editing an existing one, which matches the behavior the read-only logic was already designed for.

A minimal regression test verifies that `title` is present and editable on add, and read-only on change.

### DIFF
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -103,6 +103,7 @@ class ProgramAdmin(admin.ModelAdmin):
             "Info",
             {
                 "fields": [
+                    "title",
                     "readable_title",
                     "type",
                     "description",
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -5,7 +5,7 @@ import logging
 
 from django.contrib import admin
 from django.db.models import Count, F, Q
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 from django.urls import path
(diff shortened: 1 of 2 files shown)


gateway/api/admin.py
```diff
@@ -103,6 +103,7 @@ class ProgramAdmin(admin.ModelAdmin):
             "Info",
             {
                 "fields": [
+                    "title",
```
ElePT, line 106: This is the actual change, right? the rest is from the other PR.

ElePT (commented):
LGTM

gateway/api/admin.py
ElePT: I merged from main and can confirm it now 👍

Tansito (approved):
Tested and working, thanks @avilches !

PR #2336, opened 2026-07-13
Use format_html instead of mark_safe in admin displays

avilches:
### Summary

The admin display helpers built HTML with `mark_safe()` on f-strings that interpolate values (job status, sub-status, and event data JSON). `mark_safe` marks the whole string as safe without escaping, so any value that reached these helpers would be rendered as raw HTML in the admin. This swaps them for `format_html()` with arguments, which escapes each interpolated value while keeping the surrounding markup intact.

It also keeps `pylint` green: `pylint-django` 2.8.0 added the `mark-safe-interpolation` (W5151) check, which flags exactly these three call sites, so a fresh lint run on `main` now fails without this change.

### Details and comments

Three call sites in `gateway/api/admin.py`:

- `render_data_json` (event data JSON block)
- `pretty_status` (status/sub-status badge)
- `status_badge` (job status badge)

Rendered output is unchanged for normal values; the only difference is that interpolated values are now HTML-escaped. The `mark_safe` import is replaced by `format_html`.

### DIFF
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -5,7 +5,7 @@ import logging
 
 from django.contrib import admin
 from django.db.models import Count, F, Q
-from django.utils.safestring import mark_safe
+from django.utils.html import format_html
 from django.urls import path
 from django.shortcuts import render, get_object_or_404
 from django.contrib.admin.views.main import PAGE_VAR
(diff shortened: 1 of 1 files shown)


ElePT (approved):
LGTM. This PR will unblock CI for the rest.

PR #2337, opened 2026-07-13
Refactor integration tests

korgan00:
### Summary

Split integration tests into two folders depending on the resources that the test access.
One folder for function executions and another folder with access to databases and storages.


tests/source_files/gold_path_function.py
```diff
@@ -0,0 +1,42 @@
+# source_files/program_3.py
```
ElePT: Stale comment:
```suggestion
```

tests/integration/execution/test_files_execution.py
```diff
@@ -0,0 +1,75 @@
...
+
+resources_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../source_files")
+
+filename = "test.txt"
+filename_path = os.path.join(resources_path, filename)
+filename_not_valid = "test-img.png"
+filename_not_valid_path = os.path.join(resources_path, filename_not_valid)
+
+
+class TestFilesExecution:
```
ElePT, line 18: both of these tests  read and write COS files, so I'm pretty sure that they need storage access and belong in `storage`. Let me know if I'm mistaken.

ElePT (commented):
Thanks @korgan00, the PR overall makes sense to me, it's good to start separating the tests between those that need COS credentials and not, I just left a couple of suggestions/comments.

tests/integration/execution/test_files_execution.py
korgan00: You are right but the test is about the execution, the COS is de indirect way to see if the execution gone right. I prioritized execution because is more specific.

ElePT: I thought the idea was to separate tests by what credentials they need to run, right? so if they need COS credentials it's different than if they don't, right @korgan00?

ElePT (approved):
After an offline discussion with @korgan00, we agreed that the current test distribution is a good first approach to the refactor. It could be improved by separating tests that require no credentials at all, but so far we are separating those that require running on actual fleets vs. not (can run only in simulation)

PR #2338, opened 2026-07-14
Add more informative auth error messages

ElePT:
### Summary

Users with an unentitled resource plan (not in RESOURCE_PLANS_ID_ALLOWED) were told only "Credentials couldn't be verified", this could lead to confusion, since their credentials were valid; the plan was not entitled. This PR fixes that, together with a few other auth improvements.

### Details and comments

Changes:

- Client: surface the server's message instead of discarding it, so the real reason reaches the user.
- Gateway: reword the entitlement failure to explain the cause and next step, without leaking the plan id (kept in logs only).

On top of this, other auth UX improvements:

- Gateway: catch RuntimeFunctionsException in the auth backend and re-raise a friendly, CRN-free AuthenticationFailed instead of an opaque 500.
- Fix typos in the deactivated-user message.
- Add tests for the disallowed-plan and Runtime-API-failure paths, asserting no plan id / CRN / status code leaks to the user.


gateway/api/authentication.py
```diff
@@ -69,7 +70,17 @@ def authenticate(self, request):
...
         ).execute()
 
-        accessible_functions = FunctionAccessClient().get_accessible_functions(crn, authorization_token)
+        try:
+            accessible_functions = FunctionAccessClient().get_accessible_functions(crn, authorization_token)
+        except RuntimeFunctionsException as error:
+            # The Runtime API failed while resolving the user's accessible functions.
+            # Details (status codes, CRN) are already logged by FunctionAccessClient;
+            # surface a generic, CRN-free message instead of an opaque 500.
+            logger.warning("Could not resolve accessible functions: %s", error.message)
```
Tansito: Should we log this as an error or a warning? I don't know the expectations in this point. I have the feeling that this is not an expected behavior and we should probably show an error here (from our side).

Tansito (dismissed):
LGTM @ElePT , I only left a little comment. Apart from that everything good 😄

client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -185,7 +185,7 @@ def _verify_credentials(self):
                 )
             )
         except QiskitServerlessException as reason:
-            raise QiskitServerlessException("Credentials couldn't be verified.") from reason
+            raise QiskitServerlessException(f"Credentials couldn't be verified: {reason}") from reason
```
avilches, line 190: I think the whole try/except could be removed. I guess the original idea was specifically to hide the source of the exception behind a more generic one. But if we're going to show the reason now (which I agree, the `safe_json_request` already prints the right reason in the whole code), we can simply not catch anything and let the `QiskitServerlessException` raised in `safe_json_request` reach the user...

gateway/api/authentication.py
avilches: 1) I would use a logger.exception(error). 2) Do we have a test for this?

Tansito: I think we are commenting the [same line](https://github.com/Qiskit/qiskit-serverless/pull/2338#discussion_r3582778038) Berto haha. For me it is an `error` (I'm not aware about exception for traces 🤔)

ElePT: Sure, I didn't pay much attention to the logs, for me the key thing was raising the error below. We can have `logger.exception`

client/qiskit_serverless/core/clients/serverless_client.py
ElePT: True, it doesn't add much, but I kept it because if we see "Credentials couldn't be verified" it's easy to know where it was raised in our side of the code. I guess I am so used to seeing it that I would find it confusing to not see "Credentials couldn't be verified"

korgan00 (dismissed):
LGTM

HuangJunye (approved):

PR #2339, opened 2026-07-14
Replace GPU allowlist ConfigMap with a gpu field on Program

ItamarGoldman:
### Summary

- [x] Add a `gpu` `BooleanField` (default `False`) to the `Program` model (`gateway/core/models.py`), analogous to the existing `gpu` field already present on `Job` and `ComputeResource`.
- [x] Update `_runner_config()` in `gateway/api/use_cases/programs/run.py` to read `function.gpu` directly instead of calling `create_gpujob_allowlist()`.
- [x] Remove the now-unused `create_gpujob_allowlist()` helper, the `GATEWAY_GPU_JOBS_CONFIG` setting, and the `gpu-jobs.yaml` ConfigMap/mount from the Helm chart.
- [x] Remove from charts GPU job config files.
- [x] Update tests.
- [x] Add a Django migration for the new `Program.gpu` field.
- [x] Add a releasenotes.

**Note:**
Serializers are not going to be updated. The field will be only editable through the backoffice.


gateway/core/models.py
```diff
@@ -151,6 +151,7 @@ class Program(ExportModelOperationsMixin("program"), models.Model):
         max_length=20, choices=RUNNER_CHOICES, default=RAY, help_text="Execution backend for this program"
     )
 
+    gpu = models.BooleanField(default=False, null=False)
```
Tansito: Can we add a helper text here as we have in some new fields? I think it is being very useful to determine for what we are using the field. In this case we can specify that this logic is only for ray runners that want to execute GPU jobs in GPU nodes.

Tansito (commented):
Awesome work @ItamarGoldman ! Just a minor comment before approve it 😄

gateway/core/models.py
Tansito: Little example [here](https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/core/models.py#L151) for your reference, Itamar.

ItamarGoldman: Thats a nice addition! I added a help text as suggested 😄

ElePT (commented):
The changes LGTM, my only comment here is that we should make sure to review if there are any changes required in the internal repo, as the deployment template has changed.

gateway/core/models.py
```diff
@@ -151,6 +151,14 @@ class Program(ExportModelOperationsMixin("program"), models.Model):
         max_length=20, choices=RUNNER_CHOICES, default=RAY, help_text="Execution backend for this program"
     )
 
+    gpu = models.BooleanField(
+        default=False,
+        null=False,
```
avilches, line 157: The field must be nullable and have a default value to allow rollback in the database. Imagine we roll back to a previous version: the db will require specifying the gpu on create because there is no default value and it must not be null, so every insert would fail.

Tansito: Well, it has a default, it is `False` as a `BooleanField`. But it's true that for a rollback the correct configuration here would be:

```py
        default=False,
        null=True,
```

For old entries with no values, so the field in database can be nullable. WDYT @avilches ? This case is something I didn't think about for `BooleanField`s and we were doing it wrong until now probably 😅

ItamarGoldman: Great catch @avilches ! I will change it. @Tansito , I can change other gpu fields (like in Job model) to also accept null as part of this PR. Is it alright?

ItamarGoldman:
> The changes LGTM, my only comment here is that we should make sure to review if there are any changes required in the internal repo, as the deployment template has changed.

I have opened a PR in the internal repo with the changes needed. Merging the changes here and syncing won't break the internal repository. Full detail in the internal repo PR.

korgan00 (dismissed):
LGTM

gateway/core/models.py
Tansito: I would let that for other PR, It's not urgent now. Previous fields are just working so.

Tansito (dismissed):

Tansito:
Thanks you @ItamarGoldman , good work here!

avilches (dismissed):
A kindly reminder you have to re-generate the migration every time you change the models.py, no only the first time. The 0055 migration needs regenerating, right now it still has default=False and doesn't match the model after the null change...

But instead of `null=True`, please change it with `db_default=False`: It keeps the field NOT NULL while adding a real db default in postgress, so old code inserts during a rollback still get false, without making the flag tri-state. Verified the SQL: with db_default there's no DROP DEFAULT, so the default stays in the DB.

Tansito:
Let's do that @ItamarGoldman that Berto proposes: 

```py
default=False,
db_default=False
null=False,
```

and let's see what migration it generates. Monday we can review it and deploy it in staging to test it together 😄

gateway/api/migrations/0055_program_gpu.py
```diff
@@ -1,4 +1,4 @@
-# Generated by Django 6.0.7 on 2026-07-20 12:09
+# Generated by Django 5.2.16 on 2026-07-20 13:39
```
Tansito, line 1: Be aware Itamar that you were using 6.0.7 from Django when in our requirements we have 5.2.16. This might be important, can you review your current setup?

Tansito (approved):
LGTM, @ItamarGoldman take a look to my last comment if you can 🙏

PR #2340, opened 2026-07-14
Load .env file in Gateway/Scheduler

avilches:
### Summary

This adds optional loading of a local `.env` file into the environment on startup, controlled by an `ENV_FILE` variable, so developers can point their local server at a different configuration without editing code or exporting variables by hand every time.

It is opt-in and does nothing unless `ENV_FILE` is set. Real deployments do not set it (docker compose and Kubernetes pass their configuration through actual environment variables), so they are unaffected.

### Details and comments

- New `python-dotenv` dependency, used only by `manage.py`.
- Set `ENV_FILE` to the file you want, resolved next to `manage.py`, for example `ENV_FILE=.env.staging python manage.py runserver`.
- `load_dotenv(..., override=False)` so any variable already set in the shell still wins over the file.
- If `python-dotenv` is not installed or the file does not exist, it is a no-op.
- `.env` files are added to both `.dockerignore` files so they never get baked into the image.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -21,6 +21,11 @@ docker-compose-dev.yaml
 .idea
 .vscode/
 
+### Local env files (never bake into the image)
+
+**/.env
+**/.env.*
+
 ### Python project ignores
 
 *.pyc
--- a/gateway/.dockerignore
+++ b/gateway/.dockerignore
@@ -1,3 +1,5 @@
+.env
+.env.*
 *.pyc
 *.pyo
 *.mo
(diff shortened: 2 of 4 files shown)


ElePT (approved):
This can be very useful to do local debugging of the gateway, as it allows to spin up just the gateway + DB and add breakpoints, etc. For me, as long as it's not baked into the image or affects CI, LGTM.

avilches:
> This can be very useful to do local debugging of the gateway, as it allows to spin up just the gateway + DB and add breakpoints, etc. For me, as long as it's not baked into the image or affects CI, LGTM.

Yep. There is a rule in the `.dockerignore` to explicitly exclude any `**/.env*` file.

ismaelRozasRamallal (approved):

PR #2341, opened 2026-07-14
Arguments validations (2/3): validate against the schema and add validate_arguments endpoint

avilches:
- Depends on #2330

### Summary
Validate the arguments passed to `/programs/run` against the function's declared `arguments_schema` before creating the job, so invalid arguments are rejected with a 400 and no orphaned job or config rows are left behind. Also add a `POST /programs/validate_arguments/` endpoint that checks arguments against a function`s schema without submitting a run.

### Details and comments
- Adds `jsonschema` as a gateway dependency.
- Validation runs before the job and its config are created in `/run`, so an invalid payload never creates partial rows.
- The endpoints catch `jsonschema.ValidationError` directly so the 400 response keeps the `path` to the offending field, instead of flattening it to a generic message.


gateway/api/v1/views/programs/run.py
```diff
@@ -144,21 +146,32 @@ def run_program(request: Request) -> Response:
...
+    except jsonschema.ValidationError as exc:
+        return Response(
+            {"message": exc.message, "path": [*exc.path]},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+    except json.JSONDecodeError as exc:
+        return Response(
+            {"message": f"arguments is not valid JSON: {exc.msg}"},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
```
korgan00: Why not put this into `endpoint_handle_exceptions`? maybe creating a dedicated exception

gateway/api/v1/views/programs/validate_arguments.py
```diff
@@ -0,0 +1,112 @@
...
+    except jsonschema.ValidationError as exc:
+        return Response(
+            {"message": exc.message, "path": [*exc.path]},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+    except json.JSONDecodeError as exc:
+        return Response(
+            {"message": f"arguments is not valid JSON: {exc.msg}"},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
```
korgan00: I think this should be an use case. We decided not to access databases into views and let that to use cases.

korgan00 (changes requested):
Nice, just a couple of things here

gateway/api/v1/views/programs/run.py
avilches: You are right. I created a specific exception for this case called `InvalidArgumentsException` and it will be handled in the `exception_handler.py` module

gateway/api/v1/views/programs/validate_arguments.py
avilches: Correct. Changed.

gateway/api/v1/views/programs/validate_arguments.py
```diff
@@ -0,0 +1,73 @@
...
+    """Validates arguments against the function schema without creating a job."""
+    user = request.user
+    accessible_functions = cast(FunctionAccessResult, request.auth.accessible_functions)
+
+    serializer = InputSerializer(data=request.data)
+    if not serializer.is_valid():
+        return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    provider_name = sanitize_name(serializer.data.get("provider"))
+    function_title = sanitize_name(serializer.data.get("title"))
```
korgan00: We can sanitize inside `InputSerializer` and check if the string is valid. If we sanitize post validation, we can have an empty title for example.

```
    def validate_title(self, value: str):
        """
        Validates the function title
        """
        return sanitize_name(value)
```

I dont know if we should check for empty or the `CharField` does it after `validate_title`

korgan00 (changes requested):
In general it looks good. Just a minor comment.

gateway/api/v1/views/programs/validate_arguments.py
avilches: Changed

korgan00 (approved):

PR #2345, opened 2026-07-15
Fix/surface non json error responses

ElePT:
### Summary

A function upload against staging failed with:

`QiskitServerlessException: JSON1001: Expecting value: line 1 column 1 (char 0)`

This error is misleading. It says the server sent bad JSON, but what actually happened is that an intermediary returned a non-JSON body (a Cloudflare/ingress block page, or an empty body) and the client passed it to` json.loads()`. The real information (an HTTP 403 and a message like "Sorry, you have been blocked") was discarded and replaced with a generic decode error.

The cause is that several client paths parse the body as JSON before checking whether the response succeeded:

- `safe_json_request()` (used by `upload()`, `run()`, and most API calls) calls `json.loads(response.text`) first. Any non-JSON error body raises `JSONDecodeError` and is reported as `JSON1001`. The existing `if not response.ok` branch that would have surfaced the status code was never reached for non-JSON bodies.
- `result()` decoded the body with `json.loads(...)` and no fallback.
-` logs() / provider_logs()` returned `"Error fetching logs."` on non-OK responses, hiding the status and body.

Result: whenever Cloudflare, a proxy, or COS returned an error, the user saw `JSON1001` instead of the actual status and message. 

**IMPORTANT: I am open to suggestions if you think I'm exposing too much by sharing the error codes. I just think it would be beneficial for the debugging process, but we can find a middle ground.** 

### Details and comments

Fix: On any non-OK response, surface the HTTP status code and a snippet of the body instead of parsing it as the expected payload.

- New `raise_for_non_ok_response(response)` helper in `utils/json.py`: on a non-OK response, raises with the status code and body: JSON error values if the body is JSON, otherwise a whitespace-collapsed, length-capped text snippet (`_body_snippet`). Empty bodies report (empty response body).
- `result()`, `logs()`, and `provider_logs()` call it before decoding, so a block page surfaces as `HTTP 403: ... Sorry, you have been blocked.`
- `safe_json_request()` hardened: a non-OK, non-JSON body now surfaces status and snippet. Only a 2xx body that fails to decode still reports `JSON1001`, which is the correct signal for a server-contract


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -434,11 +449,15 @@ def result(self, job_id: str) -> Dict[str, Any]:
...
+        # Any non-OK response (from the gateway, COS, or an intermediary such as
+        # Cloudflare returning a block page) is surfaced with its status and body
+        # instead of being fed to the JSON parser as a cryptic decoding error.
+        raise_for_non_ok_response(response)
         # Not all redirects go to COS — HTTP→HTTPS redirects stay on the same host.
         # Checking the hostname detects only redirects to an external host (COS/MinIO).
         redirected_to_cos = urlparse(response.url).hostname != urlparse(gateway_url).hostname
         if redirected_to_cos:
-            return json.loads(response.text, cls=QiskitObjectsDecoder) if response.ok else {}
+            return json.loads(response.text, cls=QiskitObjectsDecoder)
```
marceloamaral, line 460: good catch!

marceloamaral (approved):
Looks good to me, let's wait for the tests

Tansito (approved):
As a short term solution for the error we found with Cloudflare I think is fine. As a major future upgrade, I think it would be nice start thinking in a better pattern to manage this. Maybe a Strategy or an Adapter pattern could fit here.

avilches (approved):
The PR is correct. Exposing information through errors is only a problem that can be resolved on the server. On the client side, the information has already arrived, so hiding it from the user doesn't make it disappear...

avilches (dismissed):
Exposing information through errors is only a problem that can be resolved on the server. On the client side, the information has already arrived, so hiding it from the user doesn't make it disappear... The PR is correct.

avilches (dismissed):
Exposing information through errors is only a problem that can be resolved on the server. On the client side, the information has already arrived, so hiding it from the user doesn't make it disappear... The PR is correct.

PR #2348, opened 2026-07-17
Arguments validations (3/3): Add validate_arguments to the client and rename QiskitFunction.schema to arguments_schema

avilches:
- Depends on #2341 

### Summary
Add client SDK support for argument validation. `RunnableQiskitFunction` gets a `validate_arguments(arguments)` method that calls the new `/programs/validate_arguments/` endpoint to check inputs without submitting a run, and the previously unused `QiskitFunction.schema` field is renamed to `arguments_schema` (a JSON Schema dict) and sent in the upload payload.

### Details and comments
- Breaking change: `QiskitFunction.schema` becomes `arguments_schema`, and its type changes from `Optional[str]` to `Optional[Dict[str, Any]]`. There is no deprecation path since the old field was never used.
- Adds an end-to-end integration test in `tests/integration/test_jobs.py`.


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -646,6 +646,40 @@ def function(self, title: str, provider: Optional[str] = None) -> Optional[Runna
...
+        title: str,
+        arguments: Optional[Dict[str, Any]] = None,
+        provider: Optional[str] = None,
+    ) -> dict:
+        """Validate arguments against a function's schema without creating a job.
+
+        Args:
+            title: function title, optionally in "provider/title" format
+            arguments: arguments dict to validate
+            provider: optional provider name
```
isma-roza, line 650: Why we add here the provider specifically if that's part of the title?

isma-roza: My fault, I see that format is optional

isma-roza (approved):
Lgtm!

korgan00:
Why do we want the schema in the client?

avilches:
> Why do we want the schema in the client?

Hi! 2 reasons: 1) the schema can only be sent to the gateway in the /upload, and the upload only happen in the SDK, so `QiskitFunction` is where a function author declares it. It works the same way as every other part of the upload metadata (dependencies, env_vars, image, working_dir): a field on the dataclass that upload() serializes into the request.
2) And this is where your comment was right: it used to be write-only, since the gateway never returned it, so a function read back always had arguments_schema=None. But now I changed it: the schema is now exposed by `get_by_title` and `list`, so the client can decode it back into a dict, and you can inspect what arguments a function expects even when you did not upload it yourself.

The full flow is written up in specs/ARGUMENTS_VALIDATION.md, including where validation happens and why it is server-side only.

korgan00:
> > Why do we want the schema in the client?
> 
> Hi! 2 reasons: 
> ...

Thank you

avilches:
@isma-roza I need your review from @ismaelRozasRamallal :)

ismaelRozasRamallal (approved):

PR #2349, opened 2026-07-20
Fix `Job.status()` ignoring `sub_status` 

ElePT:
## Summary

- `Job.status()` was always returning `RUNNING` for jobs that report a sub-status, instead of the granular sub-status value.
- Root cause: PR #2072 introduced a fast path via `get_job_data()` that only read the top-level `status` field, silently dropping the `sub_status` check that existed in `ServerlessClient.status()`.
- Fix: restore the sub-status check in the `get_job_data` fast path, mirroring the existing logic in `ServerlessClient.status()`.

## Root cause bisection

Introduced in commit `b99435b47` (2026-05-04) — "Expose fleet_id on Job and allow status polling via get_job_data" (#2072).

Before that commit, `Job.status()` always delegated to `self._job_service.status()` → `ServerlessClient.status()`, which correctly handled `sub_status`. The new fast path bypassed that method entirely.

## Test plan

- [x] Added unit tests for the fixed `Job.status()` path: sub_status returned when RUNNING, ignored for non-RUNNING states, `None` sub_status falls back to RUNNING.

### DIFF
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -195,7 +195,11 @@ class Job:
         fresh_data = self._job_service.get_job_data(self.job_id)
         if isinstance(fresh_data, dict):
             self.raw_data.update(fresh_data)
-            return _map_status_from_serveless(fresh_data.get("status"))
+            raw_status = fresh_data.get("status")
+            sub_status = fresh_data.get("sub_status")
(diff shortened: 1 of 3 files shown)


HuangJunye (dismissed):
LGTM. Thanks for the quick fix and adding new tests!

ElePT:
I added a release note, @HuangJunye can you re-review?

HuangJunye (approved):

PR #2356, opened 2026-07-21
Wipe job env vars from DB immediately after submission

ElePT:
### Summary

Fleets job env vars were stored in the DB from job creation until the job reached a terminal state, which for long-running jobs could be hours. This clears `job.env_vars` to `{}` in `execute_fleets_job` right after `runner.submit()` returns, at which point the env vars have already been read and forwarded to Code Engine, so they are no longer needed.

The field is cleared unconditionally (success and failure paths alike) and persisted via the existing `save_direct` call by adding `env_vars` to the field list.

### Details and comments

The existing wipe in `UpdateFleetsJobsStatuses.to_terminal()` (on job completion) is kept as-is and becomes a harmless no-op for newly submitted jobs.

Arguments, logs, and results for Fleets jobs already go to COS and are not stored in the DB, only env vars had this gap.


Tansito (commented):
LGTM @ElePT question here, if we are restarting in the submit the env_vars (something I see correct) how do you see to remove the override from the terminal states also?

- https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/scheduler/tasks/update_fleets_jobs_statuses.py#L111
- https://github.com/Qiskit/qiskit-serverless/blob/main/gateway/scheduler/tasks/update_ray_jobs_statuses.py#L57

So we only have one place where we are doing this cleanup.

ElePT:
Thanks @Tansito! That makes sense, I've extended the approach to Ray jobs too, so now there's no longer a terminal state wipe, for any of them, the env vars are directly wipe after submitting the executions.

avilches (approved):
Perfect. I know the Ray part is trickier than the fleets because due to wrapping the actions in tracer spans, the variable set is in one place and the save_direct in another. Thank goodness it doesn't matter because the Ray part is going to die, and the Fleets one doesn't have tracer so it's extremely simple! Approved!

PR #2357, opened 2026-07-22
Add fleet-node base image for Fleets workers

marceloamaral:
### Summary

Adds `fleet-node`, a Python 3.12 base image for Code Engine Fleets workers, mirroring `ray-node` (UBI9 + the qiskit dynamic dependencies + the client). The fleets integration-test worker now builds `FROM` this image instead of `python:3.11-slim`, so the test runs against the same base a fleet job uses. It also becomes the default Fleets runtime image, pinned to the release version like ray-node.

### Details and comments

- `make build-fleet-node` builds it alongside ray-node and gateway (folded into `build-all`/`push-all`), and also tags it `fleet-node:latest` locally so the test worker can build on it without the registry path.
- `docker compose build` does not build a `FROM`-referenced base, so the fleets integration CI job (and the local docs) build `fleet-node` first; `docker-compose-fleets-test.yaml` takes it as a `BASE_IMAGE` build arg.
- The test worker adds only what the mock harness needs on top of the base: `s3fs`/`fuse` (from EPEL, since UBI9 doesn't ship s3fs) and `boto3`. `storage.py` drops the `-o nonempty` s3fs option (removed in newer s3fs) and removes the base-created mount dir before symlinking into it.
- `fleet-node/Dockerfile` is added to the build-smoke matrix.
- `FLEETS_DEFAULT_IMAGE` now defaults to `private.icr.io/quantum-public/qiskit-serverless/fleet-node:0.34.0` (was a `helloworld` placeholder), and `update-component-versions.yaml` bumps the fleet-node tag in settings alongside ray-node on release. The locally-built `fleet-node:latest` test tags are left as-is (rebuilt each CI run, never pulled). fleet-node installs the client without the `ray` extra, since Fleets is the non-Ray engine.

### Checklist

- [x] I have added a [release note](https://github.com/Qiskit/qiskit-serverless/blob/main/CONTRIBUTING.md#release-notes) for any user-facing change (new feature, bug fix, deprecation, or breaking change), or this PR has no user-facing changes.


fleet-node/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+# ray-node/Dockerfile. This is the runtime image Code Engine fleet tasks run in;
+# the integration-test worker (tests/fleets/fleet-worker/Dockerfile) builds FROM it.
+#
+# Build from repo root (client/ and ray-node/ must be in the build context):
+#   docker build --platform linux/amd64 -f fleet-node/Dockerfile -t fleet-node:latest .
+#
+# No CUDA base: nothing here links against CUDA (qiskit-aer is the CPU build),
+# and CE's NVIDIA container runtime injects the driver + nvidia-smi into GPU
+# workers at runtime. UBI9 (glibc 2.34) satisfies the mthree manylinux_2_34 wheel;
+# all pinned deps ship linux/x86_64 wheels for 3.12, so no build toolchain is needed.
```
ElePT, line 24: This is good, did you get a chance to test the image on a real fleet in code engine to verify it?

.github/workflows/build-containers-test.yaml
```diff
@@ -20,6 +20,7 @@ jobs:
       matrix:
         dockerfile:
           - ray-node/Dockerfile
+          - fleet-node/Dockerfile
           - gateway/Dockerfile
         python:
           - "3.11"
```
ElePT: This python version is actually never used as a build arg, as far as I've seen. Having it be 3.11 is confusing when the fleet node is already 3.12, I think that we should be able to delete this.

tests/fleets/fleet-worker/Dockerfile
```diff
@@ -9,14 +9,36 @@
...
+
+# boto3 for the worker's MinIO/COS access. The client and patched pip/setuptools
+# already come from the base image (fleet-node), so no build-tool re-upgrade here.
+RUN pip install --no-cache-dir boto3
 
-FROM python:3.11-slim
-RUN apt-get update && apt-get install -y s3fs fuse && rm -rf /var/lib/apt/lists/*
-# Upgrade the base image build tools to versions without known CVEs
-# (CVE-2026-8643 in pip, CVE-2026-24049 in wheel, CVE-2026-59890 in setuptools).
-RUN pip install --no-cache-dir --upgrade "pip>=26.1.2" "setuptools>=83.0.0" "wheel>=0.46.2"
```
ElePT, line 17: The old Dockerfile had explicit floors (pip>=26.1.2, setuptools>=83.0.0, wheel>=0.46.2) to address specific CVEs because the base image wasn't ubi9. These don't match with the versions in the new one, did you verify that the installed versions in the new dockerfile meet these minimums?

fleet-node/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+COPY ray-node/requirements-dynamic-dependencies.txt /tmp/
+RUN pip install --no-cache-dir \
+        -r /tmp/requirements-dynamic-dependencies.txt && \
+    rm /tmp/requirements-dynamic-dependencies.txt
+
+# Install the client last for better layer-cache reuse. Keep this on its own RUN
+# with plain '&&' so a failed install fails the build — do NOT chain '|| true'
+# guards onto this step (that would silently ship an image with no client).
+# (No symengine lib/lib64 copy: Qiskit 2.1+ dropped symengine, so it is not
+# installed here; ray-node/gateway removed the same copy in 0.33.0.)
```
ElePT:
```suggestion
# installed here; ray-node/gateway removed the same copy in 0.34.0.)
```

fleet-node/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+# prebuilt 3.12 wheels.
+COPY ray-node/requirements-dynamic-dependencies.txt /tmp/
+RUN pip install --no-cache-dir \
+        -r /tmp/requirements-dynamic-dependencies.txt && \
+    rm /tmp/requirements-dynamic-dependencies.txt
+
+# Install the client last for better layer-cache reuse. Keep this on its own RUN
+# with plain '&&' so a failed install fails the build — do NOT chain '|| true'
+# guards onto this step (that would silently ship an image with no client).
+# (No symengine lib/lib64 copy: Qiskit 2.1+ dropped symengine, so it is not
```
ElePT:
```suggestion
# (No symengine lib/lib64 copy: Qiskit 2.2 dropped symengine, so it is not
```

fleet-node/Dockerfile
```diff
@@ -0,0 +1,70 @@
...
+USER 0
+# /data and /function_data are Ray runner PDS mount points.
+# /function_user_data, /job_user_data, /function_provider_data, /job_provider_data
+# are the Fleets runner 4-mount COS path layout.
+RUN mkdir /data /function_data /function_user_data /job_user_data /function_provider_data /job_provider_data && \
+    chown 1000:1000 /data /function_data /function_user_data /job_user_data /function_provider_data /job_provider_data && \
+    python${PYTHON_VERSION} -m ensurepip --upgrade && \
+    pip install --upgrade --no-cache-dir \
+        "pip>=24.2" \
+        "setuptools>=80.0.0"
```
ElePT: This is a fault of the original dockerfile, but we only need the new mounts in the fleets image and the old mounts in the ray image:

```suggestion
# /function_user_data, /job_user_data, /function_provider_data, /job_provider_data
# are the Fleets runner 4-mount COS path layout.
RUN mkdir /function_user_data /job_user_data /function_provider_data /job_provider_data && \
    chown 1000:1000 /function_user_data /job_user_data /function_provider_data /job_provider_data && \
    python${PYTHON_VERSION} -m ensurepip --upgrade && \
    pip install --upgrade --no-cache-dir \
        "pip>=24.2" \
        "setuptools>=80.0.0"
```

ElePT (commented):
Thanks a lot @marceloamaral! I gave the PR a first pass and left some comments, let me know if anything isn't clear.

fleet-node/Dockerfile
marceloamaral: Yes, and I could run `nvidia-smi` in a fleet!

.github/workflows/build-containers-test.yaml
marceloamaral: Good point, I removed it, I believe it was used before for job-name suffix, and nothing is really referencing `matrix.python`...

tests/fleets/fleet-worker/Dockerfile
marceloamaral: Good, point, I made the CVE floors explicit in the fleet-node base (`pip>=26.1.2`, `setuptools>=83.0.0` and `wheel` isn't need for ubi9), so that every image `FROM` it now inherits them guaranteed floor versions...

fleet-node/Dockerfile
marceloamaral: Good point! Added!

ElePT (dismissed):
From what I can see, the new base image makes sense. LGTM

gateway/main/settings.py
```diff
@@ -446,12 +446,11 @@
...
 CE_ICR_PULL_SECRET = os.environ.get("CE_ICR_PULL_SECRET", None)
-#######
-### Note: Update the default image with the default custom function ICR image
-#######
+# Default runtime image for imageless (custom-code) Fleets jobs; real deployments
+# override it via ce.fleetsDefaultImage.
 FLEETS_DEFAULT_IMAGE = os.environ.get(
     "FLEETS_DEFAULT_IMAGE",
-    "private.icr.io/qc-qiskit-functions-ce-staging/helloworld:latest",
+    "private.icr.io/quantum-public/qiskit-serverless/fleet-node:0.33.0",
```
ElePT: Agreed to use a versioned image, but then it should be the latest one:

```suggestion
    "private.icr.io/quantum-public/qiskit-serverless/fleet-node:0.34.0",
```

ElePT (commented):
Added one last suggestion, you'll have to update the register tag before merging this PR.

ElePT:
Also, note my comment in https://github.com/Qiskit/qiskit-serverless/pull/2362#issuecomment-5068535302, we can do it as a follow-up, but we should review the env variable names and regularize them with the ray ones, adding a selector image for custom fleets functions.

ElePT (dismissed):
LGTM. We will have to iterate a bit on the env variables and the image design, but this is a good starting point to have a formalized dockerfile to iterate on. Thanks a lot @marceloamaral!

.github/workflows/build-containers-test.yaml
```diff
@@ -20,9 +20,10 @@ jobs:
       matrix:
         dockerfile:
           - ray-node/Dockerfile
+          - fleet-node/Dockerfile
           - gateway/Dockerfile
         python:
-          - "3.11"
+                  - "3.11"
```
ElePT:
```suggestion
          - "3.11"
```

ElePT (commented):
It looks like the action is dependant on this python field, so let's leave it for the time being and fix in a follow-up

ElePT (approved):
LGTM

PR #2362, opened 2026-07-23
Refactor ray-node Dockerfiles into single multi-stage file, bump default to Python 3.12

ElePT:
### Summary

Bump the default ray-node Python version from 3.11 to 3.12.

### Details and comments

The `ARG PYTHON_VERSION=3.12` default replaces the previous hardcoded `python3.11` references.


marceloamaral (dismissed):
Looks good!

Should we also add two phases in https://github.com/Qiskit/qiskit-serverless/pull/2357? 
Maybe in another PR...

ElePT:
> Looks good!
> 
> Should we also add two phases in #2357? Maybe in another PR...

Thanks for the early review @marceloamaral! Could you take a look now? I was working on the internal PR in parallel (1661) and have gone back and forth a couple of times with the changes. Regarding the 2 phases in fleets: if we keep things as they are, we are probably going to have to add both a fleets node image and a fleets selector image, so not only should we add the 2 phases, we are going to have to add and reference the node and selector images in the code accordingly. The cleanup helped me get clarity on that. But yes, it can be a follow-up.

docs/custom_images/run_on_apple_silicon.rst
```diff
@@ -20,23 +20,24 @@ arm64** ``ray-node`` image that can be used as-is or for custom image developmen
...
-This repository ships an arm64 variant of the base image, ``ray-node/Dockerfile.arm64``, and a
+This repository ships an arm64 stage (``AS arm64``) inside ``ray-node/Dockerfile``, and a
 matching dependency file, ``ray-node/requirements-dynamic-dependencies-arm64.txt``. Build it from
 the repository root:
 
 .. code-block::
    :caption: Build the arm64 ray-node image
 
    docker build --platform linux/arm64 \
+     --target arm64 \
```
marceloamaral, line 31: Interesting, looks like a better approach!

marceloamaral (approved):
/LGTM

PR #2363, opened 2026-07-27
Harden Fleets job wrappers: drop capabilities and set no-new-privs

ElePT:
## Summary

IBM Code Engine Fleets has no equivalent to Kubernetes' `securityContext`,  there is no way to set `runAsGroup`, `allowPrivilegeEscalation: false`, or capability drops in the fleet spec. As a result, Fleets job containers start with a permissive capability bounding set and without `no-new-privs`, unlike Ray cluster pods which were already hardened in PR #2318.

This PR closes that gap by embedding a `_harden()` function at the top of both fleet wrapper templates. Because the wrapper is the first code that runs inside the container (before any user logic, before the subprocess is launched) calling it there achieves the same effect as a `securityContext` spec would, using only `ctypes` + `stdlib` (no extra packages or Dockerfile changes required).

**What `_harden()` does, in order:**

1. `PR_SET_NO_NEW_PRIVS` — process and all descendants can never gain privileges via setuid/setgid binaries or file capabilities
2. `setresgid` — resets GID to match UID if the container starts with `gid=0` (the CE default when `runAsGroup` is not configurable)
3. `PR_CAPBSET_DROP` (caps 0–41) — clears the capability bounding set
4. `capset(2)` — zeroes effective, permitted, and inheritable capability sets
5. `PR_CAP_AMBIENT_CLEAR_ALL` — clears ambient capabilities (silently ignored on kernels < 4.3)

The function is non-fatal on platforms that lack these syscalls (e.g. musl/Alpine, macOS) so it does not affect the test suite or non-Linux environments.

**Files changed:**
- `gateway/templates/fleet_custom_job_wrapper.py`
- `gateway/templates/fleet_provider_job_wrapper.py`

## Test plan

- [x] Existing wrapper template tests pass (log routing, truncation, exit code propagation, periodic flush)
- [x] New test `test_harden_applies_no_new_privs_and_drops_capabilities` verifies `NoNewPrivs=1` and all capability sets are zeroed inside a real container


Tansito (approved):
Thanks to looks at this @ElePT , really thank you. I was chatting with Claude in case there would be other alternatives and so on and initially I didn't find anything to be honest. I'm going to approve it so we can test it as soon as we can to see if it works the patch.

PR #2364, opened 2026-07-27
Show function artifact in job storage files admin view

avilches:
### Summary

The admin "Storage Files" page for a job lists the job's own storage (results, logs, arguments, data files) but not the uploaded function code. That code is stored per Program (`program.artifact`), separate from the job, so it never showed up there and there was no way to see it from the admin.

This adds a "Function artifact" group to that page so the uploaded code archive is visible alongside the rest.

### Details and comments

The new group is built in `JobFileExplorer.explore`, so it works for both Ray and Fleets jobs. Metadata (size, last modified) is read through the FileField's storage API, so it works whether the media volume is a local disk or a COS backed mount. When a function was uploaded as provider image (no artifact), the group is omitted. No template changes are needed since the page already renders whatever groups it gets.


ElePT (approved):
LGTM, but I have one question: given that the artifact is only used for custom functions, should we also add a way to visualize the uploaded image for provider functions? knowing the size, for example, could be handy in some cases.

PR #2367, opened 2026-07-28
Align Functions usage events with the billing service contract

ismaelRozasRamallal:
### Summary
The billing consumer bills **per metric**, but we were publishing usage events in a job-lifecycle shape with no notion of a metric — `event_type`, a raw `usage_nanoseconds` value and a `function_id`, none of which the consumer can resolve a billable metric against. This reshapes the payload so every message carries the metric and its value. Step-0 change that unblocks the consumer work on the billing side.

```diff
  "data": {
-     "event_type": "job_started",
-     "usage_nanoseconds": 5000000000,
+     "metric_type": "classical_time",
+     "metric_value": 5000,
      "instance_crn": "crn:v1:bluemix:public:quantum-computing:...",
-     "function_id": "0fa1...",
+     "resource_id": "0fa1...",
+     "job_started": true,
+     "job_completed": false
  }
```

- `metric_type` — *what* is billed. Passed in by the caller, since it derives from provider/function/size and compute profile.
- `metric_value` — *how much*, in **milliseconds** (how usage is measured today).
- `resource_id` — contract field name for the existing `function_id` value (the job id).
- `job_started` / `job_completed` — lifecycle boundaries as explicit booleans, so consumers don't parse the metric type. `job_started: true` only on the started event, `job_completed: true` only on the completed event.

CloudEvents envelope, topic (`quantum.{environment}.function-usage.v1`) and flush-and-raise delivery are unchanged.

### Details and comments
Three commits:

1. **Make event streams interface metric-aware** — `metric_type` param on the abstract interface and the no-op client; `emit_job_ended` → `emit_job_completed` to match the `function_job_*` vocabulary.
2. **Emit metric-oriented usage payload from Kafka client** — reshapes `_publish`; `_usage_ns` → `_usage_ms`. Client tests now assert the whole `data` dict, so an unexpected or missing field fails.
3. **Pass the billed metric type from the scheduler** — supplies the metric at the three lifecycle call sites via a `CLASSICAL_TIME_METRIC_TYPE = "classical_time"` constant.

| Helper | `metric_value` | `job_started` | `job_completed` |
| --- | --- | --- | --- |
| `emit_job_started` | `0` | `true` | `false` |
| `emit_job_in_progress` | elapsed ms | `false` | `false` |
| `emit_job_completed` | elapsed ms | `false` | `true` |

### Checklist

### DIFF
--- a/gateway/core/ibm_cloud/event_streams/abstract_event_streams_client.py
+++ b/gateway/core/ibm_cloud/event_streams/abstract_event_streams_client.py
@@ -19,13 +19,13 @@ class EventStreamsClient(ABC):
     """Interface for job usage event publishing."""
 
     @abstractmethod
-    def emit_job_started(self, job) -> None:
-        """Publish or log a job_started event."""
+    def emit_job_started(self, job, metric_type: str) -> None:
+        """Publish or log a function_job_started event for the given metric."""
 
     @abstractmethod
-    def emit_job_in_progress(self, job) -> None:
-        """Publish or log a job_in_progress event."""
+    def emit_job_in_progress(self, job, metric_type: str) -> None:
+        """Publish or log a function_job_in_progress event for the given metric."""
 
     @abstractmethod
-    def emit_job_ended(self, job) -> None:
-        """Publish or log a job_ended event."""
+    def emit_job_completed(self, job, metric_type: str) -> None:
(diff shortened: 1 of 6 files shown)


korgan00 (approved):
LGTM

PR #2373, opened 2026-07-30
Consolidate ray-node and fleet-node under a docker-images directory

marceloamaral:
### Summary

The dynamic dependencies list was duplicated. ray-node and fleet-node each carried their own `requirements-dynamic-dependencies.txt` with identical content, and gateway copied ray-node's, which is what let them drift. Following the review, this consolidates the base worker images into a single `docker-images/` directory so the Dockerfiles and their shared dependencies live in one place:

```
docker-images/
  ray-node/Dockerfile
  fleet-node/Dockerfile
  requirements-dynamic-dependencies.txt
  requirements-dynamic-dependencies-arm64.txt
```

There is now one source of truth for the dependency list, and both node images sit next to it.

### Details and comments

- Moved `ray-node/` and `fleet-node/` under `docker-images/`, and put the shared `requirements-dynamic-dependencies.txt` (plus the arm64 variant) at the top of `docker-images/`. Removed fleet-node's duplicate list.
- ray-node, fleet-node and gateway now build against the shared file. The build context is still the repo root, so the `COPY` lines inside the Dockerfiles keep working.
- Updated every `-f` and `dockerfile:` reference: Makefile, the docker-compose dev and arm64 files, the build-containers and fleets-integration workflows, the Apple Silicon and local-cluster docs, and tests/tox.ini.
- `gateway/api/utils.py` read the list from `../ray-node/` under tests, so it now points at `../docker-images/`. The production path is unchanged: the Dockerfile still copies the file to the same place inside the image.
- ray-node/Dockerfile moves with no logic change; fleet-node/Dockerfile only gets its header build examples repointed. Image contents do not change.
- The internal release pipeline builds these Dockerfiles by path, so it needs a matching update on the internal side before this is released. That change is prepared separately.


ElePT (commented):
Question. What's the motivation to have a separate directory just for the requirements? My idea was to unify the dirs to try to keep the info more easily accessible, I feel like this makes it more separate if that makes sense.

marceloamaral:
> Question. What's the motivation to have a separate directory just for the requirements? My idea was to unify the dirs to try to keep the info more easily accessible, I feel like this makes it more separate if that makes sense.

Hum, that is right, let me consolidate in only one dir...

marceloamaral:
@ElePT could you please review again?

ElePT (commented):
@Tansito, what do you think about this new structure? Do you think it helps?

Tansito:
I'm going to try to look at this today 🙏

Tansito:
It looks good to me, we are going to need to update these references in the internal repository and also there is a test that is not running I don't know why, @marceloamaral : "Test build containers", can you take a look before approve?

marceloamaral:
@Tansito sure, I will wait all test finish first!

marceloamaral:
@Tansito and @dpatel-ops found the problem.

Since I updated the Dockerfiles paths, it renamed the CI checks. So, the matrix in `.github/workflows/build-containers-test.yaml` uses the Dockerfile path as the matrix value, and GitHub builds the check name from it. Then, when I changed `ray-node/Dockerfile` to `docker-images/ray-node/Dockerfile`, the check name changed with it:
```yaml
jobs:
  build-containers:
    name: Test build containers
    strategy:
      matrix:
        dockerfile:
          - docker-images/ray-node/Dockerfile
          - docker-images/fleet-node/Dockerfile
          - gateway/Dockerfile
        python:
          - "3.11"
```
The github branch ruleset still requires the old name by exact string: `Test build containers (ray-node/Dockerfile, 3.11)`

So, since nothing reports that name anymore, and then it stays on "Expected - Waiting for status to be reported" and blocks the PR...

To fix, I keep the check label matching the required name and point the build at the new path:
```yaml
    name: Test build containers (${{ matrix.label }}, ${{ matrix.python }})
    strategy:
      fail-fast: false
      matrix:
        include:
          - label: ray-node/Dockerfile          # label kept to match the required check name
            dockerfile: docker-images/ray-node/Dockerfile
            python: "3.11"
          - label: fleet-node/Dockerfile         # not required in CI
            dockerfile: docker-images/fleet-node/Dockerfile
            python: "3.11"
          - label: gateway/Dockerfile            # unchanged (label == path)
            dockerfile: gateway/Dockerfile
            python: "3.11"
```
We can clean this up later, or drop it once the ray path is deprecated...

Tansito (approved):
Reviewed with @marceloamaral everything fine 🎉

PR #2377, opened 2026-08-03
Add license fee event

korgan00:
### Summary

In kafka_event_streams_client.py, add a dedicated emission for the license fee — e.g. emit_license_fee(job) alongside the existing emit_job_* methods

### Checklist


gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -133,6 +133,7 @@ def to_running(self, job: Job) -> None:
             Job.RUNNING,
         )
         self.event_streams_client.emit_job_started(job, CLASSICAL_TIME_METRIC_TYPE)
+        self.event_streams_client.emit_license_fee(job)
```
ismaelRozasRamallal: Just one thing here 😄 if we recap on the metric type format, it is:
`METRIC_VENDOR_FUNCTION_SIZE`

At this moment, we don't have sizes yet, but we have vendors and functions. So, by the moment, we need to concatenate the vendor name and function id to the metric. That's the only way we have in the other services to identify to what vendor/function we are billing the fee 👍

ismaelRozasRamallal: By other side, just for consistency, what do you think about moving the LICENSE_FEE_METRIC_TYPE to here as the classical one is?

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -83,6 +85,15 @@ def emit_job_completed(self, job, metric_type: str) -> None:
             job_completed=True,
         )
 
+    def emit_license_fee(self, job) -> None:
+        self._publish(
+            job,
+            metric_type=LICENSE_FEE_METRIC_TYPE,
+            metric_value=1,
+            job_started=True,
+            job_completed=False,
```
frankharkins: Is there is only ever one license fee event emitted? If so, `job_completed` should be `True`. We use this flag to know if the metric value will ever be updated.

```suggestion
            job_completed=True,
```

In hindsight, I guess something like `job_and_metric_updates_complete` would be more accurate, but I think it's too late to change now.

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -21,11 +21,14 @@
 from datetime import datetime, timezone
 
 from confluent_kafka import Producer
+from core.models import Job
 
 from .abstract_event_streams_client import EventStreamsClient
 
 logger = logging.getLogger("gateway.ibm_cloud.event_streams_client")
 
+LICENSE_FEE_METRIC_TYPE = "license_qesem"
```
ismaelRozasRamallal: I think now we have the join below we don't need the qesem part 😄 
```suggestion
LICENSE_FEE_METRIC_TYPE = "license"
```

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -83,6 +86,16 @@ def emit_job_completed(self, job, metric_type: str) -> None:
             job_completed=True,
         )
 
+    def emit_license_fee(self, job: Job) -> None:
+        metric_type = "_".join([LICENSE_FEE_METRIC_TYPE, job.program.provider.name, job.program.title])
```
ismaelRozasRamallal, line 90: one doubt just as double check, both provider.name and title are mandatory right?

gateway/tests/core/services/ibm_cloud/event_streams/test_event_streams_client.py
```diff
@@ -221,3 +221,47 @@ def test_emit_job_in_progress_returns_zero_usage_when_running_started_at_is_none
...
+                        client.emit_license_fee(job)
+
+        call_kwargs = mock_producer.produce.call_args[1]
+        published = json.loads(call_kwargs["value"])
+        assert published["specversion"] == "1.0"
+        assert published["type"] == "quantum.production.function-usage.v1"
+        assert published["source"] == "qiskit-serverless/scheduler/fleets"
+        assert published["subject"] == str(job.id)
+        assert published["data"] == {
+            "metric_type": "license_qesem_ibm_test-program",
```
ismaelRozasRamallal: same here pointed before 😄 
```suggestion
            "metric_type": "license_ibm_test-program",
```

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
korgan00: yeah, mistake. Thank you!

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
korgan00: provider itself is optional, so i will change it ;) good catch

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -133,6 +133,8 @@ def to_running(self, job: Job) -> None:
             Job.RUNNING,
         )
         self.event_streams_client.emit_job_started(job, CLASSICAL_TIME_METRIC_TYPE)
+        if job.program.provider:
```
ismaelRozasRamallal, line 138: one nit, can we put here a comment explaining we are doing this if to leave the custom functions out of the license fee events logic since they don't have it?

ismaelRozasRamallal (approved):

PR #2372, opened 2026-07-29
w3id secret refactor

Tansito:
### Summary

This PR is a first step in a series of PR I'm planning to do to standardize our external secrets configuration. This one specifically introduces a minor refactor in the w3id secreto to ensure:
- The secret uses the `secret` field in the values in a structure way (`secrets.w3idSso.name` & `secrets.w3idSso.secretStoreName`)
- The secret store name is dynamic
- Sync the secret to be a key-value secret instead of an arbitrary-secret

This change doesn't require modifications in the internal repository (that's the reason I'm going to open one PR per template, so we can keep track of every change that we need to apply there).

### Checklist

- [x] I have added tests to cover my changes.
- [ ] I have updated the documentation accordingly.
- [ ] I have added a [release note](https://github.com/Qiskit/qiskit-serverless/blob/main/CONTRIBUTING.md#release-notes) for any user-facing change (new feature, bug fix, deprecation, or breaking change), or this PR has no user-facing changes.

PR #2378, opened 2026-08-03
Helm kafka refactor

Tansito:
### Summary

This PR is a follow-up from #2372 and the second PR in a series of refactor I'm planning to do in our helm configuration around secrets to try to leave them as more similar as possible. In this one we are focusing in the kafka secret configuration applying two main changes:
- `kafkaCredentials` was removed in favor of `eventStreams` so we only need to maintain one variable. For reference, `eventStreams` is just being used in our internal repository. I'm just syncing this change and improving the configuration.
- Similar to the previous PR in the other change we are making dynamic the secret store reference.

This PR will require the next changes in the internal repo:
- `kafkaCredentials` refactor
- `secrets.kafka` refactor

### Checklist

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
+++ b/charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
@@ -524,7 +524,7 @@ spec:
               value: {{ .Values.application.eventStreams.enabled | default false | quote }}
             - name: ENVIRONMENT
               value: {{ .Values.application.eventStreams.environment | default "" | quote }}
-            {{- if .Values.secrets.kafka }}
+            {{- if .Values.application.eventStreams.enabled }}
             - name: EVENT_STREAMS_BOOTSTRAP_SERVERS
               valueFrom:
                 secretKeyRef:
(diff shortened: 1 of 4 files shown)


Tansito:
@korgan00 , @ismaelRozasRamallal how you are the two more involved around this effort let me ping you directly in case you can take a review here 👍

avilches (approved):
Yes, this is the way. Only one flag to wrap the secret creation in the clsuter and the env variable definition in runtime.

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -524,7 +524,7 @@ spec:
               value: {{ .Values.application.eventStreams.enabled | default false | quote }}
             - name: ENVIRONMENT
               value: {{ .Values.application.eventStreams.environment | default "" | quote }}
-            {{- if .Values.secrets.kafka }}
+            {{- if .Values.application.eventStreams.enabled }}
```
ismaelRozasRamallal, line 527: Good change, I left in the other way at the beginning to decouple secret creation from event streaming logic to test them separated but now everything is working I think we can use the same feature flag 👍

ismaelRozasRamallal (approved):
LGTM! thanks @Tansito !! 🚀

PR #2382, opened 2026-08-06
Block external references in function arguments schemas

avilches:
### Summary
A function's `arguments_schema` is supplied by whoever uploads the function, and the gateway evaluates it with `jsonschema` on every run. Reference resolution was left at the library default, which tries to retrieve `$ref` targets when the schema is used. That put an unbounded piece of work in the request path: a schema with an `http://` or `file://` reference made the gateway attempt a retrieval it never needed to do, and a reference that never answered kept a worker busy until the request timed out.

This PR removes that work. Reference retrieval is disabled outright, and external references are rejected at upload time so they cannot be stored in the first place. It also stops an invalid JSON Schema from being accepted at upload and then breaking every run of that function.

### Details and comments
On impact, this is about CPU and worker time, not data exfiltration. Outbound internet access from the gateway is restricted by egress rules, so a reference pointing at an external host cannot actually reach it. What was left is that resolving a reference occupied a worker while it tried, so the practical risk is wasted CPU and workers held up until the request timeout, not reaching internal or external services and not leaking data. The right way to read this change is that it takes an unnecessary and unbounded step out of the request path.

Also worth being explicit: argument validation is not deployed to production yet. No function has a stored `arguments_schema` today and nothing is currently exposed. This is being fixed before the feature ships, not in response to an incident.

What changed:

- `validate_arguments` now builds the validator with an empty `referencing.Registry`. With no retrieval hook, an external reference resolves to nothing and raises `Unresolvable`, which is surfaced as a `400`. Same-document references such as `#/$defs/name` are unaffected: they resolve against the schema itself, not through the registry, and a test covers that they still accept valid values and still reject invalid ones.
- `ProgramSerializer.validate_arguments_schema` walks the schema at upload time and rejects any `$ref` that does not start with `#`, with a message telling the vendor to inline the definitions or use an internal reference. Rejecting at upload means the person who can fix it is the one who gets the error.
- The same serializer method now runs `check_schema`, so a schema that is valid JSON but not a valid JSON Schema (for example `{"type": "integar"}`) is refused at upload with a `400`. Previously it was accepted and then made every run of that function return `500`. That is a small robustness fix rather than a security one, but it lives in the same two functions, which is why it is here.

Not addressed here: a schema `pattern` is still an unbounded backtracking regex applied to caller input, so a crafted pattern can burn CPU in the request path. Fixing that is a separate design decision (linear-time regex engine, an upload-time complexity check, or a wall-clock budget) and is handled in its own PR.

No release note is included on purpose. The feature is not deployed, so there is nothing here to tell users about.


avilches:
Superseded by #2383, which now contains these commits plus the rest of the bounded-cost work: the combinatorial `anyOf` + internal `$ref` blowup, and the two cases that surfaced as a 500. Keeping it all in one PR avoids two changes editing the same two functions.

PR #2383, opened 2026-08-06
Arguments validation/security: bound the cost of validating job arguments against a function's JSON Schema

avilches:
### Summary
A function's `arguments_schema` is supplied by whoever uploads the function, and the gateway evaluates it against caller-supplied arguments on every run, in the request thread. Both sides are untrusted, and neither was bounded: a `pattern` of under 100 bytes could keep a worker busy for hours, and nothing bounded memory at all.

This bounds that cost from outside `jsonschema` rather than from within it. Validation runs in a forked child with a CPU limit and an address space limit, so whatever the schema asks for, the child either finishes or dies and the parent returns a `400`. Cheap length, depth and node-count checks run first, on the text, before anything is evaluated.

Argument validation is not deployed yet and no function has a stored schema, so this lands before the feature ships rather than in response to an incident. The realistic impact was CPU and worker time, not data exposure: egress rules already prevented a schema reference from reaching an external host.

### Details and comments
This supersedes #2382 and replaces an earlier version of this PR that tried to bound the cost from inside the library, by subclassing the validator to count subschema visits, adding a wall clock deadline, matching `pattern` with RE2, replacing `uniqueItems` and refusing keywords whose cost hid in private helpers. Review refuted that four separate ways, every one of them under the declared limits: keywords whose cost lives in private module level helpers no subclass can reach, a `$schema` at the root restoring the stock backtracking class when a reference resolves, the size of a compiled regex program rather than the length of its input, and memory. `jsonschema` also documents that subclassing its validator classes will become an error, so that approach needed replacing regardless.

About 200 lines of that machinery are deleted, along with the `google-re2` dependency. Four previously reported problems go away with it rather than being fixed: a misleading "is not a 'regex'" message, false positives from the keyword scan, a step budget that rejected a legitimate array of 10001 numbers, and a metaschema check repeated on every request.

What remains in the module is four limits checked on the text: `MAX_SCHEMA_LENGTH` (64 KB), `MAX_ARGUMENTS_LENGTH` (1 MB), `MAX_DOCUMENT_DEPTH` (64) and `MAX_SCHEMA_NODES` (200). The last is new and bounds memory on every platform, which matters because the address space limit is sized from `/proc/self/status` and therefore does not fire on macOS: that half of the protection degrades in development and applies in production.

One keyword is still replaced. `uniqueItems` compares elements pairwise whenever they are not sortable, which any array of objects triggers, so it pays its full quadratic cost inside a single CPU second regardless of isolation: 1500 objects in 18 KB were refused, where a hash of a canonical form of each element answers in 0.011s and agrees with the stock keyword on all 30 equality cases tested. `pattern` is back on Python's own engine, left for the isolation to bound.

The upload also refuses a schema it can tell is unusable, rather than letting it become a `400` on every later run: it runs a trial validation inside the same child, which catches a `$ref` back to the document root, and walks the document checking that every internal JSON Pointer resolves. A plain-name anchor fragment is skipped rather than resolved, since resolving those ourselves would mean reimplementing part of the library, and a recursive schema of the ordinary tree shape is still accepted. These checks are quality of service and make no claim to completeness: the isolation is what bounds anything they miss.

Several ways of failing a request outright are now `400`s, which still matters for a schema already in the database or put there through the admin: a schema that is a JSON scalar, a `$ref` back to the root of the document, a document or an instance nested deep enough to exhaust the stack, an integer with more than 4300 digits, and a stored schema that is not JSON at all. Parsing the arguments happens inside the child for that reason. The schema `false` also stops disabling validation: it is the strictest schema there is, and treating the parsed value as a boolean turned it into no validation at all.

A rejected value no longer crowds out the reason it was rejected. `jsonschema` builds its messages with `repr(instance)` first and the reason last, so truncating to a fixed length returned half a kilobyte of base64 and no explanation, which is exactly what a caller sending an encoded circuit got. The message is now built from the failing keyword and what it expected, with a bounded excerpt of the value after it, and `path` is unchanged.

The child memory limit is `ARGUMENTS_SCHEMA_MEMORY_LIMIT_MB`, 128 by default and clamped to between 64 and 256 before use. That figure is measured rather than guessed: at 512 MB, two concurrent requests against a 1200 MB base drove the cgroup to 2047 of 2048 MB and the kernel killed a process, which is the failure the limit exists to prevent. At 128 the same shapes peak at 1567 MB with no kills, while a 1 MB batch of encoded circuits and 4000 objects under `uniqueItems` both pass even at the clamp minimum.

The CPU budget is configurable the same way, as `ARGUMENTS_SCHEMA_CPU_LIMIT_SECONDS`, 1 by default and clamped to between 1 and 5. Every legitimate schema measured validates in milliseconds, so no deployment should need to raise it, but that default is a measurement and not a proof, and a vendor whose schema turns out to need longer is better unblocked by a value change than by a code deploy. The clamp is tighter than the memory one because this is the whole bound on the work a single request can ask for, and no endpoint is rate limited. The wall clock deadline is not configured separately: `run_isolated` derives it by multiplying the budget by 5, so the default budget keeps the 5.0 seconds that were measured and the clamp maximum gets 25. The two limits are independent bounds rather than a limit and its fallback, since `RLIMIT_CPU` bounds how much CPU a request consumes while the deadline bounds how long it can hold a worker while consuming it, and which one fires depends on how much CPU the child actually gets. That is why the deadline is multiplied and not offset by a constant: a constant tolerates less and less slowdown as the budget grows, so at a budget of 5 a schema legitimately needing 3 CPU seconds would still be refused, which is the case raising the budget exists to allow. Past a 5x slowdown the deadline fires first and reports itself, which is the outcome to prefer: measured on a 16 core laptop against 40 CPU hogs, a child got 0.083 of a core and needed 60 wall seconds to spend 5 CPU seconds, and holding a worker for a minute is worse than refusing the request. In the deployed pod a child gets far more than a fifth of a core, so there the CPU budget is what fires.

Measured through the endpoints, before and after:

| input | before | after |
|---|---|---|
| `^(a+)+$` against 40 characters | hours | 1.00s, `400` |
| the same with `$schema` at the root | 7.85s and rising per character | 1.00s, `400` |
| a 6 KB regex program against 8000 characters | 24.6s | 1.00s, `400` |
| `anyOf` of 2000 branches against 500 KB | 1.55 GB of RSS | refused on node count |
| `uniqueItems` over 8000 objects | 32.2s | 0.011s for 4000, refused beyond |
| `{"$ref": "#"}`, 13 bytes | `500` | `400` |
| a schema of `123`, `null` or `1.5` | `500` | `400` |
| a schema of `false` | accepted everything | rejects everything |

Both limits are exposed through the Helm chart, so they can be tuned without a code deploy. One thing it deliberately does not do is limit the rate: the bound is per request, and the gateway has no throttling on any endpoint, so a caller can still spend a CPU second per request without creating a job. That is recorded in the spec as a known limit. The exposure went from hours per request to about a second.

`specs/ARGUMENTS_VALIDATION.md` is updated in the same PR, since it described the removed rules. It also documents two things a function author needs and that were written down nowhere: that arguments are validated in the form `QiskitObjectsEncoder` produces, so a circuit arrives as `{"__type__": "QuantumCircuit", "__value__": "<base64 QPY>"}` and the natural looking schema rejects every legitimate call, and that `/validate_arguments` deliberately does not check `disabled` while `/run` does.

No release note is included on purpose, since the feature is not deployed and there is nothing to tell users about.


ismaelRozasRamallal (dismissed):
LGTM!

charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -47,6 +47,8 @@ application:
     maxGpuResources: 1
     keepClusterOnComplete: False
     programTimeoutHours: 24
+    argumentsSchemaMemoryLimitMb: 128
+    argumentsSchemaCpuLimitSeconds: 1
```
ElePT, line 59: isn't 1 second maybe too short? I find it hard to come up with a limit but I fear this might be too tight.

korgan00 (approved):
I think is good

ElePT (commented):
I have gone through the PR and not going to lie, it goes so deep into the internals of jsonschema and all of the different safeguards that I wouldn't be able to catch a bug in it. I think that the tests are pretty extensive and that's a good sign. My only comment is if 1 CPU second is a good limit, but I am not against merging it and trying in staging

PR #2384, opened 2026-08-06
Arguments validations: release note for the feature

avilches:
### Summary
This is the release note for the arguments validation feature, split out into its own PR so @ElePT  and the team can review the wording on its own. Keeping it here means the review of the text does not block the code review, and it does not get buried under a diff of code and tests either.

### Details and comments
The feature lets a Qiskit Function declare a JSON Schema for the arguments it accepts, so the gateway can validate submitted arguments against it and reject a mismatch before creating a job. It spans three PRs: #2330, #2341 and #2348. This is the single note that covers all of it, which is why it mentions things that come from more than one PR rather than from just one.

What to look at: the wording, and in particular whether the `upgrade` entries describe the breaking changes clearly enough, since those are the ones users will actually have to act on.

Two behaviours in the note are worth a second opinion:

- A schema is removed by uploading an empty one (`{}`), because omitting the field on re-upload preserves the existing schema. The note says this, but please check that it reads unambiguously, since the two cases (omit vs empty dict) do different things.
- `QiskitFunction.schema` has been renamed to `arguments_schema`, and its type changes from a string to a dict. There is no deprecation path for the old field, I think it doesn't matter because the code use not used by anybody (it was useless)...

### DIFF
--- a/releasenotes/notes/0.34.0/arguments-validation-5b7701bd6f304c2f.yaml
+++ b/releasenotes/notes/0.34.0/arguments-validation-5b7701bd6f304c2f.yaml
@@ -1,0 +1,55 @@
+---
+features:
+  - |
+    Qiskit Functions can now declare the arguments they accept as a JSON Schema, and the gateway validates submitted
+    arguments against it before any compute is used. Previously a function given bad input failed mid-execution, after
+    the user had already waited through queue time and container startup.
+
+    Attach the schema when uploading a function through the ``arguments_schema`` field of ``QiskitFunction``::
+
+        function = QiskitFunction(
+            title="my-function",
+            entrypoint="main.py",
+            working_dir="./src",
+            arguments_schema={
+                "type": "object",
+                "required": ["shots"],
+                "properties": {"shots": {"type": "integer"}},
+            },
+        )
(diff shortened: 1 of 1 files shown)


avilches:
Draft until all PRs are merged and @ElePT validate it.

releasenotes/notes/0.34.0/arguments-validation-5b7701bd6f304c2f.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+        {"message": "'wrong-type' is not of type 'integer'", "path": ["shots"]}
+
+    The feature is opt-in and backwards compatible: functions without a schema are never validated.
+
+    Re-uploading a function without the field keeps the schema it already had, so a schema is removed by uploading an
+    empty one::
+
+        function.arguments_schema = {}
+        client.upload(function)
```
ElePT, line 33: Regarding this re-upload behavior, I am ok with keeping it as-is and seeing whether users find it comfortable or there is any concern.

ElePT (approved):
Thanks a lot @avilches! Very good documentation of the feature, I don't have any major comments :)

PR #2385, opened 2026-08-07
Stop seeding Code Engine projects on startup

marceloamaral:
### Summary
Stop running sync_ce_project on gateway/scheduler startup; it is run on demand instead.

### Details and comments
Removes the boot call from both entrypoints and the now-unused CE_PROJECTS pod env. Adds a test.


avilches (approved):
Great! Don't forget to delete the ce.projects in the production/staging ci files too.

PR #2386, opened 2026-08-10
T-shirt sizes - Adding models to the gateway

ItamarGoldman:
### Summary
Adding Models and data for for billing usage. It will help the user and provider to configure what computation resources are needed. 

### Details and comments

- [x] Add models `function_sizes` and `compute_profiles`.
- [x] Add `flavor` field to `Job` model.
- [ ] check if 'default_compute_profile' is being used and if not, remove it.
- [x] Add migration.


gateway/api/migrations/0057_computeprofiles_job_compute_profile_fk_functionsize.py
```diff
@@ -0,0 +1,89 @@
...
+    dependencies = [
+        ("api", "0056_program_arguments_schema"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="ComputeProfiles",
+            fields=[
+                (
+                    "compute_profile",
```
ismaelRozasRamallal: I would say to call it id here too? thinking again on that, it makes sense for me and it's also aligned with the other models

ismaelRozasRamallal: @Tansito , @ItamarGoldman and me were discussing about this point and we have good arguments for both calling it `id` or `compute_profile`.

- In favor of `id`: It's a naming convention for all tables primary keys and a well known practice
- In favor of `compute_profile`: all the `id` fields we have are just autogenerated (DB ids mainly), however in this case it is an string defined by us

What do you think we should do in this case?

gateway/api/migrations/0057_computeprofile_and_more.py
```diff
@@ -0,0 +1,89 @@
...
+                (
+                    "function_size",
+                    models.CharField(choices=[("S", "Small"), ("M", "Medium"), ("L", "Large")], max_length=1),
+                ),
+                (
+                    "compute_profile",
+                    models.ForeignKey(
+                        blank=True,
+                        null=True,
+                        on_delete=django.db.models.deletion.PROTECT,
```
ismaelRozasRamallal, line 78: just as curiosity, what PROTECT means here?

ItamarGoldman: `on_delete=django.db.models.deletion.PROTECT` tells Django to prevent the deletion of a referenced (parent) object if any related (child) objects still point to it. Meaning if Model B (`FunctionSize`) has a ForeignKey to Model A (`ComputeProfiles`) with `PROTECT`, deleting an instance of Model A  (`ComputeProfiles`) fails if Model B references it (so it won't left with "dangling" FK, creating a broken link in the database).

gateway/api/migrations/0057_computeprofiles_job_compute_profile_fk_functionsize.py
avilches: If compute_profile acts as a primary key (unique and used in other tables to handle the relationship), I would use `id` or `compute_profile_id` actually. It will help us to figure out the behavior in the field based on the name.

avilches: Anyway, have you considerer use a regular id with an autogenerated uuid, and set up the compute_profile as secondary field called "key" or "name" not null and unique?

ItamarGoldman: By setting `compute_profile` as PK, it will be `compute_profile_id` in the DB. Though, I can add the suffix of _id for clarity.

ItamarGoldman: > Anyway, have you considerer use a regular id with an autogenerated uuid, and set up the compute_profile as secondary field called "key" or "name" not null and unique?

It is a good question. We discussed in Slack.

But in the weekly I said it is different from the convention used now and we converged to the current design (compute_profile as PK) .

gateway/api/migrations/0057_computeprofile_and_more.py
ismaelRozasRamallal: oh, gotcha, seems correct to me, thank you!

gateway/api/migrations/0057_computeprofiles_job_compute_profile_fk_functionsize.py
ismaelRozasRamallal: I would say to keep the custom string as id, but prefix the compute_profile with `_id` too. This adjust to what we discussed in the daily 👍

ismaelRozasRamallal (commented):
LGTM! just change the compute profiles table primary key from `compute_profile` to `compute_profile_id` and you have my approval 😄

ismaelRozasRamallal (dismissed):
LGTM!!

Tansito:
From internal conversations @ItamarGoldman will apply a couple more changes here, @ItamarGoldman let us know when you consider ready the PR to review it again 😄

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
         return f"{self.project_name} ({self.region})"
 
 
+class ComputeProfiles(models.Model):
```
Tansito: By convention we are using singular no plurals in the database table names for Django ORM.

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
...
+    )
+    cpu = models.CharField(max_length=64, null=True, blank=True, default=None)
+    gpu = models.CharField(max_length=64, null=True, blank=True, default=None)
+    memory = models.CharField(max_length=64, null=True, blank=True, default=None)
+
+    class Meta:
+        app_label = "api"
+
+    def __str__(self):
+        return self.name or self.compute_profile_id
```
Tansito: initially I would only return `self.compute_profile_id` I think it's safe and also is the `id` so we can ensure it will be always populated. `self.name` will be for very specific cases and probably we will need to select it manually. For logs and so on `self.compute_profile_id` will be always desired.

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
...
+    name = models.CharField(
+        max_length=255,
+        null=True,
+        blank=True,
+        default=None,
+        help_text="Human-readable display name / tag for the profile",
+    )
+    cpu = models.CharField(max_length=64, null=True, blank=True, default=None)
+    gpu = models.CharField(max_length=64, null=True, blank=True, default=None)
+    memory = models.CharField(max_length=64, null=True, blank=True, default=None)
```
Tansito: Question for @marceloamaral here, does it make sense that cpu and memory can be null? GPU yes because a node cannot have gpus but it will always have cpu and memory right?

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
...
+    created = models.DateTimeField(auto_now_add=True, editable=False)
+    updated = models.DateTimeField(auto_now=True, null=True)
+
+    function = models.ForeignKey(
+        to=Program,
+        on_delete=models.CASCADE,
+        related_name="function_sizes",
+    )
+    function_size = models.CharField(max_length=64)
+    compute_profile_fk = models.ForeignKey(
```
Tansito: Here we don't have the constrain from the Job so I think `compute_profile` instead of `compute_profile_fk` is better.

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
...
+        to=ComputeProfiles,
+        on_delete=models.PROTECT,
+        null=True,
+        blank=True,
+        related_name="function_sizes",
+    )
```
Tansito, line 439: I would remove the nullable from the model, it doesn't have sense to have PROTECT and Nullable at the same time here. A compute profile should always exist before creating a size and we are specifying with this that a compute profile cannot be deleted if it is associated to a size so the correct configuration here I think it would be:

```
        to=ComputeProfiles,
        on_delete=models.PROTECT,
        related_name="function_sizes",
```

Without the null and blank.

Tansito (changes requested):
Regardless my comments here @ItamarGoldman , really good work here. This is complex so let's review this carefully to ensure we finish with a good model, thanks for your effort 🫡

gateway/core/models.py
marceloamaral: That is right, `name, cpu and memory` should not be `null`

Tansito: well, name is fine, it's just the marketing one

gateway/core/models.py
```diff
@@ -361,6 +361,80 @@ def __str__(self):
...
+    Maps a Code Engine compute profile to a specific CPU/GPU/memory allocation
+    (driving the classical cost we pay CE and the GPU submission payload for
+    Performance tiers). In billing, the profile identifies the classical time
+    rate (``classical_time_<compute_profile>``).
+    """
+
+    compute_profile_id = models.CharField(
+        max_length=255,
+        primary_key=True,
+        help_text="Code Engine compute profile identifier (e.g., gx3d-24x120x1a100p)",
```
marceloamaral: The `id` is the resource shape, while the class/size label is in `name`. So the `help_text` example is wrong.

`compute_profiles` will use IDs like `8x32` or `24x120x1l40` (`{cpu}x{memory}[x{gpu}]`). `gx3d-24x120x1a100p` is a Code Engine flavor, which is a different format.

Please update the help_text to use the actual `compute_profiles.id`...

gateway/core/models.py
marceloamaral: Also, Code Engine requires the GPU count and lowercase model, as defined in `_parse_compute_profile`. So `24x120xL40` should be `24x120x1l40`.

Once `cpu`, `memory`, and `gpu` are stored in this table, we won't need `_parse_compute_profile` anymore and can use those columns directly. We just need to make sure the `gpu` values (e.g., `1l40`, `8a100`, etc...) and compute profile IDs still follow the Code Engine format.

gateway/core/models.py
```diff
@@ -464,6 +538,17 @@ class Job(models.Model):
         blank=True,
     )
     program = models.ForeignKey(to=Program, on_delete=models.SET_NULL, null=True)
+    compute_profile_fk = models.ForeignKey(
+        to=ComputeProfiles,
+        on_delete=models.SET_NULL,
```
marceloamaral: To preserve the historical compute resource used by a Job, `ComputeProfiles` rows should be immutable and non-deletable while referenced.

Also, since a compute profile is required for every Job, the FK should be mandatory:

```python
compute_profile_fk = models.ForeignKey(
    to=ComputeProfiles,
    on_delete=models.PROTECT,
    related_name="jobs",
)
```

Using `PROTECT` prevents deleting a profile that is referenced by a Job, and making `ComputeProfiles` immutable ensures the resource values do not change after the Job is created.

gateway/core/models.py
ItamarGoldman: Changed as suggested.

gateway/core/models.py
ItamarGoldman: Changed as suggested.

gateway/core/models.py
ItamarGoldman: Changed as suggested.

gateway/core/models.py
ItamarGoldman: Changed as suggested.

gateway/core/models.py
ItamarGoldman: I agree set `on_delete=models.PROTECT,`, it sounds logical and correct. Regarding the other I am not sure.
Since there is no `compute_profile_fk` for older jobs stored in the DB, I think we need to keep `null=True` and since Ray jobs don't have a compute profile associate with them, I think `blank=True` is also needed.
What do you think? @Tansito

gateway/core/models.py
ItamarGoldman: I used `_fk` suffix only to emphasize it is the same type as the one in `Job` model. I will change it as suggested.

gateway/core/models.py
ItamarGoldman: changed as suggested

gateway/core/models.py
ElePT: @ItamarGoldman the latest changes may not have been pushed, I still see the plural

gateway/core/models.py
ElePT: I think @ItamarGoldman has a point. At the moment only new fleets jobs will have a `compute_profile_fk`, the field won't apply for older jobs or ray jobs.

Tansito: Sorry team, I just reached to this now. I fully agree, you are right Itamar.

gateway/core/models.py
```diff
@@ -212,6 +215,16 @@ def __str__(self):
...
         return f"{self.title}"
 
+    def clean(self):
+        """Validate that ``default_size`` belongs to this function.
+
+        The rule spans two tables (``program.id`` vs ``function_size.function_id``),
+        so it cannot be expressed as a row-local database CHECK constraint.
+        """
+        super().clean()
+        if self.default_size_id and self.default_size.function_id != self.id:
```
Tansito, line 225: Good double check, I consider we should have this kind of logic in other part but by now it's fine. We can always move it because this doesn't affect to the database.

gateway/core/models.py
marceloamaral: Good point, I agree too!

Tansito (commented):
Only one comment for the future but I see everything correct now, @ItamarGoldman good work 😄

Tansito:
@ElePT , @ItamarGoldman , @marceloamaral how do we want to manage this? Do you want to wait until the release to merge it?

marceloamaral:
@Tansito I believe this change is safe, so feel free to merge when you're ready so that Itamar can focus in the other PRs.

Tansito (approved):

PR #2387, opened 2026-08-10
PSQL secret refactor

Tansito:
### Summary

This PR creates an external secret for PSQL services managed from secrets managers in the cloud.

Additional key `createExternalSecret`: this key is temporal and I'm going to remove it as soon as we can confirm everything is working as expected in all environments. 

- [x] This PR requires also a migration in the internal repository to use the correct keys

### Checklist

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/values.yaml
+++ b/charts/qiskit-serverless/charts/gateway/values.yaml
@@ -99,6 +99,9 @@ secrets:
       host: database-host
       port: database-port
       userName: user-name
+    createExternalSecret: false
+    secretRef: ""
+    secretStoreName: qiskit-serverless
   superuser:
     name: gateway-superuser
     key:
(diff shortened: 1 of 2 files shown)


charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -99,6 +99,9 @@ secrets:
       host: database-host
       port: database-port
       userName: user-name
+    createExternalSecret: false
```
avilches, line 102: Kafka has 

```
  kafkaCredentials:
    create: false
```

can we use the same style?

Tansito: Ups, I forgot to mention it in the PR. This is just temporal to add it first in staging and if it works promote it to production. When we have everything ready I will remove it so I preferred to have it near from the rest of the secret configuration.

avilches (approved):

PR #2389, opened 2026-08-11
Arguments validation/fix: report an unusable stored schema instead of decoding it into the wrong type

avilches:
### Summary
Two small fixes from the argument validation work. A function's `arguments_schema` is stored as text in the database and nothing validates it, so reading a function back through the client SDK could hand the caller a value of the wrong type, or let a raw JSON error escape. Worse, `false`, the strictest possible schema, was reported as no schema at all. Now booleans round-trip correctly and anything else invalid raises `QiskitServerlessException`. This widens the declared type of `arguments_schema` to `Optional[Union[Dict[str, Any], bool]]`. The second change is a comment explaining a routing detail on the `get_by_title` endpoint.

### Details and comments
On the client, decoding used to treat a falsy schema as "no schema", which turned `false` into `None`. Other stored values that are not valid schemas (a bare number, string, or list) used to come back typed as that value instead of failing; they now raise `QiskitServerlessException`, matching the rule the gateway already applies at upload: a schema must be an object or a boolean. Sending `{}` on upload still removes a stored schema, and a column holding `{}` still reads back as `None`.

The other change adds a comment to `get_by_title`, recording why it reads the provider from a query parameter rather than the path: the route uses Django's `<str:title>` converter, which can't contain a slash, and a percent-encoded slash is decoded before routing, so both forms 404 before the view runs. The shared parsing helper still supports both forms because two other endpoints take the title from the request body, where a slash is meaningful, so the branch is not dead code even though it looks unreachable here.

### DIFF
--- a/client/qiskit_serverless/core/function.py
+++ b/client/qiskit_serverless/core/function.py
@@ -38,24 +38,53 @@ from qiskit_serverless.core.job import (
     Job,
     Configuration,
 )
+from qiskit_serverless.exception import QiskitServerlessException
 
 GenericType = Literal["GENERIC"]
 ApplicationType = Literal["APPLICATION"]
 CircuitType = Literal["CIRCUIT"]
 
 
-def _decode_fields(data: Dict[str, Any], field_names: set) -> Dict[str, Any]:
-    """Keep only known dataclass fields and decode the ones the gateway sends as text.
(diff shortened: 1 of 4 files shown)


Tansito (approved):
LGTM @avilches thank you! 😄

PR #2390, opened 2026-08-11
T-shirt sizes model serializer

ItamarGoldman:
<!--
-->

### Summary
Adding serializers to models in #2386 for for billing usage. It will help the user and provider to configure what computation resources are needed.


### Details and comments

- [x] Add model managers\admin for new models and fields.
- [x] Add serializer for the new Models.
- [ ] Add into `compute_profile_fk` field on job creation (Can be done in #2431).
- [ ] Add a methods to get all supported ComputerProfiles. This will use by users\providers to know what are the available legal computation profile.  _**Note**:_ maybe we should add a flag if it is active or not if we will retire some of the computation profiles, so they will be kept in the DB for historical reasons but we won't expose them to the user.
- [x] Add releasenotes.

### Out of Scope:
 - Adding a parse tp 

This PR builds the data layer only — managers, validation, and admin. It changes no runtime behavior. Reading function_size at job creation and resolving it to a profile is a follow-up (see Follow-up PR below), split out to keep this diff reviewable and because the risk profiles differ sharply: nothing here can alter how an existing job runs.

1. New manager — core/model_managers/compute_profiles.py                                                                                                                                                                                                                     
House style (match core/model_managers/providers.py exactly): Apache/Qiskit license header (as in the newest manager, code_engine_projects.py), one-line module docstring, module logger logging.getLogger("core.compute_profiles"), TYPE_CHECKING import of the model, subclass QuerySet, class docstring """Compute profile query set to transform into a manager.""".                                                            

```python
class ComputeProfileQuerySet(QuerySet):
def get_by_id(self, compute_profile_id: str | None) -> Optional["ComputeProfiles"]:
    """Return the profile row for an id, or None."""
```
This absorbs _resolve_compute_profile_fk from api/use_cases/programs/run.py:51-63, moving a raw .objects.filter(pk=...) out of the api layer. Keep the None-guard behavior exactly — the call site at run.py:121 keeps working unchanged, so this is a pure refactor.
Wire in core/models.py on ComputeProfiles, following the annotated-assignment style used at core/models.py:192,355,561:
objects: ComputeProfileQuerySet = ComputeProfileQuerySet.as_manager()

2. New manager — core/model_managers/function_sizes.py
```python
class FunctionSizeQuerySet(QuerySet):
     def function_sizes(self, function) -> Self:
         """Rows belonging to one function."""

     def get_function_size(self, function, function_size: str) -> Optional["FunctionSize"]:
         """Look up one (function, size) row. Size is lowercased before lookup."""

     def resolve_compute_profile(self, function, function_size) -> Optional["ComputeProfiles"]:
         """Resolve a size to its profile. None when the size is unset, no row
         matches, or the matched row has no profile."""
```
`resolve_compute_profile` is the entry point the follow-up PR will call — it is written and tested here, unused for now. Lowercase the size key inside these methods so callers need not. Log at warning on a miss, consistent with ProviderQuerySet.get_by_name (providers.py:17-22).
**Deliberately not here:** any provider check. Whether a size applies to a custom function is a policy question, and the run-time warning for users belongs to the follow-up. Keep this method a pure lookup.

Wire as objects: FunctionSizeQuerySet = FunctionSizeQuerySet.as_manager().


3. Validation, all in one place — core/models.py, Program

Add a single clean() holding every FunctionSize-specific rule:
```python
def clean(self):
        """Validate that ``default_size`` belongs to this function.

        The rule spans two tables (``program.id`` vs ``function_size.function_id``),
        so it cannot be expressed as a row-local database CHECK constraint.
        """
        super().clean()
        if self.default_size_id and self.default_size.function_id != self.id:
            raise ValidationError({"default_size": "default_size must be a FunctionSize belonging to this function."})
```


Open decision — the provider-only rule. Sizes are only meaningful for provider functions, but
whether creating a size row for a custom function should be blocked is undecided (it overlaps
with how custom functions are handled at upload, which is out of scope). Two options:
- (a) Leave it out for now — clean() only normalizes casing. Admin can create rows for any function; they are simply never consulted. Nothing to unwind later.
- (b) Add the rule now — clean() also raises when self.function.provider is None, so bad rows cannot be entered via backoffice:
```python
if self.function_id and self.function.provider is None:
    raise ValidationError({"function": "Function sizes apply only to provider functions."})
```
Currently going with option (a) as it is easier to change it afterwards

**Decided all function should have sizes by default. When a function is uploaded without sizes it will automatically set the sizes dict, e.g. `{"m": "64x128"}`**


4. Admin — api/admin.py

Import ComputeProfiles and FunctionSize in the core.models block (admin.py:12-24).

```python
@admin.register(ComputeProfile)
class ComputeProfileAdmin(admin.ModelAdmin):
    """ComputeProfileAdmin."""

    # search_fields is required for FunctionSizeAdmin's autocomplete on compute_profile
    search_fields = ["compute_profile_id", "name"]
    list_display = ["compute_profile_id", "name", "cpu", "gpu", "memory", "updated"]
    ordering = ["compute_profile_id"]
    readonly_fields = ["created", "updated"]


@admin.register(FunctionSize)
class FunctionSizeAdmin(admin.ModelAdmin):
    """FunctionSizeAdmin."""

    list_display = ["function", "function_size", "compute_profile", "updated"]
    list_filter = ["function_size"]
    search_fields = ["function__title", "function_size", "compute_profile__compute_profile_id"]
    autocomplete_fields = ["function", "compute_profile"]
    list_select_related = ["function", "compute_profile"]
    readonly_fields = ["created", "updated"]

```
Wire the new FK into the existing Job form: add "compute_profile_fk" to JobAdmin's "Fleets" fieldset (admin.py:296-307, next to compute_profile) and to its autocomplete_fields (admin.py:275) so it renders as a searchable widget rather than a dropdown of every profile.


### Follow-up PR: size resolution

Captured here so nothing is lost; not built in this PR.

- Read `function_size` from the arguments JSON at job creation and resolve it via `FunctionSizeQuerySet.resolve_compute_profile`, setting both `Job.compute_profile` and
`Job.compute_profile_fk` from the same source so they cannot drift.
- Warn the user when a size is passed to a custom function (no provider) and is therefore
ignored — a silently dropped kwarg hides a real mistake by whoever ran the function.
- Undecided: precedence between a resolved size and an explicit compute_profile. Arguable
either way — the size is the higher-level input, but an explicit low-level value overriding a
high-level one is the more common convention. Settle it alongside the code.
- Client support: client/ has no function_size kwarg; until added it must be passed manually
inside arguments.

Separately undecided and out of scope for both PRs:

- Custom functions at upload time — whether to block a function_size / leave it empty.
- The "is this profile supported" check — validating an explicit compute_profile against
the function's catalog. Needs a policy for functions with an empty catalog: every function has
zero rows today, so a strict check would reject all Fleets jobs.
- Whether FunctionSize keeps the function FK. If sizes become global, drop the FK, change
unique_function_size, and simplify resolve_compute_profile. The one-method clean()
encapsulation is what keeps that cheap.
- ComputeProfiles.cpu/gpu/memory are decorative. fleets_runner._parse_compute_profile
(core/services/runners/fleets_runner.py:647-682) parses CPU/mem/GPU out of the profile
string and never reads these columns. Worth reconciling so they cannot disagree with the id
they describe.

### Checklist


ElePT:
Picking this up (taking over from @ItamarGoldman). Pushed three commits:

1. **Merge `main`**: updated branch to latest status.
2. **Fix the `compute_profile_fk` seam.** `_runner_config` always resolves a Fleets job to a profile string (default `DEFAULT_COMPUTE_PROFILE`, e.g. `bx3d-24x120`), but `get_by_id` returns `None` when no matching `ComputeProfile` row exists , so the FK was silently **null on the default path**. Since `ComputeProfile` rows are a required invariant, an unregistered profile is a misconfiguration: we now **reject the job at creation** (mirroring the existing 'no Code Engine project' guard) rather than persist a null FK. Ray jobs are unaffected (no profile → FK stays null). Regression tests added.
3. **Dedupe `ComputeProfileSerializer`**: moved the copy in `list.py`/`retrieve.py` into `serializer_utils.py` to avoid duplication.

⚠️ **Rollout ordering (reviewers please note):** once this merges, every default-path Fleets job will be **rejected** until the matching `ComputeProfile` rows (`bx3d-24x120` + any explicitly-requested profiles) exist in each environment's DB. The rows must be added (backoffice now / seed migration later) **before or with** the deploy.

Scope unchanged from the author's split: this stays the data-layer PR. Reading `function_size` at run time and `resolve_compute_profile` remain the follow-up (#2431).

.github/workflows/fleets-integration-tests.yaml
```diff
@@ -39,6 +39,15 @@ jobs:
...
         run: docker compose -f docker-compose-fleets-test.yaml exec -T gateway python manage.py sync_ce_project
+      - name: Register default compute profile
+        # A Fleets job resolves to DEFAULT_COMPUTE_PROFILE and job creation is
+        # rejected unless a matching ComputeProfile row exists (staging/prod seed
+        # these via backoffice). Seed the same default the integration env uses.
+        run: >-
+          docker compose -f docker-compose-fleets-test.yaml exec -T gateway
+          python manage.py shell -c
+          "from core.models import ComputeProfile;
+          ComputeProfile.objects.get_or_create(compute_profile_id='bx3d-24x120', defaults={'cpu': '24', 'memory': '120'})"
```
ElePT: Note that this logic only exists for the fleets integration tests. In staging and prod we still need to register the compute profile manually through backoffice.

gateway/api/use_cases/programs/run.py
```diff
@@ -103,6 +104,16 @@ def execute(  # pylint: disable=too-many-locals
             trial = business_model == BusinessModel.TRIAL
 
         compute_profile, gpu = _runner_config(function, data.compute_profile)
+        # ``compute_profile`` (string) is transitional and will be removed; the FK
+        # ``compute_profile_fk`` is the source of truth going forward. We derive the FK
+        # from the same string, so they agree at creation.
+        # A Fleets job always resolves to a profile string. If no ``ComputeProfile`` row
+        # is registered for it, that is a deployment misconfiguration: we reject the job
+        # rather than store a null FK. Ray leaves ``compute_profile`` None (profiles are
+        # a Fleets concept), so the FK stays null there.
```
ElePT: Changed this part to fail explicitly instead of silently assigning "None"

ElePT (dismissed):
I think this PR is in a good state to merge. There are a few listed follow-ups that we can tackle in the future, like:

- adding the method for listing available compute profiles
- size resolution logic
- see PR description for more


But in terms of the serializer, I think this PR covers it, including the failure logic if no compute profile is found in the DB, as well as a new addition that normalizes the saved compute profile to never include the ibm cloud prefix (even though users can submit it).

@Tansito I'd appreciate your review too.

Tansito (commented):
Now that we are also adding the compute profile to the Job at some point we are going to need to bring the discussion of `select_related` from @marceloamaral .

We will need to start optimizing our queries overall those ones related with `Jobs`.

gateway/api/v1/views/programs/run.py
```diff
@@ -34,7 +34,9 @@ class InputSerializer(serializers.Serializer):  # pylint: disable=abstract-metho
     provider = serializers.CharField(required=False, allow_null=True)
     compute_profile = serializers.CharField(required=False, allow_null=True)
 
-    _COMPUTE_PROFILE_RE = re.compile(r"^[a-z]+\d+[a-z]?-\d+x\d+(?:x\d+[a-z0-9]+)?$")
+    # Instance-family prefix (e.g. "bx3d-") is optional: accepted for backward
+    # compatibility but normalized away at ingest. Canonical form is bare.
+    _COMPUTE_PROFILE_RE = re.compile(r"^([a-z]+\d+[a-z]?-)?\d+x\d+(?:x\d+[a-z0-9]+)?$")
```
Tansito, line 39: Hmm... I don't really like to have two regular expressions checking very similar things, I'm going to try to open a PR fixing this. For now I think we can leave it as it is.

.github/workflows/fleets-integration-tests.yaml
Tansito: we should probably create a migration script for testing if we think this will start to be needed so every time we want to test something we can populate the database (similar to what we are doing with the `fixtures` in Django)

Tansito (approved):

gateway/api/v1/views/jobs/retrieve.py
```diff
@@ -81,6 +83,7 @@ class Meta:
             "sub_status",
             "fleet_id",
             "compute_profile",
```
Tansito, line 85: Reviewing with @avilches we think we should remove the old `"compute_profile"`

gateway/api/use_cases/programs/run.py
```diff
@@ -39,13 +41,38 @@ def _is_trial(function: Function, user) -> bool:
...
+    Raises:
+        FunctionConfigurationException: If a resolved profile has no registered row.
+    """
     if function.runner == Function.FLEETS:
-        profile = compute_profile_requested or getattr(settings, "DEFAULT_COMPUTE_PROFILE", "cx3d-4x16")
-        return profile, False
-    if function.provider and function.gpu:
-        return None, True
-    return None, False
+        requested = compute_profile_requested or getattr(settings, "DEFAULT_COMPUTE_PROFILE", "24x120")
```
Tansito: Remove `getattr` we should access directly to the settings from Django

gateway/api/use_cases/programs/run.py
```diff
@@ -39,13 +41,38 @@ def _is_trial(function: Function, user) -> bool:
...
+        # Normalize away any instance-family prefix so the bare id is what we
+        # store and look the FK up by. Clients may still submit a prefixed value.
+        compute_profile = normalize_compute_profile(requested)
+    elif function.provider and function.gpu:
+        return None, True, None
```
Tansito: We should start using a dataclass here to keep track easily for these returns

gateway/api/use_cases/programs/run.py
```diff
@@ -39,13 +41,38 @@ def _is_trial(function: Function, user) -> bool:
...
     if function.runner == Function.FLEETS:
-        profile = compute_profile_requested or getattr(settings, "DEFAULT_COMPUTE_PROFILE", "cx3d-4x16")
-        return profile, False
-    if function.provider and function.gpu:
-        return None, True
-    return None, False
+        requested = compute_profile_requested or getattr(settings, "DEFAULT_COMPUTE_PROFILE", "24x120")
+        # Normalize away any instance-family prefix so the bare id is what we
+        # store and look the FK up by. Clients may still submit a prefixed value.
+        compute_profile = normalize_compute_profile(requested)
```
Tansito, line 63: This is related with our comment about having two regex. This should be normalized in the input, I will open a PR for this.

Tansito (changes requested):

gateway/api/use_cases/programs/run.py
Tansito: fixed in [bd113d4](https://github.com/Qiskit/qiskit-serverless/pull/2390/commits/bd113d4dd556ab2e0c23e04dbcc2171db49e252a)

gateway/api/use_cases/programs/run.py
Tansito: fixed in [1784d7d](https://github.com/Qiskit/qiskit-serverless/pull/2390/commits/1784d7d22e3688487d3ce4277a275d3cf9342815)

gateway/api/v1/views/jobs/retrieve.py
Tansito: Let's keep track for this change as it's something we will need in the future.

Tansito (approved):

gateway/api/v1/views/programs/run.py
Tansito: PR created for this: https://github.com/Qiskit/qiskit-serverless/pull/2447

gateway/api/use_cases/programs/run.py
Tansito: PR created for this: https://github.com/Qiskit/qiskit-serverless/pull/2447

PR #2394, opened 2026-08-12
Propagate business_model in the license fee event

ismaelRozasRamallal:
### Summary
The license fee event carried no indication of which business model the function is billed under, so the consumer could not tell a licensed function from a trial or consumption one. This reads `business_model` off the job and adds it to the license fee payload, mapped to the name billing expects: `SUBSIDIZED → licensed`, `TRIAL → trial`, `CONSUMPTION → consumption`.

```diff
  "data": {
      "metric_type": "license_ibm_test-program",
      "metric_value": 1,
      ...
+     "business_model": "licensed"
  }
```

License fee event only — the `classical_time` lifecycle events are byte-identical to before, so this is additive on the wire.

### Details and comments
All changes in `gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py`:

- `BUSINESS_MODEL_NAMES` — the mapping, keyed on the `BusinessModel` constants so a rename upstream surfaces here. `_business_model_name(job)` resolves through it, falling back to `job.business_model.lower()` for anything unmapped.
- `emit_license_fee` passes the resolved name into `_publish`, which gained an optional `business_model` kwarg; `data` is built as a local and the key inserted only when the argument is present. That is what keeps the other events unchanged.

No signature change to `emit_license_fee(job)`, so the abstract interface, no-op client and scheduler call sites are untouched.

Tests: `_make_job` takes a `business_model` (default `SUBSIDIZED`); the full-payload assertion now includes the key; a parametrized test covers all three mappings; and one test asserts the key is *absent* from a `job_started` payload, guarding the additive claim.

## Notes for reviewers

- **`TRIAL` and `CONSUMPTION` names are inferred** — only `SUBSIDIZED → licensed` was specified, the rest follow the lowercase convention. This new naming for subsidized was decided from the product side and we need to consolidate it across the platform

### Checklist

### DIFF
--- a/gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
+++ b/gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
@@ -21,6 +21,7 @@ import uuid
 from datetime import datetime, timezone
 
 from confluent_kafka import Producer
+from core.domain.business_models import BusinessModel
 from core.models import Job
 
 from .abstract_event_streams_client import EventStreamsClient
@@ -28,6 +29,14 @@ from .abstract_event_streams_client import EventStreamsClient
 logger = logging.getLogger("gateway.ibm_cloud.event_streams_client")
 
 LICENSE_FEE_METRIC_TYPE = "license"
+
+# Business model names as billing expects them. Subsidized functions are billed
+# as licensed; the rest keep their own name, lowercased.
+BUSINESS_MODEL_NAMES = {
+    BusinessModel.SUBSIDIZED: "licensed",
+    BusinessModel.TRIAL: "trial",
+    BusinessModel.CONSUMPTION: "consumption",
+}
 
 
 class KafkaEventStreamsClient(EventStreamsClient):
@@ -38,7 +47,8 @@ class KafkaEventStreamsClient(EventStreamsClient):
     single metric in its `data` payload: `metric_type` (what is being billed) and
     `metric_value` (how much, in milliseconds for time-based metrics), plus
     `job_started` / `job_completed` flags so consumers can detect lifecycle
-    boundaries without interpreting the metric type.
+    boundaries without interpreting the metric type. License fee events also
+    carry `business_model`.
 
     Configured from environment variables:
       EVENT_STREAMS_BOOTSTRAP_SERVERS — comma-separated broker list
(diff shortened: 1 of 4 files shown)


gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -94,8 +104,14 @@ def emit_license_fee(self, job: Job) -> None:
             metric_value=1,
             job_started=True,
             job_completed=True,
+            business_model=self._business_model_name(job),
         )
 
+    @staticmethod
+    def _business_model_name(job: Job) -> str:
+        """Map the job's business model to the name billing expects."""
+        return BUSINESS_MODEL_NAMES.get(job.business_model, job.business_model.lower())
```
Tansito: I'm a bit concern about this line because if there is a new Business model without a name is not going to fail but return a `None`.

I have the feeling that we should maybe raise an exception or as minimum log it. Something like:
```
def _business_model_name(job: Job) -> str:
    name = BUSINESS_MODEL_NAMES.get(job.business_model)
    if name is None:
        raise ValueError(f"No billing name mapped for business model: {job.business_model!r}")
    return name
```
Just an example of a solution I didn't stop a lot thinking about it.

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -29,6 +30,14 @@
 
 LICENSE_FEE_METRIC_TYPE = "license"
 
+# Business model names as billing expects them. Subsidized functions are billed
+# as licensed; the rest keep their own name, lowercased.
+BUSINESS_MODEL_NAMES = {
```
Tansito: I have the fear that this object just remains here hidden from the `domain` logic about the business_model. I have the feeling that we should move it to `business_models` in the `domain` area.

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
Tansito: Also, how do you see to move this method to the `business_model` file and call it: `_business_model_mapping` or something like that? I think is easier to understand.

ismaelRozasRamallal: totally agree, addressed here [87dce62](https://github.com/Qiskit/qiskit-serverless/pull/2394/commits/87dce62ab9fcfbb96f72152abaa5db6eaa30de1e)

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
ismaelRozasRamallal: totally agree, addressed here [87dce62](https://github.com/Qiskit/qiskit-serverless/pull/2394/commits/87dce62ab9fcfbb96f72152abaa5db6eaa30de1e)

Tansito (approved):

PR #2397, opened 2026-08-13
Fix/provider function author upload

dpatel-ops:
### Summary

Authors of provider functions cannot manage what they publish. Uploading a new version, listing that function's jobs, reading its provider logs and managing its provider files all need either membership of the provider's admin group (legacy authorization) or an explicit per-function entitlement (instance authorization). An automated publisher, such as a Functional ID that uploads provider functions, ends up owning functions it cannot then operate on.

This PR treats the author of a provider function as an admin of that one function for the six provider operations, on both authorization paths. Authorship becomes a third grant that ORs with the existing two, so nothing that worked before stops working.

Ownership stays narrow on purpose. It covers only the provider operations. It does not make the function show up in the catalog, does not grant the right to run it, and does not make the author an admin of the provider as a whole. Creating a brand new provider function still needs the upload permission, because a function that does not exist yet has no author.

### Details and comments

Ownership is matched on provider and title together, because titles are uniqlobally. Authoring `provider-a/my-function` grants nothing on`provider-b/my-function`.

The provider job list changes shape when no function filter is given: it now falls back to the caller's own functions instead of hiding the provider. Before this, an author who was neither a provider admin nor entitled got `ProviderNotFound` for a provider they publish to. Entitled titles and authored titles add up, rather than one replacing the other.

`is_provider_admin` keeps ignoring ownership. It answers "admin of the entire provider" and hands out provider-wide job scope, so owning one function must not widen the caller's scope
to every function of that provider.

The rule lives in one place. `ProviderAccessPolicy._check` owns it, and the upload use case is taken out, since it applied the same test one layer up andwould have drifted apart from `_check` over time.

Authorship is checked last because on the granular path it is the only brancwhile the permission check reads a value already cached on the request. Theordering cannot change the result, since both operands are free of side effects and the result is their OR.

Catalog filtering is left alone. `FunctionsQuerySet` still gates provider funs even for the author, so only a comment changed there.

This branch also picks up main, including #2416, which scopes provider job listings to the requested provider in the job model manager. That fix and the change here touch the same listing from different layers, so they are tested together.

**Known limitations**

Revoking access takes two steps now. No code path reassigns `Program.author` after creation, so taking a user out of `admin_groups` no longer removes their access to the functions
they authored. Offboarding also has to reassign the author in the admin backthat second step. The field cannot be cleared because it is not nullable, anddeleting the user is not a way round it, since `author` cascades and that would delete their functions.

The reconfigurable-instance acceptance suite shares one token across roles, so the cells that assert a provider operation is denied only hold for a function the caller does not own. Since `instances1-test` is uploaded by the same token that then asserts against it, those cells no longer pin down the permission list on their own. Fixing that needs a function owned
by somebody else, which the harness cannot express yet. Tracked separately.

**Tests**

New ownership tests sit at the policy layer and on the provider job list, an the author was denied are updated to match. Local run: 983 passed, coverage86.30%, pylint 10.00/10, import contracts kept.

`specs/INSTANCES_AUTH.md` picks up an "Ownership grants the provider operations" section under Design decisions. The overview paragraph that said the author has full access without consulting the client is corrected, because that is no longer true for provi

### Checklist


avilches:
Hi! After talking with @Tansito, we agreed that the owner of a function should be an admin of it, so the idea in this PR is the right one. My only concern is the fix is very specific, where the real change should be in another place.

Why not in upload.py only? Because every provider operation shares one decision point, `_check` in `api/access_policies/providers.py`, and none of them looks at who authored the function. The upload endpoint is one of seven. The others are provider jobs, the job list of a single function, retrieving a job that someone else ran, provider logs, and provider files for reading and writing...

So this PR fixes one endpoint and leaves the other six rejecting the owner. They would be able to replace the image and the entrypoint of the function, which is what runs, but still could not read the logs of those runs or list its files.

## Suggested changes

1. Add the owner check to `_check` in `api/access_policies/providers.py`, before it splits into groups and instance permissions, so it holds on both paths. All seven endpoints pick it up at once and `upload.py` goes back to what it was.

2. `tests/instances/` expects a 404 on upload, provider jobs, provider logs and provider files at the NONE and USER levels, and the user running those tests is the author of the function under test, so all of them turn into allowed. To keep testing what they mean to test, those checks need a function that somebody else owns.

3. `specs/INSTANCES_AUTH.md` needs four edits:

   * The paragraph under **When the external client is consulted** (line 29) says that if the user is the author of a function, serverless or provider, they have full access without consulting the client. That becomes true for the provider operations, but it is still too wide for the catalog and for run, where the permission list keeps deciding on its own. It should say that ownership grants the provider operations on both auth paths, and that seeing and running a provider function still go through the permission list.
   * The table under **Provider access checks** lists each `ProviderAccessPolicy` method with the permission it uses and what happens with legacy groups. It needs a line above it saying that the owner of the function passes all of them before either column comes into play, so the table reads as the rule for everybody else.
   * The matrix under **What each test verifies** has rows for upload, provider jobs, provider logs and provider files that say 404 or 403 at the NONE and USER levels. Those only hold for a function the caller does not own. The note right below the matrix, the one that already explains that every test shares the same token, is the place to say it.
   * A short entry under **Design decisions**, next to the one about provider file access, explaining why ownership is enough on its own. Without it the next person reading `_check` will take the owner check for an oversight.

4. `core/model_managers/functions.py` line 47 says that provider functions are gated by instance permissions even if the user is the author. That stays true for the code right under it, which is the listing and run path, but written like that it reads as a rule for the whole system and now contradicts `_check`. It should say that ownership is not enough to see or run a provider function, and point at `ProviderAccessPolicy` for the provider operations, where ownership is enough.

dpatel-ops:
Thanks @avilches for the review and detailed information, this makes sense. I will learn the current implementation with the above comments and start applying changes one by one. I will move this to draft for now till I learn and play with it.

ElePT (commented):
I think @avilches should review that his comments were addressed, on my side it looks like they are. Regarding the documented limitations, I think that they are fair, in particular:

1. given that we plan to use this mainly for our automated pipeline, I think there isn't a lot of risk of wanting to remove auth and not being able to because the user was an author (our flow basically blocks out that scenario, as users can't be authors, unless they are from the team, in which case I think it should be fine). 
2. I think that the testing limitation is fair, and wouldn't block the PR for it, I think we can try to address this in a follow-up

Final comment: I was thinking if we should add a release note or not. I think we can keep this one fix internal and not add a release note.

avilches (approved):

dpatel-ops:
@ElePT yes the change is mostly going to impact internal(that is base on my initial learning) and if users wont notice it then release notes could be skipped. 



Thank you @avilches and @ElePT for the reviews. 🙇

PR #2399, opened 2026-08-18
Arguments validation/backoffice: validate scheme on save job

avilches:
### Summary

An unusable `arguments_schema` stored through the Django admin makes every later run of that function fail, and makes `client.functions()` raise for everyone who can see it. The upload endpoint already turns those schemas down. The admin did not, and now it does.

### Details and comments

Opening a function whose stored schema is broken shows a warning, but still lets you save. Whoever is editing is usually not the person who put the bad schema there, and turning a broken function off has to stay possible.

Changing the schema field is a different matter. That gets validated on save, so a bad schema cannot be stored in the first place.

It is the same check the upload endpoint makes, so the admin doubles as a place to try a schema out without packaging and uploading a function. It is not a dry run though: a schema that passes gets saved.


Tansito (approved):
LGTM Alberto, thank you!

PR #2401, opened 2026-08-18
Arguments validation: measure what legitimate schemas cost against the CPU budget

avilches:
### Summary

Adds a test that holds the schemas used to check the CPU budget for validating job arguments, and measures what they cost. It comes out of a review question on #2383: is one second of CPU too short for a legitimate schema? The answer was a measurement, and a measurement that only exists in a review comment cannot be re-run, so this puts the schemas in the repository where anyone can read them, change them, and see the numbers for themselves.

No production code changes, this only adds a test file.

### Details and comments

The three profiles are the most expensive legitimate shapes found while working on that PR: a hand-written vendor schema against the largest payload a caller may send (a 1 MB batch of encoded circuits), a wide schema with internal references approaching the subschema limit against the same 1 MB, and `uniqueItems` over 4000 objects. What they measure, on a laptop:

| Profile | CPU time | Margin under the 1s budget |
|---|---|---|
| Typical vendor schema, 1 MB batch of circuits | 0.79 ms | 1267x |
| Wide schema with internal references, 1 MB of arguments | 1.85 ms | 541x |
| `uniqueItems` over 4000 objects | 15.08 ms | 66x |

Run it with `-s` to see those lines printed.

Each profile asserts two things, and the first matters more than the second. It goes through the real isolation with no limit firing, and still tells a valid payload from an invalid one: if the budget were too short, that is the assertion that would fail, and it would fail with `UnsupportedSchemaError` naming "CPU time". Then the cost is measured and has to stay under a quarter of the budget.

The cost is measured with `time.process_time` rather than the wall clock on purpose. CPU time is what `RLIMIT_CPU` counts, and unlike elapsed time it does not inflate when the runner is busy, which is what made earlier wall-clock assertions in this feature fail on CPU-throttled CI runners. The threshold is a quarter of the budget rather than something tighter so that slower hardware has room, while still catching a real regression.

Checked that the test can fail, by restoring jsonschema's own `uniqueItems` in place of the hashing replacement: the third profile then goes red with `it took more than 1 seconds of CPU time`, which is exactly the symptom the review question was about.

### DIFF
--- a/gateway/tests/api/domain/test_legitimate_schema_cost.py
+++ b/gateway/tests/api/domain/test_legitimate_schema_cost.py
@@ -1,0 +1,166 @@
+"""What legitimate schemas actually cost, measured against the CPU budget the isolation gives them.
+
+"Is one second too short?" is answered by a measurement rather than by a proof, so this file holds
+the measurement and repeats it on every run instead of leaving it in a review comment. The profiles
+are the most expensive legitimate shapes found: the largest arguments the limits allow (a 1 MB batch
+of encoded circuits), a schema wide enough to approach MAX_SCHEMA_NODES, and uniqueItems over
+thousands of objects.
+
+Each profile asserts two things, and the first matters more than the second:
+
+- It goes through the real isolation with no limit firing, and still tells a valid payload from an
+  invalid one. If the budget were too short, this is the assertion that would fail, and it would
+  fail with UnsupportedSchemaError naming "CPU time".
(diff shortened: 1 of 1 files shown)


Tansito (approved):
I don't know if this is too much but I trust on your criteria @avilches haha

ElePT (approved):
wow, a very deep reply to a very short comment. Fine with keeping the test.

PR #2404, opened 2026-08-18
Look up jobs with filter().first() instead of catching DoesNotExist

avilches:
### Summary

Every job use case looks a job up by id and turns a miss into a 404. Until now they did it by catching the ORM exception:

```python
try:
    job = Job.objects.get(id=job_id)
except ObjectDoesNotExist:
    raise JobNotFoundException(job_id)
```

This PR moves all eleven of them to the other shape we already use elsewhere in the repo (`api/use_cases/programs/get_jobs.py`, `Provider.objects.get_by_name`):

```python
job = Job.objects.filter(id=job_id).first()
if job is None:
    raise JobNotFoundException(job_id)
```

Follows up on the review discussion in #2402, and rebased on main now that it is in, so the diff here is only the sweep.

### Details and comments

Behaviour is identical, so no test changes: a missing job still raises `JobNotFoundException`, which the exception handler maps to 404. The lookup is still a single primary key query, `first()` only adds an `ORDER BY id LIMIT 1` that the index already satisfies.

Beyond the style preference, the reason to unify is that mixing both shapes is what produced the bug in #2402. `JobRetrieveUseCase` kept a `if job is None` check that stopped being reachable the moment `JobsRepository.get_job_by_id()` (which did return `None`) was replaced by `Job.objects.get()`, and the commit that added the `except` to its ten siblings gave `retrieve.py` the `ObjectDoesNotExist` import and nothing else. With one shape, a lookup that forgets its guard is visibly wrong instead of silently returning a 500.

The now unused `ObjectDoesNotExist` imports are gone from the eleven files. `Config.get()` in `core/models.py` still catches `DoesNotExist` on purpose: it converts the miss into a `KeyError` that flags a boot ordering mistake, which is a different intent from an HTTP 404, so it stays as it is.

Files touched: `retrieve`, `get_result`, `get_logs`, `provider_logs`, `save_result`, `stop`, `set_sub_status`, `create_event`, `list_events`, `get_runtime_jobs` and `associate_runtime_jobs`.

### DIFF
--- a/gateway/api/use_cases/jobs/associate_runtime_jobs.py
+++ b/gateway/api/use_cases/jobs/associate_runtime_jobs.py
@@ -4,7 +4,6 @@ import logging
 from uuid import UUID
 
 from django.contrib.auth.models import AbstractUser
-from django.core.exceptions import ObjectDoesNotExist
 
 from api.access_policies.jobs import JobAccessPolicies
 from api.domain.exceptions.job_not_found_exception import JobNotFoundException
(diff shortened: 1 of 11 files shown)


ElePT (approved):
LGTM

PR #2410, opened 2026-08-19
Arguments validation/backoffice: add an admin page to validate arguments against a function's schema

avilches:
### Summary

Checking whether some arguments match a function's `arguments_schema` currently means going through the API, which needs a token, an instance and an HTTP request. This adds a "Validate Arguments" link on the function page in the Django admin that opens a form where you can paste arguments and see whether they are accepted, along with the reason and the location when they are not.

### Details and comments

- Rule 1: Nothing is written to the database: the page only reads the function to get its schema.
- Rule 2: The validation answer must be exactly the one a client would get. It calls the same `validate_arguments` the API calls, so the size limits, the isolated evaluation and the wording of the rejection are the same, and a function with no schema comes back as valid. That case gets a note next to the verdict saying there was no schema to check against, so an affirmative answer is not read as a check that never happened.

> [!IMPORTANT]
> The page also points out that arguments are validated in the form the SDK sends them, where Qiskit objects arrive as an object with `__type__` and `__value__` keys. That is the thing most likely to surprise someone writing a schema, and without the note trying a circuit gives a confusing result.


Tansito (approved):
Pretty useful @avilches

PR #2413, opened 2026-08-20
Arguments validation/docs: fix release note + spec

avilches:
### Summary

The arguments validation docs were written before the last of the seven PRs landed, so parts of them no longer describe what the code does. Documentation only: the release note, the spec, and one SDK docstring. Everything below was verified by running it against a deployed gateway or by executing the code.

### Details and comments

Two errors in the release note, which matters because it shipped with 0.35.0. It promised the error body jsonschema used to build, but #2383 composes its own, so the real one is `{"message": "'type' validation failed, schema requires 'integer': rejected value 'wrong-type'", "path": ["shots"]}`. It also declared the field as `Optional[Dict[str, Any]]` where it is `Optional[Union[Dict[str, Any], bool]]`, and the boolean half matters since `false` rejects every instance.

Three gaps in the spec. It said messages are jsonschema's own, when `_validation_message` builds them inside the child and caps each quoted value at 150 characters. It never mentioned `format`, which is asserted for six values and ignored for everything else, so `format: "email"` rejects a bad value while `format: "uri"` accepts one silently. And it presented the memory figures as general when they were measured at two workers, while the worst case is the margin times workers times threads.

Three behaviours that were correct but unwritten, each reachable by a caller: `/validate_arguments` also skips the Fleets Code Engine project check, not just `disabled`; a function with no schema answers `200 {"valid": true}` to anything, non-JSON included; and blank `arguments` is the one `400` here with no `path`, because DRF rejects it in the serializer.

The SDK docstring now warns that the schema is matched against the arguments as the SDK encodes them, so a `QuantumCircuit` is an object with `__type__` and `__value__` and describing it as `{"type": "array"}` rejects every legitimate call.

Left out on purpose: the chart README does not list the two `argumentsSchema*` values, but its table covers only two of the eight `application.limits` keys, so completing it is a separate chore. Both already have detailed comments in `values.yaml`.

### DIFF
--- a/releasenotes/notes/0.35.0/arguments-validation-5b7701bd6f304c2f.yaml
+++ b/releasenotes/notes/0.35.0/arguments-validation-5b7701bd6f304c2f.yaml
@@ -20,9 +20,13 @@ features:
         runnable = client.upload(function)
 
     From then on, running that function with arguments that do not match the schema is rejected with a ``400`` before
-    the job is created, so no job is queued and no partial rows are left behind. The error names the offending field::
+    the job is created, so no job is queued and no partial rows are left behind. The error names the keyword that
+    failed, what the schema asked for, and the value that was rejected, and ``path`` points at the offending field::
 
-        {"message": "'wrong-type' is not of type 'integer'", "path": ["shots"]}
+        {
+          "message": "'type' validation failed, schema requires 'integer': rejected value 'wrong-type'",
+          "path": ["shots"]
+        }
 
     The feature is opt-in and backwards compatible: functions without a schema are never validated.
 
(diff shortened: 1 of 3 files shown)


releasenotes/notes/0.35.0/arguments-validation-5b7701bd6f304c2f.yaml
```diff
@@ -31,6 +35,20 @@ features:
...
+    with ``{"type": "array"}`` therefore rejects every legitimate call. The advice is to constrain the plain arguments
+    (counts, names, options, flags) and to check only the ``__type__`` tag of the Qiskit ones.
+
+    Second, a declared schema brings size limits with it: the schema itself may be at most 64 KB and is turned down at
+    upload time when it is larger, may not nest deeper than 64 levels or hold more than 200 subschemas, and the encoded
+    arguments of a function that declares a schema may be at most 1 MB. Evaluating a schema is also bounded in CPU time
+    and memory, so a schema that is too expensive to evaluate comes back as a ``400`` naming the limit it hit rather
+    than as a server error. Functions without a schema are unaffected by all of this.
+
+    Both are written up in full in ``specs/ARGUMENTS_VALIDATION.md``.
```
avilches:
```suggestion
```

ElePT (approved):
LGTM. The deployed release notes in github pages should update automatically, and we can also update the wording manually in the github release page.

PR #2414, opened 2026-08-20
Wrap job creation in a database transaction

avilches:
### Summary
A failure part way through `POST /programs/run` left an orphan `JobConfig` row behind, because that row was created and committed before the `Job` was saved. Job creation now runs inside a single database transaction.

### Details and comments
The transaction covers only the database writes. The arguments upload stays outside it and still runs first, because for Fleets it goes to COS and holding a transaction open across a network call would keep it alive for as long as the upload takes. A rollback therefore leaves the arguments file behind, same as before.


avilches:
I added a release note, not sure if it's worth it, you tell me if I should delete it @ElePT

gateway/api/use_cases/programs/run.py
```diff
@@ -136,12 +135,18 @@ def execute(  # pylint: disable=too-many-locals
             env.update(json.loads(function.env_vars))
         job.env_vars = json.dumps(env)
 
+        # Deliberately outside the transaction below: for Fleets this uploads to COS, and keeping
+        # a database transaction open across a network call holds it for as long as the upload takes.
```
avilches:
```suggestion
```

releasenotes/notes/wrap-job-creation-in-transaction-8b2f4c6d1e9a3705.yaml
```diff
@@ -0,0 +1,6 @@
+---
+fixes:
+  - |
+    Job creation on ``POST /api/v1/programs/run`` now runs inside a single
+    database transaction, so a failure part way through no longer leaves an
+    orphan ``JobConfig`` row behind.
```
ElePT: This case is a bit on the edge between keeping and removing. I think that because this wasn't really affecting users in this case we can delete the release note. But I'm sure that if you look back we do have some similar release notes we ended up merging.

ElePT (approved):
From the description of the problem, LGTM, but I will leave it unmerged in case @Tansito has any additional comment.

releasenotes/notes/wrap-job-creation-in-transaction-8b2f4c6d1e9a3705.yaml
Tansito: Yeah, same here. I think I would remove the release note, I don't see it interesting for anyone.

Tansito (commented):
Same opinion as Elena, I see everything fine I would just remove the release note to be honest. Good catch btw, @avilches !

ElePT (dismissed):
Requesting a change so that @avilches sees it and removes the reno :)

avilches:
> Requesting a change so that @avilches sees it and removes the reno :)

Np! Removed! @Tansito @ElePT :)

Tansito (approved):

PR #2415, opened 2026-08-21
Fleets: inject private URLs into job containers

ElePT:
## Summary

This PR adds 4 new gateway settings driven by env vars: 

 - `FLEETS_RUNTIME_IAM_URL`
 - `FLEETS_RUNTIME_GLOBAL_SEARCH_URL`
 - `FLEETS_RUNTIME_GLOBAL_CATALOG_URL`
 - `FLEETS_RUNTIME_EXPERIMENTAL`

When these variables are set,  `build_run_env_variables()` will inject the corresponding `qiskit-ibm-runtime` env. variables in the Fleets job container:

| qiskit-serverless | qiskit-ibm-runtime |
|---|---|
| `FLEETS_RUNTIME_IAM_URL` | `IAM_URL` |
| `FLEETS_RUNTIME_GLOBAL_SEARCH_URL` | `GLOBAL_SEARCH_URL` |
| `FLEETS_RUNTIME_GLOBAL_CATALOG_URL` | `GLOBAL_CATALOG_URL` |
| `FLEETS_RUNTIME_EXPERIMENTAL` | `QISKIT_FUNCTIONS_EXPERIMENTAL` |

This enables Fleets jobs to access Qiskit IBM Runtime via private endpoints through the VPE (Virtual Private Endpoint) setup in Code Engine. The setup is different for production and staging. The internal Helm chart will set the correct private IBM Cloud service endpoints per environment:

  - **prod**: `FLEETS_RUNTIME_IAM_URL` (prod endpoint), `FLEETS_RUNTIME_GLOBAL_SEARCH_URL`, `FLEETS_RUNTIME_GLOBAL_CATALOG_URL`
  - **staging**: `FLEETS_RUNTIME_IAM_URL` (test endpoint), `FLEETS_RUNTIME_EXPERIMENTAL=true` (this will skip global search and global catalog endpoints)

## Requirements

`qiskit-ibm-runtime >= 0.49.0` is required in the Fleets job container to support these private endpoint variables. This is satisfied via the client dependency (`qiskit-ibm-runtime>=0.47.0, <0.50.0` in `client/requirements.txt`).

### DIFF
--- a/charts/qiskit-serverless/charts/gateway/values.yaml
+++ b/charts/qiskit-serverless/charts/gateway/values.yaml
@@ -99,6 +99,10 @@ ce:
   defaultComputeProfile: "bx3d-24x120"
   fleetsDefaultImage: "private.icr.io/quantum-public/qiskit-serverless/fleet-node:0.35.0"
   fleetsDefaultMaxInstances: 1
+  fleetsRuntimeIamUrl: ""
+  fleetsRuntimeGlobalSearchUrl: ""
+  fleetsRuntimeGlobalCatalogUrl: ""
+  fleetsRuntimeExperimental: "false"
 
 secrets:
   secretKey:
--- a/gateway/core/ibm_cloud/code_engine/fleets/utils.py
+++ b/gateway/core/ibm_cloud/code_engine/fleets/utils.py
@@ -211,6 +211,7 @@ def build_run_env_variables(
             "RESULTS_PATH": paths.container_result_path,
             "LOG_FLUSH_INTERVAL_SECONDS": str(flush_interval),
             "LOG_SIZE_LIMIT_BYTES": str(getattr(settings, "FUNCTIONS_LOGS_SIZE_LIMIT", 52428800)),
+            "QISKIT_IBM_PRIVATE_ENDPOINT": "true",
         }
     )
     if paths.container_private_log_path is not None:
--- a/gateway/core/ibm_cloud/code_engine/fleets/utils.py
+++ b/gateway/core/ibm_cloud/code_engine/fleets/utils.py
@@ -194,6 +194,15 @@ def build_run_env_variables(
     if gateway_host:
         env["ENV_JOB_GATEWAY_HOST"] = gateway_host
 
+    if iam_url := getattr(settings, "FLEETS_IAM_URL", ""):
+        env["IAM_URL"] = iam_url
+    if global_search_url := getattr(settings, "FLEETS_GLOBAL_SEARCH_URL", ""):
+        env["GLOBAL_SEARCH_URL"] = global_search_url
+    if global_catalog_url := getattr(settings, "FLEETS_GLOBAL_CATALOG_URL", ""):
+        env["GLOBAL_CATALOG_URL"] = global_catalog_url
+    if getattr(settings, "FLEETS_EXPERIMENTAL", False):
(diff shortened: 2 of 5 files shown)


gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -194,6 +194,15 @@ def build_run_env_variables(
...
 
+    if iam_url := getattr(settings, "FLEETS_RUNTIME_IAM_URL", ""):
+        env["IAM_URL"] = iam_url
+    if global_search_url := getattr(settings, "FLEETS_RUNTIME_GLOBAL_SEARCH_URL", ""):
+        env["GLOBAL_SEARCH_URL"] = global_search_url
+    if global_catalog_url := getattr(settings, "FLEETS_RUNTIME_GLOBAL_CATALOG_URL", ""):
+        env["GLOBAL_CATALOG_URL"] = global_catalog_url
+    if getattr(settings, "FLEETS_RUNTIME_EXPERIMENTAL", False):
+        env["QISKIT_FUNCTIONS_EXPERIMENTAL"] = "true"
+
```
marceloamaral, line 205: Do we also need the `QISKIT_IBM_PRIVATE_ENDPOINT=true` ?
We were using it before...

gateway/core/ibm_cloud/code_engine/fleets/utils.py
```diff
@@ -202,6 +211,7 @@ def build_run_env_variables(
             "RESULTS_PATH": paths.container_result_path,
             "LOG_FLUSH_INTERVAL_SECONDS": str(flush_interval),
             "LOG_SIZE_LIMIT_BYTES": str(getattr(settings, "FUNCTIONS_LOGS_SIZE_LIMIT", 52428800)),
+            "QISKIT_IBM_PRIVATE_ENDPOINT": "true",
```
ElePT, line 214: This enables the client changes from #2424

charts/qiskit-serverless/charts/gateway/values.yaml
```diff
@@ -99,6 +99,10 @@ ce:
   defaultComputeProfile: "bx3d-24x120"
   fleetsDefaultImage: "private.icr.io/quantum-public/qiskit-serverless/fleet-node:0.35.0"
   fleetsDefaultMaxInstances: 1
+  fleetsRuntimeIamUrl: ""
+  fleetsRuntimeGlobalSearchUrl: ""
+  fleetsRuntimeGlobalCatalogUrl: ""
+  fleetsRuntimeExperimental: "false"
```
ElePT, line 105: The specific values will be set in an internal PR

marceloamaral (approved):
LGTM

gateway/core/ibm_cloud/code_engine/fleets/utils.py
avilches, line 205: I know this is already merged, and my comment is not really important. But I feel that using `getattr` with a default value here (or every time a setting is used) is redundant. There should always be a default value per field, and the place to define this default value should be where the field is created: settings.py, not in every place where the setting is used. This creates the false impression that the field could be null and that the default value is the one defined in `getattr`, when in reality it is never null and, if it were, the default value is in `os.environ.get()`.

tag people because the PR is merged, and maybe may comment is lost @ElePT @marceloamaral

PR #2416, opened 2026-08-21
Scope provider job listings to the requested provider

dpatel-ops:
### Problem

`JobsQuerySet._apply_filters` only applied `filters.provider` inside the `filter=CATALOG`
arm. `GET /v1/jobs/provider` never sets `filter`, so the provider the caller asked for was
silently dropped and scoping fell to the function-title filter alone.

Titles can't scope by provider — they're unique *per provider*, not globally
(`unique_provider_title`). So the listing returned same-titled functions belonging to other
providers, plus serverless functions of that name. For a provider admin it's worse:
`_apply_access_scope` returns `None` (provider-wide, no title filter), so with the provider
clause missing there was no filter at all and the endpoint returned every job in the system.

This is the default configuration — `gateway.runtime_instances_api.enabled` ships `"false"`,
making the legacy provider-admin branch the live path.

The same clause also compared a provider **name** against `Provider`'s UUID primary key,
raising `ValueError: badly formed hexadecimal UUID string`. `CATALOG` was the only arm that
reached it, so `GET /v1/jobs?filter=catalog&provider=<name>` was a 500 — probably why the
missing scoping went unnoticed.

### Severity

Cross-tenant **metadata** disclosure, bounded. The serializer exposes job id, status,
timestamps and function `{id, title, provider}` — no results, logs or arguments. No
escalation: `JobAccessPolicies.can_access` re-authorizes against the job's *own* provider, so
a leaked id can't be used against retrieve / results / provider-logs.

### Fix

Apply the provider clause whenever a provid `filter`, and join
on the name. Under `CATALOG` a named provider makes the existing
`exclude(program__provider=None)` redundant, so collapsing that `if/else` loses nothing.

### Behaviour changes

- Provider admins now see only the requestejob in the system.
- `GET /v1/jobs?provider=X` now actually filters; it was ignored unless `filter=catalog` was
  also passed. That endpoint was always lim, so this narrows an
  already-correctly-scoped result.
- `?filter=catalog&provider=<name>` returns

### Tests

8 new cases across `test_jobs.py::TestFilterProvider` and
`test_provider_list.py::TestProviderScope`; 6 fail without the fix, 2 are guardrails pinning
the `match` arms. Clean `main` 956 passed →1 pre-existing
environmental failures. `black` clean, `pylint` 10.00/10.

### Checklist


gateway/core/model_managers/jobs.py
```diff
@@ -85,13 +85,23 @@ def _apply_filters(
...
+        # Applied whenever a provider is named, independently of `filter`. It used to live inside the
+        # CATALOG arm above, which had two consequences:
+        #   - `jobs/provider` never sets `filter`, so the provider the caller asked for was dropped and
+        #     scoping fell to the title filter below. Titles are unique only *per provider*, so that
+        #     returned same-titled functions of other providers, and custom functions too.
+        #   - it compared against Provider's UUID primary key while callers pass a name, raising
+        #     ValueError("badly formed hexadecimal UUID string"). Every other provider lookup in the
+        #     codebase joins on `__name`.
+        # Under CATALOG a named provider makes the exclude() above redundant, so dropping the branch
+        # there loses nothing: filtering to one provider already excludes the null-provider rows.
```
avilches:
```suggestion
```

This comment doesn't comment on the code, it comments on the change: it explains where the code was and why it was wrong, and now it's better. So I would just remove it...

avilches:
Could you please reduce the amount of test? I don't feel adding two classes and 8 new test just for a filter is healthy. Try to locate the tests where the filter is checked and ensure it works with the new code.

dpatel-ops:
make sense @avilches , removed extra tests and fixtures where possible

avilches (approved):

avilches:
> make sense @avilches , removed extra tests and fixtures where possible

Thanks!

PR #2417, opened 2026-08-24
Add regions to kafka

korgan00:
### Summary

Implement region-aware event publishing for Qiskit Functions. The service is global but usage events must be published to regional Kafka buses. This change makes `KafkaEventStreamsClient` hold multiple regional producers and route each job's usage events to the correct bus based on the region encoded in the job's instance CRN.

### Details and comments

**Core**
- `kafka_event_streams_client.py`: Changed from single producer to registry of regional producers. Auto-discovers regions from `EVENT_STREAMS_BOOTSTRAP_SERVERS_<REGION>` env vars. Routes events by extracting region from CRN (6th colon-delimited segment).

**Tests** (9 new cases)
- Default region from unsuffixed vars (back-compat)
- Suffixed vars discovered by scan
- `EVENT_STREAMS_DEFAULT_REGION` respected
- Correct producer selected by region
- Unconfigured region raises error
- Null/malformed CRN falls back to default
- Missing API key raises at init
- Startup log lists regions

Environment:
```bash
EVENT_STREAMS_ENABLED=true
EVENT_STREAMS_DEFAULT_REGION=us-east
ENVIRONMENT=production
```

### Checklist


charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -558,6 +560,22 @@ spec:
                 secretKeyRef:
                   name: kafka-credentials
                   key: user
+            # Multi-region Event Streams support: optional suffixed regional secrets
+            # (kafka-credentials-eu-de, kafka-credentials-ap-sg, etc) are projected here if present
+            {{- range $region := (list "eu-de" "eu-gb" "ap-sg" "au-syd" "jp-osm") }}
```
ismaelRozasRamallal: By the moment, let's restrict it to just eu-de since that's the only region we have:
```suggestion
            {{- range $region := (list "eu-de") }}
```

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -558,6 +560,22 @@ spec:
...
+            - name: EVENT_STREAMS_BOOTSTRAP_SERVERS_{{ upper (replace $region "-" "_") }}
+              valueFrom:
+                secretKeyRef:
+                  name: kafka-credentials-{{ $region }}
+                  key: bootstrap_servers
+            - name: EVENT_STREAMS_API_KEY_{{ upper (replace $region "-" "_") }}
+              valueFrom:
+                secretKeyRef:
+                  name: kafka-credentials-{{ $region }}
+                  key: api_key
```
ismaelRozasRamallal, line 579: missing here to also build the env variable EVENT_STREAMS_USER with the region. For us-east (the default one which is not prefixed), we are defining the three (EVENT_STREAMS_BOOTSTRAP_SERVERS, EVENT_STREAMS_API_KEY and EVENT_STREAMS_USER)

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -43,38 +43,81 @@ class KafkaEventStreamsClient(EventStreamsClient):
...
+            self._producers[default_region] = self._create_producer(default_bootstrap_servers, default_api_key)
+
+        # Discover regional producers by scanning for suffixed env vars
+        for env_key in os.environ:
+            if env_key.startswith("EVENT_STREAMS_BOOTSTRAP_SERVERS_"):
+                suffix = env_key[len("EVENT_STREAMS_BOOTSTRAP_SERVERS_") :]
+                region = suffix.lower().replace("_", "-")
+                bootstrap_servers = os.environ[env_key]
+                api_key_env = f"EVENT_STREAMS_API_KEY_{suffix}"
+                api_key = os.environ.get(api_key_env)
```
ismaelRozasRamallal, line 83: Can we group this two lines? I don't think we need to separate the api key string concatenation (unless it conflicts with the `environ.get`, which I don't think so)

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -43,38 +43,81 @@ class KafkaEventStreamsClient(EventStreamsClient):
...
+        self._default_region = default_region
+
+        # Register default region from unsuffixed variables
+        default_bootstrap_servers = os.environ.get("EVENT_STREAMS_BOOTSTRAP_SERVERS")
+        default_api_key = os.environ.get("EVENT_STREAMS_API_KEY")
+        if default_bootstrap_servers and default_api_key:
+            self._producers[default_region] = self._create_producer(default_bootstrap_servers, default_api_key)
+
+        # Discover regional producers by scanning for suffixed env vars
+        for env_key in os.environ:
```
ismaelRozasRamallal, line 76: can we add some debug logs in this loop just in case it doesn't work at first to have some visibility on what envs we are parsing, how we are doing that...?

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -186,15 +229,31 @@ def _publish(
             "data": data,
         }
 
+        # Route to the appropriate regional producer
+        region = self._region_from_crn(job.instance_crn)
+        if region is None:
+            region = self._default_region
```
ismaelRozasRamallal, line 240: I would say if we don't find the region to throw an exception. It's better to keep them non emitted than send it to the wrong region 👍

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -545,6 +545,8 @@ spec:
               value: {{ .Values.application.eventStreams.enabled | default false | quote }}
             - name: ENVIRONMENT
               value: {{ .Values.application.eventStreams.environment | default "" | quote }}
+            - name: EVENT_STREAMS_DEFAULT_REGION
+              value: {{ .Values.application.eventStreams.defaultRegion | default "us-east" | quote }}
```
ismaelRozasRamallal, line 549: This default region env definition can go inside the if `{{- if .Values.application.eventStreams.enabled }}`

charts/qiskit-serverless/README.md
```diff
@@ -71,3 +71,29 @@ Gateway is the API that we offer to manage Qiskit Patterns. Scheduler is the par
...
+| application.eventStreams.enabled        | enable / disable event publishing to Kafka                                                                 |
+| application.eventStreams.environment    | deployment environment name (e.g. production, staging); passed to Kafka topic name                         |
+| application.eventStreams.defaultRegion  | default region served by unsuffixed `kafka-credentials` secret (default: `us-east`)                        |
+
+**Event Streams Secrets**
+
+The default region uses the `kafka-credentials` secret with keys `bootstrap_servers` and `api_key`. Additional regions use suffixed secrets: `kafka-credentials-eu-de`, `kafka-credentials-ap-sg`, etc. Each regional secret requires the same keys.
+
+```bash
+# Default region (required if eventStreams.enabled is true)
```
ismaelRozasRamallal, line 90: I am thinking about the wording in this PR, concretely about how we are using the "default" word.

From how we are using it, it seems that we have a region were to publish all those jobs which doesn't fit in other region, but that's not the case... in order to avoid future confusions, I think it's better to use the word "main" (main region instead of default region) and update the references we have here.

gateway/tox.ini
```diff
@@ -25,6 +25,9 @@ setenv =
   # Local-only dev secret so static analysis tools (pylint-django, import-linter)
   # can import Django settings, which now fail closed with DEBUG off and no key.
   DJANGO_SECRET_KEY=django-insecure-tox-only-change-me
+  # Event Streams test configuration for multi-region support
+  EVENT_STREAMS_DEFAULT_REGION=us-east
```
ismaelRozasRamallal, line 29: aligned with my previous comment, we should rename where having the default word:

```suggestion
  EVENT_STREAMS_MAIN_REGION=us-east
```

ismaelRozasRamallal:
`deployment.yaml` lines 568 and 574 use `upper (replace $region "-" "_")`. Sprig's signature is `replace OLD NEW STRING`, with the string last so it can be pipelined, so this replaces `"-"` inside the literal `"_"` and returns `"_"`. The region name is lost.

Rendered from this branch with `helm install --dry-run=server` against a cluster where `kafka-credentials-eu-de` exists:

```
- name: EVENT_STREAMS_BOOTSTRAP_SERVERS__
- name: EVENT_STREAMS_API_KEY__
```

Expected `EVENT_STREAMS_BOOTSTRAP_SERVERS_EU_DE` / `EVENT_STREAMS_API_KEY_EU_DE`.

The effect is not a crash. In `kafka_event_streams_client.py` the discovery loop takes the suffix `"_"` and computes `region = "-"`, so a producer registers under the region `"-"` and no producer is ever registered for `eu-de`. Any job whose CRN resolves to `eu-de` then fails to find a producer instead of publishing, and the startup log lists `regions=['-']`.

Suggested fix, verified to render `EVENT_STREAMS_BOOTSTRAP_SERVERS_EU_DE` pointing at the right secret and key:

```
{{ $region | replace "-" "_" | upper }}
```

Worth noting why CI is green: the unit tests set `EVENT_STREAMS_BOOTSTRAP_SERVERS_EU_DE` directly rather than deriving it from the chart, so both sides pass in isolation and only the seam between them is broken. A render assertion on the generated env var names would cover this.

### Minor

- The region list is hardcoded to `(list "eu-de")` while the comment above it and the PR description both describe auto-discovery of additional regions. Adding `ap-sg` needs a template edit, so it may be worth aligning the comment with the behaviour.

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -561,6 +563,22 @@ spec:
                 secretKeyRef:
                   name: kafka-credentials
                   key: user
+            # Multi-region Event Streams support: optional suffixed regional secrets
+            # (kafka-credentials-eu-de, kafka-credentials-ap-sg, etc) are projected here if present
+            {{- range $region := (list "eu-de") }}
+            {{- if (lookup "v1" "Secret" $.Release.Namespace (print "kafka-credentials-" $region)) }}
+            - name: EVENT_STREAMS_BOOTSTRAP_SERVERS_{{ upper (replace $region "-" "_") }}
```
ismaelRozasRamallal: Can you double check if this format works? as far as I could read, if we are using Helm under ArgoCD, Helm uses the Sprig template funtions where the right signature is `{{ upper (replace "-" "_" $region) }}`

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
```diff
@@ -561,6 +563,22 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: kafka-credentials-{{ $region }}
+                  key: bootstrap_servers
+            - name: EVENT_STREAMS_API_KEY_{{ upper (replace $region "-" "_") }}
```
ismaelRozasRamallal: same here 😄

charts/qiskit-serverless/charts/gateway/templates/deployment.yaml
korgan00: The merge with main solved it: {{ $region | replace "-" "_" | upper }}

ismaelRozasRamallal (approved):
LGTM!!

PR #2419, opened 2026-08-24
Set an explicit request body size limit and report going over it as 413

### LINKED ISSUE #2418: 500 Internal server error on /programs/run with a large batch of arguments
### What happens

Running a function with a large batch of arguments fails with a 500:

```
Message: Http bad request.
Code: 500
Details: Internal server error
```

The gateway log shows `django.core.exceptions.RequestDataTooBig: Request body exceeded settings.DATA_UPLOAD_MAX_MEMORY_SIZE.`

### Cause

`DATA_UPLOAD_MAX_MEMORY_SIZE` is Django's own setting, it defaults to 2.5 MB, and this project never set it. That default never reached a JSON request, because Django REST Framework read the request stream directly, until **3.17.2** added this to `Request._parse`:

```python
from rest_framework.parsers import FormParser, JSONParser
if isinstance(parser, (JSONParser, FormParser)):
    stream = io.BytesIO(self.body)
```

`HttpRequest.body` is where Django enforces the limit, so from that version on it applies to every JSON and form request. The SDK posts `/programs/run` as JSON, and one 100 qubit, depth 100 circuit is about 39 KB of base64, so a batch of the order of sixty five reaches 2.5 MB.

`endpoint_handle_exceptions` then catches `RequestDataTooBig` in its generic `except Exception` and answers 500 with "Internal server error".

### Measured

Running the same test against each version, `POST /api/v1/programs/run/` with a 3 MB body:

| Django REST Framework | Function declares a schema | Result |
|---|---|---|
| 3.17.1 | no | 200, job queued |
| 3.17.2 | no | refused |
| 3.18.0 | no | 500 `RequestDataTooBig` |
| 3.18.0 | yes | 500 `RequestDataTooBig` |

Two things follow from the version.

It is not related to arguments validation: the failure happens identically with and without a schema, and 3.17.2 was released on 2026-08-05, two weeks before that feature was merged.

And it needed no change in this repository to arrive. The pin was `djangorestframework>=3.17.1, <4`, an open range, so 3.17.2 came in on any image rebuild from that date, which is why the failures look intermittent and why they predate the explicit bump to 3.18.0 (#2393).

### Two things this uncovered

- The 500 hides a limit the caller could work around by sending less per batch. It should be a 413 naming the limit.
- `MAX_ARGUMENTS_LENGTH` is 1 MB, about twenty five encoded circuits, which is below what callers send per batch. A function that declares a schema gets a legitimate 400 for an ordinary batch.

avilches:
### Summary

- Fixes #2418. Running a Qiskit Function with a large batch of arguments failed with `500 Internal server error`.

Django 3.18.0 added a step through `HttpRequest.body` in `Request._parse` for JSON and form requests, which started applying Django's `DATA_UPLOAD_MAX_MEMORY_SIZE`, 2.5 MB by default, to a JSON body for the first time. One 100 qubit, depth 100 circuit is about 39 KB of base64, so a batch of fifty is already over that. `RequestDataTooBig` then reached the endpoint decorator's generic `except Exception` and came back as a 500.

Measured with the same test on both versions, `POST /programs/run` with a 3 MB body gives a 500 on 3.18.0 and a 200 on 3.17.1, identically whether or not the function declares a schema. So this is unrelated to arguments validation, even though the two changes landed a day apart, which is what made them look connected.

This PR does three things:

- Sets the body limit explicitly, to 50 MB, instead of inheriting a default nobody chose.
- Answers `413 Request Entity Too Large` naming the limit, instead of a 500 that tells the caller nothing about a request they could fix by sending less per batch.
- Raises the arguments length checked against a function's `arguments_schema` from 1 MB to 32 MB, and makes it configurable like the other limits around that feature. The old figure fitted about twenty five encoded circuits, below what callers send in one batch.

### Details and comments

**Every limit is now configurable from the same block**, where `MAX_ARGUMENTS_LENGTH` used to be a constant in the code:

| Setting / env var | Default | Clamped to | Chart value under `gateway.application.limits` |
|---|---|---|---|
| `MAX_REQUEST_BODY_SIZE_MB` | 50 | not clamped | `maxRequestBodySizeMb` |
| `MAX_ARGUMENTS_LENGTH_MB` | 32 | 1-64 | `maxArgumentsLengthMb` |
| `ARGUMENTS_SCHEMA_MEMORY_LIMIT_MB` | 128 | 64-256 | `argumentsSchemaMemoryLimitMb` |
| `ARGUMENTS_SCHEMA_CPU_LIMIT_SECONDS` | 1 | 1-5 | `argumentsSchemaCpuLimitSeconds` |

**Why the arguments limit is smaller than the body limit**, which is the question the numbers exist to answer: accepting a body is paid once in the worker, while validating forks a child that parses the arguments again under its own memory and CPU allowance. Measured on Linux in a container, where `RLIMIT_AS` is actually applied, that child's address space grows 1.00 MB per MB of arguments for a batch of circuits, all of it `json.loads` rather than the schema comparison, but 8.2 MB per MB for an array of tiny objects, which also spends 0.27 CPU seconds per MB. So cost follows the shape of the payload and not only its size, and no length limit can guarantee the child will fit. For the shape the platform actually receives it is the memory margin that refuses a caller first, at about 127 MB, and the length is the cheaper bound applied before anything is forked.

**Both limits are bounded by the pod, not by one request.** A body costs about 3.2 times its size in the worker holding it, and with `--threads=1` every worker can hold one at a time, so the pod needs `resources.limits.memory` above `workers x 3.2 x maxRequestBodySizeMb`. That matters because the two ways of running out of memory are not equivalent: when the child's own limit fires the caller gets a 400 and the server carries on, but when the pod runs out the kernel's OOM killer picks a victim by size and can take a gunicorn worker instead, which is a dropped connection and no status code at all. Both happened in one measured run of ten concurrent 200 MB bodies in a 4 GB container: seven clean rejections and three workers killed with `SIGKILL`.

**All of the above, the tables, and the scripts to repeat any measurement are in [`specs/ARGUMENTS_LIMIT.md`](specs/ARGUMENTS_LIMIT.md)**, added by this PR. It includes the sizing table per pod memory and worker count, the container recipe, and which figures are worth re-checking if `jsonschema`, Django or DRF move. `specs/ARGUMENTS_VALIDATION.md` now points there instead of carrying its own copy of the same numbers, and the comments in `settings.py` and the chart's `values.yaml` are one or two lines each pointing to the same place.

**Chart.** One value drives the body limit: the gateway reads it as `MAX_REQUEST_BODY_SIZE_MB`, and when the gateway ingress is enabled `templates/ingress.yaml` derives its `proxy-body-size` annotation from the same value, so a proxy cannot allow less than the gateway and cut a body off before it can answer 413. An annotation set under `ingress.annotations` still wins. A deployment fronting the gateway another way, with an Istio ingress gateway for instance, has to bound the body there itself.

Three regression tests cover the fix in `gateway/tests/api/test_request_body_size.py`: a batch over Django's old default is accepted with and without a schema, and an oversized body comes back as a 413 with the limit in the message.

### DIFF
--- a/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
+++ b/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
@@ -18,10 +18,3 @@ upgrade:
     count, since a body costs about three times its size in the worker holding it
     and every worker can hold one at once. ``specs/ARGUMENTS_LIMIT.md`` has the
     table.
-  - |
-    When the gateway ingress is enabled, the chart now derives its
-    ``nginx.ingress.kubernetes.io/proxy-body-size`` annotation from
-    ``maxRequestBodySizeMb``, so the proxy and the gateway agree on the limit.
-    Setting that annotation under ``gateway.ingress.annotations`` still overrides
-    it. A deployment that fronts the gateway another way has to bound the request
-    body there itself.
--- a/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
+++ b/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
@@ -20,10 +20,10 @@ fixes:
     evaluates the schema: five hundred circuits cost 55 MB of peak child memory.
 upgrade:
   - |
-    The chart derives the ingress annotation
-    ``nginx.ingress.kubernetes.io/proxy-body-size`` for the gateway from
+    When the gateway ingress is enabled, the chart now derives its
+    ``nginx.ingress.kubernetes.io/proxy-body-size`` annotation from
     ``gateway.application.limits.maxRequestBodySizeMb``, so the proxy and the
     gateway agree on the limit and changing it in one place changes both. A
     deployment that sets that annotation under ``gateway.ingress.annotations``
-    keeps its own value, and the proxy then rejects a large body before the
-    gateway sees it.
+    keeps its own value. A deployment that fronts the gateway some other way has
+    to bound the request body there itself.
--- a/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
+++ b/releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
@@ -14,11 +14,19 @@ fixes:
     endpoint decorator's generic handler produced.
   - |
     Raised the maximum length of the arguments validated against a function's
-    ``arguments_schema`` from 1 MB to 8 MB. The old figure fitted about twenty five
-    encoded circuits, which was under what callers send in one batch. Validating
-    the larger limit stays well inside the memory given to the forked child that
-    evaluates the schema: five hundred circuits cost 55 MB of peak child memory.
+    ``arguments_schema`` from 1 MB to 32 MB. The old figure fitted about twenty
+    five encoded circuits, which was below what callers send in one batch. The new
+    one fits about eight hundred and still leaves the validation well inside the
+    memory it is allowed: validating a batch of circuits grows the forked child by
+    2.24 MB per MB of arguments, so 32 MB uses about 72 MB of the 128 MB margin.
 upgrade:
+  - |
+    Both limits are bounded by the pod rather than by one request, so weigh them
+    against ``gateway.resources`` and ``gateway.application.httpServer.workers``.
+    A request body costs about 3.5 times its size in the worker holding it. At
+    4 Gi with 5 workers, a 50 MB body leaves roughly 2.9 GB in use at full
+    concurrency; 100 MB would reach 3.5 GB, and past the pod's memory the kernel
+    OOM-kills a worker instead of the gateway answering 413.
   - |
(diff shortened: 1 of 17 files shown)


releasenotes/notes/fix-request-body-size-limit-9c4f81ba2e5d7a36.yaml
```diff
@@ -0,0 +1,27 @@
...
+    count, since a body costs about three times its size in the worker holding it
+    and every worker can hold one at once. ``specs/ARGUMENTS_LIMIT.md`` has the
+    table.
+  - |
+    When the gateway ingress is enabled, the chart now derives its
+    ``nginx.ingress.kubernetes.io/proxy-body-size`` annotation from
+    ``maxRequestBodySizeMb``, so the proxy and the gateway agree on the limit.
+    Setting that annotation under ``gateway.ingress.annotations`` still overrides
+    it. A deployment that fronts the gateway another way has to bound the request
+    body there itself.
```
avilches:
```suggestion
```

ElePT (approved):
Thanks a lot for the PR and the specs, after our offline review, LGTM

PR #2421, opened 2026-08-24
Drop the `api_program.default_compute_profile` column

avilches:
### Summary

Do not merge yet. This is the second half of #2420, and it should only ship once no release that still declares `Program.default_compute_profile` can be deployed any more.

#2420 removed the field from Django's state but kept the `api_program.default_compute_profile` column in the database, so that the previous release stayed deployable. This PR adds migration `0058`, which drops the column for real.

### Details and comments

The field is already out of Django's state, so a plain `RemoveField` would fail: `database_forwards` looks the field up in the from state and would not find it. Instead the migration puts the field back into the state without touching the database, then removes it normally. The drop goes through the schema editor, so there is no hand-written SQL and no hardcoded table name.

Nothing reads the column. It was only ever settable through the admin backoffice, and the Fleets runner resolves its default from the `DEFAULT_COMPUTE_PROFILE` setting, so no data migration is needed.

After this migration runs, rolling back to any release up to 0.35.0 stops working, since those declare the field and Django lists every declared column in each `SELECT`. Recovering from that would mean recreating the column by hand:

```sql
ALTER TABLE api_program ADD COLUMN default_compute_profile varchar(255) NULL;
```

The migration is reversible, so `migrate api 0057` also brings the column back, empty.


releasenotes/notes/drop-program-default-compute-profile-column-9f4c2be71a3d5e08.yaml
```diff
@@ -0,0 +1,8 @@
+---
+upgrade:
+  - |
+    The ``api_program.default_compute_profile`` column is now dropped from the
+    database. The field itself was already removed from Qiskit Functions in an
+    earlier release, which kept the column in place so that release could still
+    be rolled back. Once this migration runs, any release that still declares
+    the field can no longer be deployed.
```
avilches:
```suggestion
---
upgrade:
  - |
    The ``api_program.default_compute_profile`` column is now dropped from the
    database. The field was removed from Qiskit Functions in the
    earlier release.
```

avilches:
Previous version already deployed. Safe to merge @Tansito @ElePT

ElePT (approved):
LGTM

PR #2426, opened 2026-08-25
Stabilize the flaky uniqueItems CPU budget test

avilches:
### Summary

Fixes an intermittent CI failure in `test_a_legitimate_schema_validates_well_inside_the_cpu_budget` (the "uniqueItems over 4000 objects" case), first seen on the CI run of #2419 and gone on retry. The test was introduced by #2401.

The test measures a single call to `validator.iter_errors(parsed)` with `time.process_time()` and asserts it stays under 25% of the isolation's one second CPU budget. `process_time()` bills that window for anything else the process does inside it, including a CPython garbage collection triggered by the roughly 16,000 tracked objects the block creates. That collection cost depends on how many live objects the rest of the test suite has accumulated, not on the work being measured, which is around 14ms. On a CI run this pushed the measurement to 293ms against the 250ms threshold.

The fix pauses the collector around a window of three repetitions and keeps the cheapest one, restoring it in a `finally` block, which removes the collection cost and any CPU contention spike from the measurement.

### Details and comments

This is a test-only change. It does not touch any production code, so no release note is included.

Before opening this, the fix was verified beyond just re-running the test:

- 30 runs before and after the change show a tighter spread (11.92-13.68ms before, 10.46-11.68ms after on the same machine), consistent with keeping the minimum of three repetitions with the collector paused.
- 10 more runs under real CPU contention (one CPU-bound process per core running in the background) all pass without approaching the budget.
- To make sure the fix did not leave the test unable to catch a real regression, the `uniqueItems` optimization it exercises (`_validator` in `gateway/api/domain/arguments_schema.py`, which replaces jsonschema's pairwise `uniqueItems` check with a single-pass hash comparison) was temporarily reverted to the stock jsonschema behavior. The test still fails, though earlier than the budget assertion: `validate_arguments_in_isolation` raises `UnsupportedSchemaError` because the isolation's hard `RLIMIT_CPU` limit (1 second) is hit before reaching the measurement block. That is an even stronger signal that the regression is caught, and it is the exact failure mode the module's docstring already describes as the assertion that matters most.

Pausing the garbage collector for this window is safe here: the block only builds dictionaries, tuples and a frozenset through `_comparison_key`, which are released by normal refcounting rather than needing the cyclic collector, and the largest profile in this file is about 1MB.


ElePT (approved):
LGTM, it's a test fix so we can merge it without affecting the release.

PR #2427, opened 2026-08-25
Accept licensed as the new name of the subsidized business model

avilches:
### Summary

The `business_model` value `subsidized` is being renamed to `licensed`, so that entitlements configuration uses the same name as billing, the UI and the product decision. This PR is the first step of that rename in the gateway, and it has to land before anything else changes: right now, if `GET /functions` answered `licensed` for a function, `FunctionAccessEntry` would reject it, `FunctionAccessClient` would swallow the error per entry and only log it, and the function would silently drop out of the user's access list. After this PR both names are accepted, so no client loses access whichever name it sends.

From here on, `LICENSED` is the only name the gateway writes and the only one anyone sees. New jobs are stored with it, and a job row still holding the old name is translated as it is read, so `SUBSIDIZED` is gone from API responses, from the admin backoffice and from the billing events. Billing itself does not change: both names were already billed as `licensed`.

Existing rows are deliberately left untouched. A follow-up PR migrates them and then removes the translation.

### Details and comments

`SUBSIDIZED` arrives from two places, and both are handled at the boundary rather than deeper in:

- From the Runtime API, in `FunctionAccessEntry.__post_init__`, next to the uppercasing that was already there.
- From the database, in a new `Job.from_db`. Doing it there means no layer above the model can see the old name, so the serializers and the admin needed no change of their own, and code written later cannot forget about it.

Everything about the old name lives in one new file, `core/domain/subsidized_license_mapping.py`, whose docstring says so. Removing it later is a single delete plus its two call sites.

Two consequences of translating on read that are worth knowing while the old rows are still there:

- An old job that is read and then saved with a full `job.save()` writes `LICENSED` to its row, so the data migration also happens by trickle until the follow-up PR does it in one go. The model's partial writes (`save_direct`, `update_fields`) name their columns and do not touch the field.
- Queries that never build a model instance (`values`, `values_list`, aggregates) read the column raw, and filtering still matches the stored value. Nothing in the gateway reads or filters `business_model` that way today.

`SUBSIDIZED` is dropped from the field `choices`, since no instance in memory can hold it any more. That keeps it out of the admin form dropdown and out of the enum published in the OpenAPI schema. It stays in `BILLING_NAMES`, because old rows still need a billing name until they are migrated.

Migration `0058` produces no SQL at all: `choices` and `default` are Django-side only. Checked against PostgreSQL 18.4, where `manage.py sqlmigrate api 0058` prints just `BEGIN; -- (no-op) COMMIT;` and `api_job.business_model` stays a `varchar(50) NOT NULL` with no database default. The migration file carries that note in its header.


gateway/api/use_cases/programs/run.py
```diff
@@ -98,17 +97,19 @@ def execute(  # pylint: disable=too-many-locals
 
         if business_model is None:
             trial = _is_trial(function, user)
-            business_model = BusinessModel.TRIAL if trial else BusinessModel.SUBSIDIZED
+            business_model = BusinessModel.TRIAL if trial else BusinessModel.LICENSED
```
avilches, line 227: This is only for legacy auth, with the instances auth, the business model comes from Ntc and it always comes with some value

gateway/core/domain/authorization/function_access_entry.py
```diff
@@ -5,10 +5,11 @@
 from typing import Set
 
 from core.domain.business_models import BusinessModel
+from core.domain.subsidized_license_mapping import licensed_instead_of_subsidized
 
 logger = logging.getLogger("api.FunctionAccessEntry")
 
-VALID_BUSINESS_MODELS = {BusinessModel.TRIAL, BusinessModel.SUBSIDIZED, BusinessModel.CONSUMPTION}
+VALID_BUSINESS_MODELS = {BusinessModel.TRIAL, BusinessModel.LICENSED, BusinessModel.CONSUMPTION}
```
avilches, line 12: Before validation, the subsidized business model is mapped to licensed, which is why we can afford a validation that does not account for subsidized.

riley2012 (approved):
LGTM!

PR #2431, opened 2026-08-26
T-Shirt Sizes API - uploading function with sizes

ItamarGoldman:
### Summary
The API the user will use to upload a function with T-shirt sizes and update its sizes.
~This PR is blocked until #2390 is merged~

#### Feature

- [x]  User need to be able to upload a function with size and update it (updating sizes change all sizes).
- [x] normalization to lower case text for size.
- [ ] run pipeline (getting the pieces together)
- [ ] A function can be uploaded without FunctionSize and relevant fields if `function.runner !- Fleeets`.
- [ ] Requesting for supported sizes and default size from a function.
- [ ] In the file `client/qiskit_serverless/core/function.py` add to `RunnableQiskitFunction` methods querying the backend for supported size `RunnableQiskitFunction.sizes()` and default one by `RunnableQiskitFunction.get_default_size()`.
- [ ] Tests.
- [ ] releasenotes.

#### Validation

- [x] The defualt size should be an existing function size of the function. 
- [x] Function size validation.
- [ ] Sending the information to the billing service ?

### Checklist


ElePT:
I've picked this up from Itamar and pushed two commits that close most of the open items in the description. Rebased onto `main` (post #2390 and #2447), so the blocker is cleared.

## What the new commits do

**`ca7920c` — client: `function_size` run kwarg + `compute_profile` deprecation**
- Adds a keyword-only `function_size` to the run surface (`RunService.run`, `RunnableQiskitFunction.run`, `ServerlessClient.run` and the IBM subclass) and sends it in the `/programs/run` body.
- Deprecates `compute_profile` in favour of `function_size`: the base `ServerlessClient.run` emits a single `DeprecationWarning` when it's used; subclasses delegate without re-warning. Both are still forwarded for back-compat.

**`c455ceb` — gateway: resolve runs through the size catalog + record provenance**
- **Run pipeline wired end-to-end.** A Fleets run now resolves its compute profile through the function's `FunctionSize` catalog (this is what `resolve_compute_profile` / `get_function_size` from #2390 were waiting for a caller). Precedence:
  1. both `function_size` and `compute_profile` → **400** (ambiguous),
  2. `function_size` → catalog lookup; unknown size → **400** listing the declared sizes,
  3. `compute_profile` (deprecated) → used as-is + a gateway deprecation log,
  4. neither → the function's `default_size`, else `settings.DEFAULT_COMPUTE_PROFILE`.
  All 400s fire before the `Job` is built, so no partial job is persisted.
- **Sizing provenance on `Job`.** Because the size determines the compute profile (and two sizes can map to one profile), the resolved size label was being thrown away. The `Job` now records a `size_source` enum (`REQUESTED` / `DEFAULT_SIZE` / `SETTINGS_DEFAULT` / `COMPUTE_PROFILE` / `NONE`) and a nullable `function_size` FK to the resolved row — so a job where the user explicitly chose the default size is distinguishable from one the platform filled in, and billing can attribute a job to its size tier. Migration `0060`, admin fields, and `size_source` on the run response included.

Verification: gateway suite **1078 passed** (the handful of failures are the known sandbox socket-bind cases, unrelated), client run tests green, `black`/`pylint` (10/10)/`lint-imports` clean, no migration drift.

## Status of the description checklist

Feature:
- [x] Upload a function with sizes / update sizes — *(already done before takeover)*
- [x] Normalization to lowercase for size — *(already done)*
- [x] **Run pipeline (getting the pieces together)** — done in `c455ceb`.
- [x] **A function can be uploaded without FunctionSize when `runner != FLEETS`** — ⚠️ **not addressed and needs a decision.** Today upload seeds a default `FunctionSize` regardless of runner (`_seed_default_size` isn't gated on `runner`), so a Ray function still gets a size row. At *run* time the Ray path correctly ignores sizes (`size_source=NONE`, no FK). If we want upload to skip sizes entirely for non-Fleets, that's a small follow-up. Happy to hear your feedback.
- [ ] **`RunnableQiskitFunction.sizes()` / `get_default_size()`** — ⚠️ **not implemented.** No backend query endpoint or client methods yet. This is the main remaining client-facing feature. Left for a follow-up
- [x] **Tests** — added for the run path and provenance (use-case, view, and client). Upload-side tests from the earlier work remain.
- [ ] **Release notes** — ⚠️ **not added yet.** We don't need renos until the feature is publicly documented.

Validation:
- [x] Default size must be an existing size of the function — *(done)*
- [x] Function size validation — *(done; extended to the run path: unknown size → 400)*
- [x] **Sending information to the billing service** — not in scope of these commits, but synced with billing service team and in principle the compute profile change should immediately translate to the billing events (no action required).
-
### Summary of what's left
1. `RunnableQiskitFunction.sizes()` / `get_default_size()` + the backend endpoint they query.
2. Decision on whether upload should skip `FunctionSize` for non-Fleets functions.

gateway/api/use_cases/programs/upload.py
```diff
@@ -121,23 +194,83 @@ def _create(self, data: UploadFunctionInput, user, provider) -> Function:
...
-        function.save()
+
+        # atomic() so all these writes commit together or not at all: the function
+        # row must exist before a FunctionSize can reference it, and default_size
+        # needs a second save for a UUID that did not exist at the first. Without
+        # it, a size naming an unregistered profile returns 400 with the function
+        # already committed and sizeless.
+        with transaction.atomic():
+            function.save()
+            if data.sizes is not None:
```
Tansito, line 243: It looks like we are not validating that the user can send a `default_size` without `sizes` so he can set here `default_size="s"` without `sizes` and the logic would be a bit weird. Is this expected or should we also validate that if you set `default_size` you must define a `size`?

gateway/api/use_cases/programs/run.py
```diff
@@ -41,37 +42,126 @@ def _is_trial(function: Function, user) -> bool:
...
-    elif function.provider and function.gpu:
-        return RunnerConfig(compute_profile=None, gpu=True, compute_profile_fk=None)
-    else:
-        return RunnerConfig(compute_profile=None, gpu=False, compute_profile_fk=None)
-
-    compute_profile_fk = ComputeProfile.objects.get_by_id(compute_profile)
-    if compute_profile is not None and compute_profile_fk is None:
+    # Ambiguous input is always a 400, whatever the runner, so check before the
+    # Ray short-circuit.
+    if compute_profile_requested and function_size_requested:
```
Tansito, line 105: Hmmm... Is this correct? I assume Ray doesn't care about compute profiles and so on in reality, right? Or is it just for consistency?

gateway/api/use_cases/programs/upload.py
```diff
@@ -121,23 +194,83 @@ def _create(self, data: UploadFunctionInput, user, provider) -> Function:
...
+                _replace_function_sizes(function, data.sizes)
+                self._apply_declared_default(function, data)
+            else:
+                self._seed_default_size(function)
```
Tansito, line 258: For Ray functions do we want to populate them with a size also? Just double checking here.

gateway/main/settings.py
```diff
@@ -488,6 +488,12 @@
 )
 # Compute profile settings for Fleets runner
 DEFAULT_COMPUTE_PROFILE = os.environ.get("DEFAULT_COMPUTE_PROFILE", "24x120")  # 24 CPU, 120GB RAM
+# Size seeded for a function uploaded without an explicit size catalog, so every
+# function has a size to run with while declaring sizes is still optional. The
+# profile must name an existing ComputeProfile row; when no such row exists the
+# function is created with no sizes and runs fall back to DEFAULT_COMPUTE_PROFILE.
+DEFAULT_FUNCTION_SIZE = os.environ.get("DEFAULT_FUNCTION_SIZE", "m")
+DEFAULT_FUNCTION_SIZE_PROFILE = os.environ.get("DEFAULT_FUNCTION_SIZE_PROFILE", "16x128")  # 16 CPU, 128GB RAM
```
Tansito, line 496: should we remove DEFAULT_COMPUTE_PROFILE having this now?

gateway/api/domain/function_sizes.py
```diff
@@ -0,0 +1,105 @@
...
+MAX_SIZES_PER_FUNCTION = 32
+
+# Long enough for a descriptive label, and bounded by FunctionSize.function_size
+# (CharField(max_length=64)) so an over-long name is refused with a 400 here
+# rather than a database error on save.
+MAX_SIZE_NAME_LENGTH = 64
+
+# Bounded by ComputeProfile.compute_profile_id (CharField(max_length=255)), for
+# the same reason.
+MAX_COMPUTE_PROFILE_ID_LENGTH = 255
```
Tansito, line 36: I assume in reality, this should be a set of settings?

gateway/api/domain/function_sizes.py
```diff
@@ -0,0 +1,105 @@
...
+
+# A size catalog is a hand written menu of machine shapes, so single digits are
+# the norm and this only exists to stop a caller from turning one upload into an
+# unbounded number of rows and profile lookups.
+MAX_SIZES_PER_FUNCTION = 32
+
+# Long enough for a descriptive label, and bounded by FunctionSize.function_size
+# (CharField(max_length=64)) so an over-long name is refused with a 400 here
+# rather than a database error on save.
+MAX_SIZE_NAME_LENGTH = 64
```
Tansito, line 32: I think we are going to have xs or xl, should we use just 2 by now here?

gateway/api/domain/function_sizes.py
```diff
@@ -0,0 +1,105 @@
...
+
+# Bounded by ComputeProfile.compute_profile_id (CharField(max_length=255)), for
+# the same reason.
+MAX_COMPUTE_PROFILE_ID_LENGTH = 255
+
+
+class InvalidFunctionSizesError(ValueError):
```
Tansito: This error should be probably with the rest of the exception in the API

gateway/api/use_cases/programs/upload.py
ElePT: you're right, we should be validating that and erroring if there isn't any `size` value that corresponds to the `default_size` key.

gateway/api/use_cases/programs/run.py
ElePT: It's for consistency. Ray ignores compute profiles. I think it's a matter of taste here.

gateway/api/use_cases/programs/upload.py
ElePT: This was one of the open items from my comment above: https://github.com/Qiskit/qiskit-serverless/pull/2431#issuecomment-5494071396 ( A function can be uploaded without `FunctionSize` when `runner != FLEETS `— ⚠️ not addressed and needs a decision.). I think at this point we should not populate a size for Ray, but I am open to your suggestions.

gateway/main/settings.py
ElePT: At the moment, it is still used, because we are still supporting the `compute_profile` input (although deprecated). Once we remove the deprecated compute profile maybe we can simplify the variables and resolve to default size + default size profile. I am a bit weary about this one because I know we have users already running on it, this is why I haven't plain removed it.

gateway/api/domain/function_sizes.py
ElePT: This is a guardrail, I don't think we plan to really change it much, but I don't feel too strongly about it.

gateway/api/domain/function_sizes.py
ElePT: This isn't really limiting the list of sizes. In fact, I was considering opening a follow-up and thinking whether to be more strict with the values we accept as keys in the sizes dict, as well as how many keys. This is a bit undefined at the moment.

gateway/api/domain/function_sizes.py
ElePT: Yes, true.

gateway/api/use_cases/programs/upload.py
Tansito: improved here: [9e25820](https://github.com/Qiskit/qiskit-serverless/pull/2431/commits/9e25820cce74900060605840ceb319d6a34de952)

Tansito: main changes:
- if we receive `sizes` and `default_size` the second one must be included in the first one
- `_create` now forces that `sizes` and `default_size` are together, or you send both or any of them (`_update` doesn't change)

gateway/api/use_cases/programs/run.py
Tansito: Added a test for this specifically to keep track of this: [be13f99](https://github.com/Qiskit/qiskit-serverless/pull/2431/commits/be13f997ee0e71c9c3e23b53897d27c20ae43a85)

gateway/api/use_cases/programs/upload.py
Tansito: fixed in: [70f1806](https://github.com/Qiskit/qiskit-serverless/pull/2431/commits/70f1806331223fe38d7bff3b4137eb4dccaef4d9)

Tansito: I basically separated the validation and the persistence to simplify the logic so Ray doesn't store data related with sizes:
- `_resolve_compute_profiles` -> `_validate_compute_profiles`
- `_replace_function_sizes ` is divided on: `_validate_declared_sizes ` & `_persist_function_sizes ` (last one only being called by Fleets jobs)
- `_apply_declared_default `: now `_validate_declared_default ` but it's validated against the sizes in memory, it doesn't go to DB (initially is fine because we have all the info in memory at this point)

So basically the difference between Ray and Fleets is that for both of them we run the same validations but in case of Ray we don't store sizes-related data.

gateway/main/settings.py
Tansito: after internal conversations, meanwhile we do the backward compatibility we said we were going to keep `DEFAULT_COMPUTE_PROFILE` but for consistency we will assign now `16x128` to it instead of the current `24x120` that we were inheriting from Ray.

Tansito: fixed here: [670f5df](https://github.com/Qiskit/qiskit-serverless/pull/2431/commits/670f5df4ede6c8f6b30e3a778babfa5f9a8455d4)

Tansito: there were some more changes as it was affecting the helm and some tests, I changed it in most of the scenarios

gateway/api/domain/function_sizes.py
Tansito: easy fix: [7a0aaf6](https://github.com/Qiskit/qiskit-serverless/pull/2431/commits/7a0aaf6d04467b46fa9f7bdde6713f35eb9176d2)

Tansito (approved):

PR #2433, opened 2026-08-26
Allow filtering functions by provider name in list()

ElePT:
## Summary

Closes internal issue.

As users gain access to more Qiskit Functions, the output of `catalog.list()` / `client.functions()` grows long and hard to scan, even more when `-dev` variants are included. This adds a `provider` filter so a user can narrow the list to a single provider:

```python
client.functions(provider="q-ctrl")
# -> [QiskitFunction(q-ctrl/performance-management),
#     QiskitFunction(q-ctrl/performance-management-dev), ...]
```

## Changes

- **Gateway view** (`api/v1/views/programs/list.py`): reads the `provider` query param, passes it to the use case, and documents it in the OpenAPI schema.
- **Use case** (`api/use_cases/programs/list.py`): applies `provider__name` filtering **after** permission scoping across all three branches (serverless / catalog / default), so it can only narrow the set the caller may already see — never widen access. An unknown or inaccessible provider yields an empty list. Reuses the existing `FunctionsQuerySet` filtering; no manager change.
- **Client** (`core/clients/serverless_client.py`, `core/client.py`): adds an explicit `provider` argument to `functions()` and the deprecated `list()`, for discoverability in the signature/docstring/autocomplete. Backwards compatible (the param was already forwarded via `**kwargs`).

## Tests

- Use case: provider filter narrows to one provider; provider filter does not bypass permissions.
- Integration: `GET /programs?filter=catalog&provider=q-ctrl` returns only that provider's functions.
- Client: `functions(provider="q-ctrl")` forwards `?provider=q-ctrl`.

## Release note

`releasenotes/notes/list-functions-by-provider-*.yaml` (features).

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -602,13 +602,21 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
         return function_uploaded
 
     @_trace_functions("list")
-    def functions(self, **kwargs) -> List[RunnableQiskitFunction]:
-        """Returns list of available functions."""
+    def functions(self, provider: Optional[str] = None, **kwargs) -> List[RunnableQiskitFunction]:
+        """Returns list of available functions.
+
+        Args:
+            provider: if given, only functions belonging to this provider are returned,
+                e.g. ``functions(provider="q-ctrl")``.
+        """
+        params = dict(kwargs)
+        if provider:
+            params["provider"] = provider
         response_data = safe_json_request_as_list(
             request=lambda: requests.get(
                 f"{self.host}/api/{self.version}/programs",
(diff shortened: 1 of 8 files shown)


client/qiskit_serverless/core/clients/serverless_client.py
```diff
@@ -602,13 +602,21 @@ def upload(self, program: QiskitFunction) -> Optional[RunnableQiskitFunction]:
         return function_uploaded
 
     @_trace_functions("list")
-    def functions(self, **kwargs) -> List[RunnableQiskitFunction]:
-        """Returns list of available functions."""
+    def functions(self, provider: Optional[str] = None, **kwargs) -> List[RunnableQiskitFunction]:
```
ElePT, line 605: I preferred making the arg explicit for documentation purposes.

korgan00 (approved):
LGTM

PR #2435, opened 2026-08-27
Add a command to migrate old job business models to licensed

avilches:
### Summary

Adds `migrate_job_business_model`, the management command that rewrites the old `SUBSIDIZED` business model of existing jobs to `LICENSED`. It runs from the internal deployment repo as a Kubernetes Job through the existing `qiskit-serverless-migrations` chart, the same way the earlier `migrate_old_job_*` commands do.

Set as draft because it needs the `LICENSED` field from #2427, so it targets that branch instead of `main` until it merges.

### Details and comments

`--dry-run` only reports how many rows still hold `SUBSIDIZED`, without writing. Safe to interrupt and rerun, since each run only touches rows that still hold the old name.

Once every job row is migrated, a follow-up cleanup PR removes `core/domain/subsidized_license_mapping.py` and its two call sites, drops `SUBSIDIZED` from `BusinessModel`, and turns `BILLING_NAMES` back into an identity map.


ismaelRozasRamallal (dismissed):

riley2012 (approved):
LGTM!

PR #2436, opened 2026-08-27
Assign Code Engine projects per provider

marceloamaral:
### Summary

Lets a Code Engine project be dedicated to one or more Qiskit Function providers, so a provider's Fleets workloads get their own project — with their own subnet pool and COS buckets — instead of sharing the default project with everyone else. Multiple providers (e.g. `ibm` and `ibm-dev`) can be dedicated to the same project when isolation isn't needed between them.

### Details and comments

- **A provider can now have its own dedicated Code Engine project**: each provider (e.g. `ibm`, `ibm-dev`) can be linked to a specific project from Django admin. Several providers can share the same project when they don't need to be isolated from each other.
- **Functions get routed to the right project automatically:** when a Fleets function belongs to a provider, it now runs in that provider's dedicated project instead of the shared default one. Functions with no provider keep using the shared default, as before.
- **Uploading or running a function now fails clearly** if something's misconfigured, instead of silently defaulting or breaking later:
  - If a provider has no project assigned yet → clear error naming the provider.
  - If a provider's project exists but was deactivated → clear error naming that project.
  - This second check used to be missing when running a function (only when uploading it), so a function could still run against a project that had already been turned off. That's fixed now.
- **Errors are reported more consistently**: these configuration errors (and a couple of others, like uploading a function with no code) now use a dedicated error type that always returns a clean "400 Bad Request" response, instead of a generic validation error (⚠️ discussion in the PR around this ⚠️).
- **Local test setup updated**: to match the fleets test environment now links the test provider to its project explicitly (before, it matched by name automatically), since that link now lives on the provider record.


ElePT (commented):
The implementation looks good. But I think we should review whether it makes more sense to add a `provider` field to the project or viceversa. Thinking about cases where we may end up assigning two providers to the same project (for example, `ibm` and `ibm-dev`) for testing. cc: @Tansito

Tansito:
Discussed with @ElePT and I agree, I'm going to take a look into this one

Tansito:
Working on this PR, moving it to Draft until I finish

gateway/api/use_cases/programs/run.py
```diff
@@ -85,8 +85,18 @@ def execute(  # pylint: disable=too-many-locals
...
+            if not function.code_engine_project:
+                message = "Program has no Code Engine project assigned. Contact administrator."
+            elif not function.code_engine_project.active:
+                message = (
+                    f"Code Engine project '{function.code_engine_project.project_name}' assigned to "
+                    "this function is not active. Contact administrator."
+                )
+            if message:
+                logger.warning("user_id=%s program=%s | %s", user.id, function.title, message)
+                raise FunctionConfigurationException(message)
```
Tansito, line 99: In `upload` and `run` I'm using a new `FunctionConfigurationException` to manage the new errors for providers that don't have a Code Engine project assigned. Question here @avilches , @korgan00 , @ElePT : this `FunctionConfigurationException` is configured as a `400`, is this value correct or should I return in this case a `500` because is it an internal configuration?

gateway/api/use_cases/programs/upload.py
```diff
@@ -72,7 +89,7 @@ def _create(self, data: UploadFunctionInput, user, provider) -> Function:
         logger.info("user_id=%s program=%s | Creating function", user.id, data.title)
 
         if data.entrypoint is None and data.image is None:
-            raise DRFValidationError("At least one of attributes (entrypoint, image) is required.")
+            raise FunctionConfigurationException("At least one of attributes (entrypoint, image) is required.")
```
Tansito, line 92: For example, I think this `FunctionConfigurationException` as a `400` so maybe I should implement a different exception for the provider/project configuration.

gateway/api/use_cases/programs/run.py
ElePT: From what I have been seeing, I think that the 400 is correct because it's not a problem in the server, there is a "client" action required (the thing is, the client persona here is and admin, not any client).

releasenotes/notes/ce-project-per-provider-83dea96302b691d9.yaml
```diff
@@ -0,0 +1,17 @@
...
+    shared default project.
+upgrade:
+  - |
+    A Fleets function that belongs to a provider now requires a Code Engine project
+    dedicated to that provider. It is no longer given the default project as a fallback,
+    so uploading or running such a function fails until a project is assigned to that
+    provider in Django admin.
+  - |
+    Uploading or running a Fleets function whose Code Engine project has been deactivated
+    now fails with a clear message, instead of silently using it.
```
ElePT: Given that we haven't publicly documented fleets yet, we don't need the release note here.

ElePT (dismissed):
I think that the logic is clean and the assignment is more straightforward this way (having a project stored in the provider). On my side, LGTM, we'd have to test it in staging.

releasenotes/notes/ce-project-per-provider-83dea96302b691d9.yaml
ElePT: (I think this is non-blocking though)

Tansito: Ok! Let me remove it, no worries!

ElePT (approved):
LGTM

PR #2437, opened 2026-08-27
Read the Fleets task state without pinning the COS schema version

marceloamaral:
### Summary

Code Engine is moving the COS task-store layout from `v2/queue/` to `v3/queue/`. The gateway hardcoded `v2`, so a fleet on the new layout returns no keys and its job never finishes: `status()` returns `None`, the `None` branch skipped the timeout check, and no other scheduler task touches a PENDING or RUNNING Fleets job, so it held the user's concurrency slot indefinitely.

### Details and comments

The reader now tries v3 then v2 and takes the state from the first segment after `queue/`, so no version is pinned. Nothing else in the key is parsed, because segment counts and result codes differ per state (`succeeded/exit_0` against `failed/error_exit_1`).

An unrecognised state returns no status with a warning instead of raising. Raising marked the job FAILED irreversibly and emitted a billing event, so one rename upstream would have failed every in-flight job in a single scheduler tick.

Code Engine still creates v2 fleets today, so the fallback is the live path and this can merge ahead of the switch. Checked against real staging fleets on both layouts, and the fleets integration stack now writes v3 keys.

### DIFF
--- a/gateway/core/ibm_cloud/code_engine/fleets/cos.py
+++ b/gateway/core/ibm_cloud/code_engine/fleets/cos.py
@@ -21,22 +21,56 @@ from core.ibm_cloud.cos.cos_client import COSClient
 logger = logging.getLogger("FleetHandler")
 
 
-def queue_prefix(project_id: str, fleet_id: str) -> str:
+# Task-store schema versions, newest first. Code Engine writes the task state under a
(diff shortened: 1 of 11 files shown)


ElePT (approved):
Given that you have tested this offline and it seems to work, I think at this point, the best we can do is merge, deploy in staging, and test it in staging. So this is what I am going to try.

PR #2439, opened 2026-08-30
Fix backoffice dashboard status colors and bar chart

avilches:
### Summary
Several visual and rendering fixes to the Django admin ("backoffice") dashboard.

- `QUEUED` job status badges shared the same amber color as `PENDING`. `QUEUED` now gets its own purple, in both the job list and the job events timeline.
- The "Jobs by status" and "Jobs by provider" bar charts switched some rows (`FAILED`/`STOPPED`, and the `Custom` provider) to a lighter, hardcoded blue for no stated reason. All bars now use one consistent color.
- The bar widths, which are set via an inline `style="width:X%"` attribute, were always rendering at 100% regardless of the actual percentage. The site's CSP had no `style-src-attr` directive, so it fell back to `default-src: 'none'` and silently blocked every inline `style` attribute in the backoffice, not just this one. Added a scoped `style-src-attr: 'unsafe-inline'` directive (values here are always server-computed numbers, never user input).
- The per-app admin page (e.g. `/backoffice/api/`) was rendering the home dashboard (KPI cards, empty charts) instead of the app's model list, because Django's built-in `admin/app_index.html` extends `admin/index.html`, which this project overrides with the custom dashboard template. Added a dedicated `admin/app_index.html` template so per-app pages get their normal model list back.

### Details and comments
Verified the bar chart width fix and the app_index fix by rendering both pages with real data through Django's test client.

### DIFF
--- a/gateway/api/static/admin/css/admin_dashboard.css
+++ b/gateway/api/static/admin/css/admin_dashboard.css
@@ -116,10 +116,6 @@
     background: var(--primary);
 }
 
-.qs-bar-fill--dim {
-    background: #a6c8ff; /* blue-20, for Failed/Stopped */
-}
-
 .qs-bar-val {
     font-size: 0.6875rem;
     font-weight: 600;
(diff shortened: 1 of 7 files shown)


avilches:
Before:
image
image

After:
image
image

Tansito (approved):
Nice 🫡

PR #2440, opened 2026-08-30
Fix swagger schema generation and redirect root to swagger

avilches:
### Summary
- `/swagger/` returned a 500. Two different `OutputSerializer` classes (in `get_runtime_jobs.py` and `set_sub_status.py`) implicitly shared the same drf_yasg `ref_name`, which drf_yasg rejects once both endpoints are included in schema generation. Gave each an explicit, unique `ref_name`, following the naming convention already used by the other serializers in these view modules.
- The site root (`/`) returned a 404, there was no route for it. It now redirects to `/swagger/`.

### Details and comments
Added a regression test for each: one confirming `/` redirects to `/swagger/`, one confirming `/swagger/?format=openapi` returns 200 instead of the schema generation error.

### DIFF
--- a/gateway/api/v1/views/jobs/get_runtime_jobs.py
+++ b/gateway/api/v1/views/jobs/get_runtime_jobs.py
@@ -30,6 +30,7 @@ class OutputSerializer(serializers.ModelSerializer):
     class Meta:
         model = RuntimeJob
         fields = ["runtime_job", "runtime_session"]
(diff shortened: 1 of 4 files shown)


avilches:
Not found in /
image

Error in /swagger
image

Tansito (approved):
Thanks @avilches 🙏

PR #2441, opened 2026-08-30
Fix slow admin job list caused by N+1 program filter query

avilches:
### Summary
The "By Program" sidebar filter on the Job admin changelist loaded every matching job into Python and then ran one extra query per distinct program to build the filter options. With 198 distinct programs in production, that is 199 sequential DB round trips on every page load, which is what made the admin job list feel slow.

### Details and comments
Replaced the loop with three fixed queries: distinct `program_id`s referenced by the current queryset, an `exists()` check for the "Custom" bucket, and one `select_related` query for the matching `Program` rows. Verified against a local copy seeded with 100k jobs across 203 distinct programs: the old code issued 204 queries and took 2.35s, the new code issues 4 queries and takes 0.02s.

### DIFF
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -403,23 +403,12 @@ class JobProgramFilter(admin.SimpleListFilter):
     parameter_name = "job_program"
 
     def lookups(self, request, model_admin):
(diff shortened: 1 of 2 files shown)


ElePT (approved):
LGTM

PR #2442, opened 2026-08-31
Add Fleets-specific per-user concurrent job limit

ElePT:
## Summary

The per-user concurrent running-job cap `LIMITS_JOBS_PER_USER` (default **2**) was:
- **shared across all runners**, and
- counted a user's running jobs **without regard to runner**.

Fleets (Code Engine) scales differently from Ray, it has no per-cluster capacity, so this limit is artificial and not necessary.

## Changes

- **New setting** `LIMITS_JOBS_PER_USER_FLEETS` (env var, default **50**) in `gateway/main/settings.py`.
- **`get_jobs_to_schedule_fair_share` is now runner-aware** (`gateway/scheduler/schedule.py`):
  - picks the per-user limit by runner (Fleets → `LIMITS_JOBS_PER_USER_FLEETS`, everything else → `LIMITS_JOBS_PER_USER`), and
  - scopes the running-jobs tally to the same runner, so Fleets and Ray per-user budgets are tracked **independently**.
- **Tests** added in `gateway/tests/scheduler/test_schedule.py` covering (a) a user's Ray jobs no longer blocking their Fleets jobs, and (b) Fleets using the new higher limit.

Ray behavior is unchanged (still default 2, now counted only against Ray jobs).

## For staging

Set `LIMITS_JOBS_PER_USER_FLEETS` on the scheduler, or rely on the new default of 50.

### DIFF
--- a/gateway/main/settings.py
+++ b/gateway/main/settings.py
@@ -327,6 +327,10 @@ FLEETS_GATEWAY_HOST = os.environ.get("FLEETS_GATEWAY_HOST", SITE_HOST)
 
 # resources limitations
 LIMITS_JOBS_PER_USER = int(os.environ.get("LIMITS_JOBS_PER_USER", "2"))
+# Per-user concurrent running-job limit for the Fleets (Code Engine) runner.
+# Fleets scales differently from Ray (no per-cluster capacity), so it gets its
+# own, higher default independent of the Ray-oriented LIMITS_JOBS_PER_USER.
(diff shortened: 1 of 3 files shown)


avilches:
I agree on having a different limit per user + runner, so a user can run 2 jobs in Ray and more jobs in Fleets. But I feel 50 jobs is a super high number. GPU nodes take a lot of time to spin up, so a greedy user could hog the queue for hours...

avilches (approved):

avilches:
Let's having a different limit for runner, then we can play with the numbers laters. Approved!

PR #2354, opened 2026-07-21
Remove client name

ItamarGoldman:
### Summary
Removing client name that had deprecation warning in v0.33.
This is change was planned in Phase 2 in the internal issue ([link](https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1516)).

### Details and comments
The following changes were done:

- [x] Removing `name` attribute and deprecation warning from `ServerlessClient` and `BaseClient`.
- [x] Changing `BaseClient.__eq__` and `BaseClient.__repr__`.
- [x] Updating `ServerlessClient.from_dict(...)` for backwards compatibility.
- [x] Write release notes and do a migration guide since this is a breaking change.

**Note(documentation):**
There is no need to update documentation since there is no usage of the name attribute.

**Note(`IBMServerlessClient` class):**
The name attribute in `IBMServerlessClient` is used to load an account by its name and it has no connection to the client name that is deleted here.

### Checklist

- [x] ~~I have added tests to cover my changes~~ - **no need to test** .
- [x] I have updated the documentation accordingly.
- [x] I have added a [release note](https://github.com/Qiskit/qiskit-serverless/blob/main/CONTRIBUTING.md#release-notes) for any user-facing change (new feature, bug fix, deprecation, or breaking change), or this PR has no user-facing changes.

PR #2283, opened 2026-06-30
Deprecation notice on widget and client name attribute

ItamarGoldman:
### Summary

Deprecation notice on the Widget and Name attribute of client as preparation to their removal.

### Details and comments

**Goal**: Warn users that the `name` attribute will be removed in a future release.

### 1.1 Add warning to `BaseClient.__init__()`

**File**: [`client/qiskit_serverless/core/client.py:77-81`](client/qiskit_serverless/core/client.py:77-81)

```python
if name:
    warnings.warn(
        "The 'name' attribute is deprecated and will be removed in a future release.",
        DeprecationWarning,
        stacklevel=2
    )
```

#### 1.2 Update docstrings

Mark `name` parameter as deprecated in:
- [`client/qiskit_serverless/core/client.py:67-76`](client/qiskit_serverless/core/client.py:67-76)
- [`client/qiskit_serverless/core/clients/serverless_client.py:112-122`](client/qiskit_serverless/core/clients/serverless_client.py:112-122)

Widget is also deprecated in the client.
The whole visualization folder is going to be deleted.

PR #2443, opened 2026-08-31
Remove version on client

korgan00:
### Summary

Remove version parameter on client since is not used at this moment.

### Checklist


ElePT (changes requested):
While I am pretty sure nobody was using this feature, we are trying to implement stricter rules on deprecations and removals. This means sticking to a deprecation period before submitting the removal PR as we did, for example, with https://github.com/Qiskit/qiskit-serverless/pull/2283 (deprecation) and https://github.com/Qiskit/qiskit-serverless/pull/2354 (removal). I think we should do the same here. A new PR first deprecating the argument (with its respective release note) + a follow-up on a future release with this removal (and a release note).

PR #2444, opened 2026-08-31
Add a Timeline action to the job admin

avilches:
### Summary

Adds a "Timeline" action to the Job admin's changelist: select one or more jobs, run the action, and land on a page showing a Gantt-style chart of those jobs, one row per job, colored by the state it was in during each segment of its lifetime (QUEUED, PENDING, RUNNING), with a concurrency chart above it and a job details panel below.

The point is to make it possible to see, at a glance, how a batch of jobs overlapped in time and how long each one spent in each state, without having to piece that together from the job list or the raw event history by hand.

image

### Details and comments

- The rendering logic lives in a new pure module, `gateway/api/domain/job_timeline.py`, which turns a `Job` queryset (with `job_events` prefetched) into ready to render HTML/SVG strings. The admin wiring (the action, the URL, the view) is in `gateway/api/admin.py`, and the page is `gateway/templates/admin/api/job/timeline.html`.
- Colors and status names match the job list's own status badges (QUEUED purple, PENDING orange, RUNNING light blue, SUCCEEDED/FAILED/STOPPED in the same green/red/gray), so a status reads the same everywhere.
- A job's own duration in each state is drawn inside its segment when there is room, and moves outside the bar (next to the final status) when the segment is too narrow.
- Overlap is scoped per runner: a Ray job and a Fleets job with the same running window are never flagged as overlapping each other, since they run on separate infrastructure. Clicking a job highlights it and every job it overlaps with, all the same way, for as long as its details panel stays open.
- There is a filter by runner (hides non-matching rows) and by compute profile (dims non-matching rows), and both combine.
- Jobs still in a non-terminal state (still queued, pending or running) have their last known segment extended to "now" instead of showing zero duration, and their outcome label falls back to their current status in that status's own color, rather than showing nothing or an unrelated color.
- The whole page, including the chart canvas, follows the admin's own light/dark theme toggle through CSS variables. Status colors themselves stay fixed regardless of theme, matching the job list's badges.
- The project's Content Security Policy forbids inline `<script>` blocks and inline `style` attributes, so all the page's behaviour lives in an external static file (`gateway/api/static/admin/js/job_timeline.js`) and visibility is driven by CSS classes rather than inline styles.
- The home dashboard now also embeds a timeline of the 20 most recent Fleets jobs (Ray excluded), right below the existing "Jobs by status" / "Jobs by provider" panels, reusing the same template fragment as the full page.
- Clicking a job row shows a details panel with the job's id, author, business model, account id and instance CRN, plus a separate Fleets section (fleet id, compute profile, CE project name, region).
- A job's selected id list travels in the action's redirect query string, capped at 200 jobs so a "select all" on a large changelist can't build a query string long enough for the webserver to reject outright.

### Test plan

- [x] `pytest tests/` passes (1039 passed, 2 skipped)
- [x] `pylint -rn api main core scheduler` rates 10.00/10
- [x] `lint-imports` keeps all 3 import boundary contracts
- [x] `black --check` is clean
- [x] Manually exercised against a local Postgres with a mix of Ray and Fleets jobs in different states (queued, pending, running, succeeded, failed, stopped, and overlapping running windows), in both light and dark admin theme


ElePT (approved):
As Spanish people say, an image is worth more than 1000 words. In this case I can definitely trust your screenshots and demos more than 1000 lines of code. LGTM

PR #2447, opened 2026-09-01
Compute Profile validation refactor

Tansito:
### Summary

This PR unifies the compute profile validation and formatting logic that used to be duplicated, and somewhat out of sync, between the API view and the job-execution logic. The view now validates the input format and normalizes it to a single canonical form in one step, and the rest of the code always works with that already-normalized value. The internal normalization logic is kept, but only as a safety net for jobs created before this change (it could removed after).

### Details and comments

- Previously there were two independent definitions of the compute profile format: one in the view, to validate what the client sends, and another in the job-execution logic, to strip an optional prefix (e.g. "bx3d-24x120" → "24x120"). Living in two separate places, they could drift apart from each other without it being obvious. Now a small shared domain module was created with two functions: one that checks whether a value has an acceptable format (with or without prefix), and another that normalizes it to the prefix-less form. Both share the same definition of "what counts as a prefix", so they can no longer drift apart.
- The view now does two explicit steps when it receives a request: first it validates the format, and if valid, normalizes it before it reaches the rest of the system. Previously that normalization happened later, inside the business logic.
- Since normalization now happens at the entry point, the business logic no longer normalizes on its own: it trusts that the value it receives is already in canonical form.
- The logic that provisions cloud resources for each job still normalizes the value right before using it, but now only as a safety net for jobs that were still queued right at deploy time and get processed by the new code. In the normal flow, that normalization should no longer do anything, since the value already arrives correct from the source.
- There is no visible behavior change for API consumers: formats that were previously accepted are still accepted and stored the same way. The change is purely an internal code reorganization.

### Checklist


ElePT (approved):
LGTM. I did find a small inconsistency that was already there, but I don't think we should block the PR for it. `_RESOURCE_RE` accepts a prefix with only numbers, while `_parse_compute_profile` in `fleets_runner.py` requires the prefix to start with a letter, so a profile would be considered valid first and then rejected. Again, it's a small detail and doesn't really affect functionality.

PR #2448, opened 2026-09-01
Add sizes() and get_default_size() to functions, carried in the function body

ElePT:
## What this does

Lets a function developer discover the sizes a function declares, and its default size, before choosing one for a run. This closes item 1 of the "what's left" list on the T-shirt sizes PR (#2431): `RunnableQiskitFunction.sizes()` / `get_default_size()`.

The size catalog is trivial data (a map of strings and a string), so it rides along in the function representation rather than behind a dedicated endpoint. The client exposes it through methods, and no new client or gateway operation is added.

Built on top of `t_shirt_sizes_api`, so this PR targets that branch.

## Client

- `RunnableQiskitFunction.sizes()` returns the declared catalog as `{size_label: compute_profile}`, empty when the function declares no sizes.
- `RunnableQiskitFunction.get_default_size()` returns the default size label, or `None` when the function declares no default (a run then falls back to the platform default).
- Both read the already-fetched representation, so neither makes a request. Re-fetch the function (`client.function(...)`) to pick up sizes changed after it was loaded.
- The data arrives on the function under the gateway's `sizes` key. `from_json` remaps it onto the `sizes_map` field so the field does not collide with the `sizes()` accessor.

## Gateway

- `sizes` and `default_size` are carried in the function representation, not behind an endpoint.
- A shared `SizesFieldsMixin` adds read-only `sizes` and `default_size` to the `get_by_title` and `list` serializers.
- The upload response renders the stored catalog through `to_representation`, so a caller sees back what it just stored.
- `sizes` is `{label: compute_profile_id}`; `default_size` is the default row's label, or null when the function declares no default.

This is read-only over the existing `FunctionSize` catalog and `Program.default_size`, so there is no model change and no migration.

## Why methods and not plain attributes

The data is a snapshot from fetch time, so a method is honest about staleness in a way a bare attribute is not, and it leaves room to re-fetch later without a client API change. The methods also normalize the empty-vs-absent case: `sizes()` returns `{}` rather than `None` when the representation carried no catalog.

## Tests

- Gateway: `get_by_title` and `list` carry the catalog and default (and report empty/null when none is declared); the upload response renders the stored catalog.
- Client: `sizes()` / `get_default_size()` read from the representation, empty and absent cases, `function(...)` reads the values off a `get_by_title` response, and `from_json` remaps the `sizes` key onto `sizes_map`.

All green: 163 gateway program/serializer tests and 44 client function tests pass, pylint 10/10 both sides, black clean, import contracts kept.

## Not included

- No release note (reno) yet.
- Billing integration is out of scope.


client/qiskit_serverless/core/function.py
```diff
@@ -82,6 +82,10 @@ def _decode_arguments_schema(raw_schema: Any) -> Optional[Union[Dict[str, Any],
 
 def _decode_fields(data: Dict[str, Any], field_names: set) -> Dict[str, Any]:
     """Keep only known dataclass fields and decode the ones the gateway sends as text."""
+    # The gateway sends the catalog under "sizes"; the field is sizes_map so it does not
+    # collide with the sizes() accessor. Remap before the field-name filter drops it.
```
ElePT, line 86: I am not 100% sold on `sizes()` as a method name, so feel free to suggest alternatives for the naming.

avilches (dismissed):
LGTM

Tansito:
@avilches I just request your approval again because it was going to be merged due to the stack thing. I will review it now 🙏

Tansito (approved):

PR #2450, opened 2026-09-01
Filler GPU nodes 1/2

avilches:
### Summary

Adds the pieces the filler jobs feature needs, without anything that uses them yet, so this PR changes no behaviour on its own. The scheduler task that starts and stops filler jobs comes in a second PR.

### Details and comments

1. Three new dynamic config values, so the feature can be turned on, off and retuned without a deploy: `scheduler.filler.enabled` (off by default), `scheduler.filler.program_id` and `scheduler.filler.slots`. The compute profile is read from the filler program itself, through its default size.

2. A `FILLER_AUTHOR_USERNAME` environment variable holding the username of the internal user set as the author of every filler job.

3. A `filler` boolean column on the job table, false by default, set to true when a filler job is created.

4. Two new job event contexts, `FILLER_SUBMIT` and `FILLER_STOP`, for the audit trail the second PR will write.

5. Filler jobs are excluded from billing and usage reporting.

6. Fleets are now named `{job|fil}-{function}-{compute profile}-{username}-{YYMMDDHHMMSShh}` instead of carrying the job id, so a fleet can be read at a glance in Code Engine: what it runs, on which profile, for whom and when. Filler jobs are told apart by their prefix, and both prefixes are three characters so the name has the same room either way. Each segment is lowercased, stripped of characters Code Engine rejects and capped at 14 characters, which lands the longest name exactly on 63. The name is still not an identifier: `fleet_id`, assigned by Code Engine, stays what the gateway stores and queries.

It also raises the Django floor from 5.2.17 to 6.0.8. Django 6 requires Python 3.12, which the gateway already targets, so the tox environments now declare it explicitly. The migration was regenerated under Django 6.0.8 and its content is unchanged.


gateway/core/ibm_cloud/event_streams/abstract_event_streams_client.py
```diff
@@ -12,24 +12,59 @@
...
+            return
+        self._emit_job_in_progress(job, metric_type)
 
-    @abstractmethod
     def emit_job_completed(self, job, metric_type: str | None = None) -> None:
         """Publish or log a function_job_completed event for the given metric."""
+        if job.filler:
+            logger.debug("job_id=%s filler job, skipping emit_job_completed", job.id)
+            return
+        self._emit_job_completed(job, metric_type)
```
avilches, line 47:
```suggestion
        if job.filler:
            logger.info("job_id=%s filler job, skipping emit_job_started", job.id)
            return
        self._emit_job_started(job, metric_type)

def emit_job_in_progress(self, job, metric_type: str | None = None) -> None:
"""Publish or log a function_job_in_progress event for the given metric."""
        if job.filler:
            logger.info("job_id=%s filler job, skipping emit_job_in_progress", job.id)
            return
        self._emit_job_in_progress(job, metric_type)

def emit_job_completed(self, job, metric_type: str | None = None) -> None:
"""Publish or log a function_job_completed event for the given metric."""
        if job.filler:
            logger.info("job_id=%s filler job, skipping emit_job_completed", job.id)
            return
        self._emit_job_completed(job, metric_type)
```

gateway/core/ibm_cloud/event_streams/abstract_event_streams_client.py
```diff
@@ -12,24 +12,59 @@
...
 
-    @abstractmethod
     def emit_license_fee(self, job) -> None:
+        """Publish or log a license fee event."""
+        if job.filler:
+            logger.debug("job_id=%s filler job, skipping emit_license_fee", job.id)
```
avilches:
```suggestion
            logger.info("job_id=%s filler job, skipping emit_license_fee", job.id)
```

gateway/tox.ini
```diff
@@ -12,6 +12,10 @@ skip_missing_interpreters = true
 [testenv]
 envdir = {toxworkdir}/shared
 runner = ignore_env_name_mismatch
+# The gateway is Python 3.12 only, and Django 6 requires it. Without this, every env
+# other than py312 falls back to whichever interpreter tox itself runs on, which can
+# be older and cannot install the pinned Django.
+basepython = python3.12
```
avilches:
```suggestion
basepython = python3.12
```

gateway/tox.ini
```diff
@@ -12,6 +12,7 @@ skip_missing_interpreters = true
 [testenv]
 envdir = {toxworkdir}/shared
 runner = ignore_env_name_mismatch
+basepython = python3.12
```
avilches, line 15: The gateway is Python 3.12 only, and Django 6 requires it. Without this change, every env other than py312 falls back to whichever interpreter tox itself runs on, which can be older and cannot install the pinned Django.

Tansito (dismissed):

Tansito:
I think we just need to regenerate the db migration

Tansito (approved):

PR #2454, opened 2026-09-02
Send sizes_map and default_size when uploading a function

ElePT:
## Summary

Completes the client half of the t-shirt sizes API by sending a function's
size catalog **on upload**.

The client already carried a function's catalog on the `sizes_map` and
`default_size` fields — but only when *reading* a function back (populated by
`from_json`, exposed via `.sizes()` / `.get_default_size()` from #2448). The
upload payload builders dropped them, so constructing a `QiskitFunction` with
`sizes_map` / `default_size` and calling `upload()` **silently discarded** the
catalog; it could only be declared through the raw gateway API or the admin.

This PR sends both on upload:
- `sizes_map` under the gateway's expected key `sizes` (a JSON object)
- `default_size` as a string

Each is omitted when unset, so an upload that does not touch sizes leaves any
stored catalog untouched (the gateway treats an absent field as "not sent").
`from_json` already remaps the gateway's `sizes` back onto `sizes_map`, so a
fetched function still round-trips.

```python
function = QiskitFunction(
    title="my-function",
    entrypoint="entrypoint.py",
    working_dir="./src/",
    runner="fleets",
    sizes_map={"s": "24x120", "m": "16x128"},
    default_size="m",
)
serverless.upload(function)          # now actually stores the catalog

job = function.run(function_size="s")  # (already worked before this PR)
```

## Why

Run-time selection (`run(function_size=...)`) already shipped, and the gateway
`ProgramSerializer` has always accepted `sizes` / `default_size`. Only the
client's upload path was missing, so there was no way to declare a catalog from
Python. A user hit this while validating the feature: `sizes_map` was accepted
by the constructor but never reached the gateway.

## Changes

- `_sizes_payload()` helper + wiring into both upload builders
  (`_upload_with_docker_image`, `_upload_with_artifact`).
- Docstrings for `sizes_map` / `default_size` updated to note they can now be
  set on upload, not only read back.
- Release note.

## Testing

- [x] Local runtime check: `sizes_map`/`default_size` serialize into the upload
      payload (as `sizes`/`default_size`) only when set, and `from_json`
      round-trips the gateway's `sizes` key back onto `sizes_map`.


Tansito (approved):
LGTM thanks Elena! 😄

PR #2455, opened 2026-09-02
Resolve Code Engine subnet pool id from a configured name

ElePT:
> [!NOTE]
> **Live CE call confirmed (2026-09-11).** The name→id lookup calls `GET /projects/{id}/subnet_pools`. This was run live against an active staging Code Engine project (`us-east`) and the deployed API serves the endpoint, returning `id` + `name`. `FleetHandler.resolve_subnet_pool_id` resolved a real pool name to its id end to end, and the not-found path raised as expected. See the Verification section.

## Summary

A Code Engine project can now be configured with a subnet pool **name** (`subnet_pool_name`) instead of its id. The IBM Cloud console shows a pool's name readily but not its id, so requiring the id added friction to every project onboarding.

When only a name is configured, the gateway resolves it to the id against the CE API on the **first job submission** for that project and **caches** the id back onto the `CodeEngineProject` row, so later jobs reuse it without another lookup. Supplying `subnet_pool_id` directly still works and skips resolution.

Because a pool name is **not guaranteed unique** within a project (the CE API only *advises* uniqueness; pools are identified by id), a name matching more than one pool **fails the submission** with an error naming the matching ids rather than guessing which pool to place the fleet on.

## Changes

- **Vendor** `SubnetPoolsApi.list_subnet_pools` + the `V2SubnetPool`/`V2SubnetPoolList` models into the gateway `ce_client` (only `list` is needed; reuses the already-vendored pagination models).
- **`FleetHandler.resolve_subnet_pool_id`** — pages through the project's pools, matches by exact name, raises on zero or multiple matches.
- **`FleetsRunner.submit()`** — resolve-and-cache via a new `_ensure_subnet_pool_id`; `subnet_pool_id` made nullable and `subnet_pool_name` added (migration `0062`).
- **`sync_ce_project`** — accepts either field, requires at least one, and invalidates a cached id **only** when the configured name changes (so re-sync doesn't wipe the cache every boot, given `update_or_create` overwrites all defaults).

## Verification

- **Live CE API (2026-09-11):** ran `list_subnet_pools` against an active staging project in `us-east` (equivalently `ibmcloud ce connectivity subnetpool list`, the same endpoint). Confirmed: the deployed API serves the endpoint; every returned pool carries both `id` and `name`; `resolve_subnet_pool_id(<name>)` returned the matching id end to end; the not-found path raised `ValueError` naming the available pools.
- `158/158` tests pass across `test_fleets_runner`, `fleets/` handler tests, and `test_sync_ce_project` (14 new). Covers: id-in-config skips resolution; name-only resolves once + caches; reuse without a second call; ambiguity/not-found raise; pagination; sync accept/reject and rename-invalidation.
- black ✅ · pylint 10.00/10 ✅ · import-linter ✅ · `makemigrations --check` clean ✅

## Out of scope (follow-up)

Resolving `project_name`→`project_id` and `resource_group_name`→`resource_group_id`. They're not symmetric with subnet pool: `project_id` is the row's primary key needed *before* a CE client exists (no lazy-at-submit hook), and `resource_group_id` lives in a different service (Resource Manager). Both should share the fail-loud-on-ambiguity contract used here.


marceloamaral (commented):
Looks like a good feature to have as we discussed it before!

ElePT:
Hi @marceloamaral, I managed to run the verification for this PR so in principle it should be good to merge and test in staging. I updated the PR description to reflect this.

marceloamaral:
**Update**: resolving the name when the project is seeded instead of when a job runs

**Context**: this PR let a Code Engine project be configured with a subnet pool name instead of a uuid. That part is unchanged and is what makes onboarding easier. What changed is when the name gets turned into an id.

**Updates**:

1. The name is resolved when the project is seeded (`sync_ce_project`), not on the first job submission. If a name is wrong, one sync fails and the person who just typed it finds out. Before, a typo showed up later as users' jobs failing, and the reason only reached the gateway log.
2. No new database column, and no migration. Only the id is stored, as before. The name lives in the config.
3. The job submit path is untouched. No lookup and no database write while submitting, for a value that changes about once a year.
4. A configured `subnet_pool_id` is still honoured, as a fallback. If a name cannot be resolved, the configured id is used and jobs keep running. An entry with no name makes no calls to IBM Cloud at all, so nothing changes for the projects we run today.
5. The sync now fails when an entry cannot be synced. It used to log the problem and report success, so a half-seeded database looked fine.
6. New `--dry-run` flag that resolves every name and reports, without writing. It is deliberately stricter than a real run: an unresolvable name fails even when a fallback id is configured, so a clean dry run really does mean every name resolves. That makes it safe to prove this works in staging before anything depends on it.
7. A recreated pool is picked up by re-running the sync, because resolution triggers on the name being configured rather than on a missing id. Previously a pool deleted and recreated under the same name would have kept pointing at the old one forever.

How to adopt it safely: add `subnet_pool_name` next to the existing `subnet_pool_id`, run a -`-dry-run sync` to confirm it resolves, then remove the id.

**Tested**: full gateway suite, the fleets integration stack, and live against the real Code Engine API in staging, including the failure paths (bad name, missing credential, network timeout).

releasenotes/notes/fleets-subnet-pool-name-73c596ed166536e8.yaml
```diff
@@ -0,0 +1,24 @@
...
+    unresolvable or ambiguous name is reported where it is applied instead of surfacing
+    later as a failed job, and it takes a ``--dry-run`` flag that resolves and reports
+    without writing. Resolving a name needs ``IBM_CLOUD_API_KEY``; without it, entries
+    that configure a name fail the command before anything is written.
+
+    Two notes for operators. A subnet pool deleted and recreated under the same name is
+    picked up by re-running the sync, since resolution triggers on the configured name
+    rather than on a missing id. And when adopting a name, add it alongside the existing
+    ``subnet_pool_id`` first: the id then acts as the fallback while the lookup proves
+    itself, and can be removed afterwards.
```
ElePT: Fleets is still not publicly documented so I would remove the release note

ElePT (commented):
Thanks @marceloamaral! The updates look good to me, I think it's better to fail during the sync than when jobs are submitted. Good catch. My only comment is the release note. Apart from that, LGTM

releasenotes/notes/fleets-subnet-pool-name-73c596ed166536e8.yaml
marceloamaral: Thanks @ElePT for the review! I've removed it!

marceloamaral (approved):
LGTM

PR #2458, opened 2026-09-02
Validate the compute profile settings at startup and index filler job lookups

avilches:
### Summary

Two independent preparatory changes pulled out of the filler jobs balancer work so they can be reviewed on their own.

The compute profile settings are now validated and normalized when the process starts. `DEFAULT_COMPUTE_PROFILE` and `DEFAULT_FUNCTION_SIZE_PROFILE` are read through one helper that strips an instance family prefix and raises `ImproperlyConfigured` when the value is empty or is not a compute profile at all.

`api_job` gets a partial index on `created` over the filler rows, which is the lookup the scheduler will run once per second once the balancer lands.

### Details and comments

On the settings. The request path already validates and normalizes what a client asks for: the serializer in `gateway/api/v1/views/programs/run.py:65` runs both `compute_profile.is_valid` and `compute_profile.normalize`. The configuration path ran neither, and the value is read raw in three places outside tests. `gateway/core/services/runners/fleets_runner.py:798` puts it straight into the Code Engine fleet name, `fleets_runner.py:813` gets away with it because `_parse_compute_profile` normalizes again a few lines below, and `gateway/api/use_cases/programs/run.py:164` stores it on the job as the fallback profile. So a prefixed value such as `bx3d-24x120` produced non canonical profiles on jobs and a prefix in fleet names, and each new reader had to remember to normalize.

`DEFAULT_FUNCTION_SIZE_PROFILE` is worse, because `gateway/api/use_cases/programs/upload.py:270` uses it directly as a `ComputeProfile` primary key. A prefixed or empty value there matches no row, so the function is created with no sizes at all and the only trace is a warning in the log. Validating both at import turns that into a startup failure naming the variable.

The chart also renders `DEFAULT_COMPUTE_PROFILE` with no default, so an explicitly empty chart value produced an empty environment variable. Both renderings now carry `| default "16x128"`, which keeps the new startup check from being reachable through an omitted chart value.

On the index. The scheduler loop runs about once a second (`gateway/scheduler/main.py:104`) and the balancer's query is `WHERE filler = true AND status IN (...) ORDER BY created` against `api_job`, the fastest growing table in the schema. A plain index on `created` cannot serve it: the query needs every matching row, so Postgres would walk the whole index checking `filler` on each heap row and would be better off with a sequential scan. A partial index is sized by the rows that match the predicate rather than by the table, so it is a handful of rows in a single page, and an insert that does not match the predicate does not write to it. That is why it is keyed on `created` with the predicate on `filler`, rather than the other way round.

Worth flagging for whoever applies this: Django emits a plain `CREATE INDEX`, which holds an ACCESS EXCLUSIVE lock on `api_job` for the length of a full heap scan and blocks writes while it builds. The release note says to create it by hand with `CONCURRENTLY` and apply the migration with `--fake` where the table is large.

The filler jobs balancer compares the profile derived from the filler program against the profile stored on each job, so it needs both sides in canonical form. That is why this goes in first, ahead of the balancer PR.


gateway/main/settings.py
```diff
@@ -503,14 +504,41 @@
...
+    value = os.environ.get(name, default)
+    normalized = compute_profile.normalize(value)
+    if normalized is None or not compute_profile.is_valid(normalized):
+        raise ImproperlyConfigured(f"{name} is not a compute profile: {value!r}")
+    return normalized
+
+
 # Compute profile settings for Fleets runner
-DEFAULT_COMPUTE_PROFILE = os.environ.get("DEFAULT_COMPUTE_PROFILE", "16x128")  # 16 CPU, 128GB RAM
+DEFAULT_COMPUTE_PROFILE = _canonical_compute_profile("DEFAULT_COMPUTE_PROFILE", "16x128")  # 16 CPU, 128GB RAM
```
Tansito, line 533: Good catch @avilches

Tansito (commented):
LGTM @avilches

Tansito (approved):

PR #2460, opened 2026-09-02
Filler GPU nodes 2/2

avilches:
### Summary

The scheduler can now keep a scarce Fleets compute profile occupied. When real demand does not fill the profile of a designated filler program, a new scheduler task creates filler jobs to make up the difference, and stops them again as soon as real jobs need the capacity. This is the second half of the feature: the `Job.filler` column, the config keys and the billing exclusion went in with #2450.

It is off by default. `scheduler.filler.enabled` defaults to false and `scheduler.filler.slots` to zero, and either one alone is enough to keep it off. The feature also stays off until the user row that owns filler jobs exists, so merging this changes nothing until an operator turns it on.

### Details and comments

The protected compute profile is not configured directly. It is derived from the filler program's default size, so it cannot drift away from the program it belongs to. A running filler job is stopped as soon as it stops matching both the configured program and that program's profile, which is how re-pointing either one drains the jobs left behind.

Filler jobs skip the queue that `ScheduleFleetsJobs` feeds, because going through it would put them in competition with real queued jobs for the free Fleets slots. The task creates at most one filler job per iteration, so a four slot profile fills in about four seconds while the object storage upload and the Code Engine submit that each one costs stay off the critical path of the status updates that share this single threaded loop.

Nothing stops the feature on its own. A filler function is meant to run until something needs its slot, and one that exits by itself is replaced on the next loop indefinitely, so that condition is reported through metrics rather than handled in code. This PR adds `scheduler_filler_profile_slots`, the slot target of the protected profile, and `scheduler_filler_profile_jobs`, its occupancy split by a `kind` label into real and filler jobs. Their sum staying below the target for several minutes is the alert that says the feature is not doing its job. The balancer also counts what it does, on `scheduler_filler_jobs_created_total` and `scheduler_filler_jobs_stopped_total`. The counter that catches a filler function exiting by itself is `scheduler_filler_jobs_ended_total`, which ships in #2462 together with keeping filler jobs out of the series that describe user demand.

That PR and this one can merge in either order and neither depends on the other. If this one lands first, filler jobs count as user demand in `scheduler_job_status_count` until #2462 follows. The window is harmless in practice, because the feature ships off and no filler job exists until an operator turns it on.

Before turning it on: a `User` row whose username matches `FILLER_AUTHOR_USERNAME`, a Fleets program with a default size and an active Code Engine project, and then the three config keys. Any of those missing leaves the feature off and logs the reason. Two behaviours are worth knowing about as well. Maintenance mode stops running filler jobs and not just new ones, so the platform can be drained, and they come back a second after maintenance ends. And filler jobs count against `LIMITS_MAX_FLEETS` with the configured slot count honored literally, so it has to be chosen against that budget.


gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,416 @@
...
+        current = [job for job in filler_jobs if (job.program_id, job.compute_profile_fk_id) == expected]
+        # Reported after the split and before this loop acts, so the gauge is the
+        # occupancy the decisions below were taken on.
+        self.metrics.set_filler_profile_jobs(len(current), "filler")
+
+        if stale:
+            logger.debug(
+                "[BalanceFillerJobs] stopping %s filler job(s) on another program or compute profile", len(stale)
+            )
+        self._stop_filler_jobs(stale)
```
avilches:
```suggestion
            self._stop_filler_jobs(stale)
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,416 @@
...
+        if not value:
+            return self._deactivated(f"{ConfigKey.FILLER_FUNCTION.value} is empty")
+
+        programs = Program.objects.select_related("default_size__compute_profile", "code_engine_project")
+        try:
+            if "/" in value:
+                # The provider/title convention already has a parser in api.utils
+                # (parse_title_and_provider), but the import-linter contracts forbid
+                # scheduler from importing api, so it is split by hand here. The
+                # duplication is deliberate.
```
avilches:
```suggestion
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,416 @@
...
+                parts = value.split("/")
+                if len(parts) != 2 or not all(parts):
+                    return self._deactivated(
+                        f"{ConfigKey.FILLER_FUNCTION.value} is {value!r}, expected provider/title or an id"
+                    )
+                # get(), not filter().first(): the unique_provider_title constraint on
+                # (provider, title) makes at most one row match, so there is no ambiguity
+                # to resolve here and MultipleObjectsReturned cannot happen.
```
avilches:
```suggestion
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,416 @@
...
+            return programs.get(id=value)
+        except Program.DoesNotExist:
+            return self._deactivated(f"function {value} does not exist")
+        except ValidationError:
+            # Only the id branch reaches this. Program.id is a UUIDField, so a value
+            # that is not a UUID raises ValidationError while the field is cleaned,
+            # before any query runs, and never DoesNotExist. Measured on this code:
+            # get(id="not-a-uuid") raises ValidationError('"not-a-uuid" is not a valid
+            # UUID.'). Without this the scheduler would log a traceback once a second
+            # instead of deactivating and saying which value is wrong.
```
avilches:
```suggestion
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,416 @@
...
+            return self._deactivated(f"{ConfigKey.FILLER_FUNCTION.value} is {value!r}, which is not a function id")
```
avilches:
```suggestion
            return self._deactivated(f"{ConfigKey.FILLER_FUNCTION.value} is {value!r}, which is not a valid uuid")
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,319 @@
...
+from scheduler.metrics.scheduler_metrics_collector import SchedulerMetrics
+from scheduler.schedule import execute_fleets_job
+from .task import SchedulerTask
+
+logger = logging.getLogger("scheduler.BalanceFillerJobs")
+
+# Loops to skip after a failed creation. This task keeps no memory of what it tried, so
+# a failure leaves the world unchanged and the next loop would repeat the same attempt a
+# second later, forever. Counted in loops, not seconds, and only decremented on loops
+# that reach the creation path, so the delay pauses while real jobs fill the slots.
```
avilches:
```suggestion
# Loops to skip after a failed creation to avoid create a failed job per second if CE is down.
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+class BalanceFillerJobs(SchedulerTask):
+    """Keep real plus filler jobs on one compute profile at the configured minimum.
+
+    The compute profile is derived from the filler program's default size rather than
+    configured on its own, and a filler job belongs to the feature only while it matches
+    both the configured program and that profile, so re-pointing either one drains the
+    jobs left behind. Filler jobs are submitted directly instead of through the queue
+    ScheduleFleetsJobs feeds, which would put them in competition with real queued jobs.
```
avilches:
```suggestion
    - The compute profile is derived from the filler program's default size.
    
    - A filler job belongs to the feature only while it matches the program.id and that profile
    Changing the filler function or the compute profile will stop all the previous job filler
    
    - Filler jobs are submitted directly instead of through the queue
    ScheduleFleetsJobs feeds, which would put them in competition with real queued jobs.
```

gateway/scheduler/tasks/balance_filler_jobs.py
```diff
@@ -0,0 +1,326 @@
...
+        job = Job(
+            program=program,
+            # Filler jobs are excluded from billing, metrics and the fair-share tally
+            # by Job.filler, so they need no service user of their own.
+            author=program.author,
+            filler=True,
+            runner=Program.FLEETS,
+            compute_profile=compute_profile,
+            compute_profile_fk=program.default_size.compute_profile,
+            # Nobody asked for a size, so the attribution does not apply.
```
avilches:
```suggestion
```

Tansito (approved):

PR #2466, opened 2026-09-04
Remove the transitional Job.compute_profile string field (step 1)

ElePT:
> [!IMPORTANT]
> **BLOCKED — needs a change from the billing team before this can merge.**
>
> The Kafka usage-event path builds the classical metric type from
> `job.compute_profile` (the string this PR removes):
> `_build_classical_metric_type` in
> `gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py`.
> That file is owned by another team, so **this PR deliberately does not touch it.**
>
> Before this removal can merge, the billing team needs to repoint that method:
>
> ```python
> # gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
> -        if job.compute_profile:
> -            parts.append(job.compute_profile)
> +        compute_profile = job.compute_profile_id          # FK-backed, added in this PR
> +        if compute_profile:
> +            parts.append(compute_profile)
> ```
>
> `Job.compute_profile_id` (added here) returns `compute_profile_fk.compute_profile_id`
> (or `None`), which equals the old string wherever it was set — so the emitted
> metric type is unchanged. Until that lands, merging this PR would raise
> `AttributeError` when a usage event is built. The corresponding test
> (`tests/core/services/ibm_cloud/event_streams/test_event_streams_client.py`) is
> also left for them to update alongside their change.

## Summary

Removes the transitional `Job.compute_profile` **string** field. It was introduced
alongside the t-shirt sizes work as a stopgap and superseded by `Job.compute_profile_fk`
(FK → `ComputeProfile`), which is the source of truth for the profile a job ran on. The two
were written together on every run and always agreed at creation; this PR retires the string.

This is **Step 1 of a two-step removal** that keeps every deployed release safe to roll back —
the same pattern used to drop `Program.default_compute_profile` (#2420 → #2421):

- **This PR (Step 1):** repoint every reader to the FK, then remove the field from Django's
  model **state only** — the `api_job.compute_profile` **column is left in the database**, so
  the previous release (whose `Job` model still declares the field and lists it in every
  `SELECT`) stays deployable.
- **Step 2 (separate later PR):** the real `DROP COLUMN`, once no release that still declares
  the field can be deployed.

## What changed

**Readers repointed to the FK.** A new null-safe `Job.compute_profile_id` property returns
`compute_profile_fk.compute_profile_id` (or `None` for Ray / historical rows), preserving the
existing `… or settings.DEFAULT_COMPUTE_PROFILE` fallback everywhere. Repointed:

- Fleet-name construction and resource-limit parsing — `core/services/runners/fleets_runner.py`
- Admin job timeline — `api/domain/job_timeline.py` (+ `select_related("compute_profile_fk")`
  on the two timeline querysets in `api/admin.py` to avoid an N+1)

The **billing/Kafka reader is intentionally not repointed here** — see the blocked note above.

**Writers.** The string is no longer written on the run path (`api/use_cases/programs/run.py`)
or the filler-jobs path (`scheduler/tasks/balance_filler_jobs.py`); both already set
`compute_profile_fk`.

**API shape.** `compute_profile` (string) is dropped from the job serializers; `compute_profile_fk`
(nested `ComputeProfileSerializer`, already present on the list/retrieve responses) carries the
profile. Client `Job` loses its `compute_profile` attribute.

**Admin.** The `compute_profile` field is removed from the `JobAdmin` "Fleets" fieldset
(`compute_profile_fk` stays).

**Migration.** `0064_remove_job_compute_profile_state.py` — a state-only `RemoveField` wrapped in
`SeparateDatabaseAndState` (empty `database_operations`), mirroring `0057`. Its comment also
records the billing-team dependency above.

## Out of scope

- The real `DROP COLUMN` (Step 2, follow-up PR).
- The billing/Kafka reader and its test (another team — the blocked note above is their action).
- The deprecated `compute_profile` **run input** kwarg on `function.run(...)` / the run
  `InputSerializer` — a separate surface, retired separately. `function_size` is the replacement.

## Notes

- Invariant relied on: `RunnerConfig.compute_profile` is set to `compute_profile_fk.compute_profile_id`
  in every branch of `_get_runner_config`, and Ray leaves both null — so string and FK agree
  wherever the FK is non-null.
- Rollback-safe: the DB column is retained for one release.
- Related: internal t-shirt sizes epic; the `Program.default_compute_profile` precedent (#2420/#2421).

Opened as **draft** for review.

### DIFF
--- a/client/qiskit_serverless/core/clients/serverless_client.py
+++ b/client/qiskit_serverless/core/clients/serverless_client.py
@@ -236,7 +236,6 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
                 job.get("id"),
                 job_service=self,
                 raw_data=job,
-                compute_profile=job.get("compute_profile"),
             )
             for job in response_data.get("results", [])
         ]
@@ -291,7 +290,6 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
                 job.get("id"),
                 job_service=self,
                 raw_data=job,
-                compute_profile=job.get("compute_profile"),
             )
             for job in response_data.get("results", [])
         ]
@@ -314,7 +312,6 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
             job = Job(
                 job_id=job_id,
                 job_service=self,
-                compute_profile=response_data.get("compute_profile"),
             )
 
         return job
@@ -376,7 +373,6 @@ class ServerlessClient(BaseClient):  # pylint: disable=too-many-public-methods
         return Job(
             job_id,
             job_service=self,
-            compute_profile=response_data.get("compute_profile"),
         )
 
     def get_job_data(self, job_id: str) -> Optional[dict]:
--- a/client/qiskit_serverless/core/job.py
+++ b/client/qiskit_serverless/core/job.py
@@ -168,7 +168,6 @@ class Job:
         job_id: str,
         job_service: JobService,
         raw_data: Optional[Dict[str, Any]] = None,
-        compute_profile: Optional[str] = None,
     ):
         """Job class for async script execution.
 
@@ -178,7 +177,6 @@ class Job:
         """
         self.job_id = job_id
         self._job_service = job_service
-        self.compute_profile = compute_profile
         self.raw_data = raw_data or {}
 
     @property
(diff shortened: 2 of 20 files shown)


Tansito:
@dpatel-ops is taking a look into this and he will be opening small PRs solving this to try to understand everything, we will maintain this open for his reference

marceloamaral:
I think we can close this PR. @dpatel-ops is picking this up in separate PRs, see https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1865.

ElePT:
Got it, closing in favor of https://github.com/Qiskit/qiskit-serverless/pull/2492 and follow-ups.

PR #2467, opened 2026-09-04
Surface T-shirt size elements in the backoffice admin

ElePT:
## Summary

The T-shirt size models (`ComputeProfile`, `FunctionSize`, `Program.default_size`) shipped in 0.36.0, but a function's **size catalog** (its sizes map) and its **default size** weren't surfaced in the backoffice admin where operators work. This PR adds them, and — as a second step — makes a Fleets function created *purely in the admin* actually runnable.

### Part 1 — Surface the size elements

**On the Program (Function) page**
- **`FunctionSizeInline`** — a function's whole sizes map (each size key → compute profile) is now editable inline on its own page, mirroring the catalog the upload endpoint's `sizes` payload builds. It renders as "Sizes (compute profile per size)".
- **`default_size`** field in its own "Default size" fieldset, rendered directly below the sizes catalog. Its dropdown is restricted to the function's *own* sizes (`Program.clean()` rejects a default belonging to another function). The fieldset and the inline are placed adjacent (via a small `change_form.html` override) and named apart so the two are not mistaken for duplicates.
- **`sizes_summary`** changelist column — `N (default: s)` or `-`.

**On the ComputeProfile page**
- **`sizes_using`** column — reference count, so an operator sees a profile is in use before editing/deleting it (the FK is `PROTECT`).

Both changelist columns are annotated to avoid per-row queries.

### Part 2 — Make a backoffice-created Fleets function runnable

A `Program` created purely in the admin previously got only model defaults, skipping a runnable-critical step the upload use case performs:

- **CE project assignment.** `ProgramAdminForm.clean()` assigns a Code Engine project to a Fleets function (provider's active project, or the deployment default) via `CodeEngineProject.objects.assign_to_program`, and **blocks the save** — with the upload endpoint's own message — when none is available. A Fleets function with no active CE project cannot run, so this fails loud rather than persisting a dead function.

Refactor: promoted `_no_ce_project_message` → `no_ce_project_message` so the admin reuses the endpoint's exact wording.

## Note: default-size seeding is intentionally left out

An earlier version of this PR also seeded a per-function default size (`DEFAULT_FUNCTION_SIZE` → `DEFAULT_FUNCTION_SIZE_PROFILE`) for a size-less Fleets function, mirroring the upload path. That has been removed.

The reason is a design decision that is not in the codebase yet but is coming: the fallback for a function that declared no sizes should be a single shared **`platform-default`** `FunctionSize` row (no `function` id), resolved at **job creation** rather than at function creation, so `Job.function_size` is never null and billing can always read the size off the job. The upload path's existing seeding is untouched (out of scope) and will be revisited when the platform-default work lands.

## Scope
- Admin-only. **No model or migration changes.**
- Does **not** touch `Job.compute_profile` (the deprecated string) — that's a separate branch's concern.

## Behavioral note (blocking tradeoff)
Because `clean()` blocks any Fleets save that can't get an active CE project, a legacy/misconfigured Fleets function whose provider has no active project (and no default configured) cannot be saved — even to disable it — until a project is assignable. This is the intended "fail-loud" behavior (mirrors the upload endpoint); the operator assigns a project in the same edit.

## Testing
- `gateway/tests/test_tshirt_admin.py` (16 tests): inline rendering, `default_size` dropdown scoping, `sizes_summary`, `sizes_using` count; CE auto-assign (provider + default paths), blocking when unconfigured / no project / inactive project, operator-chosen project preserved, Ray no-ops; plus two **end-to-end admin `client`** tests proving the real add form blocks a non-runnable function and creates a runnable one.
- `pylint` 10.00/10 on both source files; `black` clean.

### DIFF
--- a/gateway/api/admin.py
+++ b/gateway/api/admin.py
@@ -5,6 +5,7 @@ import logging
 import uuid
 
 from django import forms
+from django.conf import settings
 from django.contrib import admin, messages
 from django.core.cache import cache
 from django.db.models import Count, F, Q
@@ -20,6 +21,7 @@ from api.domain.arguments_schema import (
 )
 from api.domain.job_timeline import render_job_timeline
 from api.domain.exceptions.invalid_arguments_exception import InvalidArgumentsException
+from api.use_cases.programs.upload import no_ce_project_message, seed_default_size
 from api.use_cases.programs.validate_arguments import validate_arguments
 from core.models import (
     CodeEngineProject,
@@ -242,6 +244,41 @@ class ProgramAdminForm(forms.ModelForm):
 
         return value
 
+    def clean(self):
+        """Assign a Code Engine project to a Fleets function, and refuse the save if none is available.
+
+        The upload endpoint auto-assigns a CE project (provider's project, or the deployment default)
+        and rejects a Fleets function it cannot place, since one is not runnable without an active
+        project. A plain admin save skips that, so replicate it here: fail loud with the same wording
+        rather than let an operator persist a function that can never run.
+
+        Runs before ``_post_clean`` copies cleaned values onto the instance, so the assigned project
+        is written back into ``cleaned_data`` for the save to pick up. Leaving a project the operator
+        chose alone falls out of ``assign_to_program`` being a no-op when one is already set.
+        """
+        cleaned_data = super().clean()
+
+        if cleaned_data.get("runner") != Program.FLEETS:
+            return cleaned_data
+
+        # Mirror the request onto the instance so the manager sees the values being saved, not the
+        # stored ones, then let it assign (in place) exactly as the upload use case does.
+        self.instance.runner = Program.FLEETS
+        self.instance.provider = cleaned_data.get("provider")
+        self.instance.code_engine_project = cleaned_data.get("code_engine_project")
+        try:
+            CodeEngineProject.objects.assign_to_program(self.instance)
+        except ValueError:
+            # select_default() raises when CE_DEFAULT_PROJECT_NAME is unconfigured; treat it as
+            # "no project available" and fall through to the rejection below.
+            pass
+
+        if not self.instance.code_engine_project:
+            raise forms.ValidationError(no_ce_project_message(self.instance))
+
+        cleaned_data["code_engine_project"] = self.instance.code_engine_project
+        return cleaned_data
+
 
 class ValidateArgumentsForm(forms.Form):
     """Arguments to try against a function's stored schema."""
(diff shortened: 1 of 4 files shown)


ElePT:
## Verified: admin T-shirt size elements

Ran the backoffice admin locally against a seeded SQLite DB and confirmed each new element renders with real data. Full test suite for the feature passes: `pytest tests/test_tshirt_admin.py` → 21 passed.

**Function list — new "Sizes" column** (`sizes_summary`)
A function with sizes shows the count and the default; a size-less function shows `-`.
`vqe-solver` → `3 (default: m)`, `qaoa-runner` → `-`.

program_changelist

**Compute profile list — new "Sizes using" column** (`sizes_using`)
Counts how many function sizes point at each profile, so operators see usage before editing or deleting one (the FK is PROTECT). The unused `32x256` profile correctly shows `0`.

computeprofile_changelist

**Function change page — Sizes fieldset, default_size, and the sizes inline**
The `default_size` dropdown is scoped to the function's own sizes. On `vqe-solver` it lists only `vqe-solver (s/m/l)`, nothing from other functions. The "Sizes (compute profile per size)" inline below maps each size to its compute profile. The Fleets fieldset shows the assigned Code Engine project.
program_change

ElePT:
## Verified: admin T-shirt size elements

Ran the backoffice admin locally against a seeded SQLite DB and confirmed each new element renders with real data. Full test suite for the feature passes: `pytest tests/test_tshirt_admin.py` → 21 passed.

**Function list — new "Sizes" column** (`sizes_summary`)
A function with sizes shows the count and the default; a size-less function shows `-`.
`vqe-solver` → `3 (default: m)`, `qaoa-runner` → `-`.

program_changelist

**Compute profile list — new "Sizes using" column** (`sizes_using`)
Counts how many function sizes point at each profile, so operators see usage before editing or deleting one (the FK is PROTECT). The unused `32x256` profile correctly shows `0`.

computeprofile_changelist

**Function change page — Sizes fieldset, default_size, and the sizes inline**
The `default_size` dropdown is scoped to the function's own sizes. On `vqe-solver` it lists only `vqe-solver (s/m/l)`, nothing from other functions. The "Sizes (compute profile per size)" inline below maps each size to its compute profile. The Fleets fieldset shows the assigned Code Engine project.



program_change

Tansito:
I added a little change to the proposal moving the sizes and default size fieldset nearest to the "FLEETS" fieldset. I think semantically it has more than adding it at the end of the page:

Screenshot 2026-09-18 at 12 40 59 PM

Tansito (approved):

PR #2468, opened 2026-09-05
Set `running_started_at` before emitting the job-started events

ismaelRozasRamallal:
### Summary
When a Fleets job moved from `PENDING` to `RUNNING`, `to_running` emitted the classical usage and license fee events *before* writing `running_started_at`. Both events carry that field as `job_started_at`, so it was always published as `null`. This splits the single write at the end of the method: `running_started_at` is now persisted first, then the events are emitted, then the status write follows. The status write deliberately stays after the emits, so a Kafka failure still raises, leaves the job `PENDING`, and lets the scheduler retry it on the next tick — that behaviour is unchanged.

Persisting the timestamp first also fixes a drift problem. If publishing fails while the fleet is already running in Code Engine, the start time is recorded anyway, so a retry two or three ticks later reports when the job *actually* started rather than when the emit finally succeeded. The write is guarded by `if job.running_started_at is None`, so retries reuse the stored value instead of pushing the start time forward. Note this means a `PENDING` row can briefly carry `running_started_at`; the only consumer, `api/domain/job_timeline.py`, uses it as a fallback and already discards inconsistent intervals.

Three tests in `TestEventStreamsIntegration` were updated: they previously asserted that *no* DB write happened before the emit, and now assert the narrower guarantee that the **status** write comes after it. New cases cover the timestamp being visible to both emit calls, a retry reusing the stored value, and a failed publish keeping the timestamp while leaving the job `PENDING`. To confirm the original bug, revert the production change and the "persists before emitting" test fails with `job_started_at` unset.

### Checklist

### DIFF
--- a/gateway/scheduler/tasks/update_fleets_jobs_statuses.py
+++ b/gateway/scheduler/tasks/update_fleets_jobs_statuses.py
@@ -142,14 +142,21 @@ class UpdateFleetsJobsStatuses(SchedulerTask):
             job.status,
             Job.RUNNING,
         )
+        # running_started_at is persisted before the events because the events carry it as
+        # job_started_at, and because it must survive a failed emit: publishing raises, the
+        # status stays PENDING and the job is retried next tick, but the moment the fleet was
+        # first seen RUNNING is already recorded. Stamping it after a successful emit instead
+        # would push it later by however many ticks failed. It is written only once — a retry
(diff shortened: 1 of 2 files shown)


korgan00 (approved):

ElePT (approved):
LGTM

PR #2472, opened 2026-09-09
License fee function size

korgan00:
### Summary

Add function size into the license fee kafka event.
Depends on https://github.com/Qiskit/qiskit-serverless/pull/2417/

### Checklist


gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -175,6 +181,16 @@ def _emit_license_fee(self, job: Job) -> None:
...
 
+    def _resolve_function_size(self, job: Job) -> str:
+        """Resolve the function size for the job.
+
+        Returns the job's explicit size if set, otherwise the function's default size.
+        Guaranteed to find a size for any Fleets job.
+        """
+        if job.function_size:
+            return job.function_size.function_size
+        return job.program.default_size.function_size
```
ismaelRozasRamallal: At first it should be never need to fallback on the program default size since we are doing that circuit when creating the job, but let's keep it anyways just in case

gateway/tests/core/services/ibm_cloud/event_streams/test_event_streams_client.py
```diff
@@ -352,22 +354,44 @@ def test_emit_license_fee_publishes_correct_payload(self):
...
+        assert published["data"]["business_model"] == "licensed"
+
+    def test_emit_license_fee_falls_back_to_program_default_size(self):
+        job = _make_job()
+        job.program = MagicMock()
+        job.program.provider.name = "ibm"
+        job.program.title = "test-program"
+        job.function_size = None
+        default_size = MagicMock()
+        default_size.function_size = "M"
```
ismaelRozasRamallal: just a nit, but something which can be confusing in the future. Function sizes are in lower case 😄

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
ismaelRozasRamallal: By other side, if we keep this mechanism, we may be hidding an error in function size assignation so, thinking again on it, it's better to remove the fallback 👍

ismaelRozasRamallal (approved):
LGTM!

PR #2473, opened 2026-09-09
Exclude filler jobs from the Fleets status/timeout update task

avilches:
### Summary
BalanceFillerJobs is meant to own the full lifecycle of filler jobs (jobs with filler == true), but UpdateFleetsJobsStatuses.run() queried all Fleets jobs in RUNNING_STATUSES without excluding them.

Filler jobs are meant to run indefinitely, so once one crossed PROGRAM_TIMEOUT (24h default) it got marked STOPPED by stop_job_if_timeout. Once that row left RUNNING_STATUSES, BalanceFillerJobs stopped tracking it and created a replacement, leaving the original Code Engine job orphaned and running..

### Details and comments
Excludes filler=True from the queryset in UpdateFleetsJobsStatuses.run(), so filler jobs stay fully owned by BalanceFillerJobs and are never touched by this task.

PR #2475, opened 2026-09-10
Revert "Exclude filler jobs from the Fleets status/timeout update task"

avilches:
### Summary
Rolls back #2473. That PR excluded filler jobs (`filler=True`) from the queryset that `UpdateFleetsJobsStatuses.run()` iterates, meaning `update_job_status()` is never called for them at all.

The fix was not correct: `update_job_status()` is also where a Fleets job transitions from PENDING to RUNNING once Code Engine reports it running (`to_running()`), and where SUCCEEDED/FAILED are detected. With filler jobs excluded from the queryset, a filler job stuck in PENDING would never move to RUNNING, and one that exits on its own would never reach a terminal status either, since nothing else in the scheduler polls Fleets status for a job in `RUNNING_STATUSES`.

### Details and comments
This PR only reverts the change (queryset filter and its test) back to how it was before #2473. It does not attempt an alternative fix for the underlying problem #2473 was addressing (a filler job hitting PROGRAM_TIMEOUT gets marked STOPPED in the DB while the underlying Code Engine job keeps running, since BalanceFillerJobs then spins up a replacement for the same slot).


ElePT (approved):
Got it, LGTM

PR #2477, opened 2026-09-10
Emit classical time events in seconds instead of miliseconds

ismaelRozasRamallal:
### Summary
Fleets jobs publish classical compute usage to IBM Cloud Event Streams as `metric_value` in the `quantum.<env>.function-usage.v1` topic. The field name carries no unit, so the value's meaning is a convention shared with the billing service rather than something the payload states. Historically that convention moved: nanoseconds when the client was introduced, then milliseconds once the events were aligned with the billing contract. Milliseconds no longer matched what the consumer expects, and a mismatched unit is silently wrong — it bills a plausible number rather than failing.

### Details and comments
`_usage_seconds` in `kafka_event_streams_client.py` is the single place the unit is defined. It computes `now - job.running_started_at` and returns `math.ceil(delta.total_seconds())`, so the value is whole seconds with any partial second rounded up and never a decimal. It feeds the in-progress and completed events; job-started sends `0` and the license fee sends `1`, both counts rather than durations. To change the unit again, edit that method, the class docstring stating the unit, and the expectations in `test_event_streams_client.py`.

### Checklist

### DIFF
--- a/gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
+++ b/gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import json
 import logging
+import math
 import os
 import uuid
 from datetime import datetime, timezone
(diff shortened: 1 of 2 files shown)


ElePT (approved):
As you have verified that the expected metric is seconds (which you showed me),  I think this looks good.

PR #2482, opened 2026-09-11
Create kafka external secret independently of billing

ismaelRozasRamallal:
### Summary

The Kafka ExternalSecret was gated on `gateway.application.eventStreams.enabled`, so provisioning the credential was tied to turning billing on. It now has its own flag, `gateway.secrets.kafka.createExternalSecret`, so the credential can be set up and checked before events are switched on.

### Details and comments

The flag sits on the secret block, matching the `createExternalSecret` naming already used by `secret-psql.yaml`. It also reads a key that has a real default in this chart: `gateway.application.eventStreams` has none in the parent values, only in the gateway subchart, so the old gate relied on deployment values to define it.

Defaults to false. The deployment values that set it per environment go in a separate PR on the internal repo, which should merge first so staging keeps its secret.

### DIFF
--- a/charts/qiskit-serverless/values.yaml
+++ b/charts/qiskit-serverless/values.yaml
@@ -95,6 +95,10 @@ gateway:
       secretRef: ""
       secretStoreName: qiskit-serverless
     kafka:
+      # Creation of the Kafka ExternalSecret is independent of
+      # application.eventStreams.enabled, so the credential can be provisioned
+      # and verified before billing events are switched on.
+      createExternalSecret: false
       name: ""
       secretRef: ""
       secretStoreName: qiskit-serverless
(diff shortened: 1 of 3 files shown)


charts/qiskit-serverless/values.yaml
```diff
@@ -95,6 +95,10 @@ gateway:
       secretRef: ""
       secretStoreName: qiskit-serverless
     kafka:
+      # Creation of the Kafka ExternalSecret is independent of
+      # application.eventStreams.enabled, so the credential can be provisioned
+      # and verified before billing events are switched on.
+      createExternalSecret: false
```
Tansito, line 101: Discussed with @ismaelRozasRamallal , this will be temporal to test the creation of secret without enabling the feature. When we can confirm everything works we can return to the original flag.

Tansito (approved):
LGTM @ismaelRozasRamallal 🫡

PR #2483, opened 2026-09-14
Add external secret for eu-de kafka instance

ismaelRozasRamallal:
### Summary

Adds a chart template and default values for a second Kafka ExternalSecret, covering the eu-de region alongside the existing us-east one. Creation is off by default, so this changes nothing until a deployment sets up the new values.

### Details and comments

The new template matches `kafka-credentials.yaml` with `secrets.kafka` swapped for `secrets.kafkaEuDe`, so both regions follow one shape. It uses the `createExternalSecret` flag from the previous PR, which keeps secret creation separate from turning billing events on.

Defaults go in both the parent chart and the gateway subchart, since subchart defaults alone do not fill in `.Values.gateway.*`. The deployment still reads the us-east secret by a fixed name, so nothing consumes the eu-de one yet.


Tansito (approved):
Veru straighforward 👍

korgan00 (approved):

PR #2489, opened 2026-09-16
Fix Fleets jobs stuck in PENDING when Kafka is down

avilches:
### Summary
Fixes a bug where a Fleets job whose Kafka event-stream publish failed while transitioning from PENDING to RUNNING got stuck in PENDING forever: `to_running()` emitted the `job_started`/license-fee events before writing the status, and a failed emit raised, skipping the DB update entirely. Since nothing else moves a PENDING Fleets job forward, every scheduler tick retried the same failing emit and the job never reached RUNNING while Kafka was unavailable.

### Details and comments
The status write (`Job.status` + `running_started_at`) and its `JobEvent` are now committed together inside a `transaction.atomic()` block, and happen before the Kafka emit rather than after it. The emit is now best-effort: a `RuntimeError` from the event-streams client is caught, logged with the job id and the underlying error, and dropped, instead of propagating and rolling back the transition.

One side effect: since the status transition no longer depends on the emit succeeding, `to_running()` now runs at most once per job (there is no more PENDING retry to reuse an existing `running_started_at`), so that branch was simplified away.

Updated the tests in `test_update_fleets_jobs_statuses.py` that asserted the old behavior (status not written until the emit succeeds, exception propagating on a Kafka failure) to match the fix, and added regression tests for the actual bug: the job reaches RUNNING even when the emit raises, and the failure is logged rather than silently swallowed.


gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -142,27 +143,30 @@ def to_running(self, job: Job) -> None:
...
-        job.update_fields({"status": Job.RUNNING})
-        JobEvent.objects.add_status_event(
-            job_id=job.id,
-            origin=JobEventOrigin.SCHEDULER,
-            context=JobEventContext.UPDATE_JOB_STATUS,
-            status=job.status,
-        )
+        # The status and running_started_at are committed together, and independently of
+        # Kafka: an event-stream outage must not leave the job stuck retrying PENDING
+        # forever, so the emit below is best-effort and never rolls this back.
```
avilches:
```suggestion
```

marceloamaral (dismissed):
Looks good making it atomic!

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -142,27 +142,27 @@ def to_running(self, job: Job) -> None:
...
-            status=job.status,
-        )
+        with transaction.atomic():
+            event = JobEvent.objects.add_status_event(
+                job_id=job.id,
+                origin=JobEventOrigin.SCHEDULER,
+                context=JobEventContext.UPDATE_JOB_STATUS,
+                status=Job.RUNNING,
+            )
+            job.update_fields({"status": Job.RUNNING, "running_started_at": event.created})
```
avilches, line 152: @Tansito @korgan00 @ismaelRozasRamallal 
- 1 transaction for the job + job_event. 
- `running_started_at` date is the same as the job_event.created

gateway/scheduler/tasks/update_fleets_jobs_statuses.py
```diff
@@ -142,27 +142,27 @@ def to_running(self, job: Job) -> None:
...
+            # prevent custom function to emit license fee
+            # since licenses is a provider feature
+            if job.program.provider:
+                self.event_streams_client.emit_license_fee(job)
+        except RuntimeError as ex:
+            logger.error(
+                "job_id=%s error emitting job_started/license_fee event to Kafka, event dropped: %s",
+                job.id,
+                str(ex),
+            )
```
avilches, line 165: Best effort until we have the (kind of) outbox pattern to mirror Kafka events.

ismaelRozasRamallal (approved):
LGTM!! thanks @avilches !!!

PR #2490, opened 2026-09-17
Force sizes exist for a function fleet

Tansito:
### Summary

Issue related: 2488, it works as a simplification of what we were trying to do here.

Fleets functions could silently end up with no compute size in two places, which let them either get created "broken" or keep running on a deployment-wide default instead of failing clearly. This PR closes both gaps so a Fleets function either has a valid size or the request is rejected with a clear error — no more silent fallbacks.

- **Upload**: when a Fleets function is created or updated without declaring `sizes`/`default_size`, the gateway seeds it with the deployment's default size. If that default compute profile isn't registered, the function used to be created anyway with no size at all (just a log warning). Now the upload is rejected outright. This seeding logic also now runs on update, not just create — so a function that switches its runner to Fleets, or an older Fleets function that never got a size, picks one up automatically the next time it's re-uploaded without touching sizes. Functions that already declare their own size catalog are left untouched.

- **Run**: when running a Fleets function that has no default size and no size/profile was requested, the gateway used to fall back to a deployment-wide default compute profile. It now rejects the run with a clear error instead, since a Fleets job with no resolved size shouldn't be allowed to proceed silently.

### Details and comments
- Both new failure paths reuse the existing `FunctionConfigurationException` pattern already used elsewhere in these flows (e.g. missing Code Engine project, unregistered compute profile) — same error shape, same "contact administrator" messaging style, mapped to a 400 response.
- The upload-time seeding change is intentionally conservative: it only kicks in when the function's size catalog is genuinely empty, so it can never clobber a function's own custom sizes on an unrelated update.
- A lower-level fallback in the Fleets runner itself (used when executing an already-created job) was deliberately left alone — it only matters for jobs created before this change, and no new job can reach it anymore since run now rejects sizeless Fleets functions upfront.
- Added a note flagging the `SETTINGS_DEFAULT` size-source value as no longer produced going forward, kept only so existing historical job records stay readable.
- Test coverage was updated to match the new behavior (upload rejecting instead of creating sizeless, run rejecting instead of falling back) and extended with new cases for the update-seeding behavior and for "existing catalog is left alone."

Co-authored-by: @dpatel-ops , @korgan00 , @ismaelRozasRamallal 

### Checklist


marceloamaral (commented):
Looks good, LGTM

korgan00 (approved):
LGTM

ismaelRozasRamallal (approved):
LGTM!! thanks @Tansito !!!

PR #2492, opened 2026-09-18
Repoint FleetsRunner and admin timeline to Job.compute_profile_fk

dpatel-ops:
### Summary

Ref: #2466 

This PR is **Step 1** (PR 1) of our step-by-step implementation plan to safely retire the legacy, flat `Job.compute_profile` string column in favor of the normalized `Job.compute_profile_fk` relationship. 

In this step, we add the null-safe `@property compute_profile_id` to the `Job` model and repoint all internal "read-only" readers—specifically the **FleetsRunner**, the **JobTimeline**, and the **Django Backoffice Admin**—to consume this property instead of the raw string column. 

---

### Details and comments

* **Core Model Helper**: Added `@property def compute_profile_id(self) -> str | None` on the `Job` model in `gateway/core/models.py`. This proxies the `compute_profile_fk` relation null-safely and utilizes direct ID lookup to incur **zero extra SQL queries** for unset or Ray jobs.
* **Fleets Runner**: Repointed `_build_fleet_name()`, `_parse_compute_profile()`, and logging in `fleets_runner.py` to use `self.job.compute_profile_id`.
* **Admin Job Timeline**: Updated `job_timeline.py` to read `job.compute_profile_id`. Added `select_related("compute_profile_fk")` in `admin.py` timeline querysets to prevent N+1 query regressions.
* **Django Admin Layout**: Removed the legacy, editable `compute_profile` string field from the `JobAdmin` detail fieldsets.
* **Tests Updated**: Fully updated `test_fleets_runner.py` and `test_job_timeline_admin.py` to cover the property and relation. All 78 targeted tests (and the full gateway test suite) pass cleanly.
* **Backward Compatibility**: 100% preserved. External user traffic is unaffected as API views, serializers, and the Client SDK remain untouched in this step.

### Checklist


marceloamaral (dismissed):
Everything looks good to me, thanks @dpatel-ops!
Let's wait for @Tansito review and then we can merge it.

gateway/core/models.py
```diff
@@ -663,6 +663,16 @@ def in_terminal_state(self):
...
 
+    @property
+    def compute_profile_id(self) -> str | None:
+        """Bare compute-profile string from the FK (the source of truth).
+
+        Returns None for Ray jobs and any historical row with no
+        ``compute_profile_fk``. Reads ``compute_profile_fk_id`` first so an
+        unset FK costs no extra query.
+        """
+        return self.compute_profile_fk.compute_profile_id if self.compute_profile_fk_id else None
```
marceloamaral: `ComputeProfile.compute_profile_id` is the primary key, so `compute_profile_fk_id` already holds the profile string. Going through `compute_profile_fk` loads the whole row from the DB just to read back the value the Job row already has.

Can we make it just:
```python
    return self.compute_profile_fk_id
```

Note that we don't need the guard `if self.compute_profile_fk_id`, if `compute_profile_fk` is `null=True` Django already stores None

gateway/api/admin.py
```diff
@@ -625,7 +624,11 @@ def job_timeline_view(self, request):
                 continue
         # one single query: the rendering iterates the jobs again, and re-running the queryset
         # could come back empty (deleted in between) and break the rendering half way through
-        jobs = list(Job.objects.filter(id__in=id_list).select_related("author").prefetch_related("job_events"))
+        jobs = list(
+            Job.objects.filter(id__in=id_list)
+            .select_related("author", "compute_profile_fk")
```
Tansito: If we apply @marceloamaral 's change I think we can remove `compute_profile_fk` from the select_related (double check @dpatel-ops in case I would be wrong)

gateway/api/admin.py
```diff
@@ -852,7 +855,7 @@ def index(self, request, extra_context=None):
         if timeline_context is None:
             recent_jobs = list(
                 Job.objects.filter(runner=Program.FLEETS)
-                .select_related("author")
+                .select_related("author", "compute_profile_fk")
```
Tansito: Same from my comment above.

Tansito (approved):

PR #2496, opened 2026-09-19
Report whether Code Engine accepted a fleet cancel

marceloamaral:
### Summary

Send the fleet cancel and let Code Engine's response decide, instead of reading the fleet status first to work out whether to cancel at all.

Two bugs go away. A fleet in a status the generated client does not know, `standby` among them, could not be cancelled, because the status read raised `ValueError` before we ever asked. And a cancel that Code Engine rate-limited was reported as a completed stop even though nothing had been sent.

First step of the stopping epic, https://github.ibm.com/IBM-Q-Software/qiskit-serverless/issues/1867, which holds the staging measurements and the plan for the rest.

### Details and comments

Two behaviour changes worth a look:

- The filler balancer now leaves a job active when a cancel is rate-limited, instead of marking it stopped. That is what its own comment already asks for.
- The stop endpoint says the stop was requested, rather than that the job "was running and has been stopped". Code Engine answers 202 for a fleet whose task has already finished, so an accepted cancel does not mean the job was running. Step 2 of the epic makes this exact again by reading the task store before cancelling.


avilches (commented):
One more general note: the root cause behind this `standby` caveat and the `_MODEL_VALIDATION_ERROR` workaround in `cancel_job` is that the vendored `V2Fleet.status` setter (`ce_client/models/v2_fleet.py`) is missing `standby` from its `allowed_values` (currently 8 entries). Adding it there would remove most of the need for these workarounds, including `test_generated_model_rejects_a_status_code_engine_really_returns`, whose own docstring says to delete it once that happens. Worth a follow-up even if out of scope here.

gateway/core/ibm_cloud/code_engine/fleets/handler.py
```diff
@@ -340,6 +393,11 @@ def _wait_until_state(
             pending, running, canceling, canceled, deleting, failed,
             succeeded, successful
 
+        That list is incomplete: Code Engine also returns ``standby``, which the generated model
```
avilches, line 396: Since this loop below only catches `ApiException`, not `ValueError`, it would still crash on the `standby` case described here if `wait=True` were ever used. No caller passes `wait=True` today so this is latent, but it'd be worth adding an `except ValueError` here that treats the model-validation case (matching `_MODEL_VALIDATION_ERROR`) as "not yet terminal, keep polling", the same way `cancel_job` already does below.

gateway/core/ibm_cloud/code_engine/fleets/handler.py
```diff
@@ -51,10 +51,29 @@
...
+# as not delivered and retried, and the retry answers 409.
+_MODEL_VALIDATION_ERROR = "Invalid value for"
+
+
+def _is_already_canceled(exc: ApiException) -> bool:
+    """Return True if an error body carries Code Engine's already-cancelling code.
+
+    The status is not checked here. The caller pairs this with the 409 it applies to.
+    """
+    return _ALREADY_CANCELED_CODE in str(getattr(exc, "body", "") or "")
```
avilches, line 75: This matches a single hardcoded error code inside the whole 409 body. Is there a separate Code Engine error code for "still cancelling" (cancel already in flight) versus "already cancelled" (terminal)? If so, this currently falls into the `else: raise` branch in `cancel_job` instead of being treated as "nothing to cancel". Could you confirm against the Code Engine API reference (or with the CE team) and add the second code here if one exists?

gateway/api/use_cases/jobs/stop.py
```diff
@@ -115,9 +115,9 @@ def _stop_ray_job_if_active(self, job: Job):
...
         if runner.is_active():
             try:
-                was_running = runner.stop()
-                if was_running:
-                    self.status_messages.append("Serverless job was running and has been stopped.")
+                stop_accepted = runner.stop()
+                if stop_accepted:
+                    self.status_messages.append("Serverless job stop has been requested.")
                 else:
                     self.status_messages.append("Serverless job was already not running.")
```
avilches: For Fleets, this `False` branch also covers a 409 "already cancelling" response, i.e. a job that could still be running with a cancel already in flight, which is a different situation from "already not running". Suggest wording this so it doesn't imply the job had already stopped, e.g. "Serverless job was already stopped or a stop was already in progress."

gateway/core/services/runners/fleets_mock.py
```diff
@@ -358,13 +357,18 @@ def _mock_cancel_job(self, identifier, **kwargs):  # pylint: disable=unused-argu
...
+        ``True``, matching the real ``cancel_job``, which reports whether Code Engine accepted the
+        cancel. Returning nothing would make every mocked stop read as "nothing to cancel".
     """
     s3 = _get_mock_s3()
     # Resolve the job's own project bucket (like _mock_get_job_status), not the
     # first active project, so cancel writes where status() reads.
     bucket = _task_store_bucket(self.project_id)
     cancel_key = f"{queue_prefix(self.project_id, identifier)}canceled/0/{identifier}-0/canceled"
     s3.put_object(Bucket=bucket, Key=cancel_key, Body=b"")
+    return True
```
avilches, line 391: Unlike the real `cancel_job`, this always returns `True` with no existence check, so the "nothing to cancel" path (fleet gone or already cancelled) can't be exercised against the mock. Suggest checking fleet existence the same way `_mock_get_job_status` does (head_object on the archive bucket) before writing the cancel key, and returning `False` if the fleet doesn't exist or the cancel key is already there.

gateway/core/ibm_cloud/code_engine/fleets/handler.py
marceloamaral: You're right. But since the wait isn't used anywhere, it might be better to delete the path than to guard it, just to avoid the confusion. And regarding adding standby, I would prefer do a separate PR that regenerates the swagger client than patch that one list (instead of an isolated update)... Also if we ever do need to wait before deleting a fleet, it could read the COS task state rather than the fleet API. So once this PR is merged, I'll open another one deleting the wait path...

gateway/core/ibm_cloud/code_engine/fleets/handler.py
marceloamaral: Good question. I checked and the docs don't list error codes for this endpoint, and the client doesn't either :/

gateway/api/use_cases/jobs/stop.py
marceloamaral: Agreed. Changed it to `"Serverless job was already stopping or no longer running."`, which stays true for Ray too.

gateway/core/services/runners/fleets_mock.py
marceloamaral: It make sense, I added it!

avilches (approved):
Great job :)

PR #2497, opened 2026-09-19
Disable manual status edits in Job admin, add Ray-only Stop job button

avilches:
### Summary
The Job admin backoffice let staff type a new value into `status` or `sub_status` directly, which is risky: it bypasses every state-machine check the rest of the codebase relies on. This PR locks both fields to read-only and replaces that capability with a single, narrow action: a "Stop job" button, shown next to the status field only for a Ray job that isn't already in a terminal state (`STOPPED`, `SUCCEEDED`, `FAILED`).

Clicking it (after two JS confirmations) only sets the job's status to `STOPPED` and logs a `JobEvent` with `origin=BACKOFFICE` and `context=STOP_JOB`, the same context the real API stop endpoint uses. It does not cancel anything on Ray or in Qiskit Runtime, and it's a no-op (with an error message) for Fleets jobs or jobs already in a terminal state.

### Details and comments
The button posts to a dedicated admin URL via a plain JS fetch instead of submitting the Job change form. This matters for two reasons found while testing: the change form requires a `program` value even though the DB column allows null (a job whose program was deleted can't satisfy that), and submitting the full form would also silently save whatever unrelated field an operator happened to be editing on the page at the time. Neither of those extra failure modes exist with a separate, minimal endpoint that only touches `status`.

The old `save_model` override that logged a `JobEvent` whenever `status`/`sub_status` changed through the form is removed, since those fields can no longer change that way.

### Demo

A fleet jobs only shows the state. It's not possible to edit the runner or the sub-state either.

image

A (no terminal) Ray job has a `Stop` button: 

image

After 2 confirms like this, the job is stopped:
image

image


gateway/api/admin.py
```diff
@@ -717,9 +726,31 @@ def get_urls(self):
...
 
+    def stop_job_view(self, request, job_id):
+        """Stop job button target: sets a non-terminal Ray job's status to STOPPED, nothing else.
+        Hit via JS fetch, not a form submit, so it never runs the Job change form (which
+        requires fields some jobs don't have, and would save unrelated edits on the page)."""
+        job = get_object_or_404(Job, pk=job_id)
+        if not self.has_change_permission(request, job):
+            raise PermissionDenied
+        if request.method == "POST" and self._can_stop(job):
+            job.status = Job.STOPPED
```
marceloamaral, line 741: Nice! For Fleets it'll write `STOPPING` and let the scheduler send the cancel and confirm it, rather than writing a terminal status directly.

marceloamaral (dismissed):
Looks good to me, LGTM!

gateway/api/static/admin/js/job_stop_button.js
```diff
@@ -0,0 +1,30 @@
...
+            !confirm(
+                "Stop this job? This only marks it as STOPPED in the database, it does not cancel anything running on Ray."
+            )
+        ) {
+            return;
+        }
+        if (!confirm("Are you sure? This cannot be undone.")) return;
+
+        const match = document.cookie.match(/csrftoken=([^;]+)/);
+        fetch(button.dataset.stopUrl, {
```
Tansito, line 23: I miss a `catch` here for scenarios like the `PermissionDenied`. An `alert` could be more than enough.

gateway/tests/api/test_models.py
```diff
@@ -6,6 +6,8 @@
 from rest_framework.test import APITestCase
```
Tansito, line 5: Does it have sense to have these tests in `test_models`? Should we not have maybe a `test_admin.py` with specific admin tests better?

gateway/tests/api/test_models.py
```diff
@@ -141,88 +143,167 @@ def test_program_trial_instances_signal(self):
 
 class TestJobAdmin(APITestCase):
```
Tansito: I also miss a test where you change exceptions like the PermissionDenied, for example.

gateway/api/static/admin/js/job_stop_button.js
avilches: Added a catch to the fetch chain: it now checks response.ok (so a 403 from PermissionDenied is caught too, not just network errors) and shows an alert with the error.

gateway/tests/api/test_models.py
avilches: Moved TestJobAdmin out into its own gateway/tests/api/test_admin.py, leaving test_models.py with only the model/signal tests.

gateway/tests/api/test_models.py
avilches: Added test_stop_job_view_denies_a_staff_user_without_change_permission in test_admin.py: a staff user without change permission on Job now asserts a 403 and that nothing changed.

marceloamaral (approved):

PR #2498, opened 2026-09-19
WIP: Publish job billing events to Kafka through a transactional outbox

avilches:
### Summary

Job billing events are published to Kafka inline, at the moment a job changes status. That ties a job transition to an external system: when Kafka is unavailable the event is dropped and the job is never billed, and a slow broker holds up the scheduler loop.

This PR moves that publishing behind a transactional outbox. A `JobOutbox` row is created when a Fleets job with an instance CRN is submitted, and it is updated in the same transaction as every status event, so the row can never drift from the job it tracks. A new scheduler task, `PublishOutbox`, drains those rows: it sends the billing facts each job owes, marks each one as sent, and deletes the row once all of them are settled. A Kafka outage now delays billing instead of losing it, and it no longer blocks a status transition.

### Details and comments

Two facts are tracked per row. The provider license fee is owed once by any job that really ran a function that has a provider. The final usage event is owed once by any job that reached a terminal status. They are sent independently, so one of them failing does not starve the other.

`has_run` is what proves the job executed, and it is set both on the transition to RUNNING and on the one to SUCCEEDED. A job that starts and finishes between two scheduler polls is only ever observed as PENDING and then SUCCEEDED, and it still owes the fee. A job that jumps straight to FAILED or STOPPED leaves the flag false and is not charged, because those two states cannot tell a job that never started from one that ran very briefly.

`PublishOutbox` drains the outbox in successive small batches (100 rows each) within a single scheduler tick, oldest first, until either nothing is left pending or the tick's time budget (500ms by default) runs out. A healthy Kafka is expected to drain everything pending in one tick this way; the batch size only bounds a single query's size, it is not a cap on how much a tick can send.

Every status write now goes through `Job.change_status`, which creates the event and persists the status in a single transaction, always in that order. That fixed order is what keeps two concurrent writers of the same job's `Job` and `JobOutbox` rows from deadlocking on lock order. Two notes for reviewers: `api/admin.py` keeps its own explicit transaction, because it can write both a status and a sub status event before a generic save, and `stop.py` now goes through `update_fields` and so no longer gets the optimistic concurrency check it used to get from `job.save()`.

Kafka now derives a job's start time from its first RUNNING `JobEvent` instead of from `Job.running_started_at`. That field stays in the model, unused: dropping it needs its own migration and touches the admin timeline, so it is out of scope here.

A small in-memory circuit breaker backs the task off while Kafka keeps failing. Its consecutive-failure counter is not reset between scheduler ticks: failures keep accumulating across as many ticks as it takes to reach the threshold (5 by default), whether they land in one tick or are spread across several, and only a successful send resets the counter to zero. Once open it stays open for a fixed pause (60 seconds by default), measured in wall-clock time from the moment it opened regardless of how many ticks pass meanwhile, and it closes itself (resetting the counter) the next time anything checks it after that pause has elapsed. It is re-checked before every batch and before every row within a batch, not just once before the tick starts, so a failure that trips it mid-tick stops the rest of that tick's sends immediately instead of only from the next tick onward. The send counter and the breaker gauge are deliberately named without "kafka": a follow-up PR will mirror the same rows to the Runtime API over HTTP, and it will reuse this task and these metrics with a different fact instead of adding a parallel set of both.

The Ray code paths are left alone on purpose. Ray is being removed and its jobs never get an outbox row.

Still open before this comes out of draft: the combined pending query is only exercised against mocks, so its ordering and its batch limit need a test against a real database, and the branch is due one more review round.


gateway/api/migrations/0065_joboutbox.py
```diff
@@ -0,0 +1,69 @@
...
+                    "job_status",
+                    models.CharField(
+                        choices=[
+                            ("PENDING", "Pending"),
+                            ("RUNNING", "Running"),
+                            ("STOPPED", "Stopped"),
+                            ("SUCCEEDED", "Succeeded"),
+                            ("QUEUED", "Queued"),
+                            ("FAILED", "Failed"),
+                        ],
```
avilches, line 37: This is Django domain constraint, not a db contraint

gateway/api/migrations/0065_joboutbox.py
```diff
@@ -0,0 +1,69 @@
...
+                    "workload_status",
+                    models.CharField(
+                        blank=True,
+                        choices=[
+                            ("PENDING", "Pending"),
+                            ("RUNNING", "Running"),
+                            ("STOPPED", "Stopped"),
+                            ("SUCCEEDED", "Succeeded"),
+                            ("QUEUED", "Queued"),
+                            ("FAILED", "Failed"),
```
avilches, line 53: This is Django domain constraint, not a db contraint

gateway/api/migrations/0065_joboutbox.py
```diff
@@ -0,0 +1,69 @@
...
+                        null=True,
+                    ),
+                ),
+                ("license_fee_required", models.BooleanField()),
+                ("license_fee_sent_at", models.DateTimeField(blank=True, null=True)),
+                ("billing_sent_at", models.DateTimeField(blank=True, null=True)),
+            ],
+            options={
+                "indexes": [
+                    models.Index(fields=["status_changed_at"], name="job_outbox_status_changed_at_idx"),
```
avilches, line 65: Needed to process outbox by oldest first.

gateway/api/use_cases/programs/run.py
```diff
@@ -140,6 +145,16 @@ def _get_runner_config(
...
+        if function.provider_id is not None:
+            # This path never associates a FunctionSize row (see _config_for_profile_id),
+            # and a licensed function's usage event needs one to be billed correctly
+            # (the function size is part of the price key on the billing side, not just
+            # a label). Rather than accept the job and fail to bill it correctly later,
+            # reject it here so the caller can retry with 'function_size'.
+            raise FunctionConfigurationException(
+                "'compute_profile' cannot be used for a licensed function. Pass 'function_size' instead, "
+                "so the license fee can be billed correctly."
+            )
```
avilches, line 157: This would be removed when the compute_profile is deleted. For now it's just a safeguard.

gateway/api/admin.py
```diff
@@ -924,17 +926,20 @@ def _can_stop(self, job):
...
         _state.adding, not obj.pk, below: id defaults to a fresh uuid before the row is saved."""
         if obj._state.adding or not self._can_stop(obj):  # pylint: disable=protected-access
             return "-"
         stop_url = reverse("admin:job_stop_job_view", args=[obj.pk])
         button = f'<button type="button" class="button qs-stop-job-btn" data-stop-url="{stop_url}">Stop job</button>'
         return mark_safe(button)
 
+    def has_delete_permission(self, request, obj=None):
+        """Disabled: a Job with a pending outbox row must not be deleted casually from the admin."""
+        return False
```
avilches, line 941: Well, actually, it just avoids deleting the job...

gateway/core/ibm_cloud/event_streams/kafka_event_streams_client.py
```diff
@@ -33,6 +33,17 @@
...
 
 
+class UnroutableRegionError(RuntimeError):
+    """Raised when an event cannot be routed to a producer: the CRN's region could not be
+    determined, or no producer is configured for that region.
+
+    Unlike a plain RuntimeError from a failed produce()/flush() call, this is not a
+    transient Kafka outage: it is either bad data on the row or a deployment config gap,
+    and neither is fixed by pausing sends, so callers should not count it against a shared
+    circuit breaker.
```
avilches:
```suggestion
    """Raised when an event cannot be routed to a producer: the CRN's region could not be
    determined, or no producer is configured for that region.
```

